PR #9402, opened 2020-12-10
chafa: new port

herbygillot:
#### Description

New port for the [chafa](https://github.com/hpjansson/chafa) terminal graphics viewer

###### Tested on

macOS 10.15.7 19H15
Xcode 12.2 12B45b

###### Verification 
Have you

### DIFF
--- a/graphics/chafa/Portfile
+++ b/graphics/chafa/Portfile
@@ -1,0 +1,49 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        hpjansson chafa 1.4.1
+revision            0
+
+homepage            https://hpjansson.org/chafa
+
+description         Terminal graphics for the 21st century
+
+long_description    {*}${description}. Chafa is a command-line utility that \
+                    converts all kinds of images, including animated GIFs, \
+                    into sixel or ANSI/Unicode character output that can be \
+                    displayed in a terminal. It is highly configurable, with \
+                    support for alpha transparency and multiple color modes \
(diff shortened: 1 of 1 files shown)


graphics/chafa/Portfile
```diff
@@ -0,0 +1,51 @@
...
+                    port:libtool    \
+                    port:pkgconfig
+
+depends_lib-append  port:glib2 \
+                    port:ImageMagick
+
+configure.cmd       ./autogen.sh
+use_configure       yes
+
+makefile.has_destdir yes
```
neverpanic, line 45: Isn't passing `DESTDIR` the default even without the `makefile` PortGroup? In fact the Portfile works fine for me without the portgroup and this line.

neverpanic (changes requested):
I've also noticed that the build system checks for gtk-doc and xsltproc, and doesn't build documentation or a manpage when both aren't installed.

We don't want the package contents to change depending on what a user has installed on their system. Can you please check that either the documentation is always off, or always on?

graphics/chafa/Portfile
herbygillot: Thanks @neverpanic - I've removed `makefile.has_destdir`, but will keep the `makefile` portgroup.

neverpanic: Does the port not build correctly without the makefile PortGroup, or what is the rationale for that? I've never seen any autoconf ports use that, so I'm a bit surprised to see it.

herbygillot: Actually you're right - removed the portgroup as well.  Apologies for that @neverpanic

herbygillot:
> I've also noticed that the build system checks for gtk-doc and xsltproc, and doesn't build documentation or a manpage when both aren't installed.
> 
> We don't want the package contents to change depending on what a user has installed on their system. Can you please check that either the documentation is always off, or always on?

Well caught - I've changed the port file to disable building documentation, so that should sort that out. Many thanks for your keen reviews.

neverpanic (approved):
Any time. Looks good now!

PR #9407, opened 2020-12-11
waypoint: new port

herbygillot:
#### Description
New port for Hashicorp's **[Waypoint](https://www.waypointproject.io/)**

###### Tested on

macOS 10.15.7 19H15
Xcode 12.2 12B45b

###### Verification 
Have you

### DIFF
--- a/devel/waypoint/Portfile
+++ b/devel/waypoint/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/hashicorp/waypoint 0.2.0 v
+revision            0
+
+homepage            https://www.waypointproject.io
+
+description         A tool to build, deploy, and release any application on \
+                    any platform
+
+long_description    Waypoint allows developers to define their application \
+                    build, deploy, and release lifecycle as code, reducing \
+                    the time to deliver deployments through a consistent and \
+                    repeatable workflow.  Waypoint supports a number of build \
+                    methods and target platforms out of the box and more can \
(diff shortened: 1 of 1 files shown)


devel/waypoint/Portfile
```diff
@@ -0,0 +1,46 @@
...
+                    Instances, Google Cloud Run, Netlify, and many more...
+
+categories          devel
+license             MPL-2
+
+maintainers         {gmail.com:herby.gillot @herbygillot} \
+                    openmaintainer
+
+# Do not restrict Go from downloading dependencies at build time.
+build.env-delete    GOPROXY=off GO111MODULE=off
```
neverpanic, line 30: Instead of going this way, can you use go2port from the go2port port and generate a Portfile that lists the dependencies? That would have the advantage of not requiring downloads during the build phase, and allow us to cache the dependencies.

neverpanic (changes requested):

devel/waypoint/Portfile
herbygillot: As useful as `go2port` is, there are still lots of situations where it (or the `golang` port group) don't work very well.  The Hashicorp products use a massive amount of Go libraries, including quite a few that don't seem to work very well with `go2port`.  Examples of these are the Azure `autorest` libs and Google's `cloud-go` libs.  Hence the reason why I just decided to simply side-step vendored dips for this port.

herbygillot:
Being that the `golang` portgroup doesn't do well with `go.vendor` references that are more than three levels deep, like:

- github.com/Azure/go-autorest/tracing
- github.com/Azure/go-autorest/logger
- github.com/Azure/go-autorest/autorest/validation
- github.com/Azure/go-autorest/autorest/to
- github.com/Azure/go-autorest/autorest/mocks
- github.com/Azure/go-autorest/autorest/date
- github.com/Azure/go-autorest/autorest/azure/cli
- github.com/Azure/go-autorest/autorest/azure/auth
- github.com/Azure/go-autorest/autorest/adal
- github.com/Azure/go-autorest/autorest

...then we'll just check in for now since I don't see how to vendor these deps.

PR #9425, opened 2020-12-12
fzf: update to 0.24.4

cardi:
###### Tested on
macOS 10.14.6 18G6020
Xcode 11.3.1 11C504

###### Verification
Have you


herbygillot:
@cardi, I notice that your update is only updating `fzf` itself, and not its dependency Go libraries.  We shouldn't forget those or else there's a risk the software won't work as intended.

I've added a commit that adds those... you'll probably want to squash into a single commit and re-push.

PR #9437, opened 2020-12-13
optipng: fix arm64 linking error

dwatteau:
Disable `PNG_ARM_NEON_OPT` to avoid a linking error in bundled libpng for `png_init_filter_functions_neon()`.

Otherwise, the build currently fails on arm64 with the following error:

```
:info:build /usr/bin/clang -Os -arch arm64 -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX11.0.sdk -arch arm64 -o optipng optipng.o optim.o bitset.o ioutil.o ratio.o wildargs.o ../opngreduc/libopngreduc.a ../pngxtern/libpngxtern.a ../libpng/libpng.a ../zlib/libz.a ../gifread/libgifread.a ../pnmio/libpnmio.a ../minitiff/libminitiff.a   -lm
:info:build Undefined symbols for architecture arm64:
:info:build   "_png_init_filter_functions_neon", referenced from:
:info:build       _png_read_filter_row in libpng.a(pngrutil.o)
:info:build ld: symbol(s) not found for architecture arm64
:info:build clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

OptiPNG already disables a good number of libpng features, so I'm not sure that a more complicated fix restoring NEON support in optipng's limited build system would be worth it.

(An alternative solution is to make optipng use the MacPorts version of libpng, instead of using the bundled copy.)

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.0.1 20B29
Xcode 12.2 12B45b

###### Verification

Have you

### DIFF
--- a/graphics/optipng/Portfile
+++ b/graphics/optipng/Portfile
@@ -30,6 +30,8 @@ checksums       rmd160  bd6df4ef9f258474f352820eec545f482c7194cd \
 configure.cppflags
 configure.ldflags
 
+configure.cppflags-append -DPNG_ARM_NEON_OPT=0
+
 configure.args-append  --mandir=${prefix}/share/man
 
 configure.universal_args-delete --disable-dependency-tracking


neverpanic (changes requested):
Since this seems to affect all systems, even those that aren't arm64, you need to increase the revision – otherwise people that already have it installed will keep NEON support, and everybody that installs after this change is merged won't, but with the same version and revision number.

That can lead to situations that we can't debug, so the revision needs to change from 0 to 1 (and you'll have to add the revision line, since the port currently doesn't have one).

ryandesign:
> Since this seems to affect all systems, even those that aren't arm64, you need to increase the revision

I assume that the "neon" being referred to here is an [ARM-specific processor optimization](https://www.arm.com/why-arm/technologies/neon), like avx or sse on Intel CPUs. As such, this build failure only affects ARM processors and fixing it does not warrant increasing the revision.

> (An alternative solution is to make optipng use the MacPorts version of libpng, instead of using the bundled copy.)

Absolutely if optipng can successfully use MacPorts libpng we should be doing that instead. I will investigate. Making that change *would* require increasing the revision since it would affect all systems.

PR #9446, opened 2020-12-15
mame: add variant 'debug'; soft link tools in ${prefix}/bin; add tools man pages; improve link time; reduce warnings; eliminate superfluous files; remove variant 'python36'

mascguy:
#### Description

Mame fixes and enhancements:

- New variant `debug` added;
- Tools now prefixed with `mame-`, to avoid conflicts with system components;
- Tools now soft-linked in `${prefix}/bin`;
- Man pages for Tools now installed;
- Link time reduced significantly, from 10+ minutes to less than 60 seconds;
- Most compiler warnings suppressed, to limit pollution in build log;
- Superfluous files no longer installed;
- Variant `python36` removed.

Tracked by issue:
[61739](https://trac.macports.org/ticket/61739)

###### Type(s)
- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

###### Verification 
Have you

- [x] ensured the port does not create any faulty symbolic links?

### DIFF
--- a/emulators/mame/Portfile
+++ b/emulators/mame/Portfile
@@ -3,13 +3,10 @@
 #------------------------------------------------------------------------------
 # Portfile Style Notes
 #
-# * For cases where local variables are created within phases, they are
-#   explicitly unset when no longer being used. This is a conscious choice,
-#   and helps avoid any possible clashes elsewhere.
-# * To aid readability, intermediate variables, lists, etc, are often
-#   employed. While that makes the portfile longer, it also assists when
-#   trying to understand what is being done... aiding in diagnosing and
-#   debugging issues. Again, this is a conscious choice.
+# * When local variables are created within phases, they are unset when
+#   finished. This prevents clashes elsewhere.
+# * Intermediate variables, lists, etc, are often employed. While more
+#   verbose, it aids readability.
 #------------------------------------------------------------------------------
 
 PortSystem          1.0
@@ -18,11 +15,12 @@ PortGroup           github 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
 github.setup        mamedev mame 0226 mame
-revision            1
+revision            2
 
 version             [string index ${github.version} 0].[string range ${github.version} 1 end]
 categories          emulators
-maintainers         {@mascguy} openmaintainer
+maintainers         {@mascguy} \
+                    openmaintainer
 platforms           darwin
 license             GPL-2+
 homepage            https://www.mamedev.org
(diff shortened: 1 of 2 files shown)


emulators/mame/Portfile
```diff
@@ -125,9 +159,15 @@ depends_run-append  \
 # Build Options
 #------------------------------------------------------------------------------
 
-use_parallel_build  yes
+# Note: ARM64 support untested
+#supported_archs     ppc ppc64 i386 x86_64 arm64
+supported_archs     ppc ppc64 i386 x86_64
```
kencu, line 164: usually, we only exclude archs that are known not to work, rather than ones we don't know whether they work or not. Principle is to be more inclusive rather than exclusive, in general.

emulators/mame/Portfile
```diff
@@ -125,9 +159,15 @@ depends_run-append  \
...
 
+use_parallel_build  yes
```
kencu, line 166: This being the default, we don't usually bother to add it, as it's just clutter.

emulators/mame/Portfile
```diff
@@ -125,9 +159,15 @@ depends_run-append  \
...
 use_configure       no
+configure.ccache    yes
+
+build.cmd           ${prefix}/bin/gmake
```
kencu, line 170: you almost never need this, unless you're building on Tiger, which has a really old make program. If you really do need this, don't forget to have a build dep on port:gmake somewhere so it actually exists.

emulators/mame/Portfile
```diff
@@ -125,9 +159,15 @@ depends_run-append  \
...
 #------------------------------------------------------------------------------
 
-use_parallel_build  yes
+# Note: ARM64 support untested
+#supported_archs     ppc ppc64 i386 x86_64 arm64
+supported_archs     ppc ppc64 i386 x86_64
 
+use_parallel_build  yes
 use_configure       no
+configure.ccache    yes
```
kencu, line 168: don't we want the user's setting for this in macports.conf to decide this for a given installation? You don't see many (any) ports forcing ccache on...

emulators/mame/files/mame-patch-0226-scripts-genie-compile-warnings.diff
```diff
@@ -1,6 +1,13 @@
...
-+++ scripts/genie.lua	2020-11-28 17:01:38.000000000 -0500
-@@ -1002,6 +1002,7 @@
+--- scripts/genie.lua	2020-12-11 20:39:27.000000000 -0500
++++ scripts/genie.lua	2020-12-11 20:39:15.000000000 -0500
+@@ -996,12 +996,13 @@
+ 	buildoptions {
+ 		"-Wall",
+ 		"-Wcast-align",
+-		"-Wundef",
++		"-Wno-undef",
```
kencu, line 8: all the rest of MacOS and MacPorts are moving in the opposite direction, to ADD warnings and errors for undefined functions rather than suppress them... apparently for arm64 this is critical. Are we really sure that in this one case, we need to do the opposite?

emulators/mame/Portfile
mascguy: Makes sense. In this case, the Mame makefile doesn't yet support MacOS builds for ARM64... and builds are guaranteed to fail. Tackling this is next on my to-do list, though.

Given that, should this be changed, or left as-is?

emulators/mame/Portfile
mascguy: This port utilizes newer gmake option `--trace`, not available in Apple's version.

I'll add a comment to clarify that.

emulators/mame/files/mame-patch-0226-scripts-genie-compile-warnings.diff
mascguy: Good to know, as I wasn't aware. These can certainly be turned back on when ARM64 is tackled... or that can be done now.

Thoughts?

emulators/mame/Portfile
mascguy: Good point, I'll remove this.

emulators/mame/Portfile
mascguy: No problem, I'll remove.

emulators/mame/Portfile
mascguy: Since ARM64 will be tackled over the next two weeks, can we leave this as-is for now?

emulators/mame/files/mame-patch-0226-scripts-genie-compile-warnings.diff
mascguy: This has been added to the ARM64-related ticket: [61795](https://trac.macports.org/ticket/61795)

mascguy:
Ken, let me know if we're good-to-go. I'm sure you're swamped given the time of year, but these changes greatly improve the experience for Mame users. And it would be great to get these out ASAP, well before the end of December!

Cheers and Thanks,
-Chris

mascguy:
Thanks Ken, much appreciated as always!

PR #9452, opened 2020-12-17
universal-ctags: update version to 5.9.0.20201213

kaimingguo:
#### Description

Update `universal-ctags` version to latest release. and update Python default version (use 3.8).

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.2 12B45b

###### Verification 
Have you


devel/universal-ctags/Portfile
```diff
@@ -21,11 +21,11 @@ long_description \
...
 
-checksums           rmd160  e37a347213bf94639b68adacc5f66d90fe643388 \
-                    sha256  73ab06a4868f2006c2cdde2276db13f1fc010483b2711cd66ef00b51b680064d \
-                    size    1663558
+checksums           rmd160  e234b2cd0edb887df465e23bb46ecfbe17ce5c04 \
+                    sha256  db6e82f964364b829ff38ae59e49f64067f11dd9e743a8a993c9229d3ed9c0b7 \
+                    size    2024988
 
-set python_version  37
+set python_version  38
```
ryandesign, line 28: Should default to python 3.9 now to match fa147a0a5b020635eb792a94330a0cc5641cc7b3.
```suggestion
set python_version  39
```

devel/universal-ctags/Portfile
```diff
@@ -3,10 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        universal-ctags ctags feffe43a4a9d339cd6701744589746e2297e4567
+github.setup        universal-ctags ctags 0837c42c3fea8854499cf1d22f2f5de12d745f20
 name                universal-ctags
-version             20200111
-revision            1
+version             p5.9.20201213.0
+revision            3
```
ryandesign, line 9: When the version increases, the revision drops to 0.
```suggestion
revision            0
```

devel/universal-ctags/Portfile
```diff
@@ -3,10 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        universal-ctags ctags feffe43a4a9d339cd6701744589746e2297e4567
+github.setup        universal-ctags ctags 0837c42c3fea8854499cf1d22f2f5de12d745f20
 name                universal-ctags
-version             20200111
-revision            1
+version             p5.9.20201213.0
```
ryandesign, line 8:
```
$ ~/macports/macports-user-ryandesign/scripts/vercmp 20200111 p5.9.20201213.0
MacPorts considers 20200111 to be greater than p5.9.20201213.0.
```
Therefore the epoch needs to be increased.
```suggestion
epoch               1
version             p5.9.20201213.0
```
How did you determine that the version number should be p5.9.20201213.0? Specifically, beginning a version number with a letter is unusual. According to the project's configure.ac, its version is 5.9.0. So perhaps 5.9.0-20201213 would be more normal. In fact https://github.com/universal-ctags/ctags/commit/0837c42c3fea8854499cf1d22f2f5de12d745f20 was committed on 20201210. There are newer commits, should you wish to switch to one of them instead and update the date to match.
```suggestion
epoch               1
version             5.9.0-20201210
```

kaimingguo:
Sure :)

devel/universal-ctags/Portfile
kaimingguo: Okay, very thanks. default Python version to 3.9.

foamkeen:
@kaimingguo what about the patch this port has (https://github.com/macports/macports-ports/blob/master/devel/universal-ctags/files/patch-configure.ac.diff)? It adds `rst2man-3.7.py`, is it still necessary? I was able to build universal-ctags without this patch after changing python_version to 39.

devel/universal-ctags/Portfile
```diff
@@ -3,10 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        universal-ctags ctags feffe43a4a9d339cd6701744589746e2297e4567
+github.setup        universal-ctags ctags 0837c42c3fea8854499cf1d22f2f5de12d745f20
```
foamkeen, line 6: Since universal-ctags project provides git tags for releases, I suggest using it instead of commit hash. Also I think github.setup's tag_prefix (p in this case) is well suited here.
```suggestion
github.setup        universal-ctags ctags 5.9.0-20201213.0 p
```

devel/universal-ctags/Portfile
```diff
@@ -3,10 +3,10 @@
...
 name                universal-ctags
-version             20200111
-revision            1
+version             5.9.0-20201213
```
foamkeen, line 8: If we specify a version in github.setup, then we don't need it separately, however we need to increase the epoch since `vercmp` considers 2020011 greater than 5.9.0.
```suggestion
epoch             1
```

kaimingguo:
> @kaimingguo what about the patch this port has (https://github.com/macports/macports-ports/blob/master/devel/universal-ctags/files/patch-configure.ac.diff)? It adds `rst2man-3.7.py`, is it still necessary? I was able to build universal-ctags without this patch after changing python_version to 39.

To generate the man pages `rst2man` command is needed.  Installing `py-docutils` will only generate a symbolic link with the version number in `prefix` (ex. `rst2man-3.9.py`), so need patch command.

i deleted the original patch file and used `rst2man-${python.version}.py` in `prefix`.

devel/universal-ctags/Portfile
```diff
@@ -3,10 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        universal-ctags ctags feffe43a4a9d339cd6701744589746e2297e4567
+github.setup        universal-ctags ctags 5.9.20201227.0 p
 name                universal-ctags
-version             20200111
-revision            1
+epoch               2
```
foamkeen, line 8: Why is it necessary to set epoch to 2? According to docs it's 0 by default and only needs to be increased by 1.

kaimingguo: Sorry, the wrong number entered.

xdxu:
Did anyone run into the issue I mentioned here?
https://github.com/macports/macports-ports/commit/39e66a0cec14c981b44ca90fb006da61d37ef43c#r46521918

kencu:
indeed, I see that. Apologies, looked right, came from port maintainer. 

@kaimingguo  -- this block is broken. What would you like us to do to fix it?

kaimingguo:
@xdxu Could you please upload the installation log and OS version?
I think the system did not find rst2man script (like `rst2man-3.7.py`, etc) in `$prefix`. you can install `py-docutils` first:

```
sudo port install py-docutils
```

kencu:
For one, the reinplace block should only be run if the manpages variant is suggested. Probably it should be moved inside the manpages variant, I guess.

For another, you can't run "sudo port patch" unless the right deps are installed, or you get an error. So that is a problem, slightly harder to fix.

kaimingguo:
> You can see the same thing on any os version. Even on your own system, right now. Just uninstall the py39-docutils and run "sudo port -v patch".
> 
> For one, the reinplace block should only be run if the manpages variant is selected. Probably it should be moved inside the manpages variant, I guess.
> 
> For another, you can't run "sudo port patch" unless the right deps are installed, or you get an error. So that is a problem, slightly harder to fix.

Okay, the problem mainly occurs in "if no `rst2man-*.py` were found" the `glob` command throws an error. I add `-nocomplain` option can fix this problem.

I will test and resubmit the PR :)

### Reference:

- https://wiki.tcl-lang.org/page/glob

PR #9456, opened 2020-12-17
mos-devel: new port

harens:
#### Description

See https://github.com/mongoose-os/mos/issues/37

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.6 19G73
xcode-select version 2373

###### Verification 
Have you

### DIFF
--- a/devel/mos-devel/Portfile
+++ b/devel/mos-devel/Portfile
@@ -3,17 +3,21 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/mongoose-os/mos 2.18.0
-conflicts           ${name}-devel
-revision            1
+name                mos-devel
+go.setup            github.com/mongoose-os/mos 8291cd5ccb64c455ef1a17c1bee917d1a0ee794e
+conflicts           mos
+revision            0
+version             202012192021
 categories          devel
 license             Apache-2
 
 maintainers         {@harens gmail.com:harensdeveloper} \
                     openmaintainer
-description         Mongoose OS command-line tool
+description         Development version of the Mongoose OS command-line tool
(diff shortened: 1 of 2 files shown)


herbygillot:
I wonder about the situation where the devel version may require slightly different versions of Go libraries than the release version.  I suppose when that case arises you can simply add a separate `go.vendors` block inside the sub-port...

harens:
You're right @herbygillot, and I've now added a separate go.vendors block anyway to get the tests to pass (see https://github.com/mongoose-os/mos/issues/37#issuecomment-747448914)

This makes it a bit more difficult to maintain (unlike the [homebrew version](https://github.com/cesanta/homebrew-mos/blob/master/Formula/mos-latest.rb) of the devel stream where `mod download` is simply run), but it should still be feasible.

herbygillot:
> You're right @herbygillot, and I've now added a separate go.vendors block anyway to get the tests to pass (see [mongoose-os/mos#37 (comment)](https://github.com/mongoose-os/mos/issues/37#issuecomment-747448914))
> 
> This makes it a bit more difficult to maintain (unlike the [homebrew version](https://github.com/cesanta/homebrew-mos/blob/master/Formula/mos-latest.rb) of the devel stream where `go mod download` is simply run to install deps), but it should still be feasible.

This looks a little too unweildy to me @harens, and maybe might merit seperate Portfiles for `mos` and its devel version, but that's certainly up to you.

harens:
Good point @herbygillot. At this point, you're definitely right that it's better as two separate ports. I also added notes as to what manual changes need to be done to `go.vendors` following an update.

harens:
Thanks @herbygillot for your help

PR #9464, opened 2020-12-18
tilt: new port

dgilmanTORQATA:
#### Description
Unfortunately the yarn stuff requires network i/o to build, but at least it has a lockfile. The go dependencies are all vendored.

###### Type(s)

- [x] enhancement

###### Tested on
macOS 10.15.7 19H114
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/devel/tilt/Portfile
+++ b/devel/tilt/Portfile
@@ -1,0 +1,40 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/tilt-dev/tilt 0.18.6 v
+revision            0
+
+categories          devel
+maintainers         nomaintainer
+platforms           darwin
+license             Apache-2
+
+description         Run Kubernetes environments locally, for local development
+
+long_description    Tilt powers multi-service development and makes sure they \
+                    behave! Run tilt up to work in a complete dev environment \
+                    configured for your team. \
+                    \
+                    Tilt automates all the steps from a code change to a new \
+                    process: watching files, building container images, and \
+                    bringing your environment up-to-date. Think docker build \
+                    && kubectl apply or docker-compose up.
+
+homepage            https://tilt.dev/
+
+checksums           rmd160  17ff0a5a8b343760d4b719aa01337cdb0ca488cd \
+                    sha256  f4f51a362b326f3df4c315bb3c932a87b2d7a5a29e6b72bb85beb0f0d82c42df \
+                    size    23480949
+
+depends_build-append \
+                    port:yarn port:nodejs12
+use_parallel_build  no
+build.cmd           "make build-js && ${go.bin} build -mod vendor ./cmd/tilt"
+build.env-replace   GO111MODULE=off GO111MODULE=on
+
+destroot {
+    xinstall -m 0755 ${build.dir}/${name} ${destroot}${prefix}/bin/
+}
+


devel/tilt/Portfile
```diff
@@ -0,0 +1,39 @@
...
+                    bringing your environment up-to-date. Think docker build \
+                    && kubectl apply or docker-compose up.
+
+homepage            https://tilt.dev/
+
+checksums           rmd160  d2bd6001e581f54986011e10f44a58b68b6f2f80 \
+                    sha256  1a4cb78c4e880157d22d6b8e209d7fc70e77dfef0136929047e5eeee96b6841c \
+                    size    20524516
+
+depends_build       port:yarn port:nodejs12
```
harens, line 32: The golang portgroup adds go as a build dependency. You should instead use `depends_build-append` to not overwrite this.

https://github.com/macports/macports-ports/blob/1a1667c302671c5f1043338ac0bbd8e44acd3df2/_resources/port1.0/group/golang-1.0.tcl#L143

I think this is what's leading to the error ```sh: /opt/local/bin/go: No such file or directory```

devel/tilt/Portfile
```diff
@@ -0,0 +1,39 @@
...
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/tilt-dev/tilt 0.18.1 v
+revision            0
+
+categories          devel
+maintainers         nomaintainer
+platforms           darwin
+supported_archs     x86_64
```
harens, line 12: Unless you're sure i386 isn't supported, I'd use the default set by the golang portgroup.

https://github.com/macports/macports-ports/blob/1a1667c302671c5f1043338ac0bbd8e44acd3df2/_resources/port1.0/group/golang-1.0.tcl#L129

devel/tilt/Portfile
```diff
@@ -0,0 +1,39 @@
...
+homepage            https://tilt.dev/
+
+checksums           rmd160  d2bd6001e581f54986011e10f44a58b68b6f2f80 \
+                    sha256  1a4cb78c4e880157d22d6b8e209d7fc70e77dfef0136929047e5eeee96b6841c \
+                    size    20524516
+
+depends_build       port:yarn port:nodejs12
+use_parallel_build  no
+build.cmd           "make build-js && ${go.bin} build -mod vendor ./cmd/tilt"
+build.env-replace    GO111MODULE=off GO111MODULE=on
```
harens, line 35: I completely understand if this is required, but it runs into https://trac.macports.org/ticket/61192. It would be preferable to use `go.vendors`, but please correct me if this is not viable for this port.

(This doesn't matter as much anyway though since the JS deps have no choice but to be downloaded at build time)

harens (changes requested):
Thanks @dgilmanTORQATA. Just a few things from me.

devel/tilt/Portfile
dgilmanTORQATA: I had tried the port2go thingy, but there was at least one unsupported source repo (google's repo for protobuf), and I suspect there were several more. There was roughly 1000 lines of golang deps!

dgilmanTORQATA: I can leave a comment explaining the history

harens: Just looking through it, it should be possible.

Here are a few things I noticed. There will be more manual changes that will need to be made, but this should help:

|Original   |  New |  Notes |
|---|---|---|
|  go.googlesource.com/protobuf |  github.com/protocolbuffers/protobuf-go |  Set using repo |
|  gopkg.in/fsnotify.v1  |  github.com/fsnotify/fsnotify |  Set using repo |
|  v17.12.0-ce-rc1.0.20181016164535-5271c7cb484d+incompatible  |  v17.12.0-ce-rc1 |  Set using lock for docker/docker. You might have to use repo to set it to moby/moby |

I would recommend looking at the beginning comments of the [golang portgroup](https://github.com/macports/macports-ports/blob/d79aad1b301412f3ff83438f4e0c23d3c513aeb2/_resources/port1.0/group/golang-1.0.tcl) to help. Let me know if anything's not clear and I'll try and assist you :)

devel/tilt/Portfile
```diff
@@ -0,0 +1,42 @@
...
+homepage            https://tilt.dev/
+
+checksums           rmd160  d2bd6001e581f54986011e10f44a58b68b6f2f80 \
+                    sha256  1a4cb78c4e880157d22d6b8e209d7fc70e77dfef0136929047e5eeee96b6841c \
+                    size    20524516
+
+depends_build-append \
+                    port:yarn port:nodejs12
+use_parallel_build  no
+build.cmd           "make build-js && ${go.bin} build -mod vendor ./cmd/tilt"
```
harens, line 34:
```suggestion
configure.cmd       make
configure.args      build-js

build.args          -mod vendor ${worksrcpath}/cmd/${name}
```

In the worst case scenario, where mod download has to be used, we can at least simplify the configure and build process.

Note that we can take advantage of the default build.cmd
https://github.com/macports/macports-ports/blob/12a814f20461168f3cc2c1fa7b24d0580819b739/_resources/port1.0/group/golang-1.0.tcl#L148

devel/tilt/Portfile
dgilmanTORQATA: I'm going to push up a WIP. After a bunch of hacking on the go.vendors I have everything downloading and verifying, but the pre-checkout GOPATH setup stuff in the golang portgroup is failing. It seems to be due to edge cases with identical versions of packages installed under different names.

devel/tilt/Portfile
```diff
@@ -0,0 +1,1296 @@
...
+                        rmd160  d2bd6001e581f54986011e10f44a58b68b6f2f80 \
+                        sha256  1a4cb78c4e880157d22d6b8e209d7fc70e77dfef0136929047e5eeee96b6841c \
+                        size    20524516
+
+depends_build-append \
+                    port:yarn port:nodejs12
+configure.cmd       make
+configure.args      build-js
+
+build.args          -mod vendor ${worksrcpath}/cmd/${name}
```
harens, line 37: In the hopes that the go.vendors works, this can then be removed.

harens (changes requested):
Thanks @dgilmanTORQATA for your work so far. A lot of great progress has been made!

> It seems to be due to edge cases with identical versions of packages installed under different names.

Assuming this is gorethink, I see what you mean about the edge case. My suggestion would be the following:

1) Keep one of them (e.g. gopkg.in/dancannon/gorethink.v3).
2) Remove the other (e.g. gopkg.in/gorethink/gorethink.v3).
3) Copy the contents over after the extraction has finished.

```tcl
post-extract {
    # Identical versions of gorethink installed under different names
    file mkdir ${gopath}/src/gopkg.in/gorethink
    copy ${gopath}/src/gopkg.in/dancannon/gorethink.v3 ${gopath}/src/gopkg.in/gorethink/gorethink.v3
}
```

As always, let me know if anything's not clear, and thanks again for your work on this.

harens:
> - [x] referenced existing tickets on Trac with full URL?

Really? ;)

dgilmanTORQATA:
I don't see any Trac tickets for this

harens:
Sorry @dgilmanTORQATA that's just me being picky 🙂. Generally, the box is ticked if there are trac tickets.

As for the port, assuming the go.vendors work, it should be nearly done. Thanks again for your work.

dgilmanTORQATA:
Pushing another WIP. After untangling the duplicate packages I'm getting this "expects import" error around jsonpath. I am not sure why that doc.go file is breaking the build here and not with the earlier vendored build / in upstream's builds, as they all have that canonical import path explicitly specified.

dgilmanTORQATA:
I've re-pushed the initial commit that uses the entirely vendored dependencies from the distribution's tarball. The dependency tree here is complicated, I wasn't able to get it to work after spending a decent time on it, so I think it's better to be able to ship this with at least good vendoring instead of perfect vendoring.

PR #9473, opened 2020-12-19
fix errors when compiling using Xcode 12

toy:
#### Description

Fix issues in code that are causing error in Xcode 12 and caused warnings in previous versions. Namely include missing stdio.h, use int as return type for main and use return instead of exit in main.

https://trac.macports.org/ticket/61870

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H15
Xcode 12.3 12C33

###### Verification
Have you

### DIFF
--- a/math/primegen/Portfile
+++ b/math/primegen/Portfile
@@ -4,6 +4,7 @@ PortSystem          1.0
 
 name                primegen
 version             0.97
+revision            1
 categories          math
 platforms           darwin
 maintainers         yandex.com:bstj openmaintainer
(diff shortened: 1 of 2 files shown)


herbygillot:
Thanks for this submission, @toy.  However one comment: the commit message should start with the port name. So the message should look like:

`primegen: fix errors when compiling using Xcode 12`

toy:
Thanks for quick reply @herbygillot, I fixed the subject and force-pushed.

herbygillot:
> Thanks for quick reply @herbygillot, I fixed the subject and force-pushed.

Looks good.  We'll leave the PR open for a bit longer for others to review.  Thanks again.

math/primegen/Portfile
```diff
@@ -4,6 +4,7 @@ PortSystem          1.0
 
 name                primegen
 version             0.97
+revision            1
```
ryandesign, line 7: This only fixes a build failure, therefore don't increase the revision.

PR #9477, opened 2020-12-20
python39: tweak copyfile patch

kencu:
closes: https://trac.macports.org/ticket/61848

Note @jmroot -- there is one bit of this that should be upstreamed.

You can't do this on one line:
```
#if defined(__has_builtin) && __has_builtin(__builtin_available)
```
You have to do it on two lines:
```
#if defined(__has_builtin)
#if  __has_builtin(__builtin_available)
```
otherwise you get an error parsing this, at least with gcc-4.2.

tested on 10.5 PPC, does quite well compared to previous.

```

389 tests OK.

13 tests failed:
    test_asyncio test_c_locale_coercion test_concurrent_futures
    test_gdb test_multiprocessing_forkserver
    test_multiprocessing_main_handling test_multiprocessing_spawn
    test_os test_pkgutil test_readline test_socket test_subprocess
    test_zoneinfo

1 test altered the execution environment:
    test_posix

22 tests skipped:
    test_ctypes test_dbm_gnu test_devpoll test_epoll test_idle
    test_ioctl test_msilib test_multiprocessing_fork test_ossaudiodev
    test_poll test_spwd test_startfile test_tcl test_tix test_tk
    test_ttk_guionly test_ttk_textonly test_turtle test_winconsoleio
    test_winreg test_winsound test_zipfile64

13 re-run tests:
    test_asyncio test_c_locale_coercion test_concurrent_futures
    test_gdb test_multiprocessing_forkserver
    test_multiprocessing_main_handling test_multiprocessing_spawn
    test_os test_pkgutil test_readline test_socket test_subprocess
    test_zoneinfo

Total duration: 43 min 33 sec
Tests result: FAILURE then FAILURE
make: *** [test] Error 2
```

### DIFF
--- a/lang/python39/files/patch-no-copyfile-on-Tiger.diff
+++ b/lang/python39/files/patch-no-copyfile-on-Tiger.diff
@@ -41,14 +41,26 @@ diff --git Modules/posixmodule.c Modules/posixmodule.c
 index 01e8bcb..ff7fb30 100644
 --- Modules/posixmodule.c
 +++ Modules/posixmodule.c
-@@ -8,6 +8,7 @@
-    test macro, e.g. '_MSC_VER'. */
+@@ -56,7 +56,10 @@
+  */
+ #if defined(__APPLE__)
  
- #ifdef __APPLE__
+-#if defined(__has_builtin) && __has_builtin(__builtin_available)
 +#include <AvailabilityMacros.h>
-    /*
-     * Step 1 of support for weak-linking a number of symbols existing on
-     * OSX 10.4 and later, see the comment in the #ifdef __APPLE__ block
++
++#if defined(__has_builtin)  
++#if __has_builtin(__builtin_available)
+ #  define HAVE_FSTATAT_RUNTIME __builtin_available(macOS 10.10, iOS 8.0, *)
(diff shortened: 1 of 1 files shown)


jmroot (approved):

jmroot:
BTW, the logic is not quote correct in the case where `__has_builtin` is defined but `__builtin_available` is not, but it works fine when only applied on <= 10.5.

kencu:
Hmm. I added that extra #endif just for that reason...

It was this, which doesn't work properly:
```
#if defined(__has_builtin) && __has_builtin(__builtin_available)

#else
```

and I made it this...

```
#if defined(__has_builtin) 
#if __has_builtin(__builtin_available)


#endif 

#else
```

If it's wrong, I'm not seeing it immediately... should it be this?

```
#if defined(__has_builtin) 
#if defined __has_builtin(__builtin_available)


#endif 

#else
```

kencu:
Oh yes, I do see it now.

kencu:
I will have to think about how that is supposed to be done correctly then, without too much spaghetti

jmroot:
python/cpython#23873

PR #9478, opened 2020-12-20
qemu: update to 5.2.0

sirn:
#### Description

Update QEMU; now requires ninja/perl5 to build. May have error with "too many open files" on system with low maxfiles limit.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 11.0.1 20B29
Xcode 12.2 12B45b

###### Verification
Have you

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -7,8 +7,8 @@ PortGroup compiler_blacklist_versions 1.0
 PortGroup legacysupport 1.0
 
 name                    qemu
-version                 5.0.0
-revision                1
+version                 5.2.0
+revision                0
 categories              emulators
 license                 GPL-2+
 platforms               darwin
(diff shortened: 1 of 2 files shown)


emulators/qemu/Portfile
```diff
@@ -25,15 +25,17 @@ homepage                https://www.qemu.org
...
-                        size    62426192
+checksums               rmd160  2c33e773f012e333f99237e3d4ff1653ea0bc88f \
+                        sha256  cb18d889b628fbe637672b0326789d9b0e3b8027e0445b936537c78549df17bc \
+                        size    106902800
 
 patchfiles              patch-configure.diff
 patch.pre_args          -p1
 
-depends_build           port:texinfo \
+depends_build           port:perl5 \
```
raimue, line 35: As perl is not linked, I would prefer to move it to the python section with `license_noconflict` below.

raimue (commented):
Thank you for working on this! This is much appreciated as I had no time yet to test any build of the new qemu releases.

I think we should also make sure that the internal meson is used. Otherwise it could pick up `${prefix}/bin/meson` if it is installed (with unknown compatibility). If I read the configure script right, that would be `--meson=internal` as another argument.

It also looks like QEMU is now using C++11 with `-std=gnu++11`, which is very likely yet another problem for legacy systems. However, it's already failing to build on macOS <= 10.11. Just noting it here, I personally will not invest time to handle that.

sirn:
Done :)

raimue:
Thank you again!

PR #9483, opened 2020-12-20
xmedcon: update to version 0.16.3

enlf-dev:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/science/xmedcon/Portfile
+++ b/science/xmedcon/Portfile
@@ -7,6 +7,8 @@ version             0.16.3
 revision            0
 categories          science graphics
 maintainers         nomaintainer
+license             LGPL-2+
+
 description         an open source toolkit for medical image conversion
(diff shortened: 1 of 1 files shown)


herbygillot (changes requested):
Thanks for this submission @enlf-dev. However as per the Github Actions output here: https://github.com/macports/macports-ports/pull/9483/files , ....it looks like there is no license field set for `xmedcon`, which we should remedy while updating its Portfile.

Looking at the Sourceforge page, the project says it's under **LGPL-2** and later.

So as per the [list of valid licenses MacPorts supports](https://trac.macports.org/wiki/PortfileRecipes#licensekeyword), we should add the following:

```
license          LGPL-2+
```

PR #9521, opened 2020-12-24
tk: fix configure error with Xcode 12.3

lmhaffner:
#### Description

* add patch to fix a warning now elevated to an error in newer Xcode
releases, which causes a configure test to fail:

```
conftest.c:26:7: error: implicitly declaring library function 'exit'
  with type 'void (int) __attribute__((noreturn))'
  [-Werror,-Wimplicit-function-declaration]
        exit(2);
        ^
conftest.c:26:7: note: include the header <stdlib.h> or explicitly
  provide a declaration for 'exit'
```

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/x11/tk/Portfile
+++ b/x11/tk/Portfile
@@ -46,6 +46,9 @@ patchfiles-append   patch-dyld_fallback_library_path.diff
 # Remove when updating to 8.6.11
 patchfiles-append   patch-macosx-tkMacOSXInit.c.diff
 
+# Fix for impliclity defined error during configure under macOS 11.
+patchfiles-append   patch-implicitly-defined.diff
+
 configure.args      --mandir=${prefix}/share/man --with-tcl=${prefix}/lib
 # see https://trac.macports.org/ticket/58447
 configure.cppflags-replace -I${prefix}/include -isystem${prefix}/include
--- a/x11/tk/files/patch-implicitly-defined.diff
+++ b/x11/tk/files/patch-implicitly-defined.diff
@@ -1,0 +1,13 @@
+--- unix/configure.orig	2020-12-23 21:29:13.000000000 -0600
++++ unix/configure	2020-12-23 21:30:37.000000000 -0600
+@@ -3078,8 +3078,8 @@
+   for (i = 0; i < 256; i++)
+     if (XOR (islower (i), ISLOWER (i))
+ 	|| toupper (i) != TOUPPER (i))
+-      exit(2);
+-  exit (0);
++      return (2);
++  return (0);
+ }
+ _ACEOF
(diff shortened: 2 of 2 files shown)


ryandesign:
Did this problem cause the build to fail? If so, then the revision should not be increased.

Or did the problem not cause the build to fail: the build and installation succeeded but the software worked incorrectly? In that case, the revision should indeed be increased.

lmhaffner:
Nothing was obviously failing in the binary installation. However, fixing this config test failing (checking for ANSI C header files) changes the checksum of one of the installed binary files (libtkstub8.6.a). Sorry, I didn't mention that for the revision bump.

ryandesign:
The checksums of every .a file change every time you build them...

lmhaffner:
Ah yes, timestamp. Sigh. 

OK, so revision should probably not be incremented? If so, I can change and update the branch. New to submitting patch pulls here...

ryandesign:
> OK, so revision should probably not be incremented?

I don't know because I haven't analyzed this software. The way to answer the question is:

Does this change cause the software to change for anyone who already had it installed? If yes, increase revision. If no, don't.

lmhaffner:
Seeing that the installed binaries are identical, then no. I'll roll back the revision.

chrstphrchvz:
I believe the `tcl` port is also affected by this. I have reported the issue upstream: https://core.tcl-lang.org/tcl/info/43d371ac19

chrstphrchvz:
> I believe the `tcl` port is also affected by this.

Taken care of by 923068766f, which did bump the revision.

lmhaffner:
There were many of these errors in the tcl port configure, and one did affect the output binaries (esp. libtclstub.a). Fixes in tcl allows the tclx package to build successfully under Xcode 12.3. So, I think the revision bump was warranted there.

Here, I don't see any changes to the installed outputs (after accounting for my .a timestamp mistake).

chrstphrchvz:
[An equivalent fix (include <stdlib.h>)](https://core.tcl-lang.org/tk/info/4c15c03b14) will appear in Tcl/Tk 8.6.11, which should be released within the next few days.

chrstphrchvz:
Tcl/Tk 8.6.11 update PR opened: #9626

lmhaffner:
OK! Should I simply close this PR? Or is there a better way to finalize things? Sorry, still new to processes...

chrstphrchvz:
This PR should get closed automatically; no further action needed.

PR #9537, opened 2020-12-27
pandoc: install binary for BigSur

kencu:
this port installs the x86_64 binary from
Catalina on BigSur Intel and arm64

ghc/stack/cabal will work on BigSur Intel
at some point. Just when they might work on arm64
is less clear.

see: https://trac.macports.org/ticket/61422


kencu:
This pandoc binary is a 100MB self-contained monolith that has only system library dependencies. 

It could use some further real-world testing to be certain that no warts exist.

kencu:
Tested on both Intel and arm64 BigSur.

It will be some months, I believe, before these systems are fully supported in stack/cabal/ghc. 

A number of ports rely on pandoc. This gets them building, and other than masquerading as an arch that they are not, do the job.

kencu:
@jmroot @ryandesign  -- just FYI what I did here to make this work for now is a bit non-standard.

PR #9550, opened 2020-12-29
gcc-devel for ARM64: update to 11-20201227

michaelld:
#### Description

update gcc-devel for ARM64 to 11-20201227

+ use "aarch64", not "arm64" for the first entry of the build triplet, per discussion with the GCC Darwin lead

+ tweak "long_description" to be a little more explicit in the status of this port

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D5029f
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -20,7 +20,8 @@ license             {GPL-3+ Permissive}
 description         The GNU compiler collection, prerelease BETA snapshot.
 long_description    The GNU compiler collection, including front ends for \
                     C, C++, Objective-C, Objective-C++ and Fortran. \
-                    This is a prerelease BETA version! built from GCC snapshots.
+                    This is an !experimental!, !BETA! version \
+                    built from GCC snapshots.
 
 if { ${build_arch} eq "arm64" } {
     # Use lead Darwin developer branch for now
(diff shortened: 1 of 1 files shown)


kencu:
Makes sense.

not sure exactly how did got away without needing this before, but ... okay. Something no doubt changed.

It is a cryin' shame that we will forever have to monkey around with ```arm64``` ==> ```aarch64``` forever on all the open source software. It is exactly this kind of thing that generates a major PITA and dozens of tickets -- but it is what it is!

kencu (approved):

michaelld:
> Makes sense.
> 
> not sure exactly how we got away without needing this before, but ... okay. Something no doubt changed.
> 
> It is a cryin' shame that we will forever have to monkey around with `arm64` ==> `aarch64` forever on all the open source software. It is exactly this kind of thing that generates a major PITA and dozens of tickets -- but it is what it is!

I _think_ that inside the top-level `config.*` scripts `aarch64` is an "alias" for `arm64`; `aarch64` is clearly the internal-to-GCC name for the new Apple ARM64 CPU. This alias worked up until [`libcody` was added](https://github.com/gcc-mirror/gcc/commit/362303298ac4c1f93bda87535df2b726481d54bb). `libcody` uses its own `config.*` scripts separately from the top-level ones, and the `libcody` versions are old enough that they do not support `arm64` aliasing ... but they do support `aarch64` directly.

Sometime in the near-ish future I expect the GCC folks will either update the GCC build to use just the top-level scripts or update the `libcody` scripts ... and then both `arm64` and `aarch64` will supported.

Or: The above is my best guess based on a cursory reviewing of the recent changes to the top-level `config.*` scripts ...

michaelld:
I'm going to merge as-is because it works ... we can tweak as necessary down the road.

PR #9580, opened 2020-12-31
tmux: update livecheck regex

herbygillot:
Fix livecheck regex for tmux

### DIFF
--- a/sysutils/tmux/Portfile
+++ b/sysutils/tmux/Portfile
@@ -69,3 +69,6 @@ post-destroot {
 }
 
 notes "If you want integration with system pasteboard consider installing port tmux-pasteboard as well"
+
+
+github.livecheck.regex {([0-9.]+[a-z])}


tessus:
I can't merge, since I don't have the permissions, but I'm ok with the change.

Although it's not fixing, but adding a livecheck. ;-)

herbygillot:
Thanks @tessus, @kencu

ryandesign:
> Although it's not fixing, but adding a livecheck. ;-)

It's fixing the livecheck, by overriding the default livecheck.

Except that it also breaks livecheck, if the project ever releases a version that does not end with a letter, like it did previously with versions 3.0 and 3.1.

herbygillot:
> > Although it's not fixing, but adding a livecheck. ;-)
> 
> It's fixing the livecheck, by overriding the default livecheck.
> 
> Except that it also breaks livecheck, if the project ever releases a version that does not end with a letter, like it did previously with versions 3.0 and 3.1.

You're right, my fault.  This was why I made it a PR for peer review.

I've fixed it in this commit: https://github.com/macports/macports-ports/commit/ac7f63ec55f26b8356fb37c852340633e13e53d0

PR #9589, opened 2021-01-01
py-rich: update to 9.6.1

harens:
#### Description

Created with [seaport](https://github.com/harens/seaport)

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.6
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/python/py-rich/Portfile
+++ b/python/py-rich/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           python 1.0
 
 name                py-rich
-version             9.6.0
+version             9.6.1
 platforms           darwin
 license             MIT
 supported_archs     noarch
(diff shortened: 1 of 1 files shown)


herbygillot:
`seaport` looks like it's coming along nicely.

Perhaps one future feature could be that when doing a livecheck, if an alpha/beta/rc version is detected on a port that isn't `-devel`, maybe show a warning and recommend adjusting the port's livecheck regex?  Or perhaps that's out of scope?

harens:
Thanks @herbygillot (both for merging and the suggestion).

>Perhaps one future feature could be that when doing a livecheck, if an alpha/beta/rc version is detected on a port that isn't -devel, maybe show a warning and recommend adjusting the port's livecheck regex? Or perhaps that's out of scope?

Maybe...If alpha/beta/etc. explicitly appears in the livecheck then this is definitely possible. It would be harder to do if it was only mentioned in the release title and not the tag, but even then it might be feasible.

N.B. I don't think adjusting the livecheck should be necessary, since `seaport --bump VERSION` can be used to manually set the version if required. Let me know if I misunderstood and thanks for your proposal.

PR #9594, opened 2021-01-01
Babllibrsvg

kencu:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

librsvg is a big PITA due to rust, which is a heavy dependency.

As much as possible, we should get rid of it, but for now, we can at least move it into variants where it is actually needed, eg +docs variants where it is used to convert image formats, usually icons, around.


kencu:
I cannot find any use for librsvg in babl* other than to run the binary rsvg-convert optionally, to convert image formats for the docs variant.

PR #9598, opened 2021-01-02
mindforger: new port

herbygillot:
#### Description

New port for the **[MindForger](https://www.mindforger.com)** GPLv2 personal knowledge base.

###### Tested on

macOS 10.15.7 19H114
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/editors/mindforger/Portfile
+++ b/editors/mindforger/Portfile
@@ -1,0 +1,70 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github  1.0
+PortGroup               cmake   1.1
+
+github.setup            dvorka mindforger 1.52.0
+revision                0
+
+homepage                https://www.mindforger.com
+
+description             Thinking notebook and Markdown editor
+
+long_description        MindForger is an open, free, well-performing Markdown \
+                        editor which respects your privacy and enables \
+                        security.  It is actually more than an editor or IDE \
+                        - it's a human mind -inspired personal knowledge \
+                        management tool.
+
+categories              editors office
+license                 GPL-2
+platforms               darwin
+
+maintainers             {gmail.com:herby.gillot @herbygillot} \
+                        openmaintainer
+
+fetch.type              git
+
+set qmake_bin           ${prefix}/libexec/qt5/bin/qmake
+
+depends_build-append    port:pkgconfig
+
+# mindforger is a Qt5 project, but one of its dependencies, cmark-gfm, uses
+# CMake.  The qmake5/qt5 portgroups do not mix very well with the cmake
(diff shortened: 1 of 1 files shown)


editors/mindforger/Portfile
```diff
@@ -0,0 +1,64 @@
...
+categories              editors office
+license                 GPL-2
+platforms               darwin macosx
+
+maintainers             {gmail.com:herby.gillot @herbygillot} \
+                        openmaintainer
+
+fetch.type              git
+
+depends_build-append    port:ccache \
```
kencu, line 29: you don't require ccache, do you?

editors/mindforger/Portfile
```diff
@@ -0,0 +1,64 @@
...
+                        port:cmake
```
kencu, line 30: 99.999% of the time, you want the cmake portgroup instead of this...

editors/mindforger/Portfile
```diff
@@ -0,0 +1,64 @@
...
+
+configure.cmd           ${qt_qmake_cmd}
+configure.pre_args
+configure.args          -r ./mindforger.pro
+configure.env           CCACHE_DIR=${_cc_tmp_path}
+
+build.env-append        CCACHE_DIR=${_cc_tmp_path}
+
+post-extract {
+    system -W ${worksrcpath} "git submodule update --init --recursive"
```
kencu, line 48: this is possible, but will not be overly popular for all the usual reasons. No option?

kencu:
You're doing a fair amount of manual work here... macports' usual mechanisms not work for you? Looks like a 4-liner port...

kencu:
I think I see why you're running into so much trouble. This is a qt5 project that uses qmake.

You want to be using the qmake5 PortGroup here instead of most of this manual stuff.

The qmake5 PG is not the same as the qt5 PG (the qmake5 PG includes the qt5 PG automatically).

The qmake5 PG is more like the cmake Portgroup -- it sets everything up for a qt5 build using *.pro files, and then you should find this much much simpler than what has gone on here.

herbygillot:
OK thanks, didn't know about the `qmake5` portgroup, will take a look later.

kencu:
After trying it, I see why you can't just easily use the qmake5 PG... there is a cmake project embedded in the deps and it turns out to be not so simple as you might hope to build a cmake subproject with qmake.

herbygillot:
> After trying it, I see why you can't just easily use the qmake5 PG... there is a cmake project embedded in the deps and it turns out to be not so simple as you might hope to build a cmake subproject with qmake.

Yeah, there was a lot going on in that project.

kencu:
@herbygillot -- if you pull out the unnecessary ccache forcing, probably someone could see their way clear to committing this.

It's a messy port -- not 100% it deserves to be a port, to be honest -- and if it is committed you can expect lots of complaints about this one from above.

herbygillot:
OK @kencu, made a number of changes, primarily removing `ccache` and using the `cmake` portgroup. We'll see if it builds.

kencu:
yeah, if this version works it will be a lot more acceptable. Thanks.

PR #9605, opened 2021-01-03
portfolio-performance: new port

Janosch:
#### Description
New port for [Portfolio Performance](https://www.portfolio-performance.info/), a tool to calculate the performance of an investment portfolio.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.3.0.0.1.1607026830

###### Verification 
Have you

### DIFF
--- a/finance/portfolio-performance/Portfile
+++ b/finance/portfolio-performance/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+PortSystem              1.0
+PortGroup               github 1.0
+PortGroup               java 1.0
+
+github.setup            buchen portfolio 0.50.3
+name                    portfolio-performance
+categories              finance
+platforms               darwin
+license                 EPL-1
+maintainers             {@janosch mailbox.org:janosch1} openmaintainer
+description             A simple tool to calculate the overall performance \
+                        of an investment portfolio.
+long_description        An open source tool to calculate the overall performance \
+                        of an investment portfolio - across all accounts - using \
+                        True-Time Weighted Return or Internal Rate of Return.
+homepage                https://www.portfolio-performance.info/en/
+
+checksums               rmd160  2b6485f607493ffff233dedbc7f68191d442a10c \
+                        sha256  f4d8d348e28b4e8f337ddb282f0f7cbce92fbe4190bdfec5fcc4ca92a83833a4 \
+                        size    3238967
+
+supported_archs         x86_64
+java.version            11
+java.fallback           openjdk11
+depends_build-append    port:maven3
+use_configure           no
+
+# Remove auto-update feature and the corresponding dialog
+patchfiles              patch-disable-autoupdate-pref.diff \
+                        patch-remove-autoupdate-dialog.diff
+
+set maven_local_repository ${worksrcpath}/.m2/repository
+pre-build {
+    file mkdir ${maven_local_repository}
+}
+
+build.cmd               mvn3
+build.pre_args          clean verify -f portfolio-app -Ppackage-distro \
+                        -Dmaven.repo.local=${maven_local_repository} \
+                        -DskipTests
+
+destroot {
+    copy -- ${worksrcpath}/portfolio-product/target/products/name.abuchen.portfolio.product/macosx/cocoa/x86_64/PortfolioPerformance.app \
+    ${destroot}${applications_dir}
+}
--- a/finance/portfolio-performance/files/patch-disable-autoupdate-pref.diff
+++ b/finance/portfolio-performance/files/patch-disable-autoupdate-pref.diff
@@ -1,0 +1,11 @@
+--- name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/PreferencesInitializer.java.orig	2021-01-02 11:36:29.000000000 +0100
++++ name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/preferences/PreferencesInitializer.java	2021-01-24 14:13:40.000000000 +0100
+@@ -13,7 +13,7 @@
+     public void initializeDefaultPreferences()
+     {
+         IPreferenceStore store = PortfolioPlugin.getDefault().getPreferenceStore();
+-        store.setDefault(UIConstants.Preferences.AUTO_UPDATE, true);
++        store.setDefault(UIConstants.Preferences.AUTO_UPDATE, false);
+         store.setDefault(UIConstants.Preferences.UPDATE_SITE,
+                         Platform.ARCH_X86.equals(Platform.getOSArch())
+                                         ? "https://updates.portfolio-performance.info/portfolio-x86" //$NON-NLS-1$
--- a/finance/portfolio-performance/files/patch-remove-autoupdate-dialog.diff
+++ b/finance/portfolio-performance/files/patch-remove-autoupdate-dialog.diff
@@ -1,0 +1,18 @@
+--- name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/OpenPreferenceDialogHandler.java.orig	2021-01-02 11:36:29.000000000 +0100
++++ name.abuchen.portfolio.ui/src/name/abuchen/portfolio/ui/handlers/OpenPreferenceDialogHandler.java	2021-01-24 14:44:23.000000000 +0100
+@@ -27,7 +27,6 @@
+ import name.abuchen.portfolio.ui.preferences.ProxyPreferencePage;
+ import name.abuchen.portfolio.ui.preferences.QuandlPreferencePage;
+ import name.abuchen.portfolio.ui.preferences.ThemePreferencePage;
+-import name.abuchen.portfolio.ui.preferences.UpdatePreferencePage;
+ 
+ @SuppressWarnings("restriction")
+ public class OpenPreferenceDialogHandler
+@@ -64,7 +63,6 @@
+         pm.addTo("api", new PreferenceNode("divvydiary", new DivvyDiaryPreferencePage())); //$NON-NLS-1$ //$NON-NLS-2$
+ 
+         pm.addToRoot(new PreferenceNode("proxy", new ProxyPreferencePage())); //$NON-NLS-1$
+-        pm.addToRoot(new PreferenceNode("updates", new UpdatePreferencePage())); //$NON-NLS-1$
+ 
+         PreferenceDialog dialog = new PreferenceDialog(shell, pm)
+         {


finance/portfolio-performance/Portfile
```diff
@@ -0,0 +1,41 @@
...
+categories          finance
+platforms           darwin
+license             EPL-1
+maintainers         @janosch mailbox.org:janosch1 openmaintainer
+description         A simple tool to calculate the overall performance of an investment portfolio.
+long_description    An open source tool to calculate the overall performance of an investment portfolio \
+                    - across all accounts - using True-Time Weighted Return or Internal Rate of Return.
+homepage            https://www.portfolio-performance.info/en/
+
+depends_build       port:maven3 \
```
harens, line 16:
```suggestion
depends_build       bin:mvn3:maven3 \
```

Only the binary is required as part of the build process, so I think it should be safe to specify it as a file dependency.

See https://guide.macports.org/chunked/reference.dependencies.html for more details.

finance/portfolio-performance/Portfile
```diff
@@ -0,0 +1,41 @@
...
+                    bin:java:openjdk15
```
harens, line 17: I'm presuming that since this is a Java project, Java would be required at both build time and runtime. I'm also going to assume that more than just the Java binary is required.

I would recommend using the Java PortGroup to set this all up for you. If you do so, this should be changed to a `depends_build-append`.

finance/portfolio-performance/Portfile
```diff
@@ -0,0 +1,41 @@
...
+    file mkdir ${maven_local_repository}
+}
+
+build.cmd           mvn3
+build.pre_args      clean verify -f portfolio-app -Ppackage-distro \
+                    -Dmaven.repo.local=${maven_local_repository} \
+                    -DskipTests
+
+destroot            {
+                      copy -- ${worksrcpath}/portfolio-product/target/products/name.abuchen.portfolio.product/macosx/cocoa/x86_64/PortfolioPerformance.app \
```
harens, line 35: Please correct me on this if I'm wrong on this, but the folder location makes it appear that this tool is built for x86_64. If this is the case, `supported_archs x86_64` should be added.

harens (changes requested):
Thanks @Janosch for your contribution, as well as all your previous inputs to the MacPorts Project. They're greatly appreciated 👍.

Just a few small things from me. I've based my review on other maven ports such as [commons-text](https://github.com/macports/macports-ports/blob/749ad2787c30497020bdc4cca116eff9f09c800b/java/commons-text/Portfile) and [commons-lang3](https://github.com/macports/macports-ports/blob/749ad2787c30497020bdc4cca116eff9f09c800b/java/commons-lang3/Portfile), which you might find helpful to look through. Thanks again for submitting this port.

finance/portfolio-performance/Portfile
Janosch: AFAIK binary dependecies should be the exception, no? From the guide:

> But when satisfying a dependency with vendor-supplied software is preferred for special reasons, or when it is possible for more than one port to satisfy a dependency, then file dependencies may be used. 

Neither of those two mentioned cases apply in this situation, so I thought a port dependency makes more sense. No?

harens: Thanks @Janosch for your reply. I'd say that it's the second point:

> it is possible for more than one port to satisfy a dependency

There are many different maven ports that provide the required binary (e.g. [maven31](https://github.com/macports/macports-ports/blob/749ad2787c30497020bdc4cca116eff9f09c800b/java/maven31/Portfile), [maven32](https://github.com/macports/macports-ports/blob/749ad2787c30497020bdc4cca116eff9f09c800b/java/maven32/Portfile), and of course, [maven3](https://github.com/macports/macports-ports/blob/749ad2787c30497020bdc4cca116eff9f09c800b/java/maven3/Portfile)).

Janosch: Are you sure? Ony my machine it looks like this when I install maven32 and maven31 (and do the `port select --set maven`):
```
~/Dev/macports-ports ❯❯❯ ls -lah /opt/local/bin/mvn*
lrwxr-xr-x  1 root  wheel    20B 24 Jan 13:04 /opt/local/bin/mvn@ -> /opt/local/bin/mvn31
lrwxr-xr-x  1 root  admin    40B 24 Jan 13:04 /opt/local/bin/mvn31@ -> ../share/java/apache-maven-3.1.1/bin/mvn
lrwxr-xr-x  1 root  admin    29B 24 Jan 13:00 /opt/local/bin/mvn32@ -> ../share/java/maven32/bin/mvn
lrwxr-xr-x  1 root  wheel    25B 24 Jan 13:04 /opt/local/bin/mvnDebug@ -> /opt/local/bin/mvnDebug31
lrwxr-xr-x  1 root  admin    45B 24 Jan 13:04 /opt/local/bin/mvnDebug31@ -> ../share/java/apache-maven-3.1.1/bin/mvnDebug
lrwxr-xr-x  1 root  admin    34B 24 Jan 13:00 /opt/local/bin/mvnDebug32@ -> ../share/java/maven32/bin/mvnDebug
~/Dev/macports-ports ❯❯❯ mvn3
fish: Unknown command: mvn3
```
IMHO, only the port maven3 provides the binary `mvn3`. A different story would be the binary (or symlink)`mvn`. But in my case I need maven 3.x. so I cannot just use any mvn binary.

I can still change it if you insist but it feels like violating the macports guidelines.

finance/portfolio-performance/Portfile
Janosch: depends_build is definitly wrong for JDK, it should be depends_lib. I'll change that.

The java PG is currently broken for Big Sur. See https://trac.macports.org/ticket/61445. I would prefer not using it until this issue is resolved.

finance/portfolio-performance/Portfile
Janosch: Because of your comment I checked if it could be build for other archs. It doesent look like that other archs are supported from the upstream project. I will add your suggestion.

finance/portfolio-performance/Portfile
harens: I see. Thanks for testing that out, I didn't appreciate that although the `mvn` binary is provided, the `mvn3` binary isn't. Who would've known that? (You @Janosch, that's who 👍)

After doing some research, it seems that prior to Mavericks (10.9), maven3 came preinstalled. This might lead to the scenario "where the versions in macOS are often sufficient", but I don't know whether the `mvn3` binary is present or not.

Janosch:
With the latest commit I implemented changes suggested from @harens I could agree with. One open point is whether to use the java port group or not.

Also, I added patches to remove the auto-update feature of portfolio-performance.

finance/portfolio-performance/Portfile
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        buchen portfolio 0.50.2
+name                portfolio-performance
+categories          finance
+platforms           darwin
+license             EPL-1
+maintainers         @janosch mailbox.org:janosch1 openmaintainer
```
mf2k, line 10: You need to put braces around your GitHub handle and email address to link them. 

`maintainers         {@janosch mailbox.org:janosch1} openmaintainer`

mf2k (changes requested):

finance/portfolio-performance/Portfile
Janosch: Good catch. Done.

finance/portfolio-performance/Portfile
Janosch: @harens Are you okay with not using the java PG for the reason state above? If not, I will change the port to use the java PG. 

I'd like to get this PR ready to merge.

harens: Apologies for the delay. Completely understand that you want to get this merged as soon as possible.

Other people might have different opinions on this, but assuming everything works at runtime, I think what you suggest is fine for the time being.

Having said that, once the issue with the Java PG is fixed, I'd prefer if we revert this. What happens if the user has openjdk11 installed? The [relevant README section](https://github.com/buchen/portfolio#building-with-maven) says this if fine, but the portfile requires openjdk15. What if they the have Oracle JDK? This is why there's a PortGroup to simplify this process.

If we stick with your suggestion for now, I'd prefer openjdk11 simply because it's an LTS release, unless there's any particular reason for going with openjdk15.

On a quick side-note, until we can use the Java PG, do you think it's worth setting `JAVA_HOME` in the scenario where the user has it pointing to a different SDK to the one in the portfile?

Thanks again for writing this portfile. A lot of great progress has been made so far.

_(Just reading through the trac ticket, I noticed you sent a PR with your suggestion for a workaround https://github.com/macports/macports-ports/pull/9617. Thank you for your work on this 👍.)_

Janosch: @harens I changed my mind regarding the java PG. As I have maven3 as dependency (which in turn uses the java PG), I will run into the java PG bug anyway so it does not really make sense to bypass it in the port of portfolio-performance. So with the latest push I added the java PG and changed `depends_build` to `depends_build-append` as requested.

Janosch:
@harens @mf2k I resolved all changes requested so far and tested the app again. Please consider approval or let me know if you see some more issues.

mf2k (approved):

finance/portfolio-performance/Portfile
```diff
@@ -0,0 +1,44 @@
...
+
+build.cmd           mvn3
+build.pre_args      clean verify -f portfolio-app -Ppackage-distro \
+                    -Dmaven.repo.local=${maven_local_repository} \
+                    -DskipTests
+
+destroot            {
+                      copy -- ${worksrcpath}/portfolio-product/target/products/name.abuchen.portfolio.product/macosx/cocoa/x86_64/PortfolioPerformance.app \
+                      ${destroot}${applications_dir}
+                    }
```
mf2k, line 40: The spacing is odd here. The braces don't need to follow the mode line. Also, the copy command should only be indented out 4 spaces.

mf2k (changes requested):

finance/portfolio-performance/Portfile
Janosch: Thanks for the hint. It is fixed now.

finance/portfolio-performance/Portfile
```diff
@@ -0,0 +1,45 @@
...
+                        of an investment portfolio - across all accounts - using \
+                        True-Time Weighted Return or Internal Rate of Return.
+homepage                https://www.portfolio-performance.info/en/
+
+checksums               rmd160  2b6485f607493ffff233dedbc7f68191d442a10c \
+                        sha256  f4d8d348e28b4e8f337ddb282f0f7cbce92fbe4190bdfec5fcc4ca92a83833a4 \
+                        size    3238967
+
+supported_archs         x86_64
+java.version            11
```
harens, line 24:
```suggestion
java.version            11+
java.fallback           openjdk11
```

Can we allow later versions of Java? Reading throught the trac ticket, I see there are some issues with asterisk and plus notation, so please correct me if this wouldn't work with the current bug.

I would also suggest adding an LTS JDK port as a fallback.

Otherwise, lgtm. Thanks @Janosch.

harens (changes requested):

finance/portfolio-performance/Portfile
Janosch: @harens It does not compile out of the box with higher Java versions. With Java 15, for example, I had to increase the version of _tycho_, a dependency of _portfolio-performance_. As the github page of _portfolio-performance_ states Java 11 explicitly as a prerequisite, I'd like to stick with 11 (without "+").

I added the java.fallback as suggested by you.

Janosch: And yes, 11+ does not work on Big Sur at the moment. It will install the fallback even if you have JDK 15 installed.

harens (approved):
Lgtm. Thanks for your work on this @Janosch, and for your suggestions on how to workaround the Java PG issue 👍.

mf2k (approved):

harens:
Unless there's anything else, this PR has been waiting for a while and I think it looks good to merge.

kencu:
multiple approvals. Looks clean. Passes all checks. Not much more we can do before committing.

PR #9609, opened 2021-01-03
Poppler update 20.12.01 to 21.01.0

michaelld:
#### Description

update Poppler from 20.12.01 to 21.01.0 & rev-bump all dependencies as found via `port echo depends:poppler`.
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D5029f
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/aqua/qt5/Portfile
+++ b/aqua/qt5/Portfile
@@ -627,7 +627,7 @@ array set modules {
         {"Qt WebEngine"}
         "very large and relatively new"
         "variant overrides: "
-        "revision 0"
+        "revision 1"
         "License: "
     }
     qtwebglplugin {
--- a/aqua/qt513/Portfile
+++ b/aqua/qt513/Portfile
@@ -599,7 +599,7 @@ array set modules {
         {"Qt WebEngine"}
         "very large and relatively new"
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtwebglplugin {
--- a/devel/olena/Portfile
+++ b/devel/olena/Portfile
@@ -5,7 +5,7 @@ PortGroup               qt4   1.0
 
 name                    olena
 version                 2.1
-revision                0
+revision                1
 categories              devel framework graphics science
 platforms               darwin
 license                 GPL-2
--- a/editors/frescobaldi/Portfile
+++ b/editors/frescobaldi/Portfile
@@ -5,6 +5,7 @@ PortGroup           python 1.0
 PortGroup           github 1.0
 
 github.setup        wbsoft frescobaldi 3.1.2 v
+revision            1
 conflicts           ${name}-devel ${name}2
 categories          editors python
 maintainers         {gmail.com:davide.liessi @dliessi} openmaintainer
@@ -28,6 +29,7 @@ subport ${name}-devel {
     github.setup        wbsoft frescobaldi c2b5af3e1cdcde51f996407a27cebff4ad9e1d33
     conflicts           ${name} ${name}2
     version             20200719
+    revision            1
 
     github.tarball_from archive
 
@@ -38,7 +40,7 @@ subport ${name}-devel {
 
 subport ${name}2 {
     github.setup        wbsoft frescobaldi 2.20.0 v
-    revision            1
+    revision            2
     conflicts           ${name} ${name}-devel
 
     checksums           rmd160  27f121545cc7efe6a2b10fa69490a5714655f729 \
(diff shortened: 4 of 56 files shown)


michaelld:
Ping on this PR. Since the port is `openmaintainer` I'll merge this in 2 more days unless it is already merged or there is discussion otherwise (after rebasing to fix the conflict, of course). It's a pretty trivial update, but I do want to give the port owners plenty of time to handle this PR themselves.

Schamschula:
`littleutils` uses `pdftocairo`. Updates to poppler should not require a rebuild.

michaelld:
> `littleutils` uses `pdftocairo`. Updates to poppler should not require a rebuild.

According to `port info littleutils`, it has a library dependency on `poppler`, and hence if it truly links to the poppler ABI then it will require a rev-bump because the ABI changes names in this PR.
```
% port info littleutils
littleutils @1.2.3 (sysutils)
Variants:             python37, (+)python38, universal

Description:          The littleutils include a duplicate file finder (repeats), image optimizers
                      (opt-jpg, opt-png, opt-gif, recomp-jpg), file rename tools (lowercase, uppercase,
                      pren), archive recompressors (to-gzip, to-bzip, to-7zip, to-lzma, to-lzip, to-xz),
                      a tempfile utility (tempname), file property tools (filedate, filemode, filenode,
                      fileown, filesize, and lrealpath), and others.
Homepage:             https://sourceforge.net/projects/littleutils/

Extract Dependencies: xz
Library Dependencies: bash, bzip2, dash, file, gettext, ghostscript, gifsicle, gzip, img2pdf, imlib2,
                      lzip, lzma, lzop, p7zip, perl5, pngcrush, poppler, xz, zstd, python38
Platforms:            darwin
License:              GPL-3+ and ijg and permissive
Maintainers:          Email: mps@macports.org, GitHub: Schamschula
                      Policy: openmaintainer
```

michaelld:
As noted in the commit message, I look the output of `port echo depends:poppler` and did a rev-bump on those ports. This method might bring in more ports than required. If there's an issue with ports that `port` says depend on poppler, then let's fix those ports or `port` itself.

Schamschula:
The configure script checks for pdftocairo:

> checking for pdftocairo... yes

however, the pdf related utilities were never installed. I've now fixed that by adding the `install-extras` target.

I then checked if these files are binaries. No, they are bash scripts calling the `pdftocairo` binary, among others.

As the configure script checks for the presence of these binaries, I don't think moving these dependencies to `depends_run` will work.

dliessi:
As far as I understand, the argument of `depends:` is a regular expression that is matched against the dependencies, so `depends:poppler` matches ports depending on ports whose name contains `poppler` instead of only those depending on the port `poppler`.

The ports `frescobaldi` and `py-qpageview` depend on `poppler` only via `py-poppler-qt4` and `py-poppler-qt5`, so there is no need to revbump them.

I agree with the revbump of my other ports, i.e. `DiffPDF`, `py-poppler-qt4` and `py-poppler-qt5`.

NicosPavlov:
No visible problem with pdf2djvu, so the upgrade is all fine with me.

pietvo:
`diff-pdf` compiles and runs fine with the new `poppler`.
Actually, it just runs fine even with no recompilation, because it links with `libpoppler-glib.8.19.0.dylib`, which is still part of the new `poppler` port. Although the file itself is changed, the file name is still the same, but it links with the new `libpoppler.106.0.0.dylib` instead of `libpoppler.105.0.0.dylib`.
So the change is transparent to `diff-pdf`.

medoc92:
Recoll only uses pdftotext, so I guess that nothing needs to be done.

michaelld:
I've heard no direct issues with merging this once it's rebased. I will try to take into account the comments about ports that actually don't require the rev-bump, but it's likely I'll mess one or more of those up ... so thanks in advance for your forgiveness & luckily it's not that big of a deal :)

kimuraw:
@michaelld, I deleted rb-gnome and rb-poppler from ports at 820ff9c008bea9f7a2883806d26ac5d5df4a6cd0.
these ports were broken.

We need not revbump for rb-gnome and rb-poppler.

kencu:
many conflicts now. I think @dbevans is getting just too busy to manage these PRs, so either we consider just not doing any more PRs involving Dave's ports, or we actually do start committing them after the 3 day timeout (although I'm not sure we have the stomach for that either).

So -- no more PRs for ports that Dave maintains? Hate to see everyone waste their time.

michaelld:
I need to rebase this one. Given that lack of response, once this is rebased & CI passes, we can merge quickly. Beyond this PR - which is partly meant to fix rev-bumps that didn't happen last time -- I think we should in general leave Dave's ports alone per his request for doing updates.

michaelld:
Closing as this is horribly outdated & needs to be rebased. Will try again with the most recent release when I get a chance.

PR #9612, opened 2021-01-03
timewarrior: new port

herbygillot:
#### Description

New port for **[TimeWarrior](https://timewarrior.net)**, a time tracking utility from the same [author](https://gothenburgbitfactory.org) as [TaskWarrior](https://taskwarrior.org).

###### Tested on

macOS 10.15.7 19H114
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/office/timewarrior/Portfile
+++ b/office/timewarrior/Portfile
@@ -1,0 +1,42 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           cmake 1.1
+
+github.setup        GothenburgBitFactory timewarrior 1.4.2 v
+revision            0
+
+homepage            https://timewarrior.net
+
+description         Timewarrior is Free and Open Source Software that tracks \
+                    time from the command line.
+
+long_description    Timewarrior is a time tracking utility that offers simple \
+                    stopwatch features as well as sophisticated \
+                    calendar-based backfill, along with flexible reporting.
+
+categories          office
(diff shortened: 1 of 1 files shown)


Tatsh:
I would suggest adding `github.livecheck.regex  {([0-9.]+)}`. I had an issue with this project where `1.4.2-pre` was thought to be newer than `1.4.2`. So we have to ignore the `-pre` releases. I filed a ticket about this but changing the livecheck `vercmp` function isn't really desired.

How is this working without libshared? The release tarball has it included?

[My Portfile](https://github.com/Tatsh/ports/blob/master/office/timewarrior/Portfile)

herbygillot:
> I would suggest adding `github.livecheck.regex {([0-9.]+)}`. I had an issue with this project where `1.4.2-pre` was thought to be newer than `1.4.2`. So we have to ignore the `-pre` releases. I filed a ticket about this but changing the livecheck `vercmp` function isn't really desired.

Very good catch - I've updated and added the regex, thanks.

> How is this working without libshared? The release tarball has it included?

This is using the `release` tarball, which has the `libshared` source included.

office/timewarrior/Portfile
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           cmake 1.0
```
ryandesign, line 5: Usually we want to use the latest version of a portgroup, which here is cmake 1.1. If you change that, then the line `cmake.out_of_source yes` can be deleted, because that is the default of cmake 1.1.

herbygillot: Good point; fixed.

PR #9630, opened 2021-01-05
Gildas: update to 202101a

bardeau:
Can someone please merge this PR? Thanks.


ryandesign:
Thanks.

You don't need to ask us to merge each PR. Hopefully that's what we do even without being asked. It would be more helpful if you would fill out at least part of the PR template, such as what version of macOS and Xcode you tested the build on, just in case there is some problem with our CI system that makes it unable to build, which is what happened here.

PR #9631, opened 2021-01-05
fpc: Update to fix build on 11.x

kamischi:
#### Description

On 11.x, the path /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk needs to be added to prevent the error: ld: library not found for -lc

###### Type(s)

- [x] bugfix
- [x] enhancement

###### Tested on

macOS 11.1
Xcode 12.3

###### Verification 
Have you

- [x] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? https://trac.macports.org/ticket/61621

### DIFF
--- a/lang/fpc/Portfile
+++ b/lang/fpc/Portfile
@@ -7,7 +7,7 @@ version             3.2.0
 categories          lang
 platforms           darwin
 license             GPL-2 LGPL-2
-maintainers         nomaintainer
+maintainers         {@kamischi web.de:karl-michael.schindler} openmaintainer
 description         Free Pascal, an open source Pascal and Object Pascal compiler.
 long_description    Free Pascal is a 32, 64 and 16 bit professional Pascal compiler. \
                     It can target many processor architectures: Intel x86 (including 8086), \
@@ -145,6 +145,7 @@ if {${subport} eq "${name}"} {
     worksrcdir          ${name}build-${version}/fpcsrc
     build.env           PP=${workpath}/${b} \
                         PREFIX=${destroot}${fpcbasepath}
+    build.args          OPT="-ap -v0 -XR/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk"
     build.target        all
 
     destroot.env        {*}${build.env}
@@ -152,8 +153,10 @@ if {${subport} eq "${name}"} {
     # build the compiler utilities msgdif and msg2inc
     post-build {
         system -W ${worksrcpath}/compiler/utils \
-               "../ppcx64 -WM10.9 -Fu../../rtl/units/${build_arch}-darwin msgdif.pp && \
-                ../ppcx64 -WM10.9 -Fu../../rtl/units/${build_arch}-darwin msg2inc.pp"
+               "../ppcx64 -WM10.9 -Fu../../rtl/units/${build_arch}-darwin -ap -v0 \
+                  -XR/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk msgdif.pp && \
+                ../ppcx64 -WM10.9 -Fu../../rtl/units/${build_arch}-darwin -ap -v0 \
+                  -XR/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk msg2inc.pp"
     }
 
     post-destroot {


kencu:
how does fpc search for the SDK it wants to use on MacOS?

The currently accepted method is to look for and use SDKROOT, which MacPorts does set.

Hard-coding in this path might be the only option, but it is a bit blunt and fragile, and removes some flexibility that we'd prefer to maintain, if we can...

kamischi:
Do you think it is worth to have a try using SDKROOT?

kamischi:
Any opinion about me claiming maintainership, while also setting openmaintainer?

kencu:
> As much as i know about fpc, fpc itself does not really need the SDK as it passes the linking stage to ld. Therefore, it becomes a question of the default paths of ld and what fpc tells ld to use in addition.
> Do you think it is worth to have a try using SDKROOT? It looks that -XR${configure.sdkroot} works, but no luck with ${SDKROOT}.

compilers should first look to see if there is a "isysroot" set on the build line, configure args, cxxflags, etc, etc, something like that, specified for the specific build. If there is nothing set there, they are meant to next look at SDKROOT.

SDKROOT is an environment variable that MacPorts sets during most builds (when there is no SDK in "/"). Compilers are hopefully supposed to respect that, so it would be written into the FPC code to look there for link libraries. 

If nothing is set with respect to SDKROOT, then compilers might have a fallback they can use. Sorting out exactly where that should be and how that should be set requires some thought. No doubt upstream has already been thinking about this, as 11.x  has been out for six months and they are pretty "on it" in that group.

So a bit of digging would be in order to see what upstream is doing, if anything.

kencu:
so that fix is a longer term fix for fpc I guess. For today -- what systems need to be directed to look for their link libraries in this location? Is this only something shows up on BigSur and up, or has it been an issue on 10.14 and 10.15 as well? I will admit not noticing this problem before, and I have run lazarus on Catalina, so so it's only on BigSur up I guess.

So we should ( I guess ) only force the SDK on BigSur +. What do you think?

kamischi:
Yes. Correct. It definitely affects only BigSur. It is caused by some change of the linker settings in BigSur. But the extra option does not hurt on older systems.  -XR<x>  means   Prepend <x> to all linker search paths. So, if it is not present, the old paths are searched.

kamischi:
hm. I cannot obtain any log from the failed build on 11.0. what should I do? Simply revert the change to configure.sdkroot?

kencu:
> Yes. Correct. It definitely affects only BigSur. It is caused by some change of the linker settings in BigSur. But the extra option does not hurt on older systems. -XR means Prepend to all linker search paths. So, if it is not present, the old paths are searched.

On BigSur, there is no path "/usr/lib" for libraries to live in. Instead, there is a different mechanism, more complicated, to be used for opening system libraries. Software looking for system libraries is meant to "dlopen" the path and some kind of macOS magic happens and the library is found in a cache somewhere, even though it is not there. It's a security thing, I understand. So that is probably why fpc can't find the link libraries on BigSur -- no doubt it is looking in "/usr/lib" for them, and there are none there.

I see now what you mean about the linker paths. Prepending ${configure.sdkroot} onto the library search path makes sense -- however there is a problem with that plan. Every time the SDK updates (frequently now) that path ${configure.sdkroot} changes, so everything breaks. So your first plan, to use the hardcoded PATH to MacOSX.sdk is probably, in the end, the best. 

The stuff I was talking about using -isysroot and SDKROOT, is something upstream should implement at some point, but for purposes of getting the build working on BigSur right now, your initial plan will  be fine.

FYI -- there is a problem with prepending the path to MacOSX.sdk onto the library search path. If that is done, the libraries will be preferentially found in that SDK, and then there will be functions found in that SDK that are probably NOT supported on the current system you are building on, because the SDK is usually for the NEXT os version rather than the CURRENT os version.

kamischi:
Oh Oh. Reverting the use of config.sdkroot does not fix building on 11.0. The problem is, that without a log, I can only guess, what is wrong and I am out of ideas. It builds here.

kamischi:
Mistery. Second build succeeds.

kencu:
The CI system is undergoing renovations. If you look at the macports/mpbb repo you'll see heavy commits there recently as Jost tries to get it sorted out.

kencu:
OK. Let's try this. Using the MacOSX.sdk preferentially over "/" even if "/" exists might have some problems over time with library functions being found in the MacOSX SDK that that system doesn't actually support, but we'll have to learn to sort this out anyway as that is the way things are going in the macOS world.

kamischi:
thanks. I am curious about the arm64 build.

ryandesign:
This PR only fixes a build failure on macOS 11? It doesn't in any way change what the port installs or how the software works on macOS 10.x? Because if it did, you would have needed to increase the revision.

There is no arm64 build because the port states that it is only compatible with x86_64.

kamischi:
1) To the best of my knowledge, it does not change prior OSes. Therefore, I did not increase the revision.

2) With respect to a native arm build, the question is, whether the current x86_64 based bootstrap compiler works through rosetta. I would expect it. Creating the native arm build would then be fairly straight forward. This path would save from creating a separate arm bootstrap compiler. Unfortunately, i have no access to a M1 Mac for a test.

kencu:
It does slightly change installed files -- it will make newly built versions of fpc look for the MacOSX.sdk ahead of the libraries in "/" on all systems that have a MacOSX.sdk in the noted path. This is minor. We can revbump if you like.

I have access to an arm mac and will see if fpc can build arm-native. If not, we should really come up with some simple way of flagging the x86_64 build to be used on arm64 macs under Rosetta2. MacPorts has made no strides in this direction. Homebrew has asked all users to install homebrew x86_64 on Apple Silicone so that it all runs under Rosetta2 for now, and install another copy of homebrew as arm64 native in parallel. Dunno what we want to do.

Ryan could not build fpc on 10.13 for some reason. No idea why. He opened a ticket about it; perhaps you can sort him out.

kamischi:
Maybe, you can use ppcx64 as bootstrap compiler and set CPU_TARGET=aarch64. The bootstrap compiler is passed through PP=AnyPathToBootstrapCompiler. If ppcx64 fails, fpc-cross creates ppca64. That is a second option.

I'll have a look at Ryan's issue.

kencu:
It looks like upstream has arm64 support for Apple Silicon for fpc already all sorted out:

<https://wiki.freepascal.org/macOS_Big_Sur_changes_for_developers#ARM64.2FAArch64.2FApple_Silicon_Support>

kamischi:
Yep. It takes an extra step to build an arm64 bootstrap compiler first and then use this to build the actual one. Maybe the first step can be saved. Also the integration into one package description needs to be checked. Another point is that the wiki page relates to trunk and not the 3.2 release. As much as I could test the compiler ppa64 is built, but I am not sure about the rtl. If so, we would have to wait for the next release, iow 3.2.2 or 3.4.0. The current state of 3.2-fixes is another option.

kamischi:
I have a first proposal for FreePascal on arm64, but how could I test it? I have seen work in progress on the list of pull requests, but is this really the way to go?

ryandesign:
> It does slightly change installed files -- it will make newly built versions of fpc look for the MacOSX.sdk ahead of the libraries in "/" on all systems that have a MacOSX.sdk in the noted path. This is minor. We can revbump if you like.

Absolutely yes. If you change the installed files, you increase the revision, every time. Reproducible builds.

> I have access to an arm mac and will see if fpc can build arm-native. If not, we should really come up with some simple way of flagging the x86_64 build to be used on arm64 macs under Rosetta2. MacPorts has made no strides in this direction.

It should already be working with MacPorts 2.6.4 today. On an arm64 Mac, any port that has x86_64 in supported_archs but not arm64 will install as x86_64, after installing its dependencies universal. See https://trac.macports.org/ticket/61430.

> I have a first proposal for FreePascal on arm64, but how could I test it? I have seen work in progress on the list of pull requests, but is this really the way to go?

Sure, submit a WIP PR and someone with access to arm64 can test it.

kamischi:
> > It does slightly change installed files -- it will make newly built versions of fpc look for the MacOSX.sdk ahead of the libraries in "/" on all systems that have a MacOSX.sdk in the noted path. This is minor. We can revbump if you like.
> 
> Absolutely yes. If you change the installed files, you increase the revision, every time. Reproducible builds.

How about setting this option only for 11 and up with a switch on os.major?

ryandesign:
If the changes are only of benefit to macOS 11, then you could do that, yes. In any case, since that would again change the contents of files on some systems, the revision should be increased.

kencu:
The changes actually only affect systems before macOS 11 that have a MacOSX.sdk in the /CLTs path, and even then, it is fairly debatable whether they really will be affected by Karl tells me that MacOSX.sdk will be used preferentially rather than as a fallback if no libs in "/", so assuming that is in fact the case, it might possibly theoretically change something.

Therefore I revbumped them all, just to avoid any semblance of concern.

PR #9647, opened 2021-01-06
radare2: Update to 5.0.0

saagarjha:
#### Description

I'm running through `port livecheck installed` and trying to update things that have an open maintainer or no maintainer.

###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D5029f
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/devel/radare2/Portfile
+++ b/devel/radare2/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           conflicts_build 1.0
 PortGroup           github 1.0
 
-github.setup        radareorg radare2 4.5.1
+github.setup        radareorg radare2 5.0.0
 revision            0
 categories          devel
 platforms           darwin
@@ -21,9 +21,9 @@ depends_lib         port:capstone \
                     port:libzip
 conflicts_build     ${name}
 
-checksums           rmd160  fc31598daa28fe7de7b7af497eda5da7a3a6fbfd \
-                    sha256  46e96ae2a4f3f9c99aeb78317be7df0acc77ccfc30725b03ac728de7a02e8e08 \
-                    size    8265525
+checksums           rmd160  bbb5700035387bbf3d74888352a8383969ae76e7 \
+                    sha256  e6600fff27001874d7885a52e538c505ea4e2b49e3193b72ae3dc0f006a393c5 \
+                    size    7898701
 
 configure.args-append \
                     --with-syscapstone \


i0ntempest:
Thanks, the update itself looks good but it breaks the port cutter-r2, I would hold this until it get updated for radare2 5.0

saagarjha:
Hmm, I noticed that Cutter seems to be using rizin now instead of radare2: https://github.com/rizinorg/cutter. Should we create a new rizin port it can depend on?

i0ntempest:
Yeah I didn't notice that till just now. Would definitely do that but rizin doesn't have a stable release yet.

g5pw:
It looks like cutter is bundling its own rizin source...? Also, there's no mentions of `radare` in the build docs, so maybe it's actually not a dependency anymore?

i0ntempest:
Great, submodules again (bruh). Also yes, rizin has replaced radare as the backend.

g5pw:
So, it looks like Cutter 1.12.0 is still linking with radare2?! I wonder if there's a configure switch somewhere...?

g5pw:
Right, so I guess this is locked until we have a Clutter update?

i0ntempest:
Yes, unless we rename radare2 4.x and submit this as a new port.

saagarjha:
I think the best thing to do here would be to ship radare2 as-is and create a new port for rizin, which cutter will depend on? (And maybe a cutter-radare, if there is interest?)

i0ntempest:
Thats exactly what I'd do but rizin has no stable release as of now. The current Cutter port is already `cutter-r2` cuz `cutter` is taken, so the new Cutter when it got released will probably be submitted as a new port `cutter-rizin` or something and the old `cutter-r2` would be replaced with a obsolete notice.

saagarjha:
I'm curious what other package managers are doing…

g5pw:
So, does it make sense if we keep this PR open or should we close it, wait for a new Cutter release, deprecate `cutter-r2` in favor of `cutter-rizin` and update `radare2`?

IMO it's not really worth keeping a separate `radare2` port tracking the old version just for compatibility reasons, since Cutter is moving away from it.

i0ntempest:
Imo either is fine, I can just reopen and merge this after new cutter and rizin stuff is sorted.

g5pw:
Ok, closing (for now).

i0ntempest:
They now have a cutter fork that uses the latest radare2: https://github.com/radareorg/r2cutter/
Do you guys think it's better to keep the name "cutter-r2" and switch the source, or deprecate "cutter-r2" and make a new "r2cutter" port?
After these are done we can then add rizin and the new cutter in.

PR #9648, opened 2021-01-07
boxes: update to 1.3

jungx098:
* update to version 1.3
* update build.args
* add patches

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/textproc/boxes/Portfile
+++ b/textproc/boxes/Portfile
@@ -3,7 +3,7 @@
 PortSystem              1.0
 PortGroup               github 1.0
 
-github.setup            ascii-boxes boxes 1.2 v
+github.setup            ascii-boxes boxes 1.3 v
 
 categories              textproc
 license                 GPL-2+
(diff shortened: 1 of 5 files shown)


ryandesign:
I see that you added a dependency on gsed and patched the build to use gsed. What problem did you encounter when using BSD sed—which of the regular expressions was not BSD sed compatible? Did you report the problem to the developers of boxes?

jungx098:
The test script heavily uses **gnu** `sed` that is NOT compatible with the **BSD** one. In this patch I replaced `sed` with `gsed` in the test script.

This issue only affects the test; it does NOT make any trouble in building. If we skipped `port test`  we would not need this dependency.

I didn't yet report this issue to the boxes developers, but I am thinking of that including some other building issues.

ryandesign:
> This issue only affects the test; it does NOT make any trouble in building. If we skipped `port test` we would not need this dependency.

In that case, you could've used `depends_test port:gsed`.

PR #9653, opened 2021-01-07
go-mmv: new port

herbygillot:
#### Description

New port for [mmv](https://github.com/itchyny/mmv) (the Go version, since there's already a historic `mmv` port)

###### Tested on

macOS 10.15.7 19H114
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/sysutils/go-mmv/Portfile
+++ b/sysutils/go-mmv/Portfile
@@ -1,0 +1,60 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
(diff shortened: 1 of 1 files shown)


sysutils/go-mmv/Portfile
```diff
@@ -0,0 +1,60 @@
...
+description         rename multiple files with editor
+
+long_description    Rename multiple files using your \$EDITOR. The command \
+                    name is named after multi-mv.
+
+conflicts           mmv
+
+categories          sysutils
+license             MIT
+platforms           darwin macosx
```
harens, line 19:
```suggestion
platforms           darwin
```

I hadn't seen `macosx` in a portfile before, so I looked it up in the docs (thanks for introducing it to me).

From https://guide.macports.org/#reference.keywords, `darwin` is "equivalent to specifying both macosx and puredarwin". So my understanding is that either one should be used, but not both.

herbygillot: Good catch @harens - fixed

PR #9654, opened 2021-01-08
mame: update for 0.227; update launch script to improve user experience

mascguy:
#### Description

Mame Changes:
* Update for Mame 0.227 release; includes seamless fallback to 0.226, for macOS releases prior to 10.13.
* Updated launch script to improve user experience.

Fixes: [61976 - mame: update to 0.227](https://trac.macports.org/ticket/61976)
Fixes: [61743 - mame: improve user experience when running](https://trac.macports.org/ticket/61743)

###### Type(s)
- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.6.8 10K549
Xcode 3.2.2 10M2148

macOS 10.8.5 12F2560
Xcode 5.1.1 5B1008

macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

###### Verification
Have you

### DIFF
--- a/emulators/mame/Portfile
+++ b/emulators/mame/Portfile
@@ -14,52 +14,83 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
-github.setup        mamedev mame 0226 mame
-revision            2
-
-version             [string index ${github.version} 0].[string range ${github.version} 1 end]
+name                mame
 categories          emulators
-maintainers         {@mascguy} \
-                    openmaintainer
 platforms           darwin
 license             GPL-2+
-homepage            https://www.mamedev.org
+
+maintainers         {@mascguy} \
+                    openmaintainer
 
 description         Multiple Arcade Machine Emulator
-long_description    The purpose of MAME is to preserve decades of software history. As \
-    electronic technology continues to rush forward, MAME prevents this important \
-    "vintage" software from being lost and forgotten. This is achieved by documenting \
-    the hardware and how it functions. The source code to MAME serves as this \
+long_description    \
+    The purpose of MAME is to preserve decades of software history. As electronic \
+    technology continues to rush forward, MAME prevents this important "vintage" \
+    software from being lost and forgotten. This is achieved by documenting the \
+    hardware and how it functions. The source code to MAME serves as this \
     documentation.
-
-github.tarball_from archive
+homepage            https://www.mamedev.org
 
 #------------------------------------------------------------------------------
-# Patches and Checksums
+# Versions, Patches, and Checksums
 #------------------------------------------------------------------------------
 
-# Patch 'language-portuguese_brazil' applies to release 0.226; fixes
-# errors during translation build phase.
-# Fix submitted, and merged, by Mame maintainers; will be part of Mame
-# releases from 0.227 onward. Tracked by Mame issue:
-# https://github.com/mamedev/mame/issues/7510
-#
-# Patch 'src-posixfile-build-error' applies to release 0.226; fixes
-# compilation errors for MacOS 10.8 and Xcode 5.x.
-# Fix submitted, and merged, by Mame maintainers; will be part of Mame
-# releases from 0.227 onward. Tracked by Mame issue:
-# https://github.com/mamedev/mame/issues/7536
-#
-# Patch 'scripts-genie-compile-warnings' applies to all releases; greatly
-# reduces compilation warnings.
+# Mame versions > 0.226 require C++ 2017, and specifically aligned allocation
+# support. That's only supported by MacOS system libc++ from 10.13 onward.
+# However, once MacPorts provides a custom libc++ for all ports, Mame will
+# support older MacOS releases.
+if {${os.major} >= 17} {
+    set g_mame_latest yes
+} else {
+    set g_mame_latest no
+}
 
-patchfiles          mame-patch-0226-language-portuguese_brazil.diff \
-                    mame-patch-0226-src-posixfile-build-error.diff \
-                    mame-patch-0226-scripts-genie-compile-warnings.diff
+if {${g_mame_latest}} {
+    set g_mame_release \
+                    "0227"
 
-checksums           rmd160  b9e4ae321b7673790d374c63bbe966d1502d6738 \
+    revision        0
+
+    checksums       \
+                    rmd160  862df6a449329fe5b16e6d46a030fe157c3a8eb3 \
+                    sha256  95dbce00a4f05a35f66ef966fe9efad1e4e78ce62e0eba3f7031dfa6737829a5 \
+                    size    195226157
+} else {
+    set g_mame_release \
+                    "0226"
+
+    revision        3
+
+    # Patch 'language-portuguese_brazil' applies to release 0.226; fixes
+    # errors during translation build phase.
+    # Merged into Mame releases from 0.227 onward. Tracked by issue:
+    # https://github.com/mamedev/mame/issues/7510
+    #
+    # Patch 'src-posixfile-build-error' applies to release 0.226; fixes
+    # compilation errors for MacOS 10.8 and Xcode 5.x.
+    # Merged into Mame releases from 0.227 onward. Tracked by issue:
+    # https://github.com/mamedev/mame/issues/7536
+    patchfiles      \
+                    mame-patch-0226-language-portuguese_brazil.diff \
+                    mame-patch-0226-src-posixfile-build-error.diff
+
+    checksums       \
+                    rmd160  b9e4ae321b7673790d374c63bbe966d1502d6738 \
                     sha256  7c4c9ec232ba988e65fd29665c9b8e40b5ac3aa9f561eeb107cebbf08ba94baf \
                     size    196379874
+}
+
+# Note: For user-presentable version string, we add a period between the first
+# two characters of release. So '0227', for example, becomes '0.227'.
+set g_mame_version_string \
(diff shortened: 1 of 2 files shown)


mf2k:
In the commit message, the correct format to actually close the tickets is as follows:

```
Fixes: https://trac.macports.org/ticket/61976
Fixes: https://trac.macports.org/ticket/61743
```

mascguy:
> In the commit message, the correct format to actually close the tickets is as follows:
> 
> ```
> Fixes: https://trac.macports.org/ticket/61976
> Fixes: https://trac.macports.org/ticket/61743
> ```

Updated commit message. Thanks for the clarification Frank!

emulators/mame/Portfile
```diff
@@ -14,52 +14,69 @@ PortSystem          1.0
...
+                    mame-patch-0226-language-portuguese_brazil.diff \
+                    mame-patch-0226-src-posixfile-build-error.diff
+
+    checksums       \
+                    rmd160  b9e4ae321b7673790d374c63bbe966d1502d6738 \
                     sha256  7c4c9ec232ba988e65fd29665c9b8e40b5ac3aa9f561eeb107cebbf08ba94baf \
                     size    196379874
+}
+
+subport ${name}-latest {
```
mf2k, line 64: In Macports, we call such a port that tracks a newer version than the latest `-devel`. So let's call this one `mame-devel`.

emulators/mame/Portfile
```diff
@@ -14,52 +14,69 @@ PortSystem          1.0
...
+                    "0227"
+
+    checksums       \
+                    rmd160  862df6a449329fe5b16e6d46a030fe157c3a8eb3 \
+                    sha256  95dbce00a4f05a35f66ef966fe9efad1e4e78ce62e0eba3f7031dfa6737829a5 \
+                    size    195226157
+}
+
+github.setup        mamedev mame ${g_mame_release} mame
+version             [string index ${github.version} 0].[string range ${github.version} 1 end]
```
mf2k, line 78: I'm not clear what this will set the version to. But we want it to be monotonically increasing. Just set the version to be the date of the commit in year-month-day format. For example:

`version   2021-01-08`

mf2k (changes requested):

emulators/mame/Portfile
mascguy: Well, the `-latest` suffix was used, as that's what it truly is: The latest version. It's not a development version, but rather, the latest released version.

This is necessary, as Mame 0.227 now depends on C++ 2017 features, the most notable being aligned allocations. And the latter isn't supported by the system libc++ prior to macOS 10.13. (Though when we support a replacement libc++ in MacPorts - hopefully sooner, rather than later - that won't be a problem any longer.)

@kencu Ken and I have discussed this at length in ticket [61976 - mame: update to 0.227](https://trac.macports.org/ticket/61976).

I'm certainly open to a different naming convention, but should a release version use a `-devel` suffix...?

emulators/mame/Portfile
mascguy: To avoid maintaining two version strings, we take the one used for artifacts - such as `0227` - and simply insert a period between the first two characters. With the result being `0.227`, which is the true user-presentable version number. (Mame releases still have a major version of zero, after all these years. Seems odd, but hey... to each their own...)

We now have a formal variable with that string - `g_mame_release` - so we can at least reference that instead, rather than `github.version`. I'll also add a brief comment, reiterating the above explanation.

How does that sound?

mascguy: Changes pushed.

emulators/mame/Portfile
mf2k: In that case, a release version would not have a suffix and should simply be called `mame`.

mascguy: > In that case, the latest release version would not have a suffix and should simply be called `mame`. Older release versions are named with the version number.

That was one of the ideas I originally considered. The problem is, users who already have Mame installed - v0.226 - will then be prompted to upgrade the port. But since the new release (0.227) requires macOS 10.13+, while 0.226 builds and runs on macOS releases from 10.6 through 10.15, the port will be flagged as outdated - and an update attempted every time those users run a `port upgrade outdated` - but the port can only be upgraded for users running on macOS 10.13+. And the upgrade will fail with an error message, every time they run it.

That's going to become very annoying, very quickly, for all of our users on older macOS releases. And asking them to workaround it by doing a `port upgrade outdated and not mame` each time, seems less-than-ideal.

I suppose we could add logic to the portfile, and prompt the affected users to switch to `mame-0226`. But I'd really prefer not to go down that path.

In short, the `mame-latest` subport route was chosen, to provide a user-friendly upgrade path... and ensure users of 0.226 [on legacy macOS releases] don't have to deal with any pain or annoyance. All whilst still providing a pathway for users of macOS 10.13+, to use newer Mame releases.

Does that make sense?

mascguy: Oh, I'm aware of the portfile techniques that we could use, in terms of port replacement/renaming, etc. But again, I'm concerned about the user experience...

mascguy: Hmmm. Pondering this a bit more, we could simply eliminate the subport, and automatically set the version based on the macOS release the user is running on. So folks on macOS 10.13+ would see 0.227, while folks on older macOS releases would see 0.226.

This would certainly provide a seamless user experience. Though it's also nice to show that we support newer releases somewhere, even if it's only in a subport.

Thoughts?

mascguy: @kencu Adding Ken to the mix, since we discussed this a bit in the ticket...

kencu:
Looks like you're doing a fine job of trying to make this work nicely for everyone.

When an older system can't run the latest version but does run a version of the same software, we often do just silently provide an older version with the same port name but different version to older systems. It would be more clear to put a short note that the version is being held back in the description or port notes, I agree, so users know what they have received.

I appreciate all your attention to detail on this; you've learned a lot about MacPorts in a relatively short period of time ! Took me much longer :<

mascguy:
> I appreciate all your attention to detail on this; you've learned a lot about MacPorts in a relatively short period of time ! Took me much longer :<

To be fair, support and feedback from folks like yourself and Frank, have eased the learning curve. Plus there are a mountain of ports to reference and learn from - along with the MacPorts base code - which have all helped.

I'm still learning, and occasionally stubbing my toes along the way. But getting there!

> When an older system can't run the latest version but does run a version of the same software, we often do just silently provide an older version with the same port name but different version to older systems. It would be more clear to put a short note that the version is being held back in the description or port notes, I agree, so users know what they have received.

Agreed, I was thinking the same thing. I'll add info to both the port descriptions, as well as the install notes.

mascguy:
Text explaining both what macOS releases are supported, as well as mentioning alterantive port, added.

emulators/mame/Portfile
kencu: So I'm afraid I'm with Frank here.

We should just have one port named "mame". 

On current systems, the version can be the newest available. On older systems, the version can be the last one that could build.

But it is overly confusing to have an extra port named "mame-latest" that people need to know to install when they want to install mame.

You are right that there is an ld64-latest port. To be fair there, though, that is not a user-installed port in 99.99% of cases. It is automatically installed by the "ld64" metaport, and then symlinked in. So that is a bit of a special case.

What we're trying to do is make this as simple as possible for users. Somebody two years from now is going to want to install "mame" and will get either an old version, or confused. So it's best to go with the MacPorts' standard ways of doing things for consistency.

mascguy: Frank, I edited my prior comments down to the bare minimum, to ease the discussion.

Also, per Ken's suggestion, I added text to the description and notes for both ports. This provides the user with an explanation regarding what macOS release is supported for each, as well as mentioning the opposite port. In short, it helps to eliminate any confusion for the user.

Let me know if you're open to the proposed naming, particularly with that additional text added.

mascguy: Given that a single port provides a seamless user experience, I'm good with that. Thanks guys!

I'll work on the changes, and let you folks know when they're ready.

mascguy:
Changes pushed, commit message revised, and PR title/description updated.

emulators/mame/Portfile
mascguy: Changes complete.

mascguy:
I've rebuilt and tested in various macOS versions, so we're good-to-go.

kencu:
it looks like this has built on the CI system and passed the lint tests.

Just so you are aware, it has become the largest and most obfuscated Portfile in all of MacPorts in the past month. If you ever leave, we'll probably have to erase this and start over :>

mascguy:
> it looks like this has built on the CI system and passed the lint tests.

Thanks for merging this Ken!

> Just so you are aware, it has become the largest and most obfuscated Portfile in all of MacPorts in the past month. If you ever leave, we'll probably have to erase this and start over :>

Yeah, it would be a lot simpler if the project used cmake. (And at some point, I may even contribute that to upstream.) But in the interim - with all of the comments explaining what's happening, along with plenty of debug logging - I'm confident folks wouldn't need to toss it out.

You guys needn't worry though, I'm not planning on disappearing anytime soon. With all of the value I've gotten out of MacPorts over the years, I still have a tremendous amount of time to give back!

mascguy:
@jmroot Josh/Anyone, Mame doesn't appear to be getting picked up by the buildbot infrastructure. Any ideas why...?

PR #9655, opened 2021-01-08
hamlib: use legacysupport

kencu:
all builds now need legacysupport


ra1nb0w:
You are right. Thank you @kencu

PR #9661, opened 2021-01-08
OpenBLAS: on ARM64 build as +native for now

michaelld:
Based on personal testing on my ARM64 macOS box, it seems like we need to be building as +native there. That's all this PR does. Easy to add, easy to remove in the future once MacPorts and OpenBLAS get to that point.


michaelld:
We've seen some discussion from users, who also say using +native works on their M1 Macs while the default does not. Hence making +native the default for ARM64 seems like the way to go at the moment.

kencu (approved):
makes sense for now at least.

NicosPavlov:
This is a little bit unexpected, but that seems to be the simplest way for now.

PR #9662, opened 2021-01-08
fpc: Update for arm (WIP)

kamischi:
also use again configure.sdkroot and use macosx_deployment_target.

#### Description

This update is my first try to enable the native arm build. Untested. Work in Progress and i am waiting for error logs from arm machines.

###### Type(s)

- [x] enhancement

###### Tested on
macOS 11.1
Xcode 12.3

###### Verification 
Have you

### DIFF
--- a/lang/fpc/Portfile
+++ b/lang/fpc/Portfile
@@ -30,7 +30,7 @@ checksums           ${src} \
                     rmd160  7c5327aa6f4fd8616d78bc4f708463b9fb3e8844 \
                     sha256  0515ad4251c51fa7e5c0ff9b1d90ea50f41530b56276cc72b73798fae437b3b4 \
                     size    1556308
-supported_archs     x86_64
+supported_archs     x86_64 arm64
 
 # doesn't install shared libraries
 installs_libs       no
@@ -96,7 +96,7 @@ subport "${name}-cross" {
     worksrcdir      ${name}build-${version}/fpcsrc/compiler
     use_configure   no
     build.args      OPT="-ap -v0"
-    build.target    arm avr i386 i8086 jvm m68k mips mipsel powerpc powerpc64 sparc
+    build.target    aarch64 arm avr i386 i8086 jvm m68k mips mipsel powerpc powerpc64 sparc
 
     destroot {
       # delete unwanted files
@@ -124,7 +124,6 @@ if {${subport} eq "${name}"} {
         system -W ${workpath} "bzip2 -dc ${distpath}/${pp} | tar xf -"
     }
 
-
     post-patch {
         # adjust the path "codfilepath" for plex from /usr/local to MacPort's prefix.
         reinplace "s|/usr/local|${prefix}|g" \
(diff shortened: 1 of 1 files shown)


lang/fpc/Portfile
```diff
@@ -134,34 +133,50 @@ if {${subport} eq "${name}"} {
...
+            set bootstrapCompiler unsupported
+            set cpuTarget unsupported
+            set compiler unsupported
         }
     }
 
+    # os.major = 20 is macOS = 11 (BigSur)
+    if {${os.major} >= 20} {
+        set linkerOptions -XR${configure.sdkroot}
+    } else {
```
kencu, line 156: if you use ${configure.sdkroot} then every time there is a minor update to Xcode, this will break and need to be revbumped. See many similar tickets about many other ports that burn in the configure.sdkroot.

kencu: if you change this linker flage to only apply on BigSur+, we will need to revbump them all again as that will change the way os.major < 20 links back to the way it was before, which is to use "/" preferentially I believe.

lang/fpc/Portfile
```diff
@@ -134,34 +133,50 @@ if {${subport} eq "${name}"} {
...
     destroot.env        {*}${build.env}
 
     # build the compiler utilities msgdif and msg2inc
     post-build {
         system -W ${worksrcpath}/compiler/utils \
-               "../ppcx64 -WM10.9 -Fu../../rtl/units/${build_arch}-darwin -ap -v0 \
-                  -XR/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk msgdif.pp && \
-                ../ppcx64 -WM10.9 -Fu../../rtl/units/${build_arch}-darwin -ap -v0 \
-                  -XR/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk msg2inc.pp"
+               "../${compiler} -WM${macosx_deployment_target} -Fu../../rtl/units/${cpuTarget}-darwin -ap -v0 \
```
kencu, line 171: I like using the current system's deployment target as this will fix the builds on systems < 10.9, which I have had to manually fix before, as they were hard-coded to 10.9. I thought there was some reason you needed it to be 10.9, actually.

lang/fpc/Portfile
kamischi: I will take it back then.

lang/fpc/Portfile
kamischi: Actually, I am not really sure about this. 10.7 may still break for two reasons. First, the bootstrap compiler is probably for 10.9 and up. There is one for 10.7, but it has version 3.0.0, which is not sufficient to build 3.2. This would require a additional step of building version 3.0.4 first. Second, some time ago, there is an issue with the assembler (https://bugs.freepascal.org/view.php?id=32324), which shows up with certain versions of XCode and has never been resolved. End of next week, I will have access and time to check these things on an actual 10.7 system. Although I wonder about whether it is worth the effort. The reason for adding this option was to silence some warnings, which were uncritical. What were your manual fixes?

kencu: MacPorts has extremely strong support for older systems. As I recall, I only had to change 10.9 to 10.7 to make it work in the installed *.cfg (but I'd like to go back and look again).

Apple has two assemblers, the GNU "gas" and llvm's assembler driven by clang. The /usr/bin/as assembler driver defaulted to use "gas" until 10.9, and then Apple decided clang was robust enough, and changed the default, so the assembler driver would use "clang" preferentially.

MacPorts can install current clang versions on older systems, so MacPorts makes a slight tweak to the assembler driver installed by our cctools port (usually in /opt/local/bin/as) so that it will default to use some current version of clang on all systems, even back to 10.4 on Intel.

So that ticket you referenced in freepascal.org does not apply to MacPorts.

Great, eh?

lang/fpc/Portfile
kencu: > I will take it back then.

Oh, it's a debate. I considered asking you to do it like this in the first place (then I would not have had to revbump all the older installs) but in the end,  the way you did it was simpler so I just left it and committed it.

Either way has merit -- on the one hand it is better to prefer fpc to test against and use the libraries in "/" if they exist, as that represents functions that the system actually supports. On the other, the newer SDK has more features and so will generally support builds longer -- although you run the risk of testing for libraries in the SDK, finding them, thinking they exist on the system, and then finding out they don't.

So long as you pick one or other, we'll be find. You're the maintainer of fpc now, so pick the one you feel is best, and just stick with it. Uncertainty and change are the main enemies here, not the choice you decide on.

kamischi: That gives me one more idea, at least as perspective. In case assembler and linker are available for i386 for all systems, that is all, what's needed for the i386 arch and with some extra effort even ppc. But let's get arm64 done first. Crosscompilers for windows are also an easy option as they use internal assembler and linker. Only Linux and related friends need external cross binutils.

lang/fpc/Portfile
kamischi: That gives me one more idea, at least as perspective. In case assembler and linker are available for i386 for all systems, that is all, what's needed for the i386 arch and with some extra effort even ppc. But let's get arm64 done first. Crosscompilers for windows are also an easy option as they use internal assembler and linker. Only Linux and related friends need external cross binutils.

kamischi:
It breaks on arm64 when building the two extra utilities. However, the error message is not verbose enough to fix it. I am also concerned about the warning about building for iOS/iPhone instead of macOS. It could be that the target aarch64-darwin actually stands for iOS. I remember that because of this, there was the former target aarch64-darwin became aarch64-iOS and a new aarch64-darwin was introduced. I need to figure out the details. On a first glance, it is fixed in the fixes-3.2 branch of fpc, but I do not like to base a package on a moving target. If the changes are not too involved, i rather prefer to patch the 3.2 release.

lang/fpc/Portfile
```diff
@@ -96,7 +96,7 @@ subport "${name}-cross" {
     worksrcdir      ${name}build-${version}/fpcsrc/compiler
     use_configure   no
     build.args      OPT="-ap -v0"
-    build.target    arm avr i386 i8086 jvm m68k mips mipsel powerpc powerpc64 sparc
+    build.target    aarch64 arm avr i386 i8086 jvm m68k mips mipsel powerpc powerpc64 sparc
```
ryandesign, line 99: Doesn't adding a build target change what gets built and installed? If yes, then the revision of this subport needs to be increased. I assume that the other changes in this commit are also not merely cosmetic and therefore also would warrant a revision increase to whatever subport(s) they affect.

kamischi: Correct. This adds another crosscompiler. I will try to resolve the issue with building the utilities by increasing the verbosity in the next commit.

kencu:
when I get some cycles, I will see what is going on with this for the Apple silicon machine.

Karl, when you have time, please answer Ryan's question about whether your changing the target yet again changed the d*amned files installed by this software. And I guess the message is getting pretty clear now that whenever some change in a Portfile might possibly change files that a user building the software right now today would get as compared to whatever the buildbot last built, we need to bump the revision so that it all matches. A user installing the software today should get exactly what the buildbot is sending down the pipe, and you (and I ) should not need to be reminded about this every single time this Portfile is touched :>

kamischi:
OK. Got it. Sorry for that and thanks for baring with me.

The two utilities are probably not really for the compiler. So you could comment them out for a first test and try to build them after installing fpc. For increasing the verbosity, I often use -vt.

What I am really concerned about is the iPhone - MacOSX warning. I could really mean that plain vanilla 3.2 is not ready for arm64-macOS. I found a related commit (https://svn.freepascal.org/cgi-bin/viewvc.cgi?limit_changes=0&view=revision&revision=46866) and it changes 659 files. No way for a small patch. The question is switch to the fixes-3.2 branch or wait with arm64 support for the next release, whether it will be 3.2.2 or 3.4.0

kamischi:
I committed a patch to enable building of lazarus on older systems and increased the revisions of the supports. Can this pull request be reopened or should i issue a new pull request. I did not really find out, how to reopen this pull request.

kencu:
if you like we can work on this directly on an arm Mac, instead of trying to use the buildbot for this. Then we can sort it out and only update the portfile once, when we're done.

Just point me to your git repo where we can work on this.

kamischi:
Sure. This is the branch: https://github.com/kamischi/macports-ports/tree/kms-fpc-update-arm

PR #9684, opened 2021-01-11
openmpi: force use of macports libevent, to eliminate implicit use when already installed

mascguy:
#### Description

Force use of macports libevent, to ensure it is not used implicitly. Necessary for OpenMPI 4.x, as the configure script behavior changed to favor an external version (when available).

Fixes: https://trac.macports.org/ticket/59626

###### Type(s)
- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.8.5 12F2560
Xcode 5.1.1 5B1008

macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

###### Verification 
Have you

### DIFF
--- a/science/openmpi/Portfile
+++ b/science/openmpi/Portfile
@@ -9,6 +9,7 @@ PortGroup           legacysupport 1.0
 
 name                openmpi
 version             4.1.0
+revision            0
 set branch          [join [lrange [split ${version} .] 0 1] .]
 categories          science parallel net
 platforms           darwin
@@ -151,7 +152,8 @@ if {${subport} != ${name}} {
     archive_sites
 
     depends_build-append        port:pkgconfig
-    depends_lib-append          port:hwloc
+    depends_lib-append          port:hwloc \
+                                port:libevent
     depends_run                 port:mpi_select port:mpi-doc
     select.group                mpi
     select.file                 ${filespath}/${name}-${cname}
@@ -173,7 +175,8 @@ if {${subport} != ${name}} {
         --datadir=${prefix}/share/${name}-${cname} \
         --docdir=${prefix}/share/docdelete \
         --mandir=${prefix}/share/mandelete \
-        --with-hwloc=${prefix}
+        --with-hwloc=${prefix} \
+        --with-libevent=${prefix}
 
     post-destroot {
         if {[string first "-devel" $subport] > 0} {


mascguy:
Of note, the revision may not need to be bumped; I'll figure that out before marking as ready for review.

mascguy:
Thanks Ken!

ryandesign:
> Force use of macports libevent, to ensure it is not used implicitly

You forgot to increase the revision of the subports that this affects.

mascguy:
> On 2021-01-15-F, at 12:13, Ryan Schmidt <notifications@github.com> wrote:
> Force use of macports libevent, to ensure it is not used implicitly
> 
> You forgot to increase the revision of the subports that this affects.
> 
That was intentional, both to avoid impacting users of the existing port, and prevent rebuilds on our infrastructure. (And per the port file, a binary is never published, due to the use of hwloc… and each build being tied to the machine it was created on.)

Beyond that, this was my thought process:

Per the best practices, the two changes suggest a revision bump… and I get that. However, the key is that any existing user installs will be fine: If they had port libevent installed when they originally installed openmpi, then libevent was used. (And the new port file ensures users can’t simply uninstall libevent, without a warning relative to breaking openmpi.) If the user didn’t have libevent installed before, then openmpi built its own internal version… and again, the user will be fine.

Does that make sense? Thoughts?

ryandesign:
Your change changes the files on disk for any users who did not previously have libevent installed. That warrants a revision increase to get everyone on the same page. Otherwise some users of `openmpi-*-4.1.0_0` and `openmpi-devel-*-4.0.999-201905080241_1` are using its built-in libevent and other users of that same exact version and revision are using MacPorts libevent. That's not consistent. That's not [reproducible](https://trac.macports.org/wiki/ReproducibleBuilds). It makes debugging possible future problems more difficult.

We do want to avoid unnecessary rebuilds, but this change makes a rebuild necessary. If the port will be updated to a new version soon anyway, then a revision increase now could be skipped. For example, the openmpi-devel subports appear to be lagging behind the openmpi port. This could be an opportunity to update the openmpi-devel subports to a newer version:

```
openmpi-devel seems to have been updated (port version: 201905080241, new version: 202101130241-f2e023d)
```

The non-devel openmpi subports appear to be at the latest version already so I don't think there's any avoiding increasing their revisions.

mascguy:
Make sense, and clearly still learning here. I’ll go ahead and bump the revision then.

As always, thanks for the feedback… and for humoring me!

Cheers,
-Chris

PR #9694, opened 2021-01-13
poetry: install bash and zsh completion in post-destroot

lpsinger:
#### Description

Have poetry install its bash completion script in the post-activate phase.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H114
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/python/poetry/Portfile
+++ b/python/poetry/Portfile
@@ -5,7 +5,7 @@ PortGroup               python 1.0
 
 name                    poetry
 version                 1.1.4
-revision                0
+revision                1
 categories-append       devel
 platforms               darwin
 license                 MIT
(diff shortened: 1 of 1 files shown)


ryandesign:
Why can't this be installed in the destroot phase?

lpsinger:
> Why can't this be installed in the destroot phase?

Because the contents of the file depend on the value of `${prefix}`.

g5pw:
I think what @ryandesign meant is, why is this post-activate and not post-destroot for example. In this way there's no need to clean up the file, and it would show up in `post contents` and `port provides`.

lpsinger:
The contents of the generated value depend upon the absolute path of the user's MacPorts prefix. When the buildbot makes 'binary' packages for poetry, its value of the prefix may differ from the user's value. So builedbot-generated binary packages would be broken if the bash completion scripts were generated during the destroot phase.

ryandesign:
> When the buildbot makes 'binary' packages for poetry, its value of the prefix may differ from the user's value.

No. The buildbot builds binary packages for prefix /opt/local. If the user's prefix is different, they do not receive our binaries.

lpsinger:
I didn't know that! I'll rework this to do it in post-destroot.

lpsinger:
Done.

g5pw:
@lpsinger "since you're here" would it be possible to add zsh completions as well? Thanks 😄

lpsinger:
@g5pw, what directory should the zsh completions go into?

g5pw:
I edited my previous command, it should go in `${prefix}share/zsh/site-functions/_poetry`. Note the underscore in the beginning.

Thanks again.

lpsinger:
@g5pw, done.

g5pw:
Awesome, thanks for your contribution to MacPorts, @lpsinger! Merging.

PR #9695, opened 2021-01-13
qt5*-qtbase: fix build for Xcode 11.4 and later

chrstphrchvz:
Outdated Xcode-LLVM clang version mapping in qcompilerdetection.h defines `Q_CC_CLANG` too low; check for `_bit_scan_reverse()` macro instead.
~~(`qt53-qtbase` not affected but updated for consistency.)~~

Revise comment and reference to upstream clang commit to better reflect why this problem appeared with clang 9.0 and not sooner.
Fixes: https://trac.macports.org/ticket/62051

###### Type(s)

- [x] bugfix

###### Tested on

Only patch phase is tested.

###### Verification 
Have you

### DIFF
--- a/aqua/phantomjs-qt/Portfile
+++ b/aqua/phantomjs-qt/Portfile
@@ -267,11 +267,21 @@ foreach {module module_info} [array get modules] {
                     mkspecs/features/mac/rez.prf
             }
 
-            # Qt assumes _bit_scan_reverse is not provided by Clang
-            # this assumption is incorrect for recent versions of Clang
-            #     see https://github.com/llvm/llvm-project/commit/e54093fcc0a651df14040a61fc8efdd338389afa
+            # Qt assumes Clang's x86intrin.h does not provide _bit_scan_reverse
+            # this assumption is incorrect for Clang 9.0 and later
+            #     see https://github.com/llvm/llvm-project/commit/88f4054f48c5
             # see https://trac.macports.org/ticket/59364
             patchfiles-append patch-qtbase-intel_intrinsics.diff
+            if {${os.major} >= 19} {
+                # Xcode 11.4 is also based on Clang 9.0, but
+                # patch-qtbase-intel_intrinsics.diff does not work
+                # for Xcode clang; use an alternative approach
+                # (not tested on older macOS but may work)
+                # See https://trac.macports.org/ticket/62051
+                patchfiles-replace \
+                    patch-qtbase-intel_intrinsics.diff \
+                    patch-qtbase-intel_intrinsics-xcode11.4.diff
+            }
 
             # ICU now requires C++11
             # see https://trac.macports.org/ticket/59434
--- a/aqua/phantomjs-qt/files/patch-qtbase-intel_intrinsics-xcode11.4.diff
+++ b/aqua/phantomjs-qt/files/patch-qtbase-intel_intrinsics-xcode11.4.diff
@@ -1,0 +1,11 @@
+--- src/corelib/tools/qsimd_p.h.orig	2017-09-06 05:13:54.000000000 -0700
++++ src/corelib/tools/qsimd_p.h	2021-01-12 23:42:29.000000000 -0600
+@@ -329,7 +329,7 @@
+     _BitScanForward(&result, val);
+     return result;
+ }
+-#  elif (defined(Q_CC_CLANG) || (defined(Q_CC_GNU) && Q_CC_GNU < 405)) \
++#  elif ((defined(Q_CC_CLANG) && !defined(_bit_scan_reverse)) || (!defined(Q_CC_CLANG) && defined(Q_CC_GNU) && Q_CC_GNU < 405)) \
+     && !defined(Q_CC_INTEL)
+ // Clang is missing the intrinsic for _bit_scan_reverse
+ // GCC only added it in version 4.5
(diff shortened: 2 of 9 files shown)


kencu:
before we can commit this, we have to show that each version of qt5-qtbase builds with the default compilers on each default system with this patch applied. 

Having these old qt5 versions build with xcode 11 is of some interest, but having them build on the systems that actually need them is the main point, and they all already build on those systems now, so we can't break that.

chrstphrchvz:
I will not be testing this PR as requested. Reject this PR if no one else will be testing it.

kencu:
It's sometimes best to just leave certain things alone...

mleinart:
Heya, just driving by to say I was unable to compile phantomjs-qt-qtbase (a direct dependency of phantomjs) without this patch applied on Catalina 10.15.7 with Xcode 12.4

chrstphrchvz:
@mleinart Maybe this has a chance of being revived, but for now I suggest subscribing to https://trac.macports.org/ticket/62051 in case of updates.

mnaydenov:
Second what is already said. Just compiled Qt 5.6 qtbase using the qt56 version of the patch on macOS 10.15.7, MacOSX11.0.sdk.

chrstphrchvz:
Reopening since there is interest in this at least on newer macOS versions.

Is @MarcusCalhoun-Lopez available to review this approach?

chrstphrchvz:
As I remarked on the ticket, if this gets held up by lack of testing on older macOS, then I will have the patch be applied only on newer macOS.

kencu:
over the years, few ports have proven to be as touchy to build as qt5. I think having ancient qt versions building with the most current Xcodes is kinda neat. Some might find it useful, like those folks who write Zoom, which seems to run on just every device.

But not even checking once to see if it breaks the natural build on the system that needs to use it is just asking for trouble. Maybe it does seem obvious that it will be fine to you. Perhaps it will be fine.

I know that testing each system is onerous, having done it hundreds of times. I suggested perhaps at least testing the one that seems most likely to show a problem, which is the build of qt56 on 10.7 with the clang 400 compiler.

If you feel this is all unnecessary and even frustrating, go ahead and push, and hope for the best. It's not what I would do, but I do tend to be quite careful having run into so many issues over the years.

chrstphrchvz:
CI builds timed out. I think this is ready to merge.

mascguy:
> As I remarked on the ticket, if this gets held up by lack of testing on older macOS, then I will have the patch be applied only on newer macOS.

Hey Chris, if you need help with testing on older macOS releases, let us know. KenC, myself, and probably a few others, have VMs for the entire suite of older macOS releases going back to 10.6.

chrstphrchvz:
> Hey Chris, if you need help with testing on older macOS releases, let us know. KenC, myself, and probably a few others, have VMs for the entire suite of older macOS releases going back to 10.6.

Feel free to give be2a5a9deb a try.

mascguy:
> > Hey Chris, if you need help with testing on older macOS releases, let us know. KenC, myself, and probably a few others, have VMs for the entire suite of older macOS releases going back to 10.6.
> 
> Feel free to give [be2a5a9](https://github.com/macports/macports-ports/commit/be2a5a9debce80055c5418c384afdd386d782e29) a try.

Chris, are there a few specific macOS releases this needs to be tested on?

chrstphrchvz:
> Chris, can you suggest a few specific macOS releases to test this on?

See https://github.com/macports/macports-ports/pull/9695#issuecomment-759704054

I think that would entail building qt53-qtbase on 10.6; qt55-qtbase, qt56-qtbase, and phantomjs-qt-qtbase on 10.7; and qt57-qtbase on 10.8.

kencu:
I didn't notice that you had altered this PR to only apply the new patch on the newest systems. That should certainly be fine.

I took some time this morning to build qt56-qtbase on 10.7 using the new patch (I changed the PR to use the new patch on 10.7 as well). On 10.7, qt56-qtbase builds with macports-clang-9.0 I notice. That would be because we force it to do so with:
```
compiler.cxx_standard 2011
```
in the Portfile.

With the new patch, on 10.7, using macports-clang-9.0, it builds through to completion without any errors.

I usually feel this is the most delicate of the systems. I'm not sure what happens on 10.8 and 10.9 just now, but if they are also building qt5* using a macports-clang-9.0 or newer, they should also be fine then. If they fail, then we can force them to do that if they are not already.

Your PR as it is does not use the new patch on any but the newest systems. It can therefore be commited as is without any worry. Given what happens on 10.7, we can also probably just use the new patch all the time, and it would seem any wreckage is fixable.

I can see your frustration with the idea of checking things. In general, the focus of MacPorts has to be on the newest systems, and required changes that break older systems are just tough luck. I don't think there is any disagreement about that. Changes that break older systems for no reason are just frustrating, because we do have a fair following in the older systems crowd, they do seem to be passionate about Macs in general, and we believe that by supporting them we keep them in the fold and bring them into the family. 

So that's why I suggested we test the change. Limiting it as you did made testing unnecessary, but I didn't notice that you did that, or when, so here this PR sits 8 months later. I'll commit it now, and we can deal with whatever fallout comes afterwards; based on the success on 10.7, anything that comes is highly likely to be fixable.

chrstphrchvz:
Please see #12189 if interested in dropping Xcode 11.4-incompatible patches.

PR #9735, opened 2021-01-15
MPICH: Incorporate #6899 changes and upgrade

eborisch:
Maintainer update.

Updates MPICH to 3.4.

Removes the [-devel] branches (not changing frequently, and I haven't had use for them lately -- if someone still really wants them, let me know.)

I'm not the mpi PG maintainer, but this copies MCL's changes in #6899 (which no longer applies cleanly.)


eborisch:
@jmroot @ryandesign @dstrubbe: Any comments / concerns with the associated changes to the mpi PG?

jmroot:
> @jmroot @ryandesign @dstrubbe: Any comments / concerns with the associated changes to the mpi PG?

Looks OK as far as it goes, but the addition of clang10 to the disabled list for mpich should also be removed.

science/mpich/Portfile
```diff
@@ -86,6 +64,7 @@ array set clist {
     gcc6    {macports-gcc-6}
     gcc7    {macports-gcc-7}
     gcc8    {macports-gcc-8}
+    devel   {clang} # Here to prevent portindex failure on -devel
```
jmroot, line 67: This isn't the right fix. Among other things, this causes you to declare a subport called `mpich-devel-devel`. Instead, don't run the code that uses `clist` when in one of the obsolete subports.

science/mpich/Portfile
```diff
@@ -113,9 +93,55 @@ subport ${name}-default {}
 subport ${name}-devel {}
 subport ${name}-devel-default {}
 
-# Linker for Apple clang version 421.11.66 segfaults
-# See https://trac.macports.org/ticket/36654#comment:9
-compiler.blacklist-append   {clang >= 421.11.66 < 425.0.24}
+if {${name} ne ${subport}} {
```
jmroot, line 96: This is one place that should exclude the obsolete subports.

science/mpich/Portfile
```diff
@@ -155,16 +179,22 @@ configure.optflags-delete   -O2 -Os
 configure.cppflags-delete   -I${prefix}/include
 configure.ldflags-delete    -L${prefix}/lib
 
-if {${subport} != ${name}} {
+if {${name} ne ${subport}} {
```
jmroot, line 182: This is another place that should exclude the obsolete subports.

eborisch: Done

science/mpich/Portfile
eborisch: Done

science/mpich/Portfile
eborisch: Done

PR #9743, opened 2021-01-16
getdns: update to 1.6.0

herbygillot:
###### Tested on

macOS 10.15.7 19H114
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/devel/getdns/Portfile
+++ b/devel/getdns/Portfile
@@ -1,10 +1,12 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           cmake   1.1
 
 name                getdns
-version             1.5.2
+version             1.6.0
 revision            0
+
 categories          devel
 platforms           darwin
 maintainers         {l2dy @l2dy} openmaintainer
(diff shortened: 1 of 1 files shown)


devel/getdns/Portfile
```diff
@@ -19,20 +21,25 @@ long_description    getdns is a modern asynchronous DNS API. It implements DNS \
...
+homepage            https://getdnsapi.net
+
+master_sites        ${homepage}/dist/
+
+checksums           rmd160  58ad868301b78a0b6ac4770324e1d0bc6c3f9b61 \
+                    sha256  40e5737471a3902ba8304b0fd63aa7c95802f66ebbc6eae53c487c8e8a380f4a \
+                    size    673593
 
-master_sites        ${homepage}dist/
+cmake.build_type            Release
```
l2dy, line 32: Is this override necessary?

devel/getdns/Portfile
```diff
@@ -19,20 +21,25 @@ long_description    getdns is a modern asynchronous DNS API. It implements DNS \
...
 
-checksums           rmd160  85ce46a88273dbe6c38f73acc5a4d15a7c4a6392 \
-                    sha256  1826a6a221ea9e9301f2c1f5d25f6f5588e841f08b967645bf50c53b970694c0 \
-                    size    1091088
+depends_build-append        port:check \
```
l2dy, line 34: `port:check` should be `depends_test` if possible.

devel/getdns/Portfile
herbygillot: Perhaps not... removed.

devel/getdns/Portfile
herbygillot: OK, CMake was failing if `check` wasn't present.  I have added `-DBUILD_TESTING=OFF`.

Is there a way to conditionally add `-DBUILD_TESTING=OFF` or `-DBUILD_TESTING=OFF` to the configure args depending on whether or not the current port process intends to run the test phase?

l2dy: That is not possible as far as I know. `libcbor` implements this with a variant, most other ports just disable the test.

kencu: It is not currently possible, but I wish it were possible.

Perhaps running ```sudo port test``` could automatically enable a ```tests``` variant, or something like that....

herbygillot: Got it.  Maybe `port` just needs to expose some boolean variable in the Portfile context if a user specifies that the test phase is to be run.

l2dy (approved):
LGTM

herbygillot:
Thanks @l2dy , @kencu

PR #9769, opened 2021-01-20
glances: update to 3.1.6.1 and fix build on Apple Silicon

rabidf0x:
#### Description

I have updated the glances port to use the latest version 3.1.5 and shifted the underlying python dependency to version 3.8 for compatibility on Apple Silicon based m1 Macs. Python37 does not install in my testing.

This is my second attempt.  Broke my forked repo, still learning git.  I took the recommendations from my first attempt in PR 9702.

notify: @harens @g5pw

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/net/teleport/Portfile
+++ b/net/teleport/Portfile
@@ -3,12 +3,8 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/gravitational/teleport 5.1.0 v
+go.setup            github.com/gravitational/teleport 5.1.2 v
 revision            0
-
-categories          net
-license             Apache-2
-platforms           darwin
 
 homepage            http://gravitational.com/teleport
 
@@ -20,6 +16,10 @@ long_description    Teleport is a modern SSH server and CA for managing clusters
                     integration with Google Apps and other OAuth identity providers, \
                     and a Web UI.  Built on the Golang SSH library, and compatible with OpenSSH
 
+categories          net
+license             Apache-2
+platforms           darwin
+
 maintainers         { gmail.com:graemelawes @gclawes } \
                     {gmail.com:herby.gillot @herbygillot} \
                     openmaintainer
@@ -27,7 +27,8 @@ maintainers         { gmail.com:graemelawes @gclawes } \
 # Build process requires a git checkout
 fetch.type          git
 
-depends_lib         port:go port:zip
+depends_lib         port:go \
+                    port:zip
 
 build.cmd           make
 build.target        full
@@ -38,6 +39,5 @@ destroot {
         xinstall ${worksrcpath}/build/${i} ${destroot}${prefix}/bin
     }
 }
-
 
 github.livecheck.regex {([0-9.]+)}


harens:
Good to see you again @rabidf0x. Really impressed by your dedication and thanks again for doing this 👍.

Just to summarise https://github.com/macports/macports-ports/pull/9702#issuecomment-760792004 and https://github.com/macports/macports-ports/pull/9702#pullrequestreview-568059345, since glances has [entry_points](https://github.com/nicolargo/glances/blob/74a499b5782cfc0a38dd6d3367e02971c273c74a/setup.py#L119), `py-setuptools` should be added as a lib dependency unless @g5pw recommends otherwise (see those referenced comments for more info).

Like I said before, I'm really impressed at your efforts considering this is your first time using git. As always, let me know if you need assistance with anything and I'd be happy to help.

g5pw:
Thank you for this patch @rabidf0x! At first glance (pun intended) most of this looks OK. Two points:
* Since we're bumping Python, we might as well bump to the latest version (3.9)
* Is there a reason for removing the reinplace? What that does is it makes sure that the interpreter used for executing the software is actually the MacPorts one.

rabidf0x:
Thanks @g5pw!  If I am getting in the way let me know.  I went with python 3.8 as it is the highest listed python version the glances author is testing against if I am reading [this](https://github.com/nicolargo/glances/commit/6f9365bb246f6a934df4531ff514991b840aa069) correctly. 

I removed the reinplace because it now throws a warning that the command made no change.  When I run `which glances` it is found in the right location.

If I don't have this squashed right let me know.  GitKraken said I needed to pull to before pushing my squashed commits or force the push causing the merge-remote tracking mess, that I reverted.

rabidf0x:
@g5pw my mistake...I see now reinplace  does.  hmmm...it looks like glances is now running on python in `/opt/local/Framework` vice `/opt/localbin`.  Let me put those lines back and see if I can get the warning message.  This may be beyond my abilities to fix.

rabidf0x:
@g5pw here is the output with the reinplace back in:

`Warning: reinplace s|/usr/local|/opt/local| didn't change anything in /opt/local/var/macports/build/_Users_rabidf0x_ports_sysutils_glances/glances/work/glances-3.1.5/setup.py`

rabidf0x:
@g5pw after more poking around I think everything is working as expected with the inreplace removed.  Would you be able to confirm?

sysutils/glances/Portfile
```diff
@@ -18,17 +18,14 @@ long_description    Glances is a cross-platform monitoring tool which aims to pr
...
-depends_lib         port:py${python.version}-psutil \
-                    port:py${python.version}-future
-
-patch {
-    reinplace "s|/usr/local|${prefix}|" setup.py
-}
+python.default_version 38
+depends_lib-append  port:py${python.version}-psutil \
+                    port:py${python.version}-future \
+                    port:py${python.version}-setuptools
```
ryandesign, line 28: Is py-setuptools really needed by glances at runtime? Normally it is only needed at build time, and if so, it should be in `depends_build`.

ryandesign:
> everything is working as expected with the inreplace removed

setup.py used to contain this code (added in https://github.com/nicolargo/glances/commit/f1c76b40c9e6d9f1802997646f16e0083eeba6d9):

```
elif 'darwin' in sys.platform:
     etc_path = os.path.join('/usr/local', 'etc', 'glances')
```

Our `reinplace` fixed that to use the MacPorts prefix instead of /usr/local.

It looks like that code is no longer in setup.py but there is something similar in glances/config.py:

```
    elif BSD or MACOS:
        path = '/usr/local/etc'
```

and also a reference to it in docs/man/glances.1 so I think we need to be using `reinplace` to make the change in glances/config.py and docs/man/glances.1 now.

sysutils/glances/Portfile
harens: That was my call. `py-setuptools` is required at build time for all the usual reasons, but it is also required at runtime since glances is a CLI tool with [entry_points](https://github.com/nicolargo/glances/blob/74a499b5782cfc0a38dd6d3367e02971c273c74a/setup.py#L119).

I find https://github.com/macports/macports-ports/pull/2684#issuecomment-426193304 quite a succinct explanation why, and @reneeotten discusses it in https://github.com/macports/macports-ports/pull/6314#issuecomment-585490282.

rabidf0x:
So glances 3.1.6.1 has been release and I spot references to python3.9 so I made that change as well.  I also took the recommended action by @ryandesign to direct the former change in setup.py to glances/config.py.

g5pw:
@rabidf0x thanks for your work! I think the port file looks good now, the only thing that's missing is squashing all commits in one and I'll merge this. Could you do that?

rabidf0x:
@g5pw I have to admit defeat on squashing...  I've made a mess on my local repo again that I am not sure how to back out of it.  I'll work getting a better understanding of using git.  Are you able to take over?  If closing this PR and having you set up a clean commit moves this port forward I am fine with that! :)

g5pw:
@rabidf0x don't worry, I can apply your patch and give you ownership :)

PR #9770, opened 2021-01-20
openal-soft: fix build errors for older macOS releases

mascguy:
#### Description

Additional change to fix build errors for older macOS releases.

Fixes: [openal-soft @1.21.0: error: inline declaration of 'PopCount<unsigned long long>' follows non-inline definition](https://trac.macports.org/ticket/61431)

###### Type(s)
- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.8.5 12F2560
Xcode 5.1.1 5B1008

macOS 10.10.5 14F2511
Xcode 6.4 6E35b

macOS 10.11.6 15G22010
Xcode 7.3.1 7D1014

###### Verification 
Have you

### DIFF
--- a/audio/openal-soft/Portfile
+++ b/audio/openal-soft/Portfile
@@ -2,6 +2,7 @@
 
 PortSystem              1.0
 PortGroup               cmake 1.1
+PortGroup               compiler_blacklist_versions 1.0
 if {${subport} eq "openal-soft" && [variant_isset gui]} {
 PortGroup               qt5 1.0
 }
@@ -51,6 +52,13 @@ patch.pre_args          -p1
 
 compiler.cxx_standard   2014
 compiler.thread_local_storage   yes
+
+# Fix for incorrect MacPorts compiler selection, tracked by:
+# https://trac.macports.org/ticket/61418
+# Ticket specific to openal-soft build errors, caused by the former:
+# https://trac.macports.org/ticket/61431
+compiler.blacklist      \
+                        {clang < 800.0.38}
 
 configure.args-append   -DALSOFT_EXAMPLES=OFF \
                         -DALSOFT_UTILS=ON \


audio/openal-soft/Portfile
```diff
@@ -52,6 +53,9 @@ patch.pre_args          -p1
 compiler.cxx_standard   2014
 compiler.thread_local_storage   yes
 
+compiler.blacklist      \
+                        {clang < 900}
```
mascguy, line 57: We can potentially make this less restrictive, but this seemed like a reasonable compromise. Let me know if you'd prefer to broaden the clang range we allow.

ryandesign:
The error being reported in the ticket now is:

```
error: thread-local storage is not supported for the current target
```

The port already specifies:
```
```
which should be all that's needed to get MacPorts to exclude compilers that don't support thread local storage. However there is a bug in MacPorts where this doesn't work right when `compiler.cxx_standard` is set to a value greater than 2011, which it is in this port. That bug is fixed by https://github.com/macports/macports-base/pull/214. So yes we should blacklist incompatible compilers until a new release of MacPorts containing that fix is released. Providing a link to that PR as a comment near the blacklist line will help us remember to remove it later when it's no longer needed. Per that PR, the correct minimum clang version is 800.0.38.

mascguy:
Updated blacklist to 8.0.38. Also added comment, referencing both the compiler-selection ticket, as well as the one specific to openal-soft.

mascguy:
I'll retest against macOS 10.8, 10.10, and 10.11, and report back.

mascguy:
Retested on macOS 10.8, 10.10, and 10.11, and looks good.

mascguy:
Ryan, the natives are getting a bit restless, relative to this issue. Can we go ahead and commit?

mascguy:
Thanks Ken!

PR #9775, opened 2021-01-20
rust-analyzer: new Portfile submission

fabius:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1
Xcode 12.3

###### Verification 
Have you

### DIFF
--- a/devel/rust-analyzer/Portfile
+++ b/devel/rust-analyzer/Portfile
@@ -1,0 +1,211 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           cargo 1.0
+
+github.setup        rust-analyzer rust-analyzer 2021-01-18
+revision            0
+
+homepage            https://rust-analyzer.github.io
+
+description         An experimental Rust compiler front-end for IDEs
+
+long_description    rust-analyzer is an experimental modular compiler \
+                    frontend for the Rust language.  It is a part of a larger \
+                    rls-2.0 effort to create excellent IDE support for Rust.
+
+categories          devel
+platforms           darwin
+supported_archs     x86_64
+license             Apache-2
+
+maintainers         @fabius openmaintainer
+
+github.tarball_from archive
+
+checksums           ${distname}${extract.suffix} \
+                        rmd160  c0db3ee3e3f6db7a03f8b53fa45b180b760f6504 \
+                        sha256  009ab946848719bb05bcaddab1cf5a75a45c5a92ae6dc31763adb42c3229f7db \
+                        size    1435696
+
+post-extract {
+    file delete ${worksrcpath}/lib/README.md
+}
+
+destroot {
+    xinstall -m 0755 ${worksrcpath}/target/[cargo.rust_platform]/release/${name} ${destroot}${prefix}/bin/
+}
+
+cargo.crates \
+    addr2line                       0.14.1  a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7 \
+    adler                            0.2.3  ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e \
+    ansi_term                       0.12.1  d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2 \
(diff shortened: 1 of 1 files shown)


devel/rust-analyzer/Portfile
```diff
@@ -0,0 +1,34 @@
...
+master_sites        https://github.com/rust-analyzer/rust-analyzer/archive/
+distname            ${version}
+checksums           rmd160  c0db3ee3e3f6db7a03f8b53fa45b180b760f6504 \
+                    sha256  009ab946848719bb05bcaddab1cf5a75a45c5a92ae6dc31763adb42c3229f7db \
+                    size    1435696
+supported_archs     x86_64 arm64
+
+build {
+    system -W ${worksrcpath} "rm lib/README.md"
+    system -W ${worksrcpath} "cargo build --release"
```
herbygillot, line 30: The `cargo` portgroup (`PortGroup cargo 1.0`) normally does the `cargo build` step for you, which means after specifying it, you don't have to specify an explicit `build` section.  Did the normal structure not work for a particular reason?

fabius: Thanks for the review! First of all: Is there any Documentation on those Portgroups other than the _resources directory? 

The normal method unfortunately did not work as the currently packaged Rust version, for whatever reason, assumes that everything inside the lib/ directory is a directory itself. So cargo looks for lib/README.md/Cargo.toml which obviously does not exist.

fabius: fyi this is not an issue on the latest stable rust version

herbygillot: Hey @fabius, I'll get back to you - I'd like to take a look at the `lib/REAME.md` issue you called out

fabius: > Hey @fabius, I'll get back to you - I'd like to take a look at the `lib/REAME.md` issue you called out

@herbygillot Any news on this one? 

_I got a working Portfile for Rust 1.49/aarch64 if that helps you testing_

herbygillot:
So here is a commit with a Portfile for `rust-analyzer` that's written as per best practices for Rust ports in Macports:

https://github.com/herbygillot/macports-ports/commit/a428e73238db25e2b096b75e9bfd74f38bec1f57

Important points of note:
* **no v prefix in `github.setup`** - the `v` at the end of `github.setup` is the branch or tag prefix.  You would use it if the branch or tag name in the `rust-analyzer` branch looked like "**v**2021-01-18", but since it's plain "2021-01-18", you don't specify any prefix.  This is important because this is what the `github` portgroup uses in order to construct the URLs it needs to fetch the source tarballs.

* **no `master_sites` or `distname`** - MacPorts and the many portgroups know how to fetch the source tarballs for a piece of software once you've specified the name, version and optional version prefix/suffix as per the way they expect.  In this Portfile, you've done this with `github.setup`.  However since the Github releases URL for `rust-analyzer` are "archive" links, we do have to specify `github.tarball_from archive`.

* **weird lib/README.md directory** - you're definitely right that something strange is going on with the "lib/README.md" path - we delete it right after files have been extracted

* **no manual invocation of `cargo build`** - we are using the `cargo` portgroup. It knows how to build Cargo projects.  As long as we provide all the deps in the `cargo.crates` section, it will take care of the rest, so we do not need our own `build` section invoking `cargo` ourselves.

* **`cargo.crates` section** - MacPorts wants to be aware of all the dependency crates for a Cargo project.  This is so that all of the crates can be mirrored to MacPorts mirrors, enabling MacPorts to build the port even in the (unlikely) scenario that the upstream Rust repositories are down. It also allows MacPorts to check the integrity of all the dependencies and ensure that all downloads happen in the "fetch" phase of building a Port.  In this way, when it's time to actually build the source, no downloading needs to happen. This is the opposite of what happens when you directly invoke `cargo build` in a `build` block - `cargo` downloads things on its own outside of MacPorts' awareness and does the build. Generating this section can be admittedly annoying.  I use my own utility to do this, [cargo2ports](https://github.com/herbygillot/cargo2ports). Others have found it useful as well.

You can see examples of other Cargo Portfiles here:
* https://github.com/macports/macports-ports/blob/master/tex/tectonic/Portfile
* https://github.com/macports/macports-ports/blob/master/sysutils/diskus/Portfile
* https://github.com/macports/macports-ports/blob/master/editors/ox/Portfile

fabius:
Cheers man! I rebased the branch on your fixup. 
Edit: `supported_archs: arm64` got removed for now until the rust base port supports it.

herbygillot:
> Edit: `supported_archs: arm64` got removed for now until the rust base port supports it.

Makes perfect sense. Merging.

PR #9777, opened 2021-01-20
opencv/opencv4: add new variants: opencv +openmp, opencv4 +openni

mascguy:
#### Description

Add variants, missed during the recent updates to opencv/opencv4:
* opencv: add variant openmp
* opencv4: add variant openni

This is phase one, of a five-phase rollout:
1. Add variants missed during opencv/opencv4 reconciliation (opencv +openmp, opencv4 +openni).
1. Add new port opencv3, segregated like opencv4. Also include initial changes, supporting later conversion to subport. [Merged PR 9789](https://github.com/macports/macports-ports/pull/9789)
1. Update ports presently dependent upon opencv, to use opencv3. And update port opencv to be a stub port, with replaced_by opencv3, etc.
1. Add new port opencv2, using same segregation strategy.
1. Reconcile all three into single opencv port, with three subports. This step may be combined with previous.

Details provided in tickets:
- [opencv/opencv4: eliminate portfile duplication, via subport](https://trac.macports.org/ticket/62011)
- [OpenCV 2.x (please re-instate or provide a new opencv2 port)](https://trac.macports.org/ticket/49670)

###### Type(s)
- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.8.5 12F2560
Xcode 5.1.1 5B1008

macOS 10.10.5 14F2511
Xcode 6.4 6E35b

macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

###### Verification 
Have you

### DIFF
--- a/graphics/opencv/Portfile
+++ b/graphics/opencv/Portfile
@@ -277,6 +277,12 @@ variant opencl description {Enable OpenCL support} {
                             -DWITH_OPENCL=ON
 }
 
+variant openmp description {Include OpenMP support (untested)} {
+    configure.args-replace \
+                            -DWITH_OPENMP=OFF \
+                            -DWITH_OPENMP=ON
+}
+
 variant openni description {Enable OpenNI support} {
     depends_lib-append      port:openni
     patchfiles-append       patch-cmake_OpenCVFindOpenNI.cmake.diff
--- a/graphics/opencv/Portfile
+++ b/graphics/opencv/Portfile
@@ -277,10 +277,12 @@ variant opencl description {Enable OpenCL support} {
                             -DWITH_OPENCL=ON
 }
 
-variant openmp description {Include OpenMP support (untested)} {
+variant openmp description {Include OpenMP support} {
+    compiler.openmp_version 4.0
     configure.args-replace \
                             -DWITH_OPENMP=OFF \
                             -DWITH_OPENMP=ON
+
 }
 
 variant openni description {Enable OpenNI support} {
--- a/graphics/opencv4/Portfile
+++ b/graphics/opencv4/Portfile
@@ -277,6 +277,16 @@ variant openmp description {Include OpenMP support (untested)} {
                         -DWITH_OPENMP:BOOL=ON
 }
 
+variant openni description {Enable OpenNI support} {
+    depends_lib-append  port:openni
+    configure.args-replace  \
+                        -DWITH_OPENNI:BOOL=OFF \
+                        -DWITH_OPENNI:BOOL=ON
+    configure.args-append \
+                        -DOPENNI_INCLUDE_DIR:PATH=${prefix}/include/ni \
+                        -DOPENNI_LIB_DIR:PATH=${prefix}/lib
+}
+
 variant nonfree description {Include nonfree algorithms} {
     configure.args-replace \
                         -DOPENCV_ENABLE_NONFREE:BOOL=OFF \
--- a/graphics/opencv4/Portfile
+++ b/graphics/opencv4/Portfile
@@ -271,7 +271,9 @@ variant gdal description {Include GDAL support} {
                         -DWITH_GDAL:BOOL=ON
 }
 
-variant openmp description {Include OpenMP support (untested)} {
+variant openmp description {Include OpenMP support} {
+    compiler.openmp_version \
+                        4.0
     configure.args-replace \
                         -DWITH_OPENMP:BOOL=OFF \
                         -DWITH_OPENMP:BOOL=ON


graphics/opencv/Portfile
```diff
@@ -277,6 +277,12 @@ variant opencl description {Enable OpenCL support} {
                             -DWITH_OPENCL=ON
 }
 
+variant openmp description {Include OpenMP support (untested)} {
+    configure.args-replace \
+                            -DWITH_OPENMP=OFF \
+                            -DWITH_OPENMP=ON
```
mascguy, line 283: Note: Unlike most other variants, `openmp` doesn't involve additional dependencies.

mascguy:
@mf2k Frank, if you're interested in merging another PR, this one is also battle-tested and ready-to-go...

graphics/opencv/Portfile
kencu: I didn't merge this because you're not specifying an openmp compatible compiler using MacPorts' facilities for this:

```compiler.openmp_version```

see this page:

<https://trac.macports.org/wiki/CompilerSelection>

Comment (kencu) [reply graphics/opencv/Portfile:283]: and also because you say it is "untested" and I just don't see how you can really enable that without testing whether it works, to be honest...

So -- seems to me you have a bit more work to do here before this is ready for prime-time.

Comment (mascguy) [reply graphics/opencv/Portfile:283]: > and also because you say it is "untested" and I just don't see how you can really enable that without testing whether it works, to be honest...

Well, this variant isn't untested - I've certainly installed with it enabled - but it's essentially a plaseholder at this point: OpenMP isn't found during the configure phase, so OpenCV is built without it.

It would probably be more accurate to change the variant description to "Not yet implemented." But my priority was to bring both ports up to parity first, and tackle this type of cleanup during phase five. (That'll be when all OpenCV-related ports are converted to subports, duplicated logic removed, etc.)

Comment (mascguy) [reply graphics/opencv/Portfile:283]: Updated variant descriptions to state "Not Yet Implemented," rather than untested.

Comment (kencu) [reply graphics/opencv/Portfile:283]: if you add this in the variant, with the proper setting, then it will probably be found during the configure phase, and actually might work:
```
compiler.openmp_version
```

mascguy: > if you add this in the variant, with the proper setting, then it will probably be found during the configure phase, and actually might work:
> 
> ```
> compiler.openmp_version
> ```

Hmmmmm. Well, given that it'll need to tested against macOS 10.8 through 10.15 - plus other unforeseen issues, like incorrect compiler selection on certain macOS/Xcode combinations (per [Base PR 214](https://github.com/macports/macports-base/pull/214), which won't be in-play until the next MacPorts release... I'm really not feeling enthused about going through that right now.  :-O

mascguy: Updated description for OpenCV reconciliation tracking ticket, specifically calling out the OpenMP variant:

[opencv/opencv4: eliminate portfile duplication, via subport](https://trac.macports.org/ticket/62011)

mascguy: Ken, I don't suppose you'd be willing to give me a pass on this one, given that it's going to be tackled in short order?

mascguy: Okay, since I have to fire up my VMs for other testing anyway, I'll give your suggestion a try, and report back.

mascguy: Ken, is specifying `compiler.openmp_version` enough? Or do I also need to set `compiler.mpi`? (For the latter, I'd probably select `openmpi`... unless there are any other gotchas I need to be aware of...?)

mascguy: When `compiler.mpi` is specified too, that switches the compiler to `openmpi-clang90`. Which is arguably a good thing, assuming that's what's needed for true OpenMP support...?

However, there's a problem: I haven't installed `openmpi-clang90` anywhere yet. And given that it requires a source build, it takes a LONG time. Even on my physical macOS instances, with eight real CPU cores, it takes forever to build. On my VMs - which are limited to four vCPUs (I don't have Parallels Professional) - it's horrific.

In any case, let me know if `compiler.mpi` is required. If not, the changes are still feasible. But if it is required, I'm going to punt on this for now.

Ken, let me know your thoughts.

mascguy: I'm currently testing with `compiler.openmp_version` set to 4.0, and that does seem to do the trick. However, running through my entire suite of regression tests - which would involve rebuilding downstream ports, and testing their functionality - just isn't feasible right now.

So, if we want to add this change - with the caveat that it's not well-tested, beyond building and simple binary testing of opencv/opencv4 - we can do it.

How shall we proceed?

mascguy: Pushed the changes, to save time. We can always back them out if that's not the approach we want to take.

mascguy:
Given that the addition of `compiler.openmp_version` seems to work for variant `openmp`, let's commit these changes.

kencu:
that looks workable.

mascguy:
Thanks Ken!

PR #9788, opened 2021-01-22
ipcalc: support perl 5.32. Default perl to 5.28.

xeron:
#### Description

Support perl 5.32 and default perl to 5.28 for `ipcalc`.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
CLT 12.3.0.0.1.1607026830
Xcode N/A

###### Verification 
Have you

### DIFF
--- a/net/ipcalc/Portfile
+++ b/net/ipcalc/Portfile
@@ -5,7 +5,7 @@ PortGroup           perl5 1.0
 
 name                ipcalc
 version             0.41
-revision            5
+revision            6
 
 categories          net
 license             GPL-2+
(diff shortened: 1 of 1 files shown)


kencu:
MacPorts current default PERL is 5.28.

ryandesign:
True, but this port already uses 5.30. If there is a good reason why it needs > 5.28, I suppose we can switch it to 5.32. However doing so changes the files that get installed, so the revision must be increased.

xeron:
Sorry I wasn't aware that default perl is 5.28. I can change it to 5.28 and keep 5.30 and 5.32 as variants. Does it sound like a good solution?

kencu:
I think that is meant to be the policy. The idea is to not require a user have too many different perl installations, by having a generally-recommended one. 

No doubt we will be updating the recommended PERL from 5.28 to 5.30 in due course ... could be tomorrow, for all I know :>

ryandesign:
Thanks. You can know what the current default version of perl is by lookin at this file:

https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/perl5-1.0.tcl#L17

PR #9820, opened 2021-01-26
shtools: update to 4.8, and switch from Accelerate framework to openblas

MarkWieczorek:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H15
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/science/shtools/Portfile
+++ b/science/shtools/Portfile
@@ -19,11 +19,11 @@ long_description    SHTOOLS is a library of Fortran 95 software that can be \
                     LAPACK compatible libraries.
 homepage            https://shtools.github.io/SHTOOLS/
 
-github.setup        SHTOOLS SHTOOLS 4.7.1 v
-github.tarball_from   archive
-checksums           sha256  6ed2130eed7b741df3b19052b29b3324601403581c7b9afb015e0370e299a2bd \
-                    rmd160  ef9a2ca09e7d9703ebb037e223c9d200f0031e39 \
-                    size    35165654
+github.setup        SHTOOLS SHTOOLS 4.8 v
+github.tarball_from   releases
+checksums           sha256  c36fc86810017e544abbfb12f8ddf6f101a1ac8b89856a76d7d9801ffc8dac44 \
+                    rmd160  226bc9852f55eac632d0e43b657c285a10e37e51 \
+                    size    40408932
 
 use_configure       no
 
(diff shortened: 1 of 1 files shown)


science/shtools/Portfile
```diff
@@ -36,10 +36,10 @@ variant openmp description {Add OpenMP support} {
...
 test.target         fortran-tests-no-timing
-test.args           PREFIX=${prefix} F95="${prefix}/bin/gfortran-mp-10" LAPACK="-framework Accelerate" BLAS=""
+test.args           PREFIX=${prefix} F95="${prefix}/bin/gfortran-mp-10" LAPACK="-lopenblas" BLAS=""
 
 destroot.args       PREFIX=${prefix}
 
 depends_build       port:gcc10
 depends_test        port:gcc10
-depends_lib         port:fftw-3
+depends_lib         port:fftw-3 port:openblas
```
mf2k, line 45: Please put this new dependency on openblas on the next line.

mf2k (changes requested):

science/shtools/Portfile
MarkWieczorek: thanks! done...

mf2k (approved):
In the commit message, you should mention switching from the Accelerate framework to openblas.

kencu:
there are quite a few thing in this Portfile that are not being done in the usual MacPorts way...

MarkWieczorek:
Hi @kencu, I'd be happy to make changes. Could you let me know what the problems are?

kencu:
First off, this seems to work, so we can commit it.

(Aside - why are we changing from Accelerate to openBLAS?)

this port hard-codes gcc-10 and the compiler into the build, but we don't usually do it that way. For example, some OS versions can't use gcc-10, eg Apple Silicon, so it is broken for all new users. So what we do instead is use a compiler selection mechanism that tells MacPorts what the port needs, and then MacPorts works out the right compiler for the given software on a given OS.

There are built-in mechanisms in base for most things, and then a "compilers" portgroup that adds more compiler selection tools for things that need fortran, etc.

Also, there is a "linear algebra" portgroup that I have used occasionally that has functionality to work out certain features that seem relevant to this port.

MarkWieczorek:
For openblas, this is just to avoid using the default version that comes with macOS. The homebrew project advises to use openblas instead of the default, so I figured that I should do the same with macports. I seem to recall that there is some kind of bug in the default version.

As for the hardcoding gcc and and the gfortran compiler name: I will definitely look into how to do this better. If there is a particular project that you know of, that would help me to at least get started. This will likely take me a while, so either you can merge this PR, or people can continue using the last release (which unfortunately has the same problem...). Thanks for the comments!

PR #9823, opened 2021-01-27
opencv3/opencv4: add/update python subports; refine file layouts for consistency

mascguy:
#### Description

* opencv4: Refine python subports, which were just added via [merged PR 6261](https://github.com/macports/macports-ports/pull/6261). Also re-add variants 'gdal' and 'opencl', which were inadvertently removed.
* opencv3: Migrate python variants to subports, for consistency with opencv4.
* Add notes to both opencv3 and opencv4, pointing out change to users.
* gerbil: Update for opencv4 file layout change.
* py-pytorch: Update for opencv4 file layout change; add ccache support.
* py-imutils: Update for renamed opencv4 python subport.

**Note to Maintainers:**
* opencv3: Since this port is brand-new as of a week ago, the switch from python variants to subports is unlikely to impact any users.
* opencv4: Python binding support is still quite new, having been added within the past two weeks. So again, user impact is minimal.
* Nonetheless, notes have been added to both, mentioning the change.

###### Type(s)
- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.8.5 12F2560
Xcode 5.1.1 5B1008

macOS 10.10.5 14F2511
Xcode 6.4 6E35b

macOS 10.11.6 15G22010
Xcode 7.3.1 7D1014

macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

###### Verification 
Have you

### DIFF
--- a/graphics/opencv3/Portfile
+++ b/graphics/opencv3/Portfile
@@ -59,11 +59,14 @@ platform macosx {
 #
 # This is needed to support Python subports, whose names have no relation to
 # the actual physical file layout.
+#
+# NOTE: Presently, 'parent_subport_name' matches 'name'. HOWEVER, this will
+#   change once the various OpenCV-related ports are merged.
 #------------------------------------------------------------------------------
-set physical_subport_name \
+set parent_subport_name \
                     ${name}
 
-worksrcdir          ${physical_subport_name}-${version}
+worksrcdir          ${parent_subport_name}-${version}
 
 compiler.cxx_standard \
                     2011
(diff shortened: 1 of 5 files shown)


mascguy:
Maintainers: Since port 'py-pytorch' is included, there's a good change the CI jobs may timeout and fail. Hopefully not, but having just gone through this with another PR... there's a good chance that will happen.

kencu:
see the cmake PG options that are meant to be used instead of setting the args directly, such as 

https://github.com/macports/macports-ports/blob/a7327c3ee8d4e3243fcccadc85983c55c513d6db/_resources/port1.0/group/cmake-1.1.tcl#L45

mascguy:
> see the cmake PG options that are meant to be used instead of setting the args directly, such as
> 
> https://github.com/macports/macports-ports/blob/a7327c3ee8d4e3243fcccadc85983c55c513d6db/_resources/port1.0/group/cmake-1.1.tcl#L45

Unfortunately that only covers a few of the umpteen CMake flags needed for these ports. But that type of cleanup will be looked at when opencv3 and opencv4 are finally merged together.

mascguy:
Added use of `cmake` portgroup, to cleanup tracking ticket:

[opencv/opencv4: eliminate portfile duplication, via subports](https://trac.macports.org/ticket/62011)

essandess:
Would you also please add a commit for `py-imutils` that changes this dependency to `port:py${python.version}-opencv4`?

https://github.com/macports/macports-ports/blob/2b2006b205f221549d2a568fa0178893547d7952/python/py-imutils/Portfile#L34

graphics/opencv3/Portfile
```diff
@@ -53,9 +54,19 @@ platform macosx {
...
 }
 
-worksrcdir          ${subport}-${version}
+#------------------------------------------------------------------------------
+# Subport name used for staging files, etc.
+#
+# This is needed to support Python subports, whose names have no relation to
+# the actual physical file layout.
+#------------------------------------------------------------------------------
+set physical_subport_name \
```
essandess, line 63: Why not just use `${name}` everywhere? The subport names are `${subport}`, so this is extra variable name is unnecessary and not consistent with standard Portfile practice, to the best of my knowledge.

graphics/opencv4/Portfile
```diff
@@ -39,9 +39,20 @@ checksums           ${distname}${extract.suffix} \
                     sha256  03d9f7d777d8a62a5481934dbba2fd692b6cb43f1690a8546e45f9bdb1ee0863 \
                     size    90119002
 
+#------------------------------------------------------------------------------
+# Subport name used for staging files, etc.
+#
+# This is needed to support Python subports, whose names have no relation to
+# the actual physical file layout.
+#------------------------------------------------------------------------------
+set physical_subport_name \
```
essandess, line 48: Why not just use `${name}` everywhere? The subport names are `${subport}`, so this is extra variable name is unnecessary and not consistent with standard Portfile practice, to the best of my knowledge.

graphics/opencv3/Portfile
mascguy: Ports opencv3 and opencv4 will soon be combined into a port named 'opencv', with both converted to subports. And once that happens, `physical_subport_name` will disambiguate which one we're dealing with for the Python subports: It'll be set to either opencv3 or opencv4. And we won't be able to use `name`, as that will be `opencv`.

Make sense?

mascguy:
> Would you also please add a commit for `py-imutils` that changes this dependency to `port:py${python.version}-opencv4`?
> 
> https://github.com/macports/macports-ports/blob/2b2006b205f221549d2a568fa0178893547d7952/python/py-imutils/Portfile#L34

Ah, that might help too...  :-)

Done.

graphics/opencv3/Portfile
mascguy: Of note, I'm not enamored with the name of variable `physical_subport_name`, but couldn't decide on a better choice at the time.

Now that I've had a chance to sleep on it, perhaps it might be more logical to be called `parent_subport_name`? (As the python subports will essentially be logical child subports of opencv3 and opencv4, even though MacPorts technically doesn't support subports of subports. Lol)

mascguy:
Alas, the last CI build failed due to timeout, for port `py-pytorch`... as expected. No surprise though.

mascguy:
Ken, are you okay with deferring cleanup via use of portgroup `cmake`? As much as I'd love to tackle that ASAP, I'm hesitant to make any more changes to this PR...

kencu:
Well, here's the deal. The cmake 1.1. PortGroup adds a whole bunch of:
```
configure.pre_args-append -DCMAKE_THINGAMABOB=MacPorts-default-thingamabob
```

When you also add:
```
configure.args-append -DCMAKE_THINGAMABOB=my-preferred-thingamabob
```

the cmake 1.1 PortGroup has no clue you did that, so it adds it's own things as well.

Now -- the cmake 1.1 PortGroup puts it thingamabobs in configure.pre_args, and you put your in configure.args, so if we're lucky, yours come after, and trash the earlier ones added by the PortGroup. 

But this is just plain ugly, right?

SO I would prefer if you take 5 minutes, and for args that the cmake PortGroup controls with those options I linked to, just set them that way instead, which is just better, cleaner, shorter, nicer, etc.

If there is also some additional arg you really want to set that there is no cmake PortGroup option for, then add those by hand in the configure.args, and you won't be competing with the cmake portgroup.

mascguy:
> SO I would prefer if you take 5 minutes, and for args that the cmake PortGroup controls with those options I linked to, just set them that way instead, which is just better, cleaner, shorter, nicer, etc.

No problem. I'll spend some time on that, and report back when it's ready for review.

mascguy:
Ken, cmake-related changes completed. While it only eliminated a few of the configure arguments, it also gave me the opportunity to eliminate path copy-paste... which was long-overdue anyway.

Take a look, and let me know what you think.

graphics/opencv3/Portfile
```diff
@@ -104,40 +118,62 @@ post-patch {
...
+                    -DOPENCV_3P_LIB_INSTALL_PATH=${opencv_install_lib}/3rdparty \
+                    -DOPENCV_LIB_ARCHIVE_INSTALL_PATH=${opencv_install_lib}/static \
+                    -DCMAKE_BINARY_DIR=${opencv_install_prefix}/bin \
+                    -DCMAKE_INSTALL_LIBEXECDIR=${opencv_install_prefix} \
+                    -DCMAKE_INSTALL_BINDIR=${opencv_install_prefix}/bin \
+                    -DCMAKE_INSTALL_SBINDIR=${opencv_install_prefix}/sbin \
+                    -DCMAKE_INSTALL_SHAREDIR=${opencv_install_prefix}/share \
+                    -DCMAKE_INSTALL_INCLUDEDIR=${opencv_install_include} \
+                    -DCMAKE_INSTALL_LIBDIR=${opencv_install_lib} \
+                    -DCMAKE_INSTALL_NAME_DIR=${opencv_install_lib} \
```
mascguy, line 176: Note to Reviewers:

The cmake portgroup does not expose an override for `CMAKE_INSTALL_NAME_DIR`, and it's hard-coded to `${cmake.install_prefix}/lib`. So we have no choice but to explicitly specify it ourselves.

graphics/opencv4/Portfile
```diff
@@ -86,24 +80,62 @@ depends_lib-append  \
...
+                    -DOPENCV_3P_LIB_INSTALL_PATH=${opencv_install_lib}/3rdparty \
+                    -DOPENCV_LIB_ARCHIVE_INSTALL_PATH=${opencv_install_lib}/static \
+                    -DCMAKE_BINARY_DIR=${opencv_install_prefix}/bin \
+                    -DCMAKE_INSTALL_LIBEXECDIR=${opencv_install_prefix} \
+                    -DCMAKE_INSTALL_BINDIR=${opencv_install_prefix}/bin \
+                    -DCMAKE_INSTALL_SBINDIR=${opencv_install_prefix}/sbin \
+                    -DCMAKE_INSTALL_SHAREDIR=${opencv_install_prefix}/share \
+                    -DCMAKE_INSTALL_INCLUDEDIR=${opencv_install_include} \
+                    -DCMAKE_INSTALL_LIBDIR=${opencv_install_lib} \
+                    -DCMAKE_INSTALL_NAME_DIR=${opencv_install_lib} \
```
mascguy, line 138: Note to Reviewers:

The cmake portgroup does not expose an override for `CMAKE_INSTALL_NAME_DIR`, and it's hard-coded to `${cmake.install_prefix}/lib`. So we have no choice but to explicitly specify it ourselves.

kencu: Someday I have to look into this more. I think this does what you want, even though it looks like it is ```@rpath``` related:
```
cmake.install_rpath
```
but ... that is not for today :>

thanks!

kencu: I think you only have to set this one like this:
```
cmake.install_prefix  ${prefix}/libexec/${parent_subport_name}
```

and 100 of your other options could all be removed, like the libdir, includedir, module dir, bin dir, libexecdir, etc, etc, etc, etc

mascguy: At least some probably can, yep! But can that wait for the next release...?

mascguy:
Thanks Ken!

PR #9833, opened 2021-01-27
mpich: Use legacysupport PG for _SC_PHYS_PAGES

eborisch:
Hopefully fixes https://trac.macports.org/ticket/62160

("Hopefully" as I don't have a legacy system to test.)

### DIFF
--- a/science/mpich/Portfile
+++ b/science/mpich/Portfile
@@ -4,6 +4,7 @@ PortSystem          1.0
 PortGroup           select 1.0
 PortGroup           compilers 1.0
 PortGroup           muniversal 1.0
+PortGroup           legacysupport 1.1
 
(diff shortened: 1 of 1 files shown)


eborisch:
Not planning to revbump, as it _should_ only impact versions that were not building. Or do we need a revbump to get the buildbots to retry those failed builds?

kencu:
> Or do we need a revbump to get the buildbots to retry those failed builds?

The buildbots will retry failed builds if the Portfile is hit with a commit.

I really have no idea if this will actually work, though. gcc won't build with the legacysupport PG added, as gcc does some deep-in-the-guts header tweaking, and it expects standard headers to be found. Somebody should test this.

kencu:
user in ticket indicates this works. Good enough for me.

PR #9834, opened 2021-01-27
cmake: Fix universal build on BigSur

kencu:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69 Intel
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/cmake-1.1.tcl
+++ b/_resources/port1.0/group/cmake-1.1.tcl
@@ -85,7 +85,7 @@ default destroot.target             install/fast
 # make sure cmake is available:
 # can use cmake or cmake-devel; default to cmake if not installed
 depends_build-append                path:bin/cmake:cmake
-
+depends_skip_archcheck-append       cmake
 
 proc cmake::rpath_flags {} {
     global prefix
--- a/devel/cmake/Portfile
+++ b/devel/cmake/Portfile
@@ -162,7 +162,7 @@ if {${subport} eq ${name}} {
     configure.universal_args
 
     # CMake's configure script doesn't recognize `--host`.
-    array set merger_host {i386 {} x86_64 {} ppc {} ppc64 {}}
+    array set merger_host {i386 {} x86_64 {} ppc {} ppc64 {} arm64 {}}
 
     platform darwin 8 {
         configure.ldflags-append -Wl,-framework -Wl,ApplicationServices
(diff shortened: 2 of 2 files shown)


kencu:
With these commits as they are, you can do this on BigSur Intel:
```
sudo port install fish +universal
```
and you get this:
```
% port -v installed fish 
  fish @3.1.2_0+universal (active) platform='darwin 20' archs='arm64 x86_64' date='2021-01-27T15:31:17-0800'

% file /opt/local/bin/fish
/opt/local/bin/fish: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
/opt/local/bin/fish (for architecture x86_64):	Mach-O 64-bit executable x86_64
/opt/local/bin/fish (for architecture arm64):	Mach-O 64-bit executable arm64
```
and 
```
% port -v installed cmake
  cmake @3.19.3_0+universal (active) platform='darwin 20' archs='x86_64' date='2021-01-27T15:29:14-0800'

% file /opt/local/bin/cmake
/opt/local/bin/cmake: Mach-O 64-bit executable x86_64
```


So there is a bit of lie there, that cmake is +universal, but +universal only = one arch.

On the other hand, using the ```merger must run binaries``` approach, cmake when built universal **on Intel** will say it is +universal with arm64/x86_64, but in the end it will only be x86_64. So that is a bigger lie, I submit.

ryandesign:
If cmake already says `installs_libs no`, adding `depends_skip_archcheck-append cmake` to ports that depend on cmake should be entirely unnecessary and redundant.

> So there is a bit of lie there, that cmake is +universal, but +universal only represents one arch.

Yeah, don't do that.

If the cmake build system builds a bootstrap cmake program that it then uses to build itself, that bootstrap cmake program should be built without arch flags.

kencu:
> If cmake already says `installs_libs no`, adding `depends_skip_archcheck-append cmake` to ports that depend on cmake should be entirely unnecessary and redundant.

> 
> > So there is a bit of lie there, that cmake is +universal, but +universal only represents one arch.
> 
> Yeah, don't do that.

I think we have to -- so far, there is no choice other than to disable the universal build completely on BigSur intel, which would be a terrible waste and disappointing to our users I believe, who want to build universal arm64/x86_64 on BigSur Intel, but will never be able to.
> 
> If the cmake build system builds a bootstrap cmake program that it then uses to build itself, that bootstrap cmake program should be built without arch flags.

Using the muniveral PG, on Intel, it builds one arm64, but can't run it, and errors out. There is no solution to this so far, other than not using the muniversal PortGroup (not an option) or deep hacking of some kind.

Any ideas?

kencu:
Or the even sillier "merger_must_run_binaries", which as I said, claims to build universal for arm64/x86_64, but does not, and only builds the x86_64 part. That is an even bigger lie.

These are just failings of base -- when building software, there is no need for cmake to ever be universal, and yet, base insists.

kencu:
Of course, cmake has been playing "universal" tricks for years. If you ask it to build universal on 10.5, which should be PPC/i386, it switches itself to PPC/PPC64, which you have absolutely no interest in. It has never built cross-arch before, and still won't now.

This is just to quiet down the need to satisfy "+universal", even though the +universal you get is not what you asked for, and not what you need.

PR #9835, opened 2021-01-28
c-ares: Add cmake build and .cmake files

essandess:
c-ares: Add cmake build and .cmake files
* Fixes https://trac.macports.org/ticket/62167#ticket
* Related: https://github.com/macports/macports-ports/pull/7791

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/grpc/Portfile
+++ b/devel/grpc/Portfile
@@ -4,7 +4,7 @@ PortSystem                                     1.0
 PortGroup       github                         1.0
 
 github.setup    grpc grpc 1.22.0 v
-revision        2
+revision        3
 categories      devel
 maintainers     nomaintainer
 license         Apache-2
--- a/games/bzflag/Portfile
+++ b/games/bzflag/Portfile
@@ -5,7 +5,7 @@ PortGroup           app 1.0
 
 name                bzflag
 version             2.4.18
-revision            1
+revision            2
 categories          games
 platforms           darwin
 maintainers         {allejo.io:me @allejo} openmaintainer
--- a/net/aria2/Portfile
+++ b/net/aria2/Portfile
@@ -5,7 +5,7 @@ PortGroup           compiler_blacklist_versions 1.0
 PortGroup           github 1.0
 
 github.setup        aria2 aria2 1.35.0 release-
-revision            2
+revision            3
 github.tarball_from releases
 categories          net
 platforms           darwin
--- a/net/c-ares/Portfile
+++ b/net/c-ares/Portfile
@@ -1,29 +1,37 @@
-PortSystem 1.0
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
-name             c-ares
-version          1.17.1
-categories       net
-maintainers      {toby @tobypeterson}
-description      asynchronous DNS library
-long_description \
-    c-ares is a C library that performs DNS requests and \
-    name resolves asynchronously. c-ares is a fork of the \
-    library named 'ares', written by Greg Hudson at MIT.
-homepage         https://c-ares.haxx.se/
-platforms        darwin
-license          MIT
-master_sites     ${homepage}download/
-checksums        rmd160 4e7f83785735951f2c22145743c238d78332f4cf \
-                 sha256 d73dd0f6de824afd407ce10750ea081af47eba52b8a6cb307d220131ad93fc40 \
-                 size   1518701
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           github 1.0
 
-configure.args   --mandir=${prefix}/share/man \
-                 --enable-shared \
-                 --disable-silent-rules
+github.setup        c-ares c-ares 1.17.1 cares-
+set version_        [join [split ${github.version} .] _]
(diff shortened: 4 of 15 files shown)


tobypeterson:
Would you like to maintain this port? 😅

tobypeterson:
Also, can you ensure that this still builds when gtest is installed? I just pushed a commit to try to fix that issue.

essandess:
> Would you like to maintain this port? 😅

I'll update it when I see that's necessary, but I'm unable to commit to port maintenance.

tobypeterson:
This does appear to build with gtest installed, but the cmake build doesn't seem to build the tests which were failing.

Note, since the version didn't change, will probably need to set dist_subdir - by default, it doesn't include the revision, so it'll conflict with existing distfiles.

tobypeterson:
I added `configure.args      -DCARES_BUILD_TESTS=ON` and verified that the gtest-related issue doesn't occur. So my only real feedback (unless you want to enable tests) is the `dist_subdir` issue.

essandess:
Thanks for the great and prompt feedback! I added `dist_subdir` and tests, and am able to run these successfully.

tobypeterson:
Still a slight tweak, I think. dist_subdir should be something starting with `${name}/` so that the files at least go inside the same directory

essandess:
Thanks, changed. No  special reason to use the git tarball, except simplifying and standardizing things like a specialized livecheck.

tobypeterson:
Ugh, I just noticed that this changes the library's compat version from 7 to 2.

essandess:
> Ugh, I just noticed that this changes the library's compat version from 7 to 2.

I know just enough to be dangerous about this stuff. Please advise.

tobypeterson:
Just means that we _may_ need to chase the update with revbumps on dependents. I'm not sure how the rev-upgrade functionality works (tries to automatically rebuild to handle changes like this).

essandess:
Is this an issue that we should report upstream? I honestly don’t know enough about this to characterize the issue and post it.

> Just means that we _may_ need to chase the update with revbumps on dependents. I'm not sure how the rev-upgrade functionality works (tries to automatically rebuild to handle changes like this).
> 
> But it's also concerning, it's a major inconsistency between the project's configure/make build system and the cmake files.

essandess:
There's also this issue with `livecheck` because [c-ares/c-ares](https://github.com/c-ares/c-ares) uses tags like `1_17_1`:
```
port livecheck re2 c-ares
re2 seems to have been updated (port version: 2021-02-01, new version: 2021-02-02)
Error: livecheck failed for c-ares: extracted version '1_17_1' is older than livecheck.version '1.17.1'
```
If anyone knows the `livecheck` fix, please post.

tobypeterson:
> Is this an issue that we should report upstream? I honestly don’t know enough about this to characterize the issue and post it.

Pretty complicated issue. It's not really a bug in c-ares, it's a long-standing bug in GNU libtool; I'm doubtful that the maintainers of GNU libtool agree with me. Arguably the cmake build does the right thing.

To be safe, it probably makes sense to bump the revision for the relatively few ports that depend on c-ares.

tobypeterson:
One thing you can do is add `github.tarball_from releases` (_after_ git.branch, it's a little finicky). Then you can revert to the previous checksums and remove `dist_subdir`. For some reason, they are generating two tarballs per release, what you're currently fetching is github's automatically-generated one, not the real release tarball.

As for the livecheck issue; no immediately obvious solution stands out to me. Could use the previous livecheck, or maybe try extracting from `${github.homepage}/releases` instead of `${github.homepage}/tags`

kencu:
As opposed to re2, #9836, it looks like c-ares has indeed done all the legwork to make it properly build with cmake and provide all the needed tweaks <https://github.com/c-ares/c-ares/blob/master/INSTALL.md>

essandess:
Fixed both the tarball and livecheck issues.

tobypeterson:
Sorry, dropped the ball on this. Can you prepare another PR to bump the revision for the various c-ares dependents? I think, to be safe, we should bump them all to ensure that they rebuild against this.

essandess:
Should this another commit to this PR? Can it be a single commit? Or one commit per port? How does one determine all the ports? `port echo depends:c-ares` looks like it only shows the ones I have installed.

> Sorry, dropped the ball on this. Can you prepare another PR to bump the revision for the various c-ares dependents? I think, to be safe, we should bump them all to ensure that they rebuild against this. This change looks good now.

essandess:
What about adding a single commit to this PR that revamps these ports in a single commit?

```bash
find ${prefix}/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/ -type f -name Portfile -exec fgrep -l port:c-ares {} ';' | sed -E -e 's|^.+/([^/]+)/Portfile|\1|'
grpc
mosquitto
wireshark2
wireshark3
aria2
etherape
wireshark22
curl
libexosip2
wireshark30
py-gevent
py-grpcio
nghttp2
bzflag
```

essandess:
@tobypeterson ping I believe this is ready to go.

tobypeterson:
CI doesn't seem happy...

essandess:
The number of builds in these two commits almost certainly broke it, so I wouldn’t view its unhappiness as diagnostic.

> CI doesn't seem happy..

tobypeterson:
Alright here we go!

PR #9837, opened 2021-01-28
cmake: disable universal on BigSur+

kencu:
there is no real need for cmake to build universal
it is a tool that works well in native arch
and it sets installs_libs no

cmake has many deps, all of which need to be universal
if cmake is universal

cmake does not presently build universal on BigSur Intel
and there is no compelling reason for it to do so

disable the universal build on BigSur+  Intel

allows the many ports that build with cmake to build
universal without requiring cmake to be universal

This is a much better idea than the previous.

### DIFF
--- a/devel/cmake/Portfile
+++ b/devel/cmake/Portfile
@@ -1,7 +1,14 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           muniversal 1.0
+
+# cmake does not build universal on BigSur Intel
+if { ${os.platform} eq "darwin" && ${os.major} >= 20 && ${build_arch} eq "x86_64"} {
+	universal_variant no
+} else {
+	PortGroup           muniversal 1.0
+}
+
 PortGroup           xcodeversion 1.0
 PortGroup           gitlab 1.0
 PortGroup           legacysupport 1.1


kencu:
even though cmake is not universal, it works just fine to build universal ports that require cmake to build:
```
% port -v installed fish llvm-10
  fish @3.1.2_0+universal (active) platform='darwin 20' archs='arm64 x86_64' date='2021-01-27T17:52:12-0800'
  llvm-10 @10.0.1_1+universal (active) platform='darwin 20' archs='arm64 x86_64' date='2021-01-27T16:58:36-0800'
```

kencu:
I have found that for ```installs_libs no``` to really function as intended, disabling the universal variant is needed.

Otherwise, attempts to install ports that require cmake as +universal still tries to build cmake as a +universal port, even though that is not needed and ```installs_libs no``` has been set.

jmroot:
What's the reason for the build failure with universal? Is there a ticket or log somewhere?

kencu:
You guys don't have any BigSur systems, I guess.

I documented it all in the previous PR I closed <https://github.com/macports/macports-ports/pull/9834> but in essence, there are two issues:

1. by making cmake build +universal, you force all of cmake's copious deps to be built universal as well, and that is troublesome at present, and unnecessary.

2. during the build of cmake with the muniversal PG, it builds it's bootstrap cmake as arm64, then fails when it tries to run it on Intel.

There is no need to have cmake build universal. We marked it as "installs_libs no" so it wouldn't build universal, in fact.

(If you ask me, all cmake builds should be non-universal, as it is a PITA and takes a long time to build, with all it's deps, but i have no air for that argument.)

jmroot:
So sounds like a general fix should be made to the muniversal portgroup then? Also, this means it's impossible to cross-compile cmake?

kencu:
cmake bootstraps with a hefty shell script and picks up the flags from the environment set by the muniversal PG there. That bootstrap shell script would have to be modified to strip the arch flags after they are picked up from the environment, before setting them to be used during the cmake bootstrap build.

# If exist compiler flags, set them
cmake_c_flags=${CFLAGS}
cmake_cxx_flags=${CXXFLAGS}
cmake_ld_flags=${LDFLAGS}

If that is not too hard todo, there could be a way there.


Or we could use the existing cmake-bootstrap port I made instead of bootstrapping cmake with it's bootstrap script...and then cmake builds like any other cmake project, universal or any arch you want.

So there are ways...just a question of need, desired outcome, complicating the existing situation, and so forth.

michaelld:
"the previous" being #9834 , yes? I have no objections to either way, but I do think this version makes more sense so long as it works in the sense of still allowing ports that use cmake to build as +universal ... meaning that `port` gets that this port doesn't provide `+universal` but that's OK.

kencu:
If someone knows the regex incantation to use in the bash bootstrap script to strip the ```-arch XYZ``` flags out of the ${CFLAGS}, etc, I'm happy to test that.

I found this in libiconv, but I couldn't spot anywhere in the ports tree where there was a sed command that did a generic stripping.

```
configure.cmd-prepend   mv Makefile.devel Makefile.devel.orig \
                        && sed -E {"s|^(CFLAGS *=).*|\1 $CFLAGS|"} < Makefile.devel.orig \
                        | sed -E {"/^CFLAGS *=/s| -arch +[^ ]+||"} > Makefile.devel &&
```

what we would have to do is take this bit of bash/zsh script
```
cmake_c_flags=${CFLAGS}
```
and make ```cmake_c_flags``` have everything except and ```-arch XYZ``` flags.

If that is doable, then the bootstrap script could be patched to that, and I'd be happy to test it.

kencu:
> so long as it works in the sense of still allowing ports that use cmake to build as +universal

It 100% does do that, and in a lean way

kencu:
here a reinplace incantation that strips arch flags:
```
reinplace -E {s|-arch [a-z0-9_]+||g}
```
perhaps we/I can mod that to something that can be used in a bash script?

kencu:
perhaps something like this?
```
cmake_c_flags=`echo ${CFLAGS} | sed 's|-arch [a-z0-9_]+||g'`
```

jmroot:
With #9845 you should be able to just set `merger_must_run_binaries yes`.

kencu:
> With #9845 you should be able to just set `merger_must_run_binaries yes`.

can try that. previously when I added that, I got this:
```
Installed:
cmake @3.19.3_0+universal (active) platform='darwin 20' archs='arm64 x86_64'

but:
$ file /opt/local/bin/cmake
/opt/local/bin/cmake: Mach-O 64-bit executable x86_64
```

kencu:
as per Michael's approval, let's get this much done so people aren't DOA building anything universal on BigSur Intel.

If I can fix it better as time goes by, I will.

ryandesign:
> perhaps something like this?
> 
> ```
> cmake_c_flags=`echo ${CFLAGS} | sed 's/-arch [a-z0-9_]+//g'`
> ```

Looks reasonable, except that you would need `sed -E` to get extended regular expression syntax (to make the `+` operator available).

PR #9843, opened 2021-01-28
cargo portgroup: use [cargo.rust_platform] for the executable's destroot path

herbygillot:
Cargo will build executables into the ./target directory under a subdir
that matches the host's current OS and architecture.  This can be
retrieved using [cargo.rust_platform].  We update the destroot example
provided by the Cargo portgroup that is shown when a Portfile author
forgets to include a destroot block to include this so that the correct
thing happens regardless of the build platform.

### DIFF
--- a/_resources/port1.0/group/cargo-1.0.tcl
+++ b/_resources/port1.0/group/cargo-1.0.tcl
@@ -29,7 +29,7 @@ destroot {
     ui_msg "Here is an example destroot phase:"
     ui_msg
     ui_msg "destroot {"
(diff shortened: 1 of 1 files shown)


harens:
This seems to run into https://trac.macports.org/ticket/62071, where the tests fail if no ports have been modified.

Maybe update a port with the example to show it works? (and for the tests to pass)

herbygillot:
This change simply modifies a string being printed to standard out. That what `ui_msg` does: you can use it to print whatever arbitrary message you want to standard out.  So tests failing don't mean anything here, really.  We're changing the suggestion printed to a Portfile author if he wrote a Portfile that uses the `cargo` portgroup without specifying a `destroot` section.

PR #9850, opened 2021-01-29
pango: Fix +universal

Gcenx:
#### Description

Currently `pango` +universal only builds a single arch. The following fix was taken from another Port.

Here's what happens currently when pango +universal is required.
```
:info:build ld: warning: ignoring file /opt/local/lib/libpangocairo-1.0.dylib, file was built for x86_64 which is not the architecture being linked (i386): /opt/local/lib/libpangocairo-1.0.dylib
:info:build ld: warning: ignoring file /opt/local/lib/libpango-1.0.dylib, file was built for x86_64 which is not the architecture being linked (i386): /opt/local/lib/libpango-1.0.dylib
```

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.14.6 18G7016
Xcode 11.3.1 11C505

###### Verification 
Have you

### DIFF
--- a/x11/pango/Portfile
+++ b/x11/pango/Portfile
@@ -3,6 +3,7 @@
 PortSystem              1.0
 PortGroup               active_variants 1.1
 PortGroup               gobject_introspection 1.0
+PortGroup               muniversal 1.0
 
 # Please keep the pango and pango-devel ports as similar as possible.
 


ryandesign:
> Currently `pango` +universal only builds a single arch.

That certainly has never been my experience. For years I have had pango and most other ports I use installed with the +universal variant with no problems, but on macOS 10.13 and earlier. I see you are on macOS 10.14, on which it is not possible to build universal.

Gcenx:
I'm forcing the 10.13 SDK and deployment target to 10.13 so I can build +universal (mostly for wine)

ryandesign:
No doubt. But I don't understand why your proposed fix would be necessary, or if it is, why it would be needed just for pango and not also for all other ports.

ryandesign:
The code that you've copy-pasted into the pango portfile should already be there thanks to the gobject_introspection 1.0 portgroup that's included at the top of the portfile. (See https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/gobject_introspection-1.0.tcl) 

It has lately been copy-pasted into many ports because many ports have switched to the meson build system and the gobject_introspection 1.0 portgroup doesn't support meson yet. I have asked that the gobject_introspection 1.0 portgroup be enhanced to support meson rather than copy-paste this code endlessly, but so far this has not been done. This version of pango does not use meson so it can use the portgroup.

ryandesign:
So the only change you've made that I guess is relevant to your issue is that you've used the muniversal portgroup. I would like to see a log showing how the build failed to produce a universal binary when that portgroup is not being used.

Gcenx:
Here is the [main.log](https://github.com/macports/macports-ports/files/5891526/main.log)

And here is the output from file
```
/opt/local/lib/libpango-1.0.0.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

Ah yeah that's true it lightly only needs the muniveral group

Gcenx:
Tested using only muniversal and it still only provides x86_64 arch dylib

Gcenx:
Here is the output when using only muniversal group

```
sh-3.2# port install pango +univeral
--->  Computing dependencies for pango
--->  Fetching distfiles for pango
--->  Verifying checksums for pango
--->  Extracting pango
--->  Applying patches to pango
--->  Configuring pango
--->  Building pango
--->  Staging pango into destroot
--->  Installing pango @1.42.4_2+quartz
--->  Activating pango @1.42.4_2+quartz
--->  Cleaning pango
--->  Found 2 broken files, matching files to ports      
--->  Found 1 broken port, determining rebuild order
You can always run 'port rev-upgrade' again to fix errors.
The following ports will be rebuilt: gstreamer1-gst-plugins-base @1.18.3+ogg+universal
Continue? [Y/n]: 
```

Gcenx:
Having muniversal group is enough to fix this issue on my end.

kencu:
@gcenx -- before we can commit this, we need to understand why, after 20 years without troubles, you now (uniquely) have this issue whereas it never existed before. We accept that is what happenned to you, but we need to understand how.

So @ryandesign wants a log showing the build of the universal pango that winds up with only one arch in it to see what went wrong.

Gcenx:
@kencu completely understandable and I attached the requested main.log above showing what happens and the results from using `file` 

What’s strange is this has only happened with my recent rebuild (swapped back form using SDKROOT=“/“ to 10.13.SDK as I need a refreshed runtime for Wineskin. @gverm provided me with a similar log early yesterday that shown gstreamer1-base failing due to pango being only x86_64.

The prior rebuild a couple of weeks ago don’t have any issues.

kencu:
the log you posted looks like it has a build of pango with only one arch requested, then a build of gstreamer-something, then other builds in it...

Something here is not quite right, and although adding muniversal fixes it for you, that may not in the end be the proper fix for the general case...

could you just clean any partial builds, remove the muniversal PG from pango, and run a

sudo port -v destroot pango +universal

and post the log of only that please?

gverm:
Here's the log I got when running `sudo port install gstreamer1-gst-plugins-base +universal -x11`, which used to work fine until recently. I'm also running 10.14 and forcing the use of the 10.13 SDK.
[main.log](https://github.com/macports/macports-ports/files/5894809/main.log)

Gcenx:
@ken sure but it won’t be until I’m home from work I just got here 26mins ago.

gverm:
@kencu Here is a log from running `sudo port -v destroot pango +universal -x11`
[log.txt](https://github.com/macports/macports-ports/files/5895916/log.txt)

kencu:
It looks like MacPorts is not even trying to build it +universal.

Gcenx:
@kencu sorry didn't get around to doing this but here [main.log](https://github.com/macports/macports-ports/files/5898721/main.log) when using the current macports-port version.

And this one [main.log](https://github.com/macports/macports-ports/files/5898727/main.log) using this pull-request.

kencu:
OK, I can see it is trying to build it universal at least. 

At various times the -arch i386 is added, but at other times (like with libtool, in particular) it is not added.

libtool could need regenerating, I think....

There are also times where the "i386 architecture is deprecated" warning is showing up, and sometimes, as you know, that causes builds to error when they shouldn't. That you get around with our usual friend, ```-Wl,-w``` (IIRC)  or by using the ld64-latest linker where I disabled it.

ryandesign:
> At various times the -arch i386 is added,

...when using CXX

> but at other times (like with libtool, in particular) it is not added.

...when using CC

Suggesting there may be a problem with CFLAGS, but I don't see it yet.

ryandesign:
Some problem with CFLAGS continues to be suggested by `-pipe` and `-Os`, which should accompany the `-arch` flags, also being present only when compiling with CXX and not being present when compiling with CC.

kencu:
This used to be one of the old-time [universal fixes](https://trac.macports.org/wiki/UniversalDevelopment) recommended:

```
Adding universal build flags to the libtool command e.g. reinplace "s|CC -dynamiclib|CC -dynamiclib ${configure.universal_ldflags}|g" ${worksrcpath}/libtool
```

Gcenx:
@ryandesign & @kencu did ether of you get chance to figure out what’s actually happening here since this version broken the older behavior?

I’ve added the current reversion to my overlay so myself and @gverm aren’t affected in the short term.

kencu:
Haven't worked on it yet. One thing I do know is that on my own systems, pango installs +universal as it is:
```
$ port -v installed pango
  pango @1.42.4_2+quartz+universal+x11 (active) platform='darwin 10' archs='i386 x86_64' date='2020-04-10T13:13:23-0700'
```

Gcenx:
Even on Mojave?, it’s very strange that it only started to have issues building very recently.

kencu:
I didn't try Mojave yet...I thought I would try it on BigSur Intel next -- that's a good, timely test.

I'm just working through upgrading one of my slightly older, no-longer-supported MacBookPros to allow it to run BigSur right now, in fact, -- it's booting on the other system as I'm typing this.

Gcenx:
I’m kinda stuck on Mojave on the Intel MacBook Pro unless Macports can be forced to build 32/64 universal.

Didn’t test +universal on the M1 Mac Mini as I know a lot of the dependencies I need don’t currently build for arm64 yet.

kencu:
I have most of them fixed -- at least all the ones I tried to fix. Been working on universal Intel/arm64 lately, and esp building universal on BigSur Intel, which is harder. 

Let me know if you need something. Fixing them is usually pretty simple, except for the ones that use the guts of the kernel headers, and similar, like the qt5 webkit fork... that's hard.

kencu:
I did get BigSur Intel to build +universal all the way through to xorg-server yesterday, which is a ton of deps...

Gcenx:
Nice, hum need to see if I can move stuff of my 512SSD so I can do an external boot then as the 256SSD in the M1 Mac Mini is mostly filled.

Gcenx:
@kencu I notice your showing +x11 where I don’t, this shouldn’t be relevant but still might be worth mentioning.

Just tried this again this morning after running `port selfupdate` to ensure nothing was outdated still failed to built with muniversal.

kencu:
@Gcenx -- IMHO you might as well close this. It won't be committed.

Gcenx:
@kencu the last comment was a typo it still failed without muniversal.

If anything this will become another permanent part of my Portoverlay.

kencu:
I'll check back in three months and see how this is doing :>

seriously, not worth the PR -- just adding the 'muniversal PortGroup is too trivial to let this fester here for months and months with no possibility of resolution.

The only way you are going to get this committed would be to:

1. have somebody see I said it wouldn't be committed, and therefore commit it (that actually could happen :> ).
2. explain why you can't build it universal without the muninversal PortGroup when everyone else can. And that, after a month, you can't do. 

Just eat it. I have closed literally dozens and dozens of perfectly wonderful PRs because I could see that for whatever reason, they were just never going to get committed. Who cares? It's just a trivial one-liner PR, not your reputation on the line.

Gcenx:
> @Gcenx -- IMHO you might as well close this. It won't be committed.

This no longer matters now since glib2 is updated, the new version uses meson.

Gcenx:
@ryandesign & @kencu well I found the cause of this and all other Portfiles that use gobject_introspection-1.0.tcl that now fail when building for me as +universal

https://github.com/macports/macports-ports/commit/90ac6621ecbbe96695afff100490e75bc9b0a4ec

Reverting that commit fixes +universal builds, no matter that is always sets sdk_root to "/"

jmroot:
> Reverting that commit fixes +universal builds, no matter what that is always sets sdk_root to "/"

Then `configure.sdkroot` must be empty, which is of course going to be a problem if you are trying to use an SDK.

jmroot:
I looked at the attached logs and they all show MacOSX10.13.sdk being used as expected. So setting sdk_root to / can't be the problem.

Gcenx:
> I looked at the attached logs and they all show MacOSX10.13.sdk being used as expected. So setting sdk_root to / can't be the problem.

The resulting binaries still end up being x86_64 only, revering that commit or adding muniversal I get the expected universal dylib.

kencu:
It must have something to do with this:

https://github.com/macports/macports-ports/blob/90ac6621ecbbe96695afff100490e75bc9b0a4ec/_resources/port1.0/group/gobject_introspection-1.0.tcl#L58

The gobject-introspection PG specifically does not add multiple arch flags to CC during builds that do not use the muniversal portgroup.

Gcenx:
Could the modifications to CFLAGS and LDFLAGS done in the commit be causing some other required options being dropped causing the breakage?

kencu:
Indeed, coming up with the elegant, all-encompassing fix is difficult sometimes. Fix one thing, break something else, is usually the rule of the day.

But we can see why you had to add the muniversal PG at least. As above, I did not https://github.com/macports/macports-ports/pull/9850#issuecomment-776903436 but things may have changed there.

ryandesign:
Let's please discuss this issue in its own ticket (https://trac.macports.org/ticket/59078) not in this closed PR.

ryandesign:
> This no longer matters now since glib2 is updated, the new version uses meson.

glib2-devel was updated to a new version that uses meson. glib2 has not yet been updated.

Gcenx:
> Let's please discuss this issue in its own ticket (https://trac.macports.org/ticket/59078) not in this closed PR.

I’d received the email for the new ticket you opened already, thanks for the cc.

> glib2-devel was updated to a new version that uses meson. glib2 has not yet been updated.

I’m not using `glib2` Portfile from Macports-port but my own updated Portfile, the latest version of pango required a newer version of glib2 and modification to harfbuzz.

kencu:
> Let's please discuss this issue in its own ticket (https://trac.macports.org/ticket/59078) not in this closed PR.

Well that ticked was closed two months ago, so we can hardly be crucified for conversing here.

But now that you opened https://trac.macports.org/ticket/62410 you can take it from there.

ryandesign:
Sorry, I must've pasted the wrong link. Yes, I meant https://trac.macports.org/ticket/62410.

PR #9853, opened 2021-01-29
libgcrypt: update to 1.9.1

l2dy:
SECURITY: https://lists.gnupg.org/pipermail/gnupg-announce/2021q1/000455.html

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [x] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you


michaellass:
This fails to build on my system because the hardware architecture is not detected during `configure`.

It works when removing the line `configure.args  --disable-asm` but I assume that was added for a reason at some point in time?

l2dy:
Upstream said they were fixing some build problems in this release...

@Schamschula Do you know how to fix this? Upstream advice is to stop using 1.9.0, so we may consider reverting to 1.8.x.

michaellass:
Looks like we always compiled with this flag: 6e1bd1632e29c65c98d90fae70d86a677691a96d

I suggest that we just remove it as it breaks the build and its only function seems to be to reduce the amount of things that are built.

Schamschula:
First of all, I got the same email from Werner when I woke up this morning. The MacPorts version of 1.9.0 had been patched for this. Version 1.9.1 includes this patch, which is removed by this PR. I don't want to revert to the older version and introduce an epoch to the Portfile.

By hardware architecture I presume you mean arm64? or is PPC?

Schamschula:
Unfortunately there is a new build issue with 1.9.1:

> bin/sh ../libtool  --tag=CC   --mode=compile /usr/bin/clang -DHAVE_CONFIG_H -I. -I..  -I../src -I../src -I../mpi -I../mpi -I/opt/local/include -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk -I/opt/local/include -pipe -Os -std=gnu89 -fheinous-gnu-extensions -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk -arch x86_64 -fno-delete-null-pointer-checks -Wall -MT blake2.lo -MD -MP -MF .deps/blake2.Tpo -c -o blake2.lo blake2.c
> keccak.c:907:23: error: use of undeclared identifier 'HWF_INTEL_FAST_SHLD'
>   else if (features & HWF_INTEL_FAST_SHLD)
>                       ^
> 1 error generated.

michaellass:
That is the one I am talking about and that can be easily fixed by removing `--disable-asm`.

> Version 1.9.1 includes this patch, which is removed by this PR

I don' understand. What is removed by this PR? The PR updates to 1.9.1 which includes the patch. 1.9.0 does not.

Version 1.9.1 introduces some configure changes to better detect the hardware architecture and available features. This was done in a couple of commits, one of them is this one:
https://dev.gnupg.org/rC8d404a629167d67ed56e45de3e65d1e0b7cdeb24

When built with `--disable-asm`, the corresponding flags are not set.

l2dy:
> First of all, I got the same email from Werner when I woke up this morning. The MacPorts version of 1.9.0 had been patched for this. Version 1.9.1 includes this patch, which is removed by this PR. I don't want to revert to the older version and introduce an epoch to the Portfile.

No, our version is not patched for this *critical security bug*. See https://dev.gnupg.org/T5275.

Schamschula:
It looks like removing `--disable-asm` does the trick, even on an Intel machine. It also build correctly on arm64. I'll push it.

michaellass:
Thanks!

I created an upstream bug report for this: https://dev.gnupg.org/T5277

Schamschula:
Fixed by https://github.com/macports/macports-ports/commit/f75b194866a959fab2c3e633b5b50d058c5a1712

PR #9855, opened 2021-01-29
cmake: Remove support for python27

rubendibattista:
Fix: https://trac.macports.org/ticket/62178

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/cmake/Portfile
+++ b/devel/cmake/Portfile
@@ -47,7 +47,7 @@ if {${subport} eq ${name}} {
     checksums       rmd160  f06e1e943457f79715bae77f347e2417b72a3326 \
                     sha256  5bff94ab6638995ffd029c9e319e96572fcec21abc9264ccdd03df97adac1113 \
                     size    7138448
-    revision        0
+    revision        1
 
     gitlab.livecheck.regex {([0-9.]+)}
     compiler.cxx_standard 2011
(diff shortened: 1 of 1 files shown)


kencu:
If it makes no difference to anyone, we could leave the python27 support in for all these ports. I am not looking forward to a massive purge of python27 from all ports, if that is the plan.

All macOS systems currently still come with python 27, and it makes bootstrapping much easier if we leave it to all the ports, rather than stripping it out.

If python27 no longer works for cmake, well that is different, of course. Then of course take it out.

michaelld:
I'm in agreement with @kencu  here. That said, looking at `cmake +python27` shows a bunch of Sphinx modules that don't exist in MP. I think cmake still supports Py27, so why do we not (still?) have those Sphinx Py27 ports?

rubendibattista:
Well, I'd argue because it reached its EOL (finally!)? It has been a pain the transition from Python2 to Python3, most of the third party libraries are removing support for it, and I think we should ease this out, not hindering it. 

Anyway before opening this PR I looked around in the commits and there were quite a few removing support for python27 on other ports. 

I would personally prefer to not restore those ports and remove Python2 altogether. The upstream developers have stopped supporting it, that means providing bugfix or security updates. I guess this might also mean potential security issue in the long run.

kencu:
OK, for now at least we're going to leave python27 support in cmake. If it becomes such that it is not useful any more, and can't be made to work, Michael will delete it.

PR #9859, opened 2021-01-30
gcc10: use generic SDK if needed

kencu:
on the newest systems, where the SDK is
changing rapidly, use a generic SDK reference

this changes MacOSX11.[0-9].sdk to MacOSX.sdk

### DIFF
--- a/lang/gcc10/Portfile
+++ b/lang/gcc10/Portfile
@@ -9,7 +9,7 @@ PortGroup           conflicts_build              1.0
 epoch               5
 name                gcc10
 version             10.2.0
-revision            4
+revision            5
 subport             libgcc10 { revision 3 }
 platforms           darwin
 categories          lang
@@ -146,7 +146,9 @@ pre-configure {
         #
         # https://trac.macports.org/ticket/53726
         # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79885
-        configure.args-append --with-sysroot="${configure.sdkroot}"
+
+        # if the sdkroot is one of the current, rapidly changing SDKS, use the generic one
+        configure.args-append --with-sysroot="[regsub {MacOSX1[1-9]\.[0-9]+\.sdk} ${configure.sdkroot} {MacOSX.sdk}]"
     }
 
 }
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -33,7 +33,7 @@ if { ${build_arch} eq "arm64" } {
 
     # Version must follow same scheme as with GCC snapshots below <version>-<commit-date>
     version         11-20210122
-    revision        0
+    revision        1
     subport         libgcc-devel { revision 0 }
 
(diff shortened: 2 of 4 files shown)


kencu:
I'm not certain if this approach is suitably robust -- but perhaps this or something like this would take care of our current troubles with rapidly moving SDKs on the newest systems, without causing troubles for the older, set systems.

jmroot:
This is reasonable as a temporary workaround on macOS 11 only. With the latest Xcode on 10.14 and 10.15, it would be using the SDK for the next OS version up (and of course that will almost certainly happen on Big Sur too when macOS 12 is released.) Long term, gcc needs to support using both `--with-sysroot` and `--with-build-sysroot` set to different values. (At least I assume that doesn't work now, or this would be easy to fix correctly?)

kencu:
it looks like the gcc folks had lots of fun trying to sort out just what those options were supposed to mean ... sort of a "Princess Bride" moment. "I don't think that means what you think that means..."

https://gcc.gnu.org/legacy-ml/gcc-help/2007-10/msg00118.html

kencu:
the official documentation of what those options mean is here (in Greek):

https://gcc.gnu.org/install/configure.html

jmroot:
OK, so `--with-build-sysroot` isn't really what we want, and `--with-sysroot` does more than we want, since "This option affects the system root for the compiler used to build target libraries (which runs on the build system) and the compiler newly installed with make install". We want to separate the two parts of that.

I wonder if we could use `--with-sysroot=${configure.sdkroot}`, but patch whatever file contains the default for the installed compiler to hardcode MacOSX.sdk. (Even better would be to search for a fallback at runtime if the compile-time SDK is not present.)

kencu:
to be noted that this regsub will not change ${configure.sdkroot} if it contains 

MacOSX10.[0-9].sdk

so all older systems without an 11.X sdk should be unaffected.

lang/gcc10/Portfile
```diff
@@ -146,7 +146,9 @@ pre-configure {
         #
         # https://trac.macports.org/ticket/53726
         # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79885
-        configure.args-append --with-sysroot="${configure.sdkroot}"
+
+        # if the sdkroot is one of the current, rapidly changing SDKS, use the generic one
+        configure.args-append --with-sysroot="[regsub {MacOSX1[1-2]\.[0-9]+\.sdk} ${configure.sdkroot} {MacOSX.sdk}]"
```
cjones051073, line 151: Why limit to macOS 11 and 12. When 13 comes out in a couple of years it will break again if no one updates it.

kencu: Yes, true, easy enough.

kencu:
OK -- I applied the same fix to all the gcc versions (8+) that are currently building on BigSur, where the SDK is changing rapidly. It will change MacOSX11.x.sdk or higher to MacOSX.sdk.

When BigSur stabilizes it's SDK, we can adjust the test if we like to hold the last SDK version with it's full numerical identifier.

PR #9867, opened 2021-01-30
libfec: fix build for arm64 machines

hexagonal-sun:
Add a pattern match in config.sub so that arm64 darwin machines are
recognised.

#### Description

Fix the build of libfec to support arm64 machines.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/science/libfec/Portfile
+++ b/science/libfec/Portfile
@@ -19,6 +19,9 @@ checksums rmd160 aa5f9e55f535d4ac27503effefd8d8ff3af8fd27 \
           size   130532
 revision  0
 
+
+patchfiles          patch-config-sub.diff
+
 use_autoreconf  yes
 
 configure.args-append \
--- a/science/libfec/files/patch-config-sub.diff
+++ b/science/libfec/files/patch-config-sub.diff
@@ -1,0 +1,11 @@
+--- config.sub.orig
++++ config.sub
+@@ -260,7 +260,7 @@ case $basic_machine in
+ 	# FIXME: clean up the formatting here.
+ 	vax-* | tahoe-* | i*86-* | i860-* | ia64-* | m32r-* | m68k-* | m68000-* \
+ 	      | m88k-* | sparc-* | ns32k-* | fx80-* | arc-* | c[123]* \
+-	      | arm-*  | armbe-* | armle-* | armv*-* | strongarm-* | xscale-* \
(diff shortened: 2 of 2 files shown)


ra1nb0w:
Sent upstream https://github.com/fblomqvi/libfec/pull/7
Waiting three days otherwise merge this PR.

michaelld:
There's a better way to go this ... see for example: https://github.com/macports/macports-ports/commit/25c865e2511982b427904b661f0f6302f3edf430

ra1nb0w:
Upstream accepted the patch. @michaelld do you suggest to use the ryan fix?

michaelld:
> Upstream accepted the patch. @michaelld do you suggest to use the ryan fix?

yes. It's important that upstream accepted the patch, too. but autotools is kept more up to date than most upstreams do for the `config.*` files, so why not use them instead of whatever the project provides?

hexagonal-sun:
@michaelld it seems a shame to have to pull in automake in order to build these packages, especially when a small single line change can fix the issue. I understand that it's a temporary fix until the package author re-relases with the latest automake/autoconf changes, but I think a single line patch feels like a better solution than pulling in automake and replacing the autogenerated files.

ra1nb0w:
@hexagonal-sun try out the new port that should fix the build on arm64.

kencu:
fixed

https://build.macports.org/builders/ports-11_arm64-builder/builds/10863/steps/install-port/logs/stdio

ra1nb0w:
Oh nice. We have an arm64 build server :)
Thank you @kencu

michaelld:
As an interim step, I'd recommend the `automake` tweak. Since upstream "fixed" the issue, it makes sense to use whatever they are doing since it minimizes the Portfile in MP. My global preference for MP would be to force using a common set of `config.*` scripts for ports that use them ... just like the noted tweak. That way we have consistency & if there's an issue it could be fixed globally rather than port by port. There are benefits and drawbacks in such a global setting, but it's honestly not much different than what we do for CMake ports or Qt ports or ...

kencu:
Don't need to worry for this particular PR, @michaelld -- the port has been updated to the upstream fix, and this PR is no longer needed or relevant, as the build on arm64 succeeds now.

But - in general -- I'm 100% on board with using Ryan's fix when we need to fix this issue.

PR #9869, opened 2021-01-30
qrupdate: Update Portfile for arm64

woachk:
Adding support for Apple Silicon Macs.

###### Type(s)
- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.2


kencu:
Looks like this works. 

The universal archs setting is not right any longer - BigSur needs to (somehow) become arm64 / x86_64.

So we'll see if we can get this right before committing it.

PR #9870, opened 2021-01-30
arpack: Update Portfile for Apple Silicon

woachk:
#### Description

Support Apple Silicon Macs.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

Apple M1


ryandesign:
Note that this can be simplified using the `in` operator:

```
        if {${build_arch} in [list arm64 ppc64 x86_64]} {
```

PR #9879, opened 2021-01-31
meson: update to 0.56.2

herbygillot:
- switch to Python 3.9 as that is now the MacPorts default Python
  version
- remove powermacintosh patch, changes are present upstream

###### Tested on

macOS 10.15.7 19H114
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/meson/Portfile
+++ b/devel/meson/Portfile
@@ -4,9 +4,8 @@ PortSystem          1.0
 PortGroup           python 1.0
 PortGroup           github 1.0
 
-github.setup        mesonbuild meson 0.55.3
-
-revision            1
+github.setup        mesonbuild meson 0.56.2
+revision            0
 
 github.tarball_from releases
 license             Apache-2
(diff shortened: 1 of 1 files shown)


kencu:
please don't commit at present.

Herby, there is a pegged version/section of meson for Tiger.

It still needs the old checksums, the same revision, and the powermacintosh patch and patchfile.

herbygillot:
> please don't commit at present.
> 
> Herby, there is a pegged version/section of meson for Tiger.
> 
> It still needs the old checksums, the same revision, and the powermacintosh patch and patchfile.

OK. The pegged version is untouched here, save for the removal of the powermac patch.  Probably the right thing to do here would be to restore that patch into the pegged version's block?

kencu:
I think I'll add all the python versions that meson supports to the list, to avoid bootstrapping headaches.

These build tools need to be very forgiving in what they accept, and meson says it is fine with any python >= 3.5, at least in the Portfile note.

The default can still be 39.

kencu (approved):

PR #9887, opened 2021-02-01
camlp5: update to 7.14

codesmythe:
#### camlp5: update to 7.14

An update is needed to eventually support more recent OCaml compilers (e.g., 4.10.2).

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
Xcode 12.4 12D4e
###### Verification 
Have you

### DIFF
--- a/lang/camlp5/Portfile
+++ b/lang/camlp5/Portfile
@@ -3,7 +3,7 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-set _version        7.08
+set _version        7.14
 github.setup        camlp5 camlp5 [string map {. ""} ${_version}] rel
 version             ${_version}
 revision            0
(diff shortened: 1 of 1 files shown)


pmetzger:
We probably need to update the entire OCaml sourcecode chain. camlp5 is the least of the obstacles.

codesmythe:
Agreed about camlp5. I was looking around at compiler dependent OCaml packages that could be upgraded to support both the current 4.08.1 and also 4.10.2, and camlp5 was one (another was my [ppx_tools](https://github.com/macports/macports-ports/pull/9891) PR). They can be upgraded 'asynchronously' to a big source code chain update. 

I recognize that a full OCaml compiler update will be a lot of work. Where is the best place to discuss the bigger picture of a compiler update. Is it https://trac.macports.org/ticket/59429 or somewhere else?

pmetzger:
That ticket seems fine. We should be targeting 4.12, and it would be best if we opened a ticket for everything that needs an update. Assistance is cheerfully accepted.

pmetzger:
FYI, now that we've updated ocaml to 4.12.0, it will be necessary to have a version of camlp5 that supports 4.12.0; so far as I can tell, none exists yet. Also, I'll note we are down to only three old packages that require camlp5.

pmetzger:
I seem to have been mistaken, in that camlp5 8 doesn't support 4.12, but 7.14 does. I've now updated to 7.14; I'm not sure what to do about camlp5 v8 but I suppose we'll just wait.

PR #9897, opened 2021-02-02
cairo: add fixes to allow building on older MacOS

michaelld:
#### Description

cairo: add fixes to allow building on older MacOS
One is already upstream; the other is obvious in the error log how to fix.

Closes: https://trac.macports.org/ticket/62156
Ref: https://gitlab.freedesktop.org/cairo/cairo/-/issues/454

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D64 // Xcode 12.3 12C33
macOS 10.5.8 9L31a // Xcode 3.1.4 DevToolsSupport-1186.0 9M2809 (PPC)
macOS 10.4.11 8S165 // Xcode DevToolsSupport-794.0 (PPC)

###### Verification 
Have you

### DIFF
--- a/graphics/cairo-devel/Portfile
+++ b/graphics/cairo-devel/Portfile
@@ -11,7 +11,7 @@ name                        cairo-devel
 conflicts                   cairo
 set my_name                 cairo
 version                     1.17.4
-revision                    1
+revision                    2
 checksums                   rmd160  7a440359be60f8cb971e9d538973ac29613e143e \
                             sha256  74b24c1ed436bbe87499179a3b27c43f4143b8676d8ad237a6fa787401959705 \
                             size    41834076
@@ -55,6 +55,9 @@ patchfiles-append           dont-check-stderr.patch
 # Fix crash on macOS Big Sur and newer.
 # https://gitlab.freedesktop.org/cairo/cairo/-/issues/420
 # https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/52
+# https://gitlab.freedesktop.org/cairo/cairo/-/commit/ba2afdcacf42eccf263b39efc77b85f3a65dcd74
+# https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/119
+# https://gitlab.freedesktop.org/cairo/cairo/-/commit/38e486b34d435130f2fb38c429e6016c3c82cd53
 patchfiles-append           patch-cairo-quartz-surfaces.diff
 
 # https://trac.macports.org/ticket/34137
--- a/graphics/cairo-devel/files/patch-cairo-quartz-surfaces.diff
+++ b/graphics/cairo-devel/files/patch-cairo-quartz-surfaces.diff
@@ -20,7 +20,7 @@
 -    /* the imageSurface will be destroyed by the data provider's release callback */
      CGImageRelease (surface->image);
 -
-+    cairo_surface_destroy (surface->imageSurface);
++    cairo_surface_destroy ( (cairo_surface_t*) surface->imageSurface);
      return CAIRO_STATUS_SUCCESS;
  }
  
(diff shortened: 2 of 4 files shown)


ryandesign:
You've reported these changes to the developers of cairo and they agree these are the correct changes to make?

If so, then please also make the same changes to cairo-devel.

michaelld:
> You've reported these changes to the developers of cairo and they agree these are the correct changes to make?

Not yet. That said, the required change is obvious as it follows what the other similar changes of recent did for getting the error returned with the correct type.

> If so, then please also make the same changes to cairo-devel.

I didn't even realize there was such a port. My fix here likely will require a tweak as part of it is already in the upstream GIT master ... assuming `cairo-devel` is following that but I honestly haven't checked ...

michaelld:
As an explanation: the function `cairo_quartz_image_surface_create` has a return value `cairo_surface_t *` as [noted here in the code](https://gitlab.freedesktop.org/cairo/cairo/-/blob/master/src/cairo-quartz-image-surface.c#L304-305). Errors such as [`SURFACE_ERROR_TYPE_MISMATCH`](https://gitlab.freedesktop.org/cairo/cairo/-/blob/master/src/cairo-quartz-image-surface.c#L320) are returned via [this macro](https://gitlab.freedesktop.org/cairo/cairo/-/blob/master/src/cairo-quartz-image-surface.c#L48), which "converts" a `_cairo_error` into a `cairo_surface_t *` via the converter `_cairo_surface_create_in_error`. A little later in the code the error `CAIRO_STATUS_NO_MEMORY` is to be returned but the conversion isn't done & hence the return type doesn't immediately match that declared in the function. Modern compilers "know" how to do the conversion, or at least don't generate an error in trying to do it. But, really, it should be converted in the same way as the other returned errors are, which is done by adding in the `_cairo_surface_create_in_error` as I've done in this PR. This tweak makes compilers that don't somehow "know" to do the conversion happy.

ryandesign:
> Not yet. That said, the required change is obvious as it follows what the other similar changes of recent did for getting the error returned with the correct type.

I don't have time this month to evaluate the correctness of your change. I rely on the expertise of the developers of cairo to determine the suitability of including this change, since they're more familiar with their own software, and since if the change is correct they will want to include it in their code.

> I didn't even realize there was such a port. My fix here likely will require a tweak as part of it is already in the upstream GIT master ... assuming `cairo-devel` is following that but I honestly haven't checked ...

The cairo project releases both stable versions and periodic development snapshots. The cairo port follows the former and cairo-devel follows the latter.

michaelld:
I rebased to current master and added in `cairo-devel` as well as tweaked the commit message to note the MP issue to close. Should be good to go now. I will report this upstream & reference the noted issue there too.

michaelld:
Trying to get a PR in upstream, but having issues. Wondering since they are disabling their gitlab instance whether they are allowing PRs to it until after the transition is done. I made recommendations on [the issue](https://gitlab.freedesktop.org/cairo/cairo/-/issues/454#note_787639).

michaelld:
FYI there is now a MR upstream for this ... they go even further than I did! Our fixes end up with the same code; I realized after I pushed this latest PR that things could be done better, which is what their MR is doing (per my suggestion on their issue). I can easily change this PR to be their MR once it is merged, since I'm guessing that will make @ryandesign confident that these are the changes we're looking for :)

michaelld:
Here's the MR: https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/119

kencu (approved):
Builds properly on 10.6.8 for me.

kencu:
rebased it while I was here

michaelld:
Rebased and updated with the actual merged code backported into this PR. I've verified that this code works on MacOSX 10.4 and 10.5 PPC as well as 11.3-beta. All of the changes make sense from a C-language perspective; I was in favor of more changes, but upstream stuck with these minimal ones & hence so do I for this PR.

ryandesign:
Thanks @michaelld!

PR #9936, opened 2021-02-05
VisualVM: new port

lhaeger:
###### Tested on

macOS 10.15.7 19H512
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/java/VisualVM/Portfile
+++ b/java/VisualVM/Portfile
@@ -1,0 +1,39 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           java 1.0
+
+name                VisualVM
+
+github.setup        oracle visualvm 2.0.6
+github.tarball_from releases
+
+categories          java
+platforms           macosx
+supported_archs     x86_64
+license             GPL-2
+maintainers         {gmx.net:lothar.haeger @lhaeger} openmaintainer
+
+description         All-in-One Java Troubleshooting Tool
+long_description    VisualVM is a visual tool integrating commandline JDK tools and lightweight profiling capabilities. \
+                    Designed for both development and production time use.
+
+homepage            https://visualvm.github.io
+
+distname            [strsed ${name}_${version} {g/[.]//}]
+use_dmg             yes
+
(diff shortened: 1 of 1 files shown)


java/VisualVM/Portfile
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+name                VisualVM
```
Janosch, line 6: This line could be omitted if you replace _visualvm_ with _VisualVM_ in the github.setup line

lhaeger:
Maybe, but I live in a case-exact world and would like to keep that principle up, if you don't mind. :-)

java/VisualVM/Portfile
lhaeger: Maybe, but I live in a case-exact world and would like to keep that principle up, if you don't mind. :-)

Janosch:
> Maybe, but I live in a case-exact world and would like to keep that principle up, if you don't mind. :-)

I don't mind :)

The only two other things which caught my attention are the fact that you did not use the java port group and by doing so did not provide a minimum JDK version that VisualVM requires. The other thing is, that this port is a binary port as nothing gets compiled. For me as a user this goes against my expectation of what a macports port usually is. However, I dont know if there is some sort of a convention in macports to make users aware of the fact that a port is precompiled.

I am a macports newbie, so both comments maybe completly irrelevant :)

lhaeger:
> The only two other things which caught my attention are the fact that you did not use the java port group and by doing so did not provide a minimum JDK version that VisualVM requires. 

Good catch! Thanks, fixed that.

> The other thing is, that this port is a binary port as nothing gets compiled. For me as a user this goes against my expectation of what a macports port usually is. 

While MacPorts *can* compile from source in most cases this is not possible or feasible in others. MacPorts is also not only about building software from source. That's a key feature, yes, but it's also a software catalogue, helping people to identify suitable software. And it's also an software management tool that helps people to easily install/uninstall MacPorts as well as keep their software collection up to date.  Those use cases remain regardless of how the code was build and are the main reason why ports like these exist in the first place. 

> However, I don't know if there is some sort of a convention in macports to make users aware of the fact that a port is precompiled.

There are ongoing discussions about that, which have not yet produced applicable results. Until that happens, I'll stick to how it's been done with similar ports.

PR #9937, opened 2021-02-05
zathura: add variant with synctex support

ryanakca:
#### Description

zathura needs synctex support to be able to synchronize latex-generated PDFs with text editors.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 11.1 20C69
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/office/zathura/Portfile
+++ b/office/zathura/Portfile
@@ -42,6 +42,11 @@ depends_lib         port:desktop-file-utils \
                     port:libmagic \
                     path:lib/libgirara-gtk3.dylib:girara
 
+# meson will automatically detect if synctex is available and link
+# against it. Explicitly disable it to avoid breaking zathura if
+# ever users uninstall texlive-bin
+configure.args-append   -Dsynctex=disabled
+
 # blacklist compilers that don't support -std=c11
 compiler.blacklist  *gcc-4.* {clang < 300}
 
@@ -75,6 +80,12 @@ subport ${name}-docs {
     }
 
     unset notes
+}
+
+variant synctex description "Enable synctex support" {
+    depends_lib-append  port:texlive-bin
+
+    configure.args-replace -Dsynctex=disabled -Dsynctex=enabled
 }
 
 livecheck.type      regex


office/zathura/Portfile
```diff
@@ -77,6 +77,11 @@ subport ${name}-docs {
     unset notes
 }
 
+variant synctex description "Enable synctex support" {
+    depends_lib-append  port:texlive-bin \
+                        port:dbus
```
mf2k, line 82: Nitpick: Please sort the dependencies alphabetically.

mf2k (changes requested):

office/zathura/Portfile
```diff
@@ -77,6 +77,11 @@ subport ${name}-docs {
...
+}
```
mf2k, line 83: What happens if those ports are already installed? Is there a configure setting to turn them off so they don't get used.

mf2k (changes requested):

office/zathura/Portfile
ryanakca: If those ports are already installed, then the build system automatically detects it and enables synctex support.

There's no setting, so there's no way to turn them off AFAICT unless you want to patch meson.build to disable the part that checks if synctex is available.

mf2k: The problem is when it picks them up, they become a hidden dependency. So if `dbus` gets uninstalled, `zathura` won't prevent that because it is not a listed dependency (unless this variant is selected). And then the port could be in a broken state.

mf2k: Or does that scenario not break the port?

mf2k (changes requested):

office/zathura/Portfile
ryanakca: It definitely breaks the port. I see two alternatives:

1. Patch the build system to disable the automatic check so that the default variant doesn't automatically pick up dbus/synctex, and then manually pass `-DWITH_SYNCTEX` when we want it
2. Fork out synctex into a separate subport from texlive-bin and make the default zathura installation depend on it. (This is what Debian does.)

mf2k: I think that option 1 is the simplest because the changes remain in this port and so that should be implemented. Option 2 requires coordination with the texlive maintainer - which certainly can be done in the future.

ryanakca: I figured out how to make meson disable synctex without having to patch the meson.build file. I also dropped the dependency on dbus, because it turns out that it is a transitive dependency and that zathura does not directly link against it.

office/zathura/Portfile
```diff
@@ -77,6 +82,12 @@ subport ${name}-docs {
     unset notes
 }
 
+variant synctex description "Enable synctex support" {
+    depends_lib-append  port:texlive-bin
+
+    configure.args-append   -Dsynctex=enabled
```
mf2k, line 88: You want to remove the other argument. This is how to do that:

`configure.args-replace   -Dsynctex=disabled -Dsynctex=enabled
`

mf2k (changes requested):

office/zathura/Portfile
ryanakca: Oh, that's exactly what I was looking for, thanks!

mf2k (approved):

PR #9955, opened 2021-02-07
qemu: fix building for apple silicon (WIP)

sirn:
#### Description

WIP for getting QEMU to work under Apple Silicon.

For [#62116 Unable to install QEMU on Apple M1 Silicon](https://trac.macports.org/ticket/62116), this was caused by `--cpu` configure option being set to `arm64` while QEMU expects `aarch64`. However, this fix alone is not enough as Big Sur now enforces W\^X under M1. [tcg patch](https://patchwork.kernel.org/project/qemu-devel/patch/20210113032806.18220-1-r.bolshakov@yadro.com/) is also required to get this working. However aforementioned patch is currently broken under macOS 11.2 due to [mprotect bug](https://lore.kernel.org/qemu-devel/161266423509.2846.13019124961581445864.malone@gac.canonical.com/) (so it's not yet included in this PR). 

The only way to get QEMU to work on Apple Silicon currently is via Hypervisor.framework which only works for `qemu-system-aarch64` and requires [hvf patch series](https://patchwork.kernel.org/project/qemu-devel/list/?series=418581). For [#61996 qemu hvf support on Big Sur requires signing security entitlement](https://trac.macports.org/ticket/61996), the hvf patch series also add codesign to QEMU, but this codeside will broke during destroot due to meson modifying the binary during install (strip/install_name_tool?) which happens after codesign so I need to add adhoc codesign in post-destroot anyway. 

I'm not sure how to approach this. Some option I can think of:

* Only fix `--cpu` (can now be built on Apple Silicon, but broken due to W\^X)
* Fix `--cpu` and include tcg patch (works on Apple Silicon without acceleration, but broken on macOS 11.2)
* Fix `--cpu` and include hvf patch (only aarch64 works on Apple Silicon with acceleration) (this PR) 
* Set `supported_arches x86_64 ppc` to disable building for M1 for the time being?

Related tickets:

* https://trac.macports.org/ticket/62116
* https://trac.macports.org/ticket/61996

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 11.2 20D64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -8,7 +8,7 @@ PortGroup legacysupport 1.0
 
 name                    qemu
 version                 5.2.0
-revision                0
+revision                1
 categories              emulators
 license                 GPL-2+
 platforms               darwin
@@ -57,8 +57,13 @@ if {${os.platform} eq "darwin" && ${os.major} < 11} {
     configure.cflags-append     -femulated-tls
 }
 
+set arch ${configure.build_arch}
+if {${configure.build_arch} eq "arm64"} {
+    set arch "aarch64"
+}
+
 # Select compiler
-configure.args          --cpu=${configure.build_arch} \
+configure.args          --cpu=${arch} \
                         --cc=${configure.cc} \
                         --objcc=${configure.objc} \
                         --host-cc=${configure.cc} \
@@ -114,6 +119,24 @@ pre-configure {
     configure.args-append --extra-ldflags="${configure.ldflags}"
 }
 
+# out of tree hvf support for Apple Silicon
+# https://patchwork.kernel.org/project/qemu-devel/list/?series=418581
+patchfiles-append       patch-hvf-Add-hypervisor-entitlement-to-output-binaries.diff \
+                        patch-hvf-x86-Remove-unused-definitions.diff \
+                        patch-hvf-Move-common-code-out.diff \
+                        patch-hvf-Introduce-hvf-vcpu-struct.diff \
+                        patch-arm-Set-PSCI-to-0.2-for-HVF.diff \
+                        patch-hvf-Simplify-post-reset-init-loadvm-hooks.diff \
+                        patch-hvf-Add-Apple-Silicon-support.diff \
+                        patch-arm-Add-Hypervisor.framework-build-target.diff \
+                        patch-arm-hvf-Add-a-WFI-handler.diff \
+                        patch-hvf-arm-Add-support-for-GICv3.diff \
+                        patch-hvf-arm-Implement--cpu-host.diff
+
+post-configure {
+    system "chmod +x ${worksrcpath}/scripts/entitlement.sh"
+}
+
 # The qemu build system bugs out when using in-source-tree builds.
 # During builds, files are generated in directories such as pc-bios, which are
 # prerequites to the config-host.mak file. The latter gets generated during a
@@ -142,7 +165,14 @@ foreach t {i386 x86_64 alpha {arm aarch64} cris hppa lm32 m68k {microblaze micro
            nios2 moxie or1k {ppc ppc64} riscv32 riscv64 rx s390x {sh4 sh4eb} {sparc sparc64} tricore unicore32 {xtensa xtensaeb}} {
     variant target_[lindex $t 0] description "Add target support for [join $t {, }]" "append target_list \",[join $t -softmmu,]-softmmu\""
 }
-default_variants-append +target_i386 +target_x86_64
+
+if {${configure.build_arch} eq "x86_64"} {
+    default_variants-append +target_i386 +target_x86_64
+} elseif {${configure.build_arch} eq "arm64"} {
+    default_variants-append +target_arm
+} elseif {${configure.build_arch} eq "ppc" || ${configure.build_arch} eq "ppc64"} {
+    default_variants-append +target_ppc
+}
 
 if {![variant_isset curses]} {
     default_variants-append +cocoa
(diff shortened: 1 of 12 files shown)


emulators/qemu/Portfile
```diff
@@ -114,6 +119,24 @@ pre-configure {
...
+                        patch-arm-Set-PSCI-to-0.2-for-HVF.diff \
+                        patch-hvf-Simplify-post-reset-init-loadvm-hooks.diff \
+                        patch-hvf-Add-Apple-Silicon-support.diff \
+                        patch-arm-Add-Hypervisor.framework-build-target.diff \
+                        patch-arm-hvf-Add-a-WFI-handler.diff \
+                        patch-hvf-arm-Add-support-for-GICv3.diff \
+                        patch-hvf-arm-Implement--cpu-host.diff
+
+post-configure {
+    system "chmod +x ${worksrcpath}/scripts/entitlement.sh"
```
raimue, line 137: No need to spawn a shell for this, this can be done from Tcl directly:
`file attributes ${worksrcpath}/scripts/entitlement.sh -permissions +x`

emulators/qemu/Portfile
```diff
@@ -250,6 +280,27 @@ platform darwin 8 {
     build.cmd ${prefix}/bin/gmake
 }
 
+platform darwin 20 {
+    post-destroot {
+        # meson modify binary during install which broke codesign for hvf-enabled binaries
+        # we need to re-sign it again
+        set codesign_targets {}
+
+        if {${configure.build_arch} eq "x86_64"} {
```
raimue, line 289: These blocks match the binaries created by the default variants. However, if any other `+target_*` variant was chosen, they must also be broken. Could we just sign anything matching a `qemu-system-*` wildcard?

For example:
```
foreach target [glob -directory "${destroot}${prefix}/bin/" "qemu-system-*"] {
    ...
}
```

raimue (commented):
Thanks a lot for working on this! It makes perfect sense to support qemu on arm64. I would rather get it to some working state than to disable the build. I do not have access to an M1 system, so I cannot test the proposed patches myself.

Aren't the HVF patches only necessary on arm64? As reviewers already raised, some of the changes do not seem to be in the right place, like ARM-specific change to `include/sysemu/hvf.h`. I am a bit concerned these might cause issues on x86_64. Therefore I suggest to put the `patchfiles-append` into an arch-specific conditional block for arm64 only.

Please follow [CommitMessages](https://trac.macports.org/wiki/CommitMessages) for the commit message format. In particular, please use "Closes: <Trac URL>" to automatically close referenced tickets when this is merged.

emulators/qemu/Portfile
sirn: There seems now to be a patch to fix codesign without requiring to manually sign via MacPorts. I think we can use this instead: https://lore.kernel.org/qemu-devel/20210225000614.46919-1-akihiko.odaki@gmail.com/

hexagonal-sun:
Hello,

I've just tried this out on my machine and I'm getting:

```
matthew@matthews-air ~/V/arch [1]> qemu-system-aarch64 -M virt -kernel boot/Image -hda out.img
Could not allocate dynamic translator buffer
```

I'm running MacOS version 11.2.3.

trodemaster:
This is an error with your qemu command line and not the binary.
You can check that the hvf acceleration is available on your system with
the specific binary by running this command. If hvf doesn't show up then
the binary is not good.

qemu-system-aarch64 -M virt -accel help



Have a look at this doc for some good info on running qemu on arm
https://gist.github.com/niw/e4313b9c14e968764a52375da41b4278#file-readme-md

Blake

hexagonal-sun:
@trodemaster Oops, ineed I was at fault, not passing `-accel hvf`.  Sorry for the noise.

phirestalker:
I don't know how far along this is, but there is a formula that works on homebrew, but I believe they are using a version higher than 5.2.0. Maybe there is some work already done that you can borrow?

https://github.com/knazarov/homebrew-qemu-virgl

anahata0108:
Coincidentally, the qemu project just released version 6.0.0 a few days ago. I’ve been tracking the RCs on my Apple Silicon Macs and they build unmodified without problem. I can’t say I’ve tested them thoroughly, but they do work for my purposes (mostly RISC-V, x86_64, ARM, and AVR emulation). I don’t have an Intel Mac to test on, I’m afraid, but I’d be happy to test this on ARM.

trodemaster:
My quick test of 6.0 built on applesilicon did not support the hvf
accelerator. I’m not seeing the needed patches to enable that for 6.0 yet
either.

It’s my view that most users are looking for aarch64 guests on applesilicon
systems. For that to work well the hvf acceleration is required.

Not a lot of work remains on the existing 5.2.0 PR to get things usable for
people. Getting that out the door would be ideal. Once the hvf patches are
incorporated into qemu bumping the port to 6.x should be easy.

Blake

sirn:
Yup sorry. Didn't have a time to take a look at this for quite some time. Going to find some time and rebase this...

trodemaster:
I’ll keep an eye on the PR and can do some testing as needed.

reneeotten:
@raimue do you somehow want to move forward with this PR or should it be closed?

raimue:
Personally I have no way to test this on Apple M1 hardware...

I hope upstream QEMU will eventually have a better solution for this than backporting proposed patches. So I guess we have to abandon this.

sirn:
Yup, please close this because patches are now starting to broke with newer QEMU versions...

PR #9964, opened 2021-02-08
py-cipheycore: Fix python3 configuration

harens:
#### Description

_Apologies for all these bug fixes for cipheycore. Loads of Ciphey users who use macOS have issues with cipheycore (e.g. https://github.com/Ciphey/Ciphey/issues/256 https://github.com/Ciphey/Ciphey/issues/362), which is why I think it's so important to have a working MacPorts portfile. Hopefully, this should be the last bug fix before adding ciphey._

cipheycore builds fine if only it's equivalent python version is active (i.e. py38-cipheycore builds fine if only python38 is active). This is fine for the buildbots, but not if the user has to build from source and has different python versions active. If the wrong python version is found, a segmentation fault occurs at runtime. This sort of issue is very similar to https://trac.macports.org/ticket/59946, and a temporary fix before this PR is merged would be to use trace mode.

The reason why this happens is explained here https://github.com/Ciphey/Ciphey/issues/238#issuecomment-688630158.

This bug fix manually sets the build dependency python to be used during the configure stage. It sets `-DCIPHEY_CORE_PYTHON` to the folder containing the python headers (see https://github.com/Ciphey/CipheyCore#linuxosxother-unices).

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69
xcode-select version 2384.

###### Verification 
Have you

### DIFF
--- a/python/py-cipheycore/Portfile
+++ b/python/py-cipheycore/Portfile
@@ -6,7 +6,7 @@ PortGroup           python 1.0
 
 github.setup        ciphey cipheycore 0.3.2 v
 name                py-cipheycore
-revision            1
+revision            2
 python.versions     38 39
 
 description         Some cryptanalysis tidbits written in a proper language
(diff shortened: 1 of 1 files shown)


harens:
Thank you very much @cjones051073.

PR #9972, opened 2021-02-09
opencv4: generate pkg-config file

Schamschula:
fix double paths in opencv4.pc file and move to default location

#### Description

Several packages require a pkg-config file to be properly configured. E.g. gmic and py-gmic, among others. opencv4, by default does not generate this file. This is a blocker in updating those packages to use opencv4 rather than opencv(3).

###### Type(s)

- [x] enhancement

###### Tested on

macOS 10.14.6 18G8012
Xcode 10.3 10G8

###### Verification 
Have you

### DIFF
--- a/graphics/opencv4/Portfile
+++ b/graphics/opencv4/Portfile
@@ -7,7 +7,7 @@ PortGroup           github 1.0
 
 github.setup        opencv opencv 4.5.0
 name                opencv4
-revision            3
+revision            4
 categories          graphics science
 platforms           darwin
 license             BSD
@@ -227,6 +227,12 @@ configure.args-append \
                     -DOPENCV_PYTHON_SKIP_DETECTION:BOOL=OFF \
                     -DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python \
                     -DPYTHON_DEFAULT_EXECUTABLE:FILEPATH=/usr/bin/python
+
+    if {${name} ne ${subport}} {
+        # generate pkg-config file
+        configure.args-append \
+                    -DOPENCV_GENERATE_PKGCONFIG=YES
+    }
 
 universal_variant   no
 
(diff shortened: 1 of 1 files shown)


mascguy:
Marius, I'm definitely fine with enabling generation of this file.

However, if we place it in the standard location - and to be consistent, opencv3's pkgconfig should be placed there too - is there any chance that ports will choose the incorrect one?

This is in the context of users with both installed simultaneously.

I ask, as I certainly dealt with this issue with CMake: Ultimately it was critical to cleanly segregate the CMake module defs - keeping them within the segregated `${prefix}/libexec/opencvX/...` area - rather than place them in the shared location.

mascguy:
Alternatively, in the interest of progress, this is something I'm willing to tackle later.

It's up to you, I'm good either way.

Schamschula:
There is no collision, as opencv4's file is called `opencv4.pc`

mascguy:
Marius, have you tested any of the py??-opencv4 subports? It looks like the CI jobs are failing to build those for some reason.

Haven't looked at the gory details yet, but could be a problem...

Schamschula:
That could be. There is no need for the .pc file for the Python subports. I'll have a look.

mascguy:
I think you'll need to disable pkgconfig generation entirely for the python subports, as those are allowed to be installed side-by-side.

So while your change will fix the case for one py??-opencv4 port, installing a 2nd subport will fail with a similar error.

graphics/opencv4/Portfile
```diff
@@ -226,7 +226,13 @@ configure.args-append \
                     -DBUILD_opencv_python3:BOOL=OFF \
                     -DOPENCV_PYTHON_SKIP_DETECTION:BOOL=OFF \
                     -DPYTHON_EXECUTABLE:FILEPATH=/usr/bin/python \
-                    -DPYTHON_DEFAULT_EXECUTABLE:FILEPATH=/usr/bin/python
+                    -DPYTHON_DEFAULT_EXECUTABLE:FILEPATH=/usr/bin/python \
```
ryandesign, line 229: A backslash shouldn't be added at the end of this line.
```suggestion
                    -DPYTHON_DEFAULT_EXECUTABLE:FILEPATH=/usr/bin/python
```

Schamschula: Good catch!

mascguy (approved):

karo11:
One remark of a gmic user and tester. 
gmic sub port gmic-qt still use opencv (3)! Both ports activated leads with either clang++ or g++ (MacPorts) to a conflict with include files during local build of gmic cli using opencv4; include/opencv2 is used, not include/opencv4/opencv2. I have during build opencv deactivated to overcome that problem.

Schamschula:
Now that opencv4 provides a pkg-config file, this is about to change.

PR #9975, opened 2021-02-09
freac: add missing codecs

Janosch:
#### Description
freac: Added missing audio codecs and DSP libraries
smooth: Removed gtk dependency as it is not needed for macOS

Note: Both changes have been suggested by fre:ac's author.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.1
Xcode 12.3 Build version 12C33

###### Verification 
Have you

### DIFF
--- a/audio/freac/Portfile
+++ b/audio/freac/Portfile
@@ -6,6 +6,7 @@ PortGroup           makefile 1.0
 
 github.setup        enzo1982 freac 1.1.3 v
 github.tarball_from releases
+revision            1
 
 categories          audio
 platforms           darwin
@@ -24,6 +25,21 @@ checksums           rmd160  2f1f9ea871793cec1a8fe02a96fb9c03b1fa0c49 \
 
 depends_lib         port:boca \
                     port:smooth
+
+depends_run         path:lib/libavcodec.dylib:ffmpeg \
+                    port:flac \
+                    port:lame \
+                    port:libfdk-aac \
+                    port:libogg \
+                    port:libopus \
+                    port:libsamplerate \
+                    port:libsndfile \
+                    port:libvorbis \
+                    port:mp4v2 \
+                    port:mpg123 \
+                    port:rubberband \
+                    port:speex \
+                    port:wavpack
 
 makefile.prefix_name    prefix
 
--- a/aqua/smooth/Portfile
+++ b/aqua/smooth/Portfile
@@ -33,7 +33,6 @@ checksums           rmd160  b53e8720b3e3e8fad48e7c1f6f1dd55a910758f8 \
 depends_lib         port:bzip2 \
                     port:curl \
                     port:fribidi \
-                    port:gtk3 \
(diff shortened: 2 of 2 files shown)


audio/freac/Portfile
```diff
@@ -25,6 +25,21 @@ checksums           rmd160  2f1f9ea871793cec1a8fe02a96fb9c03b1fa0c49 \
 depends_lib         port:boca \
                     port:smooth
 
+depends_run         port:lame \
+                    port:ffmpeg \
```
mf2k, line 29: Can `ffmpeg-devel` also be used? If so, use this specifications to allow it:

`path:lib/libavcodec.dylib:ffmpeg`

audio/freac/Portfile
```diff
@@ -25,6 +25,21 @@ checksums           rmd160  2f1f9ea871793cec1a8fe02a96fb9c03b1fa0c49 \
...
+                    port:speex \
+                    port:libsndfile \
+                    port:flac \
+                    port:libopus \
+                    port:wavpack \
+                    port:libfdk-aac \
+                    port:mpg123 \
+                    port:mp4v2 \
+                    port:libsamplerate \
+                    port:rubberband
```
mf2k, line 41: Please sort these alphabetically.

mf2k (changes requested):

kencu:
freac will need to be rev-bumped I believe, so the new Portfile is picked up by current MacPorts' installations (if any).

It's odd that these are only run-time deps. Does this software not need to see them during the build, to know where to find them, how to configure, etc?

Usually things like this are "lib" deps, as they are needed at both build time and run time.

Janosch:
@kencu Those dependencies are all optional. The software builds and runs without them but is of limited use as there is only the Core Audio encoder available. freac tries to find those dependencies on runtime and loads them if available.

As macports has no optional/recommended dependencies, I also thought about adding a +minimal variant without those runtime dependencies. If you think that makes sense I can add it. 

I just would like to have the default install as close as possible to the upstream version, therefore I added all those encoder/decoders as runtime depedencies.

audio/freac/Portfile
Janosch: Yes. I tested it and ffmpeg-devel works as well.

audio/freac/Portfile
Janosch: Done

kencu:
Thanks for clarifying.

mf2k (approved):

kencu:
Frank, does this need a revbump? No installed files are changing, but I recall Josh mentioning cached  portfiles need refreshing if things like the runtime deps change.

mf2k:
@kencu : Yes, freac should have the `revision` increased by one.

mf2k:
And it is not needed for smooth.

Janosch:
Increased revision to 1

PR #9990, opened 2021-02-11
libreoffice: move pkg-config calls to pre-configure

Tatsh:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/office/libreoffice/Portfile
+++ b/office/libreoffice/Portfile
@@ -166,18 +166,20 @@ use_xcode           yes
 configure.env-append  \
     "LCMS2_CFLAGS=-I${prefix}/include -DCMS_NO_REGISTER_KEYWORD=1" \
     PYTHON=${prefix}/bin/python${pyver}
-# PYTHON_CFLAGS and PYTHON_LIBS have to be non-empty
-set python_cflags [exec pkg-config --cflags python-${pyver}]
-set python_libs [exec pkg-config --libs python-${pyver}]
-if {${python_cflags} == ""} {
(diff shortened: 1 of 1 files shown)


Tatsh:
@kencu @herbygillot This should fix the portindex issue.

herbygillot:
I'll merge it in, then rebase my candidate branches on top of this and see if the issues no longer appear in Azure CI + GitHub Actions.

herbygillot:
Things are looking a bit better now. Thanks @Tatsh.

kencu:
It is worth checking if the pkg-config calls ever do anything at all. I have a feeling they may not ever be returning anything useful -- throw the output into a ui_debug message so you can at least see if  the who process is a waste of time an effort.

I'd say just use the right calls -- I'm changing all the llvm ports to do that -- they tried to call pkg-config too, since 10 years ago now, and I'm fairly certain they never returned anything useful.

Tatsh:
They are returning the include path for Python and nothing for lib (2.7 gives the library path, but not 3.x). Because of the latter point, I am setting the library path to something generally harmless but non-empty.

The `configure.ac` from LibreOffice requires that `PYTHON_LIBS` *and* `PYTHON_CFLAGS` be non-empty, otherwise `PYTHON` (path to interpreter) will be ignored.

kencu:
I realise they need to be set to something -- but I'm suggesting just set them and skip the pkg-config stuff. 

if you can.

PR #9995, opened 2021-02-11
ninja: fix cross-arch universal build

kencu:
there are two issues when building ninja universal, using the muniversal PortGroup on BigSur Intel.

1. configure.py: error: option --host: invalid choice: 'aarch64-apple-darwin20.3.0'

This is set by the muniveral PG but to an option the script does not accept.
Simplest solution appears to be to just delete the --host option.

2. :info:build sh: ./ninja: Bad CPU type in executable

The portfile has been written to bootstrap ninja and then rebuild ninja
with itself. This is interesting to do, and proves the built ninja
works, but it is not necessary to do so.

So we adjust the Portfile to only do this for non-universal builds.

Some complicated algorithm of figuring out if the arch would or would not
be runnable by the build system seems both cumbersome and unnecessary.

closes: https://trac.macports.org/ticket/62259

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/ninja/Portfile
+++ b/devel/ninja/Portfile
@@ -2,6 +2,7 @@
 
 PortSystem          1.0
 PortGroup           github 1.0
+
 # see https://trac.macports.org/ticket/56494
 PortGroup           muniversal 1.0
 
@@ -55,9 +56,27 @@ configure.post_args --bootstrap \
                     --verbose
 configure.universal_args
 
-build.cmd           ./ninja
-build.target
-build.args          -v
+if {![info exists universal_possible]} {
+    set universal_possible [expr {${os.universal_supported} && [llength ${configure.universal_archs}] >= 2}]
+}
+
+if {${universal_possible} && [variant_isset universal]} {
+
+    foreach my_arch ${configure.universal_archs} {
+        # strip the automatic setting of host
+        set merger_host(${my_arch}) ""
+    }
+
+    # can't reliably rebuild ninja with itself when cross-arch compiling
+    # so install the cross-compiled bootstrap version
+    build {}
+
+} else {
+
+    build.cmd           ./ninja
+    build.target
+    build.args          -v
+}
 
 destroot {
     xinstall -m 0755 ${worksrcpath}/${name} ${destroot}${prefix}/bin


kencu:
```
% sudo /opt/buildX11/bin/port -v installed ninja                                                                                                            
Password:
  ninja @1.10.2_0+universal (active) platform='darwin 20' archs='arm64 x86_64' date='2021-02-10T23:09:53-0800'
```

kencu:
@ryandesign  -- will push this tomorrow unless objections.

ryandesign:
> 1. configure.py: error: option --host: invalid choice: 'aarch64-apple-darwin20.3.0'
> 
> This is set by the muniveral PG but to an option the script does not accept.
> Simplest solution appears to be to just delete the --host option.

Ok

> 1. :info:build sh: ./ninja: Bad CPU type in executable
> 
> The portfile has been written to bootstrap ninja and then rebuild ninja
> with itself. This is interesting to do, and proves the built ninja
> works, but it is not necessary to do so.

I wasn't aware of that. That's what the port has done since before I took it over and I didn't realize it could be done any other way.

> So we adjust the Portfile to only do this for non-universal builds.
> 
> Some complicated algorithm of figuring out if the arch would or would not
> be runnable by the build system seems both cumbersome and unnecessary.

I'm not sure what the consequences are of the change you've made. Is there any difference between what gets installed now in the universal version vs. what gets installed in the non-universal version? (Besides that the one is universal.)

If there is an undesirable difference, then the current bootstrapping method could possibly be preserved by having it run the `ninja` that was built for `$build_arch` instead of running `./ninja`. But that would require the muniversal portgroup to build the one for `$build_arch` before the other archs, and I don't know if the muniversal portgroup guarantees that. If not, that would be a good improvement, as I've mentioned elsewhere.

kencu:
thanks for input.

I believe that the bootstrapped ninja and the rebuilt ninja are the same, and I know the bootstrapped-only version did well on the tests when I ran them...but I would also like to know if there are any differences at all.

Let me check on that.

kencu:
I'll reopen this once I've completed the detailed comparison.

PR #10027, opened 2021-02-14
ninja: fix cross-arch universal build

kencu:
Apple's gcc compilers do not allow dependency file generation
when multiple arch flags are used

This resulted in the muniversal portgroup being used to repair the
universal build of ninja several years ago, however this has it's own
issues by not allowing a cross-arch universal binary to be built,
due to the way the ninja bootstraps itself during compilation.

if we strip the (unnecessary) dependency file generation by removing
the dependency flags, then the muniversal PortGroup is no longer
needed and the cross-arch universal build can succeed without
using the muniversal portgroup.

see: https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html
see: https://github.com/macports/macports-base/commit/701e7b2597f2da954a390b1b63aa90d6f7aaba20
see: https://trac.macports.org/ticket/56494

closes: https://trac.macports.org/ticket/62259

### DIFF
--- a/devel/ninja/Portfile
+++ b/devel/ninja/Portfile
@@ -2,8 +2,6 @@
 
 PortSystem          1.0
 PortGroup           github 1.0
-# see https://trac.macports.org/ticket/56494
-PortGroup           muniversal 1.0
 
 epoch               1
 github.setup        ninja-build ninja 1.10.2 v
@@ -40,7 +38,8 @@ github.tarball_from archive
 
 installs_libs       no
 
-patchfiles          patch-configure.py-bootstrap-only.diff
+patchfiles          patch-configure.py-bootstrap-only.diff \
+                    patch-ninja-configure.py-remove-mmd.diff
 
 depends_build-append \
                     port:re2c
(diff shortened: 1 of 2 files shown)


kencu:
@ryandesign -- simpler, neater, easier, better fix.

ryandesign:
> if we strip the (unnecessary) dependency file generation

Does upstream confirm that this is unnecessary? Do they agree to include this change in their code?

kencu:
would never dare trouble them with an error noted on a defunct Apple fork of gcc from 2005.

but, this is similar to "disable dependency tracking" and just like other patches I referenced in MP, all on a throwaway bootstrap ninja that passes all tests.

So this is it, my friend.

kencu:
For a software package, having some form of dependency tracking during development is required. So they won't remove it from their code base, of course.

For MacPorts, it's meaningless to us, as we just build it through (and in this particular case, this only applies to the bootstrap ninja, not even the final built ninja).

So for **us** it's unnecessary (which is why for years we disabled dependency tracking during builds, to speed them up a bit).

So I closed the upstream question as irrelevant (to them).

kencu:
Perhaps a more desirable method of building ninja universal might be unearthed in due course of time.

kencu:
I plan to commit this unless objections, to continue my quest to fix universal building of as many ports as possible on BigSur. I can't see any good reason to keep it broken on MacPorts.

I'm sure you can see that I can't possibly take a complaint about a 2005 apple gcc compiler upstream.

If you  feel strongly about it, I will restrict this patch to times when apple gcc compilers are being used, but that is not necessary in general.

PR #10046, opened 2021-02-16
libvirt: update to 7.0.0

phrdina:
#### Description

Update libvirt to latest upstream release 7.0.0. In version 6.7.0 libvirt moved from autotools to meson as build system so the update is a bit more complex.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.1 20D75
Xcode CLI 12.4.0.0.1.1610135815

###### Verification 
Have you

### DIFF
--- a/sysutils/libvirt/Portfile
+++ b/sysutils/libvirt/Portfile
@@ -3,17 +3,18 @@
 PortSystem          1.0
 PortGroup           compiler_blacklist_versions 1.0
 PortGroup           legacysupport 1.0
+PortGroup           meson 1.0
 
 # O_CLOEXEC
 legacysupport.newest_darwin_requires_legacy 10
 
 # Remember to update libvirt and py-libvirt at the same time.
 name                libvirt
-version             6.6.0
+version             7.0.0
 revision            0
-checksums           rmd160  72294aa39f53386cfcd8812ddf243bd7aa9ee29c \
-                    sha256  94e52ddd2d71b650e1a7eb5ab7e651f9607ecee207891216714020b8ff081ef9 \
-                    size    9305836
+checksums           rmd160  1601c5f6ef46b048968c88a0bbb534ac95723caa \
+                    sha256  ca3833844d08c22867f1d1a46edc36bda7d6fe1a4f267e7d77100b79fc9ddd89 \
+                    size    8567648
 
 categories          sysutils
 license             LGPL-2.1+
@@ -29,13 +30,14 @@ homepage            https://libvirt.org
 master_sites        ${homepage}/sources/
 use_xz              yes
 
-depends_build       port:autoconf \
-                    port:automake \
+set python_version  39
+set python_branch   3.9
+
+depends_build-append \
                     port:bash-completion \
-                    port:libtool \
                     port:pkgconfig \
-                    port:python38 \
-                    port:py38-docutils \
+                    port:python${python_version} \
+                    port:py${python_version}-docutils \
                     port:perl5
 
 depends_lib         port:curl \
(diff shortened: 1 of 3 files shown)


sysutils/libvirt/Portfile
```diff
@@ -50,97 +48,77 @@ depends_lib         port:curl \
...
-
-# As of 5.10.0, an out-of-source build is required.
-configure.cmd       ../${worksrcdir}/autogen.sh --no-git
-configure.dir       ${workpath}/build
-build.dir           ${configure.dir}
-post-extract {
-    file mkdir ${configure.dir}
-}
+configure.env       PATH=${prefix}/bin:$env(PATH) \
+                    PKG_CONFIG_PATH=${prefix}/lib/pkgconfig
```
ryandesign, line 57: ${prefix}/bin is assumed to be in PATH already so this line isn't needed. ${prefix}/lib/pkgconfig is already in the default pkg-config search path so you don't need to specify it. If you did need to specify it, you should use `configure.pkg_config_path` rather than putting PKG_CONFIG_PATH into the environment manually.

sysutils/libvirt/Portfile
```diff
@@ -3,17 +3,18 @@
 PortSystem          1.0
 PortGroup           compiler_blacklist_versions 1.0
 PortGroup           legacysupport 1.0
+PortGroup           meson 1.0
 
 # O_CLOEXEC
 legacysupport.newest_darwin_requires_legacy 10
 
 # Remember to update libvirt and py-libvirt at the same time.
```
ryandesign, line 11: Please update py-libvirt to the same version.

sysutils/libvirt/Portfile
```diff
@@ -29,14 +30,11 @@ homepage            https://libvirt.org
...
-                    port:automake \
-                    port:bash-completion \
-                    port:libtool \
-                    port:pkgconfig \
-                    port:python38 \
-                    port:py38-docutils \
-                    port:perl5
+depends_build-append    port:bash-completion \
+                        port:pkgconfig \
+                        port:python38 \
```
ryandesign, line 35: python39 is now the default version in MacPorts. If this port will work with python39 then it should be switched to do so. If you don't want to do it as part of this commit I can do it later.

I would like to keep the alignment of this column as it was before. I suggest:

```
depends_build-append \
                    port:bash-completion \
                    port:pkgconfig \
                    ...
```

sysutils/libvirt/Portfile
```diff
@@ -50,97 +48,77 @@ depends_lib         port:curl \
 # error: You need at least XCode Clang v5.1 to compile QEMU
 compiler.blacklist-append apple-gcc* gcc-3.3 *gcc-4.* macports-clang-3.3 {clang < 503}
 
+patchfiles-append   patch-meson.diff
```
ryandesign, line 51: Instead of patching, `${frameworks_dir}/Python.framework/Versions/3.8/bin` could be prepended to PATH. That would make it easier to change the python version in the future, since only the Portfile (and not also the patchfile) would need to be changed. By abstracting 3.8 into a variable and using it everywhere, it could be reduced to a single portfile line that would need to be changed when switching the python version, ensuring that we don't forget to switch something.

ryandesign (commented):
libvirt's switch to meson and my unfamiliarity with it is why I hadn't updated the port so thanks for doing it for me. I've made a few suggestions. Also, there are a few typos in your commit message: "where" should be "were" and "virtaulport" should be "virtualport".

sysutils/libvirt/Portfile
phrdina: Thanks for the info, indeed it works without these two lines. For some reason the first time it did not work for me. I'll remove it.

sysutils/libvirt/Portfile
phrdina: Will do, I noticed that comment and made a mental note to update py-libvirt as well but as we can see it did not happen :).

sysutils/libvirt/Portfile
phrdina: This helps a lot. I looked at the failed github check and was not sure how to solve the missing python3 link and did not like adding another patch to fix meson.build for that as well. Will do both.

phrdina:
> libvirt's switch to meson and my unfamiliarity with it is why I hadn't updated the port so thanks for doing it for me. I've made a few suggestions. Also, there are a few typos in your commit message: "where" should be "were" and "virtaulport" should be "virtualport".

Glad to help. I understand that new build system might be challenging. I saw your comment in one of our issues in libvirt and decided to help with the update as the meson rewrite in libvirt was done by me.

Thanks for the review and all suggestions. I'll fix the issues and push another version.

ryandesign:
This caused the library minor versions to decrease. See https://trac.macports.org/ticket/63466. It is important to watch out for these kinds of mistakes any time a project switches from one build system to another.

PR #10066, opened 2021-02-20
cask: python versions changes

wyuenho:
* offers python version variants
* adds revision
* cleans up 807ce990556a1669d894d41feff27a5286098a6b

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2
Xcode 12.4

###### Verification 
Have you

### DIFF
--- a/devel/cask/Portfile
+++ b/devel/cask/Portfile
@@ -6,6 +6,7 @@ PortGroup           github 1.0
 PortGroup           elisp 1.0
 
 github.setup        cask cask 0.8.5 v
+revision            1
 description         Project management tool for Emacs
 long_description    Cask is a project management tool for Emacs that helps \
                     automate the package development cycle\; development, \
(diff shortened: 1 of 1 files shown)


amake:
If it works on Python 3.9 then we might as well just make that the default and remove the corresponding variant.

wyuenho:
Most ports are still relying on Python 3.8, I don't want to force another unnecessary Python version onto people's machines.

amake:
We'll never get most ports onto Python 3.9 if everyone is waiting for everyone else. 

This is also not a core port widely depended on by other ports; if it was then I might be more sympathetic.

ryandesign:
> Most ports are still relying on Python 3.8, I don't want to force another unnecessary Python version onto people's machines.

3.9 is now the default version of python in MacPorts (fa147a0a5b020635eb792a94330a0cc5641cc7b3); any ports using python38 should be updated to use python39 if possible.

PR #10067, opened 2021-02-20
neovim: fix crash on Big Sur with Apple Silicon

funasoul:
Closes: https://trac.macports.org/ticket/61550

#### Description
Fix crash on Big Sur with Apple Silicon.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.1 20D74
Xcode 12.4 12D4e

###### Verification 
Have you


raimue:
Thank you for the pull request!

As of now, neovim uses `port:lua` as dependency and not LuaJIT. This pull request seems to want to switch to LuaJIT instead, but still keeps the `port:lua` dependency. We also explicitly select Lua instead of LuaJIT with `-DPREFER_LUA`. This does not seem right... Are you sure these changes actually fix a problem? As far as I can see, most of these changes are in dead code.

Is the actual problem in lua itself? Shouldn't we then fix the `lua` port instead?

If the only viable solution is to switch to LuaJIT, we should use `port:luajit` as a dependency instead of adding a separate bundled copy that is only used by the neovim port.

Overall, be aware that we do not want to download external dependencies during build. I guess this would fail if networking is prohibited by the sandbox. If any additional external files are required, they should be downloaded during the [fetch phase](https://guide.macports.org/#reference.phases.fetch). That is usually achieved by setting `distfiles` and `master_sites`.

mojca:
Something is seriously wrong with the git history here.

funasoul:
Thank you so much for your precious comment.
I completely agree with you that the issue does not relate to lua, and LuaJIT should not be included in this pull request.
I have fixed this PR and pushed to this branch, but I accidentally made a mistake that I ran `git rebase` before the commit, which made this PR messy.

Should I close this PR and create a new PR only for this neovim issue?

funasoul:
I accidentally made a mistake that I ran `git rebase` before the commit, which made this PR messy.
I will create a new PR only for this neovim issue.

PR #10072, opened 2021-02-20
swig-java: fix build on Big Sur with Apple Silicon

funasoul:
* Changed to use zulu-jdk8 instead of openjdk8 on arm64.

#### Description
Fixed a problem where swig-java build would fail with the following message on Big Sur with Apple Silicon.

```sh
$ sudo port install swig-java
--->  Computing dependencies for swig-java
Error: Cannot install swig-java for the arch 'arm64' because
Error: its dependency openjdk8 only supports the arch 'x86_64'.
Error: Follow https://guide.macports.org/#project.tickets to report a bug.
Error: Processing of port swig-java failed
```

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.1 20D74
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/swig/Portfile
+++ b/devel/swig/Portfile
@@ -94,8 +94,13 @@ subport swig-java {
     PortGroup               java 1.0
 
     java.version            1.4+
-    # Set fallback to an LTS Java version
-    java.fallback           openjdk8
+    if { ${build_arch} eq "arm64"  } {
+        # Set fallback to an LTS Java version
+        java.fallback           zulu-jdk8
+    } else {
+        # Set fallback to an LTS Java version
+        java.fallback           openjdk8
+    }
 }
 
 subport swig-php {
@@ -306,6 +311,7 @@ if {${swig.lang} eq ""} {
 # doesn't exist and swig has only Python
 subport swig-python3 {
     PortGroup       obsolete 1.0
+
     replaced_by     swig-python
     version         4.0.1
     revision        2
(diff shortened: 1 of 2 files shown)


michaelld (changes requested):
Interesting change. I'll have to try `openjdk` on ARM64 ... guessing it doesn't build at the moment, yes? Do you know if `swig-java` builds with any newer version of `*jdk*` than 8? That version seems quite outdated. If so, then it would be optimal to provide variants for the different versions that are compatible ... or, maybe just use `zulu-jdk` which currently is at version 11? Same thing holds true for `openjdk*` ...

funasoul:
Thanks @michaelld for your comment. 
Before getting into the different version of openjdk, let me mention that the build fails not by swig nor swig-java, but swig-gcj, just like the other previous PR (https://github.com/macports/macports-ports/pull/8030#issuecomment-671204926).
I created this PR because I needed to build libsbml with Java bindings, and with this PR, I confirmed that libsbml with Java bindings worked with `sudo port test libsbml +java`.

I only found `zulu-jdk` has arm64 binaries. In the Portfile of `openjdk` it has the following line:
```
supported_archs  x86_64
```
which seems that `openjdk` only supports x86_64 at this time.
`zulu-jdk8` is not outdated, I guess it was released on January 19, 2021.
https://docs.azul.com/zulu/zulurelnotes/ZuluReleaseNotes/ReleaseDetails1529-1337-1336-1145-1144-852-851-744-743-638.htm

michaelld:
Excellent explanation. I have no issues with trying this change out & seeing what happens. I don't think a lot of folks use SWIG or JAVA, and even fewer use SWIG-JAVA (I don't; it just comes with SWIG ...). Thx!

PR #10088, opened 2021-02-22
go: change bootstrap process and support arm64

foamkeen:
Switching from using Go 1.4 as a build toolchain to binary Go in order to be able to build for darwin-arm64

* add go-bootstrap subport that uses binary Go release
* add support for arm64

Closes: https://trac.macports.org/ticket/60942

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.1 20D74
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/lang/go/Portfile
+++ b/lang/go/Portfile
@@ -176,7 +176,7 @@ if {${os.platform} eq "darwin" && ${os.major} < 11} {
 subport ${name}-bootstrap {
     # unset build dependencies
     depends_build
-    distfiles           ${distname}.darwin-${build_arch}${extract.suffix}
+    distfiles           ${distname}.darwin-${GOARCH}${extract.suffix}
     build {}
     destroot {
         set grfdir ${destroot}${GOROOT_FINAL}
--- a/lang/go/Portfile
+++ b/lang/go/Portfile
@@ -9,6 +9,7 @@ legacysupport.newest_darwin_requires_legacy 13
 name                go
 epoch               2
 version             1.16
+revision            1
 categories          lang
 platforms           darwin freebsd linux
 license             BSD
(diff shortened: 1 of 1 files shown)


herbygillot:
Thank you for contributing this, @foamkeen. This seems to do the job.

However since it downloads a full binary Go installation into the bootstrap directory as a subport, we end up with 194-or-so MB installed that will see no use after the MacPorts Go installation is built and installed. This was the case with `go-1.4` as well, but it was lighter-weight and also MacPorts is able to build `1.4` purely from source.

So perhaps it may probably make sense to just have this binary Go install exist only within the build process' temporary work directory and delete it afterwards, instead of having it as a formal sub-port?

But if we don't care about any of the above, then we can just merge this PR so that Go can build on arm64.

Thoughts, @ci42, @ryandesign, @kencu?

foamkeen:
Just realised that this change will break i386 support. But since Golang doesn't ship `darwin-386` binaries anymore for latest releases we would need to keep `go-1.4` build dependency just for i386.

foamkeen:
As I understand this port "requires OS X 10.7 or greater" which in turn requires x86-64 CPU (or Apple Silicon) so i386 support can be safely dropped. @herbygillot, @ci42 what do you think? Can I remove i386 from `supported_archs`?

`supported_archs` in golang portgroup needs to be changed as well, should I open a separate pull request for that or can it be changed in this one?

ryandesign:
> this port "requires OS X 10.7 or greater"

It doesn't even work on 10.9. https://trac.macports.org/ticket/58935

herbygillot:
Go ARM64 support has been added with [this commit](https://github.com/macports/macports-ports/commit/1f387629e89699a2bc594cfdd0d101cbdacff056).

For ARM hosts, the binary dist is used in a temporary capacity to do the build, but works as it does historically (bootstrapping using Go 1.4) for all other architectures.

@ryandesign, this won't work for universal builds, but it's a start at least for ARM64 users.

Thank you for your submission here @foamkeen; we can probably close this PR.

herbygillot:
(Looks like the buildbot is currently down, so folks won't see this change via `port selfupdate` until it's back online.)

ryandesign:
> Looks like the buildbot is currently down

Uh oh. Let me see what's up.

herbygillot:
> > Looks like the buildbot is currently down
> 
> Uh oh. Let me see what's up.

Thanks @ryandesign.  Go has now built successfully on the ARM64 buildbot: https://build.macports.org/builders/ports-11_arm64-builder/builds/13138/steps/install-port/logs/stdio

Closing this PR.

PR #10109, opened 2021-02-24
dialog: update to 1.3-20210117

iay:
Allows the port to compile under Big Sur.
Addresses https://trac.macports.org/ticket/61498.

#### Description

Allows the port to compile under Big Sur.
Addresses https://trac.macports.org/ticket/61498.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.1 20D74
Xcode 12.4 12D4e

###### Verification 
Have you

- [x] tested basic functionality of all binary files? (Got it to show a dialog with `dialog --yesno "Hello" 10 30`, but I'm not familiar with dialog in general.)

### DIFF
--- a/devel/dialog/Portfile
+++ b/devel/dialog/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                dialog
 epoch               20150920
-version             1.3-20180621
+version             1.3-20210117
 categories          devel
 license             LGPL-2.1
 maintainers         nomaintainer
@@ -24,14 +24,18 @@ master_sites        ftp://ftp.invisible-island.net/pub/dialog/ \
 #worksrcdir ${name}-${version}
 extract.suffix      .tgz
 
-checksums           rmd160  af3134df81324175bc95f9bcc4ac691a2a04d094 \
-                    sha256  4a4859e2b22d24e46c1a529b5a5605b95503aa04da4432f7bbd713e3e867587a \
-                    size    529123
+checksums           rmd160  e55b51d9c9e8836328be7f1dcddcdfe739f86fcb \
+                    sha256  3c1ed08f44bcf6f159f2aa6fde765db94e8997b3eefb49d8b4c86691693c43e1 \
+                    size    558091
 
+depends_build       port:pkgconfig
 depends_lib         port:ncurses
 
 configure.args      --mandir=${prefix}/share/man \
                     --with-ncursesw
+
+test.run            yes
+test.target         check
 
 if {${name} == ${subport}} {
 


iay:
Rebased

kurthindenburg:
A few comments in case you were not aware of:
1. If you want to close that ticket it needs to be "Closes" or any of the words listed below
2. make sure you're building/testing/installing subports - in this case libdialog
3. if you have time, try to get the tests to work (test.run yes)
4. it looks like it wants pkgconfig - investigate if you have time.
   configure: WARNING: pkg-config is not installed

otherwise looks OK to me - I can commit it as-is if you want.

https://trac.macports.org/wiki/CommitMessages

iay:
> A few comments in case you were not aware of:
> 
> 1. If you want to close that ticket it needs to be "Closes" or any of the words listed below

That was intentional, as I wanted to get the other person who had run into problems with something dependent on `dialog` to verify that his problem was gone before closing. But what you say is sensible, I can always comment on a closed Trac ticket. I'll change the commit message.

> 2. make sure you're building/testing/installing subports - in this case libdialog

I wasn't aware of that. Subports are beyond my experience, unfortunately, and I can't see anything in the documentation to say how I'd do that from my branch.

Normally, I'd say `sudo port` in the source directory and then `build` (or whatever). It seems as if adding the subport name (in this case, `build libdialog` actually works off the pristine tree, not my branch, and therefore the old sources)

If you have a clue to offer, I can look into this. I'm not trying to be the port maintainer here (I don't have any direct knowledge or interest in `dialog` as such, it's a dependency for something I use, and I have no idea whether the subport has ever worked.  But I'd be happy to expand my level of knowledge. 

> 3. if you have time, try to get the tests to work (test.run yes)

Seems to need both of the following:

```
test.run yes
test.target check
```

It passes (or is at least silent) on the main `dialog` port.

> 4. it looks like it wants pkgconfig - investigate if you have time.
>    configure: WARNING: pkg-config is not installed

It wasn't complaining anywhere visible on my build, so I missed that. In fact some other port had installed it for me so the message wasn't even in the log. Uninstalling a few things I could reproduce this, and then make it go away by adding:

```
depends_build port:pkgconfig
```

> otherwise looks OK to me - I can commit it as-is if you want.

I'll revise my commit for the things above. If you can get back to me on the subport question, I can look into that too.

Thanks for your help.

iay:
Changed commit message to close the Trac ticket. Added test configuration and build dependency on `pkgconfig`.

kurthindenburg:
'sudo port -td build subport' - in the case libdialog - you might check if tests work when only building libdialog

'port info dialog' will show subports: Sub-ports:            libdialog

iay:
> 'sudo port -td build subport' - in the case libdialog - you might check if tests work when only building libdialog

I'm still missing something, I think.

I'm in the `dialog` directory, on my branch. `Portfile` contains: `version 1.3-20210117`.

I execute `sudo port -td build libdialog`. Quite early on, the output includes:

```
DEBUG: Changing to port directory: /opt/local/var/macports/sources/rsync.macports.org/macports/release/tarballs/ports/devel/dialog
```

So, it's not using the `Portfile` from my current directory. Later this is confirmed by:

```
Executing:  cd "/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_dialog/libdialog/work/dialog-1.3-20180621" && ./configure --prefix=/opt/local --mandir=/opt/local/share/man --with-ncursesw 
```

So, it's (failing to) build `dialog-1.3-20180621`, not my revised port.

kurthindenburg:
Simpliest way is to be in the directory that contains your Portfile; then doing 'sudo port -td build' will pick up that Portfile.

iay:
> Simpliest way is to be in the directory that contains your Portfile; then doing 'sudo port -td build' will pick up that Portfile.

Right, that's what I have always done in the past, and it works just fine. It only seems to build the main port, though, and (as far as I can tell) does not build or do anything else with the subport.

As I say, I am completely out of my depth here because I don't have any experience with subports. So I am probably missing something.

I will try just overwriting the rsync'ed sources version of the Portfile with mine and see if I can get that to work.

kurthindenburg:
If you're in the directory w/ your Portfile and 'sudo port build' works then 'sudo port build libdialog' should also work.
This doesn't need to hold up this pull request - let me look at it again

iay:
> If you're in the directory w/ your Portfile and 'sudo port build' works then 'sudo port build libdialog' should also work.

You'd think so, but as I said above the latter actually builds from the rsync'ed tree, not from the current directory. Maybe that's some kind of bug?

Anyway, I tried copying the revised `Portfile` over the one in the rsync'ed tree, cleaned everything and `sudo port build libdialog` now works, as does `sudo port test libdialog`. Weirdly, `sudo port install libdialog` appears to work but and `port  installed|grep dialog` shows the right version number but `port list requested|grep dialog` shows the one from 2018.

> This doesn't need to hold up this pull request - let me look at it again

I think the above gives me confidence that merging this should be fine, now, in that `libdialog` will install.

kurthindenburg:
I'll commit once the builds have finished - thanks

What does 'port file libdialog'  say?  I wonder if you need to do a 'sudo portindex' to pick up your original Portfile

port installed | grep dialog  
  dialog @1.3-20210117_0 (active)
  libdialog @1.3-20210117_0 (active)

iay:
> I'll commit once the builds have finished - thanks

Cheers.

>I wonder if you need to do a 'sudo portindex' to pick up your original Portfile

That seems to have been at least part of my problem, thanks. Having indexed my tree and set the sources configuration, asking to build `libdialog` build the right thing.

There is still some weirdness, but I guess what `requested` does it look in each of the source trees for the current version seen there:

```
$ port installed |grep dialog
  libdialog @1.3-20210117_0 (active)
$ port list requested |grep dialog
Warning: The 'list' action only shows the currently available version of each port. To see installed versions, use the 'installed' action.
libdialog                      @1.3-20210117   devel/dialog
libdialog                      @1.3-20180621   devel/dialog
```

PR #10117, opened 2021-02-27
dart-sass: new port

queensferryme:
#### Description

[dart-sass](https://github.com/sass/dart-sass) is "the reference implementation of Sass, written in Dart".

Closes: https://trac.macports.org/ticket/57549
Closes: https://trac.macports.org/ticket/58223

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.2 20D80
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/www/dart-sass/Portfile
+++ b/www/dart-sass/Portfile
@@ -1,0 +1,34 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        sass dart-sass 1.32.8
+categories          www textproc
+platforms           darwin
+supported_archs     x86_64
+license             MIT
+maintainers         @queensferryme openmaintainer
+description         The reference implementation of Sass, written in Dart.
+long_description    Dart Sass is the primary implementation of Sass, which means \
+                    it gets new features before any other implementation. It's fast, \
+                    easy to install, and it compiles to pure JavaScript which makes it \
+                    easy to integrate into modern web development workflows.
+homepage            https://sass-lang.com/dart-sass
+
+checksums           rmd160 cd25a9414aee360c246a375cc7e3237815e881f3 \
+                    sha256 bff9997005e6548a3f8588b39eccdf47bde9b68dca4ff70896ce4df7102cd71d \
+                    size   354010
+
+use_configure       no
+
+depends_build       port:dart-sdk
+build               {
+    system -W ${worksrcpath} "pub get"
+    system -W ${worksrcpath} "dart compile exe -Dversion=${version} bin/sass.dart"
+}
+
+destroot {
+    xinstall ${worksrcpath}/bin/sass.exe ${destroot}${prefix}/bin/sass
+}
+


harens:
Hi there @queensferryme 👋, thank you for your first contribution to MacPorts! 🥳 It's great to have you.

For your first MacPorts PR, you did a great job on the portfile. If possible, it would be preferable if we could build it from source following the [README instructions](https://github.com/sass/dart-sass#from-source) rather than repackaging the binary. [`dart-sdk`](https://ports.macports.org/port/dart-sdk/summary) can be added as a lib dependency and we'll have to download the dart deps at build time, which isn't ideal but I don't think we have a choice.

Pinging @amake since he has a lot more experience with dart portfiles than I do.

A small nitpick from me, you might find it useful to take a look at the [example portfile docs](https://guide.macports.org/#development.examples). Generally things like `maintainers` and `categories` are closer to the top, but that's easily fixable.

If you have any questions or need help on anything, feel free to ask. Thanks for your work on this @queensferryme.

_(I read your very interesting [article](https://qufy.me/post/从-homebrew-迁移到-macports/) on macOS package managers. Thank you for choosing MacPorts and for helping to make it even better. 👍)_

queensferryme:
Hi @harens, thanks very much for your welcoming and informative comment🎉
I will try build `dart-sass` from source and read the [example portfile docs](https://guide.macports.org/#development.examples) to fix the `maintainers` and `categories` fields, as you suggested.
BTW, as I am new here and lack due experience, I am expecting a few more commits with trial and error. Should I use `git commit --amend` and force push those new commits (in an effort to minimize commits), or just let it go?

amake:
- I'm not aware of any ports that build Dart source. If the output is a standalone binary then dart-sdk should probably be `depends_build` rather than `depends_lib`.
- As @harens mentioned we discourage fetching sources outside of the fetch phase. See the golang and cargo portgroups for examples of how far we go to avoid doing so. However since we lack any such infrastructure for Dart I think we can allow it here for now.
- I generally don't bother caring about the order of metadata as long as it's reasonably consistent with other ports.
- Feel free to manage your branch however you like. I often amend and force-push. In the end the changes should be squashed (by whoever does the final commit).

www/dart-sass/Portfile
```diff
@@ -0,0 +1,28 @@
...
+platforms           darwin
+license             MIT
+maintainers         @queensferryme openmaintainer
+description         The reference implementation of Sass, written in Dart.
+long_description    ${description}
+homepage            https://sass-lang.com/dart-sass
+
+fetch.type          git
+git.url             https://github.com/sass/dart-sass.git
+git.branch          ad4a1692a8166a16fc57fc6f16fc5f0b15dba165
```
harens, line 17:
```tcl

github.setup        sass dart-sass 1.32.8
```

I recommend looking at the [GitHub PG](https://guide.macports.org/#reference.portgroup.github), which can help simplify things. This also deals with the `name` and `version`, which can then be removed.

www/dart-sass/Portfile
```diff
@@ -0,0 +1,28 @@
...
+PortSystem          1.0
+
+name                dart-sass
+version             1.32.8
+categories          www textproc
+platforms           darwin
+license             MIT
+maintainers         @queensferryme openmaintainer
+description         The reference implementation of Sass, written in Dart.
+long_description    ${description}
```
harens, line 12:
```suggestion
long_description    Dart Sass is the primary implementation of Sass, which means \
                    it gets new features before any other implementation. It's fast, \
                    easy to install, and it compiles to pure JavaScript which makes it \
                    easy to integrate into modern web development workflows.
```

nitpick: Maybe something along these lines from the homepage?

www/dart-sass/Portfile
```diff
@@ -0,0 +1,28 @@
...
+
+fetch.type          git
+git.url             https://github.com/sass/dart-sass.git
+git.branch          ad4a1692a8166a16fc57fc6f16fc5f0b15dba165
+
+use_configure       no
+
+depends_build       port:dart-sdk
+build.cmd           pub get && dart compile exe
+build.target        bin/sass.dart
```
harens, line 23: At least for me, `sass --version` doesn't provide any output. I find adding `build.args          -Dversion=${version}` helps.

Also if we do make use of `dart2native`, I recommend just defining your own build stage similar to some of the yarn ports, which also download deps at build time (e.g. [hyper](https://github.com/macports/macports-ports/blob/ddf03cd26ce1985f2c2837c649c65ad58cbecbcb/aqua/hyper/Portfile)).

harens (commented):
Thank you for your work on this @queensferryme! *(you nailed the force push btw)* 😎

I left a few minor comments below. I think we might find it helpful to look at the dart-sass [Homebrew formula](https://github.com/sass/homebrew-sass/blob/master/sass.rb). For instance, they make use of `dart2native` on x86_64 (see [`build_arch`](https://guide.macports.org/#internals.configuration-files.macports-conf)), which we might find useful to do as well.

As always, let me know if you have any questions. Thanks again for this portfile. 👍

www/dart-sass/Portfile
queensferryme: fixed, thx for the advice.

www/dart-sass/Portfile
queensferryme: fixed, thx for the advice.

www/dart-sass/Portfile
queensferryme: According to the latest dart official [documentation](https://dart.dev/tools/dart-compile), `dart2native` is now replaced by `dart compile`. Adding the `-Dversion=${version}` argument does resolve the problem, however.

queensferryme:
I've resolve the issues mentioned above in this commit. However, I believe a tiny drawback exists: `dart compile exe` only works on 64-bits machine. I have explicitly add the `supported_archs x86_64` line to indicate this situation.

As the [`dart-sdk`](https://github.com/macports/macports-ports/blob/269a72eda15e93ca2b8ea8f0794f6b181852ef06/lang/dart-sdk/Portfile) provided by MacPorts only support x86_64 too at the moment, I think it's acceptable to drop support for `dart-sass` on other architectures currently.

www/dart-sass/Portfile
harens: > dart2native is now replaced by dart compile

Didn't know that. Thanks @queensferryme.

> Adding the -Dversion=${version} argument does resolve the problem, however.

Yay! 🥳

harens (approved):
Lgtm. 👍 Thanks @queensferryme for your contribution! I don't know if @amake has any final thoughts, but this looks fine to me.

amake (approved):
LGTM!

PR #10123, opened 2021-02-28
textproc/qpdf: trivial update to 10.2.0

adfernandes:


textproc/qpdf/Portfile
```diff
@@ -3,11 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        qpdf qpdf 10.1.0 release-qpdf-
-revision            1
-checksums           rmd160  8d20c952e323800174533b3cb7a6c42860d530fe \
-                    sha256  862c144e4516302327cea908f2879131cc8198b10d3d3a90ef7bc006a915120d \
-                    size    18528427
+github.setup        qpdf qpdf 10.2.0 release-qpdf-
```
kurthindenburg, line 6: you should keep the revision line and change it to 0

ryandesign:
Let's make sure in the future to squash such changesets into a single commit before they go to master. The "Squash and merge" button in the web interface can do it, or you can use the git command line.

adfernandes:
No problem - just surprised it wasn’t the default!

ryandesign:
I believe the default is whichever one you most recently used.

Sometimes PRs are submitted that deliberately provide multiple commits (for example, a PR that updates several related ports, one port per commit) and those types of PRs should be rebased, not squashed. The person merging the PR needs to push the correct button depending on the situation. The end goal has always been the commits that go to master should each represent a single logical change. We usually don't want intermediate steps that occurred in a PR in response to a review for example to appear on master.

PR #10127, opened 2021-03-01
add new port: janet

tsujp:
#### Description

Janet is a dynamic lisp-like language and bytecode vm, tiny, embeddable.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2 20D64
Xcode 12.4 12D4e

###### Verification

- [x] tried a full install with `sudo port -vst install`? (the `t` flag is breaking the build on M1 machines causing tar to have failed permissions)

### DIFF
--- a/lang/janet/Portfile
+++ b/lang/janet/Portfile
@@ -1,0 +1,28 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           makefile 1.0
+
+github.setup        janet-lang janet 1.15.3 v
+categories          lang
+platforms           darwin
+license             MIT
+
+maintainers         {@tsujp wz.ht:jc+macports} openmaintainer
+description         A dynamic language and bytecode vm
+long_description    Janet is a functional and imperative programming \
+                    language and bytecode interpreter. It is a lisp-like \
+                    language, but lists are replaced by other data structures. \
+                    The language also supports bridging to native code written in C, \
+                    meta-programming with macros, and bytecode assembly.
+
+homepage            https://janet-lang.org/
+
+checksums           rmd160 dfe40985067b851b15519fbb278e44a1254c4553 \
+                    sha256 061ca6dec7684d824b9a74f738d546ff620c41608e87ebe988bb68f548f39b88 \
+                    size   598864
+
+destroot.keepdirs   ${destroot}${prefix}/lib/${name}
+
+test.run            yes


harens:
Hi there @tsujp. 👋 Thank you for your first contribution to MacPorts! 🥳 It's great to have you here.

Just looking through the portfile, you did a much better job than my very first MacPorts PR https://github.com/macports/macports-ports/pull/7925. I thought I'd just make a few changes [on my fork](https://github.com/harens/macports-ports/blob/janet/lang/janet/Portfile) which you might find interesting.

- Would you be ok adding your GitHub handle to the maintainers? This then alerts you if someone modifies the file.
- Generally, the long description gives a bit more detail on what the port actually is, so I thought I'd add some more info.
- I recommend using the [GitHub PG](https://guide.macports.org/chunked/reference.portgroup.html#reference.portgroup.github), which simplifies things (no need for `distname` or `master_sites`). It also fixes this port's livecheck.
- Can I ask how the `destroot.keepdirs` helps in this scenario, since for me it builds fine without it. Obviously though, you know more about this port than I do so let me know if I'm missing anything.
- nitpick: Could you change the commit title to something like `janet: new port`?

I'd also be interested to see the log file relating to your trace mode error on M1 machines to see if I can help. I don't have an M1 machine to test on but I'll see if I can give you a hand.

As I said before, for your first PR you've done an amazing job on this portfile. ✨Thank you for your contribution and let me know if you have any questions (I'd be happy to help).

tsujp:
Hey @harens thanks for the feedback. Sure thing I'll use the GitHub PG. Perhaps I shall try and contribute to getting a SourceHut PG going too.

1) Sure, I tend to prefer SourceHut as a hedge but I am still active on GitHub for notifications (that I don't throttle).
2) Nice long description.
3) See above.
4) Janet also provides a package manager `jpm` which has a default path specified within the [Makefile](https://git.sr.ht/~bakpakin/janet/tree/master/item/Makefile#L29). This is empty upon build so unless `destroot.keepdirs` is given it will be removed and when you attempt to install a package, say `jpm install mendoza` you'll get a missing folder error. Keeping this empty directory prevents that.
5) Will do.

Trace with `port install -vstd janet` I don't have anymore and have changed my macports prefix to a non-root one and `-t` complains about no user `macports` anymore. It was with `tar` though and it being unable to make a directory (despite there being correct permissions).

harens:
> It was with tar though and it being unable to make a directory (despite there being correct permissions).

In terms of getting this port merged (since you've done an amazing job 👍), I'm not too concerned about this error since it's probably not arm-specific, and both myself and the tests build it fine.

> Janet also provides a package manager jpm which has a default path specified within the Makefile. This is empty upon build so unless destroot.keepdirs is given it will be removed and when you attempt to install a package, say jpm install mendoza you'll get a missing folder error. Keeping this empty directory prevents that.

Thanks for the explanation and the example `jpm install mendoza`, which was very useful. Just to note that if the users install any packages, they will not be removed if janet is uninstalled since the packages weren't around during the destroot stage. It might be useful to do something similar to the [npm](https://github.com/macports/macports-ports/blob/aa41f0eb8a57f2cafa9aac004f1c8a89c8191c25/devel/npm6/Portfile) ports and make a note of this.

Also just so you're aware, on a standard installation of MacPorts, sudo is required for jpm install of global packages. This is similar to https://trac.macports.org/ticket/33175, and is noted in their [docs](https://janet-lang.org/docs/jpm.html):

> On Windows, sudo is not required. Use of sudo on POSIX systems depends on whether you installed janet to a directory owned by the root user.

 This makes sense but we should warn against installing global packages in the notes (alike to npm).

cjones051073:
Is this PR ready to merge ?

harens:
I think we should just add some "npm-like" notes and then we're done here.

Example:

https://github.com/macports/macports-ports/blob/aa41f0eb8a57f2cafa9aac004f1c8a89c8191c25/devel/npm6/Portfile#L92-L97

tsujp:
Will add this later today @harens (forgot to do this).

PR #8259, opened 2020-08-29
rust 1.45.2: Bugfix - use rust internal llvm - #61082

fredowski:
#### Description

This patch removes the current configure option to use the
default llvm which is installed by macports. Instead the
rust internal llvm which is provided and maintained by the
rust team is used. The problem

https://trac.macports.org/ticket/61082

is fixed by using the rust internal llvm. Probably there
are more changes which the rust team considers worth maintaining
its own fork of llvm. So maybe it is in general a good idea to use the rust internal llvm instead of the default llvm.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G14019
Xcode 10.1 10B61 

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL?
- [x] checked your Portfile with `port lint`?
- [x] tried existing tests with `sudo port test`?

there are two tests failing but the same tests fail without the patch
```
failures:
    [ui] ui/macros/restricted-shadowing-legacy.rs
    [ui] ui/wait-forked-but-failed-child.rs

test result: FAILED. 10219 passed; 2 failed; 72 ignored; 0 measured; 0 filtered out
```
- [x] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?
Run "Hello World!"...

PR #10163, opened 2021-03-06
rust: update to 1.50.0

herbygillot:
#### Description

Updates Rust to 1.50.0 and uses its internal LLVM instead of MacPorts'. Build time is increased, but Rust will build and run in the way it's meant to.

When we previously tried to do this in #8259, Rust installed LLVM into lib, making it possible that other software would compile against that Rust-specific LLVM, but that no longer looks to be the case today.

Includes @fabius ' ARM64 changes from #9904 

###### Tested on

macOS 10.15.7 19H524
Xcode 12.4 12D4e

###### Verification 
Have you


herbygillot:
Build failed for 10.15 while building `grafana`.  We are missing a ton of prebuilt packages for 10.15 (probably because the 10.15 buildbot has been down for a while), so Azure CI had to rebuilt everything it didn't have binary packages for.  It spent more than 6 hours building dependencies and failed before even getting to Rust.

Builds did pass for macOS 11, so this looks good to merge.  Testing one other change locally and will do so shortly.

g5pw:
Thanks for this, @herbygillot!

herbygillot:
> Thanks for this, @herbygillot!

No problem at all @g5pw.  Just added ARM64 cargo changes [here](https://github.com/macports/macports-ports/commit/912990978bd30cc420f083457af559fd33c59b51), so hopefully this all works. Then we can begin enabling ARM64 for Rust ports.

PR #10169, opened 2021-03-07
py-igraph: update dependencies

szhorvat:
#### Description

 * update dependencies given the changes in python-igraph 0.9
 * build with external (instead of bundled) dependencies when appropriate
 * remove py27 subport as Python 2.7 is no longer supported
 * remove variants
 * See: https://trac.macports.org/ticket/62399

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C504

###### Verification 
Have you

### DIFF
--- a/python/py-igraph/Portfile
+++ b/python/py-igraph/Portfile
@@ -6,11 +6,12 @@ PortGroup           python 1.0
 name                py-igraph
 python.rootname     python-igraph
 version             0.9.0
+revision            1
 categories-append   math science
 platforms           darwin
 license             GPL-2+
 
-python.versions     27 36 37 38 39
+python.versions     36 37 38 39
 
 maintainers         {snc @nerdling} {gmail.com:szhorvat @szhorvat} openmaintainer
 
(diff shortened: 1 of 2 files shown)


python/py-igraph/Portfile
```diff
@@ -22,49 +23,44 @@ long_description    Python interface to the igraph library for network analysis
...
+                            port:glpk \
+                            port:libxml2 \
+                            port:SuiteSparse_CXSparse
 
     depends_build-append    port:py${python.version}-setuptools \
-                            path:bin/cmake:cmake \
-                            port:libtool \
-                            port:autoconf \
-                            port:automake \
+                            port:cmake \
```
mf2k, line 38: This should be left as a path dependency so that cmake-devel can also be used.

mf2k (changes requested):

mf2k:
The Trac ticket reference should be added to the commit message. Add this to the end:

`See: https://trac.macports.org/ticket/62399
`

szhorvat:
@mf2k Thanks for the comments! The cmake dependency change was accidental, sorry about that.  Please take another look.

Regarding the `bison`/`flex` build dependencies, I added these for consistency, but these are technically not necessary as macOS does ship old (but compatible) versions of these.

mf2k (approved):

PR #10170, opened 2021-03-07
igraph: make compatible with old clang

szhorvat:
#### Description

 * make compatible with old clang: do not error on unknown diagnostic options

This PR attempts to make this port compatible with OS X 10.7, where the included compiler does not support the `-Wno-varargs` option, which causes the build to fail.  https://build.macports.org/builders/ports-10.7_x86_64-builder/builds/45964/steps/install-port/logs/stdio

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C504

###### Verification 
Have you

### DIFF
--- a/math/igraph/Portfile
+++ b/math/igraph/Portfile
@@ -5,6 +5,7 @@ PortGroup           cmake 1.1
 PortGroup           github 1.0
 
 github.setup        igraph igraph 0.9.0
+revision            0
 github.tarball_from releases
 
 categories          math science devel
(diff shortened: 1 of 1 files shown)


math/igraph/Portfile
```diff
@@ -5,6 +5,7 @@ PortGroup           cmake 1.1
 PortGroup           github 1.0
 
 github.setup        igraph igraph 0.9.0
+revision            1
```
mf2k, line 8: It sounds like this change only fixes the a build issue on older versions of MacOS (OS X)? If so, the revision should stay at 0 so people who have built it successfully don't have to needlessly rebuild it.

mf2k (changes requested):

math/igraph/Portfile
szhorvat: This is fixed now.

szhorvat:
@mf2k I just noticed a problem: when running `port test igraph` _without_ igraph being already installed, the tests fail due to not finding the shared library. This happens only when running it through MacPorts, but not when building and testing igraph with the same options that I set in the Portfile. Does this kind of issue sound familiar? Perhaps it occurred with other CMake-based ports as well?

mf2k:
Note that we now prefer to keep the revision line and set it to 0 rather than remove it. 

To answer your question, it is my understanding that the port must be installed for the test phase to work. So I may not understand the problem you are trying to address.

mf2k:
IOW - your result sounds like the expected result.

szhorvat:
> Note that we now prefer to keep the revision line and set it to 0 rather than remove it.

Re-added (the only change in the last force push).

PR #10178, opened 2021-03-08
ripgrep: enable ARM64-native build

herbygillot:


textproc/ripgrep/Portfile
```diff
@@ -5,11 +5,14 @@ PortGroup           github 1.0
...
 
 github.setup        BurntSushi ripgrep 12.1.1
+revision            1
+
 categories          textproc
 platforms           darwin
 maintainers         {raimue @raimue} \
                     openmaintainer
 license             MIT
+supported_archs     x86_64 arm64
```
raimue, line 15: But this removes support for all unlisted archs... That seems like an uninteded side effect?

raimue (commented):
Can you provide a build log that attempts a build with arm64, but then switches to x86_64 instead? Is your `build_arch` in /opt/local/etc/macports/macports.conf configured to `arm64`?

If `supported_archs` is not set, MacPorts will use whatever is supported by the macOS SDK. For SDK >= 11.0 that should be `arm64 x86_64` and from my understanding there should be no need to set any additional Portfile options.

kencu:
@herbygillot -- using supported_archs doesn't add support for a new arch in some way.

It removes support for all archs that are not specifically listed.

So as @raimue says, if this is not building on arm64 now, adding this will not fix it.

herbygillot:
@kencu @raimue My fault here for misunderstanding `supported_archs`' purpose - closing this PR.  Thank you.

PR #10180, opened 2021-03-08
py-h5py: Restore 27 and 36 (down-rev) for deps.

eborisch:
Maintainer update.

Addressing ticket [#62359](https://trac.macports.org/ticket/62398), but still leaving py35 (EOL) flavors off. I don't really want to have the 27 version anymore, but until tools using h5py all upgrade, we're stuck.


eborisch:
Hrmm. The (old) py27 compatible version of h5py doesn't work with the latest hdf5.

eborisch:
Closing. Opens a can of worms to keep multiple hdf5 versions around.

PR #10182, opened 2021-03-09
kitty: update to 0.19.3

jpeeler:
Update Kitty to latest version of v0.19.3, released 12/18/2020.

The dropped patch doesn't have to go away necessarily, but I [agree with the upstream author](https://github.com/kovidgoyal/kitty/pull/2477) that there's no reason to use the older C standard.


jpeeler:
As far as I can tell, it does still build without C11. For example, OpenBSD doesn't use that flag:
https://github.com/kovidgoyal/kitty/commit/11942ed6dc394f4efc890f5624df198bb57738f8.

Are you suggesting it's cleanest to update with the line you suggested or should I put back the c99 flag (or leave it as is)?

PR #10189, opened 2021-03-09
igraph: require C++11

szhorvat:
#### Description

 * Require C++11 in order to fix building on old systems

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Verification
Have you

### DIFF
--- a/math/igraph/Portfile
+++ b/math/igraph/Portfile
@@ -29,6 +29,9 @@ checksums           rmd160  297c06d217d6c6059b721ec0aa5d59e096004585 \
 test.run            yes
 test.target         check
 
+compiler.cxx_standard \
+                    2011
+
 # Build options for igraph:
 #  - Do not use ccache to build this port unless MacPorts tells it to.
 #  - Build a shared library.


jmroot:
This may not be sufficient, as based on the 10.6 log, it looks like cmake doesn't think that macports-clang-* compilers support LTO. AFAIK they do. Not sure if there's a way to convince cmake of this, or it you just need to disable LTO in that case.

jmroot:
OK, I'll merge this as-is then.

szhorvat:
@jmroot Perhaps this can also be improved in MacPorts's compiler selection mechanism, but I don't know if it's worth the trouble. In order to make LTO work, we need to tell CMake to use a matching ar and ranlib:

https://cmake.org/cmake/help/v3.20/variable/CMAKE_LANG_COMPILER_AR.html

https://cmake.org/cmake/help/v3.20/variable/CMAKE_LANG_COMPILER_RANLIB.html

(Note that using CMAKE_AR and CMAKE_RANLIB didn't work for me.)

For example, if I used `cmake-mp-11`, I had to use `llvm-ar-mp-11` and `llvm-ranlib-mp-11` in order for to LTO to work.

kencu:
I will point out one thing that can sometimes be an issue.

I have found through experimentation that for some and perhaps for all builds, cctools needs to be built against the llvm version that is the highest one you intend to use. 

If cctools is built against llvm3.7, say, and you are building with clang-11/llvm-11, the objects being produced by llvm-11 might not be able to be fully handled by cctools built against llvm-3.7.

I have not been able to automate this easily, but in the libtapi port, for example, I found a way to force the LLVM I needed to use for LTO using DYLD_LIBRARY_PATH.

What I do is rebuild cctools against the newest llvm I have installed.

kencu:
See https://github.com/macports/macports-ports/blob/ae3822fd6583d2b633cf5bac79d406bad520c948/devel/libtapi/Portfile#L88

PR #10193, opened 2021-03-09
AutoRaise: new port

lhaeger:
###### Tested on

macOS 10.15.7 19H524
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/sysutils/AutoRaise/Portfile
+++ b/sysutils/AutoRaise/Portfile
@@ -1,0 +1,35 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=portfile:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           xcode 1.0
+
+github.setup        lhaeger AutoRaise 1.9.0 launcher-v
+revision            0
+
+categories          sysutils
+platforms           macosx
+
+license             GPL-3+
+maintainers         {@lhaeger gmx.net:lothar.haeger} openmaintainer
+
+description         AutoRaise raises application windows to the front and gives them focus on mouse hover
+long_description    ${description} after a configurable delay. There is also an option to warp the mouse \
+                    to the center of the activated window, using the cmd-tab key combination for example.
+
+github.tarball_from archive
+
+checksums           rmd160  1add5f43cb447dd20074ba2ec84ca4079340f2c7 \
+                    sha256  3e8da5ecee6f7ab244942bf451e06f1425da74ed909076b082bb80e0a5a88b47 \
+                    size    374254
+
+xcode.configuration Release
+
+variant cli \
+    requires        universal \
+    description     installs the AutoRaise cli as ${prefix}/bin/${name} \
+    {
+    post-destroot {
+        xinstall -m 0755 -W ${worksrcpath} &{name} ${prefix}/bin
+    }
+}


sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,32 @@
...
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        sbmpost AutoRaise 1.5 v
+
+categories          sysutils
+platforms           macosx
+
+license             {GPL-2 GPL-3}
```
ryandesign, line 11: The license header in the source code file says:
```
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 2 of the License, or
 * (at your option) any later version.
```
So this should be:
```suggestion
license             GPL-3+
```

sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,32 @@
...
+maintainers         {@lhaeger gmx.net:lothar.haeger} openmaintainer
+
+description         AutoRaise raises application windows to the front and gives them focus on mouse hover
+long_description    ${description} after a configurable delay. There is also an option to warp the mouse \
+                    to the center of the activated window, using the cmd-tab key combination for example.
+
+homepage            htps://github.com/sbmpost/AutoRaise
```
ryandesign, line 18: You misspelled https, but also the github portgroup already sets this correctly for you.
```suggestion
```

sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,32 @@
...
+
+checksums           rmd160  6569b47fe2570fe8f079ea06c96c43314a9eb4df \
+                    sha256  a42fe824a3e1f8cb51529b52321ba138bef966a5fa496ac538c466046ced3969 \
+                    size    28613
```
ryandesign, line 22: The default file type of the github portgroup is "tarball" which is deprecated. Please use "archive" type instead. This will change the checksums.
```suggestion
github.tarball_from archive

checksums           rmd160  3a343bf8c281f5ea0ed0013ec747b7357d743fed \
                    sha256  19fd38c7e898437b8a310f4039f0d059c657436164f24c0d154a33189dbcfcf0 \
                    size    28566
```

sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,32 @@
...
+
+use_configure       no
+
+build.cmd           g++
```
ryandesign, line 26: [Use the right compiler](https://trac.macports.org/wiki/UsingTheRightCompiler), please.
```suggestion
build.cmd           ${configure.cxx}
```

sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,32 @@
...
+build.args          -O2 -Wall -fobjc-arc -o AutoRaise AutoRaise.mm -framework AppKit
```
ryandesign, line 27: Use the right flags, please.
```suggestion
variant universal {}
build.args          ${configure.cxxflags} [get_canonical_archflags cxx] -Wall -fobjc-arc -o AutoRaise AutoRaise.mm -framework AppKit
```

sysutils/AutoRaise/Portfile
lhaeger: So if a source allows GPL-2 or higher, at MacPorts we choose GPL-3 by default? Or should that rather be "GPL-2+"?

sysutils/AutoRaise/Portfile
lhaeger: Yep, double mistake: a typo and redundancy after I changed the source back to upstream instead of my fork. Thank!

sysutils/AutoRaise/Portfile
lhaeger: Shouldn't we change the default then?

lhaeger:
Can someone please have a look at why https://github.com/macports/macports-ports/pull/10193/checks?check_run_id=2325013330 failed to install MacPorts?

ryandesign:
All PRs fail right now because the BinTray service that we rely on is [conducting scheduled maintenance](https://status.bintray.com/incidents/4mzph0y8l6hk).

sysutils/AutoRaise/Portfile
ryandesign: My mistake; I meant `GPL-2+`.

sysutils/AutoRaise/Portfile
herbygillot: > Shouldn't we change the default then?

That will affect a ton of ports.  It's doable, but still a big impact.  About 1,680 ports seem to use the `tarball_from` default right now. We could update them all to set `github.tarball_from` to `tarball`, then change the _portgroup_ default to `archive` right after that. But ideally you would still want to verify that each port still works the same after changing them all.

After doing that, over time as ports get updated to new versions, then you change the source from `tarball` to `archive` (or even `releases`) as you update a port to a new version (or other routine change).

herbygillot:
Closing and re-opening to re-run CI.

sysutils/AutoRaise/Portfile
ryandesign: There may also be portfiles in third-party repositories using the github portgroup. If we change the defaults of the portgroup that would break those portfiles. An option is to introduce a new github 1.1 portgroup with archive as the default type, and portfiles can migrate to that version of the portgroup over time. However there are many other aspects of the github 1.0 portgroup that I'm no longer delighted with that it might be a good idea to fix at the same time. Even better would be to integrate fetch from github functionality into MacPorts base. I've thought about how best to do this many times over the years but have not arrived at a satisfactory answer. So for now we have what we have.

sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=portfile:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        sbmpost AutoRaise 1.6 v
+
```
herbygillot, line 7: Can you add a line for revision?
```
revision            0
```

lhaeger: Sure.

sysutils/AutoRaise/Portfile
lhaeger: Well, small steps may be better than not moving at all. Do you have a list of the piled-up github issues somewhere? Maybe enough items could be implemented in a v1.1 portgroup with reasonable effort to justify the version bump, leaving the not-so-easy stuff for a later v1.2 or for integration into macports base.

sysutils/AutoRaise/Portfile
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=portfile:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        sbmpost AutoRaise 1.6 v
+revision            0
+
+categories          sysutils
+platforms           macosx
```
herbygillot, line 10: Shouldn't this be `platforms darwin`?

lhaeger: Maybe, but since I do not have open-source Darwin to verify and the docs say `platforms` is "not interpreted in any way by MacPorts at this time; it is purely informational for users.", I'd stick with the confirmed "macos", rather than guessing or hoping "darwin" would also work.
If you can confirm AutoRaise runs on pure Darwin as well, please go ahead and update the Portfile (it's marked `openmaintainer` after all)

lhaeger: The Portfile reference has the following:

> Ports for software that does not require macOS-specific features should use darwin. Most ports use this value on the presumption that they would work on Pure Darwin, even if that has not been attempted. Ports for software that is known to require macOS-specific features should set this to macosx.

I have no idea which part of macOS is open-sourced and how to determine if a port requires macOS-specific features. If you know the details, please feel free to change the platform to darwin any time.

lhaeger: I guess this is now obsolete with the switch to xcode...

herbygillot:
@lhaeger can we squash this into a single commit?

lhaeger:
> @lhaeger can we squash this into a single commit?

Yes we can!

herbygillot:
OK, the build failure is because of an usupported SDK version:

```
<unknown>:0: error: Swift does not support the SDK 'MacOSX10.14.sdk'
```

I'm not sure if MacPorts has a method for failing the build on a macOS SDK that is too old...

lhaeger:
> OK, the build failure is because of an usupported SDK version:
> 
> ```
> <unknown>:0: error: Swift does not support the SDK 'MacOSX10.14.sdk'
> ```
> 
> I'm not sure if MacPorts has a method for failing the build on a macOS SDK that is too old...

According to https://developer.apple.com/de/support/xcode/ Swift should be fine with macOS 10.14.3 and XCode 10.2 or later. Are the buildbots on earlier versions of 10.14?

herbygillot:
> According to https://developer.apple.com/de/support/xcode/ Swift should be fine with macOS 10.14.3 and XCode 10.2 or later. Are the buildbots on earlier versions of 10.14?

Unfortunately the CI jobs that run pull requests here in Github are not the same as the buildbots that officially built the ports when the PR is committed to master.  Let's just clean this up back into one commit and get it merged.

ryandesign:
> @lhaeger can we squash this into a single commit?

Note that you can do this in the GitHub web interface using the "Squash and merge" button. If you use this, remember to edit the commit message to remove extraneous information.

ryandesign:
> remember to edit the commit message to remove extraneous information.

The commit message for 1fc9790430b52d4ecc91ba123b3259e6a908b94d should have been just "AutoRaise: New port, version 1.9.0".

PR #10213, opened 2021-03-13
libarchive: drop unneeded patch

chrstphrchvz:
#### Description
@kencu added this patch in 625eeb4906 with a note asking to drop it when updating. But the port got updated to 3.5.1 (b46f9f7b9b) without the patch being dropped.

Is the port fine for affected users who installed the patched version of 3.5.1, such that the `revision` does not need incrementing?

###### Verification 
Have you


cjones051073:
@kencu are you happy with this ?

kencu:
It does appear that upstream committed a strnlen workaround here https://github.com/libarchive/libarchive/commit/c581b317cc71f964b2446b5534ebcc07be2f7cab that made it into 3.5.1.

I haven't tested it yet, but presumably our patched-in strnlen would therefore be no longer needed, indeed. Let me try it and confirm success before you commit this, just because I'm careful.

kencu:
And by the way, can I say what a beautiful example that is about how to check for and work around a missing function, in both cmake and autotools.

kencu:
confirmed on 10.5 Leopard that our patch is indeed no longer needed.

I cannot 100% say that no files are changed on disk by removing this patch, and although I suspect that our replacement was optimized out by the compiler I can't prove it.

So I would say a revbump would be simpler than trying to prove a revbump is not needed.

kencu:
IMHO this falls below the level of change that would usually require maintainer approval, as it only affects < 10.7 where Toby doesn't live. It's an openmaintainer port, so should be good to go with it.

PR #10227, opened 2021-03-14
libfaiss: Update to version 1.7.0, add Python 39

essandess:
libfaiss: Update to version 1.7.0, add Python 39

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H524
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/math/libfaiss/Portfile
+++ b/math/libfaiss/Portfile
@@ -4,9 +4,10 @@ PortSystem          1.0
 PortGroup           cmake 1.1
 PortGroup           github 1.0
 
-github.setup        facebookresearch faiss 1.6.4 v
+github.setup        facebookresearch faiss 1.7.0 v
 name                libfaiss
 epoch               1
+revision            0
 
 categories          math
 license             BSD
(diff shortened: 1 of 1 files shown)


math/libfaiss/Portfile
```diff
@@ -4,9 +4,9 @@ PortSystem          1.0
 PortGroup           cmake 1.1
 PortGroup           github 1.0
 
-github.setup        facebookresearch faiss 1.6.4 v
+github.setup        facebookresearch faiss 1.7.0 v
 name                libfaiss
-epoch               1
+epoch               0
```
cjones051073, line 9: The epoch can never be reduced. Please set this back to 1.

cjones051073 (changes requested):

math/libfaiss/Portfile
essandess: Whoops thanks that was supposed to be `revision`. Fixed.

math/libfaiss/Portfile
```diff
@@ -4,9 +4,10 @@ PortSystem          1.0
 PortGroup           cmake 1.1
 PortGroup           github 1.0
 
-github.setup        facebookresearch faiss 1.6.4 v
+github.setup        facebookresearch faiss 1.7.0 v
 name                libfaiss
 epoch               1
+revision            0
```
cjones051073, line 10: If the revision is not set it is implicitly 0, so this line is unnecessary.

cjones051073 (changes requested):

math/libfaiss/Portfile
essandess: Unnecessary but convenient to keep around? Lots of ports set `revision 0`, and it's a nice reminder in the workflow. Personally, I prefer it there.

cjones051073 (approved):

PR #10238, opened 2021-03-14
macports-libcxx

kencu:
The macports-libcxx port uses the libc++*.dylibs and headers installed with clang-11, modifies them slightly to use full path id's and linking (rather than the rpath links clang-11 makes for them) and then installs them in a reasonable location for ports to use them.

The mkvtoolnix demonstrator is just the first port that needed these features that I have tried.The minor mods made in the mkvtoolnix portfile could easily be a PortGroup once we sort out a reliable incantation that works broadly.

Perhaps consider this a technology preview for the moment, technical improvements or feedback welcome. I have no doubt there are other methods that might be used to do this better or improve this.

### DIFF
--- a/lang/macports-libcxx/Portfile
+++ b/lang/macports-libcxx/Portfile
@@ -1,0 +1,60 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                macports-libcxx
+categories          lang
+platforms           darwin
+maintainers         {kencu @kencu} openmaintainer
+license             MIT
+homepage            http://macports.org
+
+description         provides a newer libc++ from llvm for older systems
+long_description    This port installs a recent libc++ from llvm \
+                    to use on older systems instead of the system libc++.
+version             11
+revision            0
+
+depends_build       port:clang-${version}
+
+master_sites
+fetch {}
+checksum {}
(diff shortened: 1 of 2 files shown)


kencu:
Screen Shot 2021-03-14 at 12 58 15 PM

kencu:
see https://trac.macports.org/ticket/62426 for further background and possible issues that might arise

kencu:
FYI I have not as yet tried this on macOS < 10.14.

kencu:
Here's the (maybe) magic block to make it work:

```
if {${os.platform} eq "darwin" && ${os.major} < 19} {

    # requires a newer libc++ than the system can provide
    depends_lib-append        port:macports-libcxx
    configure.cppflags-append -D_LIBCPP_DISABLE_AVAILABILITY
    configure.cxxflags-append  -nostdinc++ -I${prefix}/include/libcxx/v1
    configure.ldflags-append  -L${prefix}/lib/libcxx
    compiler.blacklist-append {clang < 1100} {macports-clang-3.[3-9]} {macports-clang-[5-8].0}
}
```

kencu:
10.11 is happy. So far looks like a win.

Screen Shot 2021-03-14 at 8 24 51 PM

kencu:
I suppose I might patch ```#define _LIBCPP_DISABLE_AVAILABILITY``` right into the copied headers in the ```__config``` file, as the whole idea is to use them with the availability disabled. I do that on 10.6.8 already.

then there would be one less issue to always have to remember to do.

kencu:
I will look at this a bit more, and then decide whether to just push it directly.

PR #10248, opened 2021-03-15
macports-libcxx: new port

kencu:
Use like this, as an example:
```
if {${os.platform} eq "darwin" && ${os.major} < 19} {

    # compiler floor currently set at clang-9.0 and Xcode 11 from Mojave, as those are tested and work
    compiler.blacklist-append {clang < 1100} {macports-clang-3.[3-9]} {macports-clang-[5-8].0}

    # requires a newer libc++ than the system can provide
    depends_lib-append        port:macports-libcxx
    configure.cxxflags-append -nostdinc++ -I${prefix}/include/libcxx/v1
    configure.ldflags-append  -L${prefix}/lib/libcxx
}
```

### DIFF
--- a/lang/macports-libcxx/Portfile
+++ b/lang/macports-libcxx/Portfile
@@ -1,0 +1,70 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                macports-libcxx
+categories          lang
+platforms           darwin
+maintainers         {kencu @kencu} openmaintainer
+license             NCSA
+homepage            https://libcxx.llvm.org
+
+description         provides a newer libc++ from llvm for older systems
+long_description    This port installs a recent libc++ from llvm \
+                    to use on older systems instead of the system libc++.
+
+# for now, we will leverage the already-built libc++ in the appropriate clang port
+# later, we can build this independently if we choose to do so, much like libtapi
+
+# the clang-11 version in use when this port is updated will be used
+version             11.1.0
+set clangversion    11
+revision            0
+
+depends_build       port:clang-${clangversion}
+
+master_sites
+fetch {}
+checksum {}
+use_configure no
+build {}
+
+destroot {
+
+    xinstall -d ${destroot}${prefix}/include/libcxx
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/c++/v1 ${destroot}${prefix}/include/libcxx/
+
+    # disable Apple libc++ availability tests, as we're using a new libc++ with these headers
+    system -W ${destroot}${prefix}/include/libcxx/v1 "patch -p0 < ${filespath}/patch-disable-availabilty.diff"
+
+    xinstall -d ${destroot}${prefix}/lib/libcxx
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/libc++.1.0.dylib ${destroot}${prefix}/lib/libcxx/libc++.1.0.dylib
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -id ${prefix}/lib/libcxx/libc++.1.0.dylib libc++.1.0.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -delete_rpath @loader_path/../lib libc++.1.0.dylib"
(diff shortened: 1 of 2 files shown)


kencu:
@cjones051073 -- can you run your eyes over this to triple-check for any obvious flaws?

I don't think you have any machines that can actually test this out, but it works from 10.11 to 10.14 to build at least some software that needs std::filesystem, std::optional, etc, etc..

As time goes by perhaps further refinements will be noted.

lang/macports-libcxx/Portfile
```diff
@@ -0,0 +1,70 @@
...
+
+    # disable Apple libc++ availability tests, as we're using a new libc++ with these headers
+    system -W ${destroot}${prefix}/include/libcxx/v1 "patch -p0 < ${filespath}/patch-disable-availabilty.diff"
+
+    xinstall -d ${destroot}${prefix}/lib/libcxx
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/libc++.1.0.dylib ${destroot}${prefix}/lib/libcxx/libc++.1.0.dylib
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -id ${prefix}/lib/libcxx/libc++.1.0.dylib libc++.1.0.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -delete_rpath @loader_path/../lib libc++.1.0.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -change @rpath/libc++abi.1.dylib ${prefix}/lib/libcxx/libc++abi.1.dylib libc++.1.0.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "ln -s libc++.1.0.dylib libc++.1.dylib"
```
cjones051073, line 45: small nit-pick, but tcl has a built-in `ln` command, so these lines don't need to be run through `system`.

lang/macports-libcxx/Portfile
```diff
@@ -0,0 +1,70 @@
...
+    system -W  ${destroot}${prefix}/lib/libcxx/  "ln -s libc++.1.dylib libc++.dylib"
+
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/libc++abi.1.0.dylib ${destroot}${prefix}/lib/libcxx/libc++abi.1.0.dylib
```
cjones051073, line 48: Can't these be a link instead of a copy ? As port providing the original is a dependent, its guaranteed to remain available.

lang/macports-libcxx/Portfile
```diff
@@ -0,0 +1,70 @@
...
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -id ${prefix}/lib/libcxx/libc++abi.1.0.dylib libc++abi.1.0.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "install_name_tool -delete_rpath @loader_path/../lib libc++abi.1.0.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "ln -s libc++abi.1.0.dylib libc++abi.1.dylib"
+    system -W  ${destroot}${prefix}/lib/libcxx/  "ln -s libc++abi.1.dylib libc++abi.dylib"
+
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/libc++.a ${destroot}${prefix}/lib/libcxx/libc++.a
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/libc++abi.a ${destroot}${prefix}/lib/libcxx/libc++abi.a
+    copy  ${prefix}/libexec/llvm-${clangversion}/lib/libc++experimental.a ${destroot}${prefix}/lib/libcxx/libc++experimental.a
```
cjones051073, line 56: likewise, these I think could be just a link.

lang/macports-libcxx/Portfile
cjones051073: forget it... its a build dep. I guess thats reasonable.

lang/macports-libcxx/Portfile
cjones051073: forget comment...

kencu:
Thanks for looking. Sounds like the "big stuff" is decipherable, and makes enough sense to you to be reasonable. This one is a bit different, but hopefully will fill a need.

I may get to building it directly from source in this Port, but that is a whole re-do of building clang-11, so ... for today I'll just leverage clang-11's build.

I expect there will be more opinions on this as time goes by -- probably 30s after it's committed. But I think this is needed, and I don't think I can fix it up more than this.

I will try to see if I can make tcl's build-in ```ln``` command obey my bidding. It is sometimes a bit of a beast. I tried it first, of course, and could not (within my patience time span) sort out how to make it do what I wanted (make a very simple link to the next lib without any path or other info in it, just a link from A->B in that folder from one lib to the other.

But for another day, I'll look at that again.

PR #10326, opened 2021-03-18
gssdp, gupnp, gupnp-igd: update to fix building gupnp-igd on Big Sur

michaellass:
#### Description

**CAUTION:** This pull request is probably not ready to merge. However, it fixes building gupnp-igd on Big Sur (https://trac.macports.org/ticket/61713), so I'd like to share it here.

Some insight: All of these projects switched to meson as a build system. Available configuration options can be seen in the meson_options.txt files of the projects:
- https://gitlab.gnome.org/GNOME/gssdp/-/blob/master/meson_options.txt
- https://gitlab.gnome.org/GNOME/gupnp/-/blob/master/meson_options.txt
- https://gitlab.gnome.org/GNOME/gupnp-igd/-/blob/master/meson_options.txt

Why I don't consider this ready to merge:
- I had to remove the gobject_introspection port group because it tries using a configure flag not understood by the new meson build system. However, from above links it looks like it's turned on by default. So maybe adding a dependency on gobject-introspection is sufficient now? But the port group seems to do much more than that.
- I removed patches and configuration flags without knowing what they were supposed to do. The patches looked to me like fixes to the build system, so maybe this is fine.
- I removed the python27 variant of gupnp-igd without providing any replacement.
- I have no way of testing the functionality. I only fixed it because gupnp-igd it is an indirect dependency of meld (and other ports).
- A revbump of dependent ports is probably required. On my system `port rev-upgrade` told me that libnice needs a rebuild but there may be more.

Feel free to request changes or to just close this PR if a different implementation is preferred.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

- [ ] tried existing tests with `sudo port test`? (no tests available)
- [ ] tried a full install with `sudo port -vst install`? (without trace mode only)
- [ ] tested basic functionality of all binary files? (**no!**)


Gcenx:
Yes you'd need to add `gobject-introspection` to depends_lib

michaellass:
I'll close this PR since the bug has been fixed differently by now, there never has been any response by the maintainer and I have no way to properly test this change.

PR #10329, opened 2021-03-19
orcus: change to C++17 standard

Tatsh:
Based on the author's update
See https://gitlab.com/orcus/orcus/-/commit/1579c12111074a50f2307b8fa749a68749af11ba

Also see https://trac.macports.org/ticket/62484

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/orcus/Portfile
+++ b/devel/orcus/Portfile
@@ -24,7 +24,8 @@ checksums           rmd160  8c18cf3459d8054dd36215db16c99720dcc55a3c \
                     size    7945943
 
 use_autoreconf      yes
-configure.cxxflags-append -std=c++14
+compiler.cxx_standard 2017
+configure.cxxflags-append -std=gnu++17
 configure.args      --disable-python \
                     --disable-static \
                     --disable-werror \


ryandesign:
You must also specify `compiler.cxx_standard 2017` if you want MacPorts to guarantee that a C++17-compatible compiler is used.

devel/orcus/Portfile
```diff
@@ -24,7 +24,7 @@ checksums           rmd160  8c18cf3459d8054dd36215db16c99720dcc55a3c \
                     size    7945943
 
 use_autoreconf      yes
-configure.cxxflags-append -std=c++14
+configure.cxxflags-append -std=gnu++17
```
cjones051073, line 27: The c++ standard needs to be specified with a line like

```
compiler.cxx_standard 2017
```

That should also automatically add the cxxflag, so it likely will no longer be required.

cjones051073 (changes requested):

cjones051073 (approved):

devel/orcus/Portfile
Tatsh: It needs both to work as far as I can tell.

ryandesign: Specifying `compiler.cxx_standard` does not automatically add any `-std` flag. It is expected that the project's build system will do so. If it does not, we add it manually in the portfile.

PR #10332, opened 2021-03-19
fix conflict between libshout & libshout2

cooljeanius:
#### Description

I have fixed the conflict between libshout and libshout2 marked by @ryandesign in 3b4af07 by adjusting the configure arguments passed to libshout. This means that libshout now installs into different directories, but as no ports depend upon it, that ought to be okay. I also disabled the livecheck for libshout while I was at it.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 11.2.2 20D80
Xcode 12.4 12D4e

###### Verification 
Have you

- [x] checked your Portfile with `port lint`? (Note that `port -v lint --nitpick` suggests renaming the patchfiles, and while I've done that with my local version of the port, I didn't feel like including that in this pull request)
- [x] tried existing tests with `sudo port test`? (N/A)
- [x] tested basic functionality of all binary files? (N/A; just libraries)

### DIFF
--- a/audio/libshout/Portfile
+++ b/audio/libshout/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                libshout
 version             1.0.9
-revision            0
+revision            1
 checksums           rmd160  b2369212c732bff9565602a47a2a68f9fd4f6e98 \
                     sha256  191215b71ca6b5a6e999e91acd699dbbcdf53c5be1d928a25acb395d8442a426 \
                     size    169553
@@ -13,7 +13,6 @@ categories          audio net
 platforms           darwin
 license             GPL-2+
 maintainers         nomaintainer
-conflicts           libshout2
 
 description         Data and connectivity lib for the icecast server
 
@@ -38,6 +37,12 @@ patchfiles          apple-patches \
 # Fix implicit declaration of functions and fix build with ccache.
 use_autoreconf      yes
 
-configure.args      --enable-shared \
+configure.args-append \
+                    --enable-shared \
                     --enable-static \
+                    --includedir=${prefix}/include/shout1 \
+                    --libdir=${prefix}/lib/shout1 \
                     --with-pic
+
+# This is an old version, so disable livecheck
+livecheck.type      none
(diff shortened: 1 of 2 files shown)


audio/libshout/Portfile
```diff
@@ -38,6 +37,15 @@ patchfiles          apple-patches \
...
+                --enable-shared \
+                --enable-static \
+                --program-suffix=1 \
+                --libdir=${prefix}/lib/shout1 \
+                --includedir=${prefix}/include/shout1 \
+                --x-includes=${prefix}/include \
+                --x-libraries=${prefix}/lib
+
+# This is an old version, so disable livecheck
+livecheck.type  none
```
ryandesign, line 51: Ideally the indentation of the entire portfile should be consistent. The rest of the lines are indented to 20 spaces but your new lines are indented to 16 spaces. Could the new lines be changed to 20 spaces to match?

Often we like to sort configure args alphabetically. The existing args were sorted. Can we sort the new args?

You added `--program-suffix=1` but I don't think this is necessary because this port does not install any programs.

You added `--x-includes=${prefix}/include` and `--x-libraries=${prefix}/lib` but I don't think this is necessary because this port does not depend on any X11 libraries.

```suggestion
configure.args-append \
                    --enable-shared \
                    --enable-static \
                    --includedir=${prefix}/include/shout1 \
                    --libdir=${prefix}/lib/shout1 \
                    --with-pic

# This is an old version, so disable livecheck
livecheck.type      none
```

ryandesign (commented):
Thanks.

Your commit message should be reformatted to follow our guidelines: the first line should list the port name(s) followed by a colon and then a short description up to a max of 50 characters total, then a blank line and subsequent paragraphs as needed wrapped at 72 characters.

cooljeanius:
Could you fix up the commit message for me? I'm having some trouble doing it from my computer...

mascguy:
Eric, I squashed your commits, and updated the comment per Ryan. Merging.

cooljeanius:
> Eric, I squashed your commits, and updated the comment per Ryan. Merging.

Thanks!

PR #10336, opened 2021-03-19
libffi: use Apple-provided sources on Big Sur

danchr:
Apple added some symbols to the `libffi` included with Big Sur, and PyObjC uses some of them. To fix this, I just switched the port to use that on Big Sur. I'm not sufficiently familiar with `libffi` to know whether there's a more clever fix.

Fixes: https://trac.macports.org/ticket/62475

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/libffi/Portfile
+++ b/devel/libffi/Portfile
@@ -41,6 +41,11 @@ if {${os.platform} eq "darwin" && ${os.major} >= 20} {
     # Remove once the included config.guess correctly identifies arm64
     use_autoreconf      yes
     autoreconf.args     -fvi
+
+    # based on homebrew, needed for PyObjC
+    patchfiles-append   patch-libffi-3.3-arm64.diff
+
+    set revision [expr ${revision} + 1]
 }
 
 if {${os.platform} eq "darwin" && ${os.major} < 11} {
--- a/devel/libffi/files/patch-libffi-3.3-arm64.diff
+++ b/devel/libffi/files/patch-libffi-3.3-arm64.diff
@@ -1,0 +1,754 @@
+Only in libffi-3.3.new: autom4te.cache
+diff -ur configure.ac libffi-3.3.new/configure.ac
+--- configure.ac	2019-11-23 06:58:22.000000000 -0700
++++ libffi-3.3.new/configure.ac	2021-01-04 19:48:35.000000000 -0700
+@@ -176,6 +176,35 @@
+     ;;
+ esac
+ 
++AC_CACHE_CHECK([whether compiler supports pointer authentication],
++   libffi_cv_as_ptrauth, [
++   libffi_cv_as_ptrauth=unknown
++   AC_TRY_COMPILE(,[
++#ifdef __clang__
++# if __has_feature(ptrauth_calls)
++#  define HAVE_PTRAUTH 1
++# endif
++#endif
++
++#ifndef HAVE_PTRAUTH
++# error Pointer authentication not supported
++#endif
++		   ],
++		   [libffi_cv_as_ptrauth=yes],
++		   [libffi_cv_as_ptrauth=no])
++])
++if test "x$libffi_cv_as_ptrauth" = xyes; then
++    AC_DEFINE(HAVE_PTRAUTH, 1,
++	      [Define if your compiler supports pointer authentication.])
++fi
++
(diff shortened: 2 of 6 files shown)


cjones051073:
Its not ideal to use different sources on different OSes. Do you know if Apple have submitted their changed back to upstream ?

danchr:
> Its not ideal to use different sources on different OSes. Do you know if Apple have submitted their changed back to upstream ?

AFAICT they have at least two pending pull requests:

https://github.com/libffi/libffi/pull/620
https://github.com/libffi/libffi/pull/621

They changes are rather involved, so extracting them individually to patches isn't straightforward. One option is to simply generate a patch containing all of Apple's changes.

kencu:
https://raw.githubusercontent.com/Homebrew/formula-patches/06252df03c68aee70856e5842f85f20b259e5250/libffi/libffi-3.3-arm64.patch

cjones051073:
I would propose then changing this PR to stick with the original source, but use the homebrew patch to apply the update. It should also, assuming it builds on all OSes be applied across the board, not just to big sur.

kencu:
Someone should run the test suites on at least a couple of systems.. 

This patch is right from Jeremy's PRs (and another issue post is useful too). It is likely to work on arm64 (that is the only place Homebrew is using it). The other systems? Who knows. 

Last time libffi was updated I ran the test suites on 10.4 up through every system, and moved several fixes upstream.

danchr:
I've updated this with a quick attempt at applying the patch from HomeBrew. As far as I can tell, importing PyObjC does work, and no longer gives this error:

```
ImportError: dlopen(/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/objc/_objc.cpython-39-darwin.so, 2): Symbol not found: _ffi_find_closure_for_code_np
  Referenced from: /opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/objc/_objc.cpython-39-darwin.so
  Expected in: flat namespace
 in /opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/objc/_objc.cpython-39-darwin.so
```

The reason the error passed silently was that MacPorts' `libffi` is installed as a bare `ffi.h`, but PyObjC picked up the system header by including `ffi/ffi.h`.

ryandesign:
The patch is enormous, presumably primarily because of unrelated changes due to regenerating with automake 1.16.3 instead of the 1.16.1 used originally. Could the patch be made as small as possible, please, for example by generating with automake 1.16.1?

danchr:
> The patch is enormous, presumably primarily because of unrelated changes due to regenerating with automake 1.16.3 instead of the 1.16.1 used originally. Could the patch be made as small as possible, please, for example by generating with automake 1.16.1?

I've removed the autogenerated parts, and enabled `use_automake` instead.

coderprodigy:
> I've updated this with a quick attempt at applying the patch from HomeBrew. As far as I can tell, importing PyObjC does work, and no longer gives this error:
> 
> ```
> ImportError: dlopen(/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/objc/_objc.cpython-39-darwin.so, 2): Symbol not found: _ffi_find_closure_for_code_np
>   Referenced from: /opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/objc/_objc.cpython-39-darwin.so
>   Expected in: flat namespace
>  in /opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/objc/_objc.cpython-39-darwin.so
> ```
> 
> The reason the error passed silently was that MacPorts' `libffi` is installed as a bare `ffi.h`, but PyObjC picked up the system header by including `ffi/ffi.h`.

I am still facing this issue. I have updated PyObjC but the error is the same. What shall I do?

danchr:
Closing; I managed to force using the system libffi.

danchr:
> Closing; I managed to force using the system libffi.

Fix applied in 217d8731dcc6247f67004de48d3f4760b07e33db.

PR #10339, opened 2021-03-19
py-astropy-healpix: update to 0.6

lpsinger:
#### Description

Update py-astropy-healpix to version 0.6.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H524
Xcode 12.4 12D4e

### DIFF
--- a/python/py-astropy-healpix/Portfile
+++ b/python/py-astropy-healpix/Portfile
@@ -4,19 +4,20 @@ PortSystem 1.0
 PortGroup python 1.0
 PortGroup github 1.0
 
-github.setup        astropy astropy-healpix 0.5 v
-name                py-${name}
+name                py-astropy-healpix
+version             0.6
+distname            astropy_healpix-${version}
 categories-append   science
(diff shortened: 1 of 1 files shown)


python/py-astropy-healpix/Portfile
```diff
@@ -4,19 +4,20 @@ PortSystem 1.0
 PortGroup python 1.0
 PortGroup github 1.0
 
-github.setup        astropy astropy-healpix 0.5 v
-name                py-${name}
+name                py-astropy-healpix
```
cjones051073, line 7: This change seems unnecessary. What was wrong with using the name variable as before ?

lpsinger: I'm dropping the GitHub port group for all but the older Python 2.7-3.5 subport because the way that Astropy affiliated packages use setuptools_scm does not play nicely with MacPorts.

PR #10344, opened 2021-03-20
shellcheck: update livecheck

Tatsh:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/shellcheck/Portfile
+++ b/devel/shellcheck/Portfile
@@ -1,7 +1,9 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           github 1.0
 
+github.setup        koalaman shellcheck 0.7.1 v
 categories          devel haskell
 
 license             GPL-3+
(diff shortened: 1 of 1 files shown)


Tatsh:
Related: https://github.com/macports/macports-ports/commit/3f55e163e05ed4c0a7595bfb5b394e0766039c86

devel/shellcheck/Portfile
```diff
@@ -30,7 +30,7 @@ if { ${os.platform} eq "darwin" && ${os.major} >= 20 } {
     homepage        https://github.com/koalaman/shellcheck
     livecheck.type  regex
     livecheck.url   https://github.com/koalaman/shellcheck/tags
-    livecheck.regex "archive/v(\[^\"]+)\\.tar\\.gz"
+    livecheck.regex "archive/refs/tags/v(\[^\"]+)\\.tar\\.gz"
```
cjones051073, line 33: Does this port need to specify the regex at all, the above looks like the default should work ok ?

cjones051073: Please try removing all the custom livecheck settings and see if it works. If not, then setting them via the github PG would be better.

Tatsh: I have refactored the port to use the PG including for livecheck and share the `github.setup` line.

PR #10388, opened 2021-03-24
py-tensorflow-metadata: Update to version 0.29.0, Use new bazel port …

essandess:
…group

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/python/py-tensorflow-metadata/Portfile
+++ b/python/py-tensorflow-metadata/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           python 1.0
 
-github.setup        tensorflow metadata 0.28.0 v
+github.setup        tensorflow metadata 0.29.0 v
 revision            0
 name                py-${github.author}-${github.project}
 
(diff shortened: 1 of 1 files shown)


python/py-tensorflow-metadata/Portfile
```diff
@@ -58,35 +56,46 @@ if {${name} ne ${subport}} {
...
     post-extract {
-        xinstall -d ${workpath}/bin ${bazel_build_proto_path}
+        xinstall -d \
+                    ${workpath}/bin \
+                    ${bazel_build_proto_path}
         ln -s ${python.bin} ${workpath}/bin/python
     }
 
     # replace with patch files after this repo is a little more stable
+    set bazel_build_opts [bazel::get_opts]
```
cjones051073, line 66: `bazel::get_opts` no longer exists in the bazel PG. It is split into `bazel::get_cmd_opts` and `get_build_opts`. Please make sure you have the latest version of the PG from git, and update accordingly.

cjones051073 (changes requested):

python/py-tensorflow-metadata/Portfile
cjones051073: I suspect the way it currently is in the port in git, following my last pushes is in fact correct here.

essandess: Lol, it was working because I was using your version. Sorry, I'll fix this commit.

essandess: P.S. I like the clever way you avoid having to redefine `build.cmd` and others so that Python building works.

PR #10390, opened 2021-03-25
bazel: Bugfix

essandess:
Fixes: https://trac.macports.org/ticket/62551

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/bazel/Portfile
+++ b/devel/bazel/Portfile
@@ -90,6 +90,9 @@ if { ${name} eq ${subport} } {
     compiler.blacklist-append {clang < 710}
     compiler.blacklist-append {macports-clang-[7-9].0}
 
+    # https://trac.macports.org/ticket/62551
+    build.env-append      EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk"
+
 } elseif { "bazel-2.0" eq ${subport} } {
     
     # Specific version for py-tensorflow - DO NOT CHANGE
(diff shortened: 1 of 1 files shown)


devel/bazel/Portfile
```diff
@@ -230,6 +230,9 @@ build.env-append  CC=${configure.cc} \
               TMPDIR=${workpath}/tmp \
            JAVA_HOME=${java.home}
 
+# https://github.com/bazelbuild/bazel/issues/7940#issuecomment-553447868
+build.env-append      EXTRA_BAZEL_ARGS="--host_javabase=@local_jdk//:jdk"
+
```
cjones051073, line 235: I would prefer if instead of appending this to the build env here, below where the build command is defined you only added it there, and also please only do this for the bazel-3.1 subport and not he others. Finally can you also please add a comment listing the trac ticket, instead of the upstream report.

cjones051073 (changes requested):

devel/bazel/Portfile
essandess: I tried that, but the build failed in the same way. I haven’t tracked it down but it appears that this must be an environment variable, not a shell variable (consistent with the issue comment). I’ll add this to `bazel-3.1` specifically.

essandess: I haven’t checked, but I’m pretty certain that the other `BAZEL_EXTRA_ARGS` isn’t set correctly in the original Portfile.
The bazel build docs [say](https://docs.bazel.build/versions/0.23.0/install-compile-source.html) this must be an environment variable.

essandess: Also, that same bazel documentation isn’t specific to version 3.1, so it’s likely this must be set for other versions as well, but I haven’t tracked them all down, only version 3.1.

cjones051073: OK thanks. I've merged and will take a look myself.

essandess:
@cjones051073 A couple of comments:

1. The bash variables defined in tcl's `bazel_build_env` are being provided as shell variables to the `bazel` command. It may be, like the environment variable `EXTRA_BAZEL_ARGS`, that all these should be environment variables too (not shell variables) so that bazel's java subprocess has access to them.
2. Ultimately, it would be good to add the `bazel` port to the `bazel` port group because bazel is used to build itself. See [haskell_cabal-1.0.tcl](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/haskell_cabal-1.0.tcl) and [haskell_stack-1.0.tcl](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/haskell_stack-1.0.tcl) for comparable examples.

cjones051073:
They already are set as environment variables. See lines 108 and 119.

PR #10404, opened 2021-03-25
clang-11: add lld build

kencu:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C504

Initial commit. Needs further testing, system coverage verification, probably a select file addition as well. 

But should be enough to get ZIG building, for those who need that immediately.


gpanders:
I can confirm that zig 0.7.1 build successfully with this change

kencu:
I have a successful install on 10.6, 10.14, and 10.15.

I think we can go ahead with this, and I'll fix up any buildbot wreckage that shows up, if any.

PR #10415, opened 2021-03-26
py-igraph: update to 0.9.1

szhorvat:
#### Description

 * update to 0.9.1
 * add test dependencies
 * only use LTO when available
 * remove obsolete patches
 * bison and flex build-time dependencies are no longer needed

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C504

###### Verification 
Have you


szhorvat:
I need a bit of help here. The `igraph` port has `compiler.cxx_standard   2011`, and uses the cmake portgroup.

The `py-igraph` port effectively has a bundled version of a _matching_ igraph, which it will compile. That requires C++11. But unlike with the `igraph` port, here cmake is not managed by MacPorts. It's called manually by `setup.py`. It's unclear to me if adding `compiler.cxx_standard   2011` will do anything here.

nerdling:
> It's unclear to me if adding `compiler.cxx_standard 2011` will do anything here.

If it's like the other compiler directives, this is mostly environment variables which should get passed along.

szhorvat:
> I'd recommend a quick glance at the portgroup file and grepping around to see if other python group Portfiles have `cxx_standard` in them.

Yes, many seem to have it, some without anything extra to try to control compiler selection. If this just sets `CC`, `CXX` (and perhaps other envvars) on old systems where Apple doesn't support C++11, then I think everything should be fine.

If you see no issues, I'd suggest merging this, and seeing how it fares on old systems (e.g. 10.6). Hopefully it works.

PR #10417, opened 2021-03-26
tmuxp: Update to 1.7.2 and add select port

judaew:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you


python/py-tmuxp/Portfile
```diff
@@ -2,15 +2,18 @@
 
 PortSystem          1.0
 PortGroup           python 1.0
+PortGroup           select 1.0
 
 name                py-tmuxp
-version             1.7.0
+version             1.7.2
+revision            1
```
mf2k, line 9: The `revision` needs to be set back to 0 when increasing the `version`.

python/py-tmuxp/Portfile
```diff
@@ -22,15 +25,35 @@ python.versions     27 36 37 38 39
 
 if {${name} ne ${subport}} {
     depends_build-append    port:py${python.version}-setuptools
-    depends_lib-append      port:py${python.version}-setuptools \
-                            port:py${python.version}-colorama \
+    depends_lib-append      port:py${python.version}-colorama \
                             port:py${python.version}-click \
```
mf2k, line 29: Let's keep these sorted alphabetically.

mf2k (changes requested):

python/py-tmuxp/Portfile
judaew: In the first commit, I update `py-tmux`, and the next commit I add support for the select port. It's why do I use the revision. Maybe does it not affect since this is one pull-request?

mf2k: That is not relevant. My comment still stands. The `revision` should be 0.

mf2k: The `revision` only gets incremented in specific cases. This is not one of them.

mf2k: Please read the [wiki reference](https://guide.macports.org/#reference.keywords) for it for the specifics.

mf2k (changes requested):

python/py-tmuxp/Portfile
judaew: Thanks for the explanation.

mf2k (approved):

PR #10442, opened 2021-03-27
nim: submission

gpanders:
#### Description

Add [nim](https://nim-lang.org) port.

Nim's install script wants to install everything under a `nim` subdirectory under whatever directory you provide to `install.sh`. So for example if we use `./install.sh /opt/local` this would create an `/opt/local/nim` directory, which we obviously don't want. So I copied what Go does and installed this directory under `/opt/local/lib` so that it creates `/opt/local/lib/nim`. Binaries (`nim`, `nimble`, `nimgrep`, and `nimsuggest`) are symlinked from that directory (e.g. `/opt/local/bin/nim -> /opt/local/lib/nim/bin/nim`).

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C505

###### Verification 
Have you

### DIFF
--- a/lang/nim/Portfile
+++ b/lang/nim/Portfile
@@ -1,0 +1,57 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                nim
+version             1.4.4
+revision            0
+license             MIT
+categories          lang
+platforms           darwin
+supported_archs     i386 x86_64 ppc64
+maintainers         {@gpanders gpanders.com:greg} \
+                    openmaintainer
+
+description         Nim programming language
+long_description    Nim is a statically typed compiled systems programming \
+                    language. It combines successful concepts from mature \
+                    languages like Python, Ada, and Modula. Its design focuses \
+                    on efficiency, expressiveness, and elegance (in that \
+                    order of priority).
+
+homepage            https://nim-lang.org
+
+master_sites        ${homepage}/download/
+
+checksums           rmd160  df2568109792112495aaed4273f222bf8623f95b \
+                    sha256  9f43f5da8f7130b137d59e49f3911f0c35cc1e94e68c333c50c59ef8b65dd6f6 \
+                    size    159095112
+
+use_configure       no
+
+variant doc description {Build HTML docs} {}
+
+build {
+    # See https://nim-lang.github.io/Nim/packaging.html
+    system -W ${worksrcpath} "./build.sh --os ${os.platform} --cpu ${os.arch}"
+    system -W ${worksrcpath} "./bin/nim c koch"
+    system -W ${worksrcpath} "./koch boot -d:release"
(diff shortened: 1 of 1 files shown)


lang/nim/Portfile
```diff
@@ -0,0 +1,56 @@
...
+
+PortSystem          1.0
+
+name                nim
+version             1.4.4
+revision            0
+license             MIT
+categories          lang
+platforms           darwin
+supported_archs     i386 x86_64 ppc64
```
herbygillot, line 11: You would only set `supported_archs` if you know for sure that a port only works with a given subset of architectures.

In this case, it looks like `nim` supports the whole hog, including **arm64**. So remove this and the right thing should get passed automatically as per the local architecture.

lang/nim/Portfile
```diff
@@ -0,0 +1,56 @@
...
+checksums           rmd160  df2568109792112495aaed4273f222bf8623f95b \
+                    sha256  9f43f5da8f7130b137d59e49f3911f0c35cc1e94e68c333c50c59ef8b65dd6f6 \
+                    size    159095112
+
+use_configure       no
+
+variant doc description {Build HTML docs} {}
+
+build {
+    system -W ${worksrcpath} "./build.sh --os ${os.platform} --cpu ${os.arch}"
```
herbygillot, line 35: Per my current understanding, you may want to use `${configure.build_arch}` here.

See https://github.com/macports/macports-ports/commit/84dd03127da4647acd1fea8d612dedde07c7e3f3

lang/nim/Portfile
```diff
@@ -0,0 +1,56 @@
...
+description         Nim programming language
+long_description    Nim is a statically typed compiled systems programming \
+                    language. It combines successful concepts from mature \
+                    languages like Python, Ada, and Modula. Its design focuses \
+                    on efficiency, expressiveness, and elegance (in that \
+                    order of priority).
+
+homepage            https://nim-lang.org
+
+master_sites        https://nim-lang.org/download/
```
herbygillot, line 24:
```
master_sites ${homepage}/download
```

herbygillot (changes requested):

lang/nim/Portfile
herbygillot: Actually I may be wrong on this one, disregard.

lang/nim/Portfile
gpanders: I was going off of their [README](https://github.com/nim-lang/Nim#compiling) which lists only x86, x86_64, and ppc64 for macOS. That’s for stable/tested versions only, so I suppose it does technically build for arm64 but it’s not officially supported by the Nim project (yet). What’s the call here in this case? Do we list *all* archs even if not officially supported, or do we stick with what upstream says?

lang/nim/Portfile
gpanders: Fixed

lang/nim/Portfile
herbygillot: Ah, in that case we should stick with upstream, but still remember to open things up (remove `supported_archs`) if the next version has official support for ARM64.

herbygillot (approved):

PR #10479, opened 2021-03-29
py-dload: Submission

essandess:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/python/py-dload/Portfile
+++ b/python/py-dload/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           python 1.0
+
+name                py-dload
+version             0.6
+platforms           darwin
+license             MIT
+maintainers         nomaintainer
+
+description         A multipurpose downloader for python > 3.6
+long_description    {*}${description}
+
+homepage            https://github.com/x011/dload
+
+master_sites        https://files.pythonhosted.org/packages/fc/30/41da7ccf08389e9e262dfef55d68acc5e066ab85bfdd26c9039fef57ead9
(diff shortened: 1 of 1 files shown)


python/py-dload/Portfile
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           python 1.0
+
+name                py-dload
+version             0.6
+platforms           darwin
+license             NULL
```
herbygillot, line 9: Please correct the license.

python/py-dload/Portfile
```diff
@@ -0,0 +1,50 @@
...
+maintainers         nomaintainer
+
+description         A multipurpose downloader for python > 3.6
+long_description    {*}${description}
+
+homepage            https://github.com/x011/dload
+
+master_sites        https://files.pythonhosted.org/packages/fc/30/41da7ccf08389e9e262dfef55d68acc5e066ab85bfdd26c9039fef57ead9
```
herbygillot, line 17: Is there a reason why we have to specify `master_sites` explicitly? Is the `python` portgroup failing to do something right here?

I see in this Portfile many things manually specified that are usually automatically picked up by the portgroup, such as `livecheck.url` and `extract.suffix`.

herbygillot (changes requested):

python/py-dload/Portfile
essandess: Thanks done.

python/py-dload/Portfile
essandess: The auto settings don't detect that download hash at PyPi or the whl download name, so both are necessary.

`livecheck` appears to work without these settings, so I deleted these lines.

herbygillot (approved):

PR #10523, opened 2021-03-31
nextdns: new port

gpanders:
#### Description

Add NextDNS client port.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C505

###### Verification 
Have you


net/nextdns/Portfile
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/nextdns/nextdns v1.11.0
```
herbygillot, line 6: Can we add `revision` here:
```
revision              0
```

We are wanting to add that to all ports by default.

net/nextdns/Portfile
```diff
@@ -0,0 +1,79 @@
...
+categories          net sysutils
+maintainers         {@gpanders gpanders.com:greg} openmaintainer
+license             MIT
+
+description         NextDNS CLI client
+
+long_description    NextDNS protects you from all kinds of security threats, \
```
herbygillot, line 13: This project has its own homepage, so would suggest adding a:

```
homepage             https://nextdns.io
```

net/nextdns/Portfile
herbygillot: Also, please fix the version prefix here.  It should be:
```
go.setup            github.com/nextdns/nextdns 1.11.0 v
```

and not
```
go.setup            github.com/nextdns/nextdns v1.11.0
```

gpanders: Whoops, I knew that and just forgot. Fixed.

herbygillot (approved):

herbygillot:
Thanks @gpanders.

PR #10528, opened 2021-03-31
sendmail_select: new port

gpanders:
#### Description

Add a new 'sendmail' select group for programs that implement the sendmail interface (e.g. msmtp).

If no option is selected, then this is essentially a no-op as the `sendmail` command will simply fallback to the default macOS version (`/usr/sbin/sendmail`).

###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H524
Xcode 12.4 12D4e

###### Verification 
Have you


gpanders:
@ra1nb0w

ra1nb0w:
I am not sure about the usefulness of this; specially if it is connected only with msmtp

gpanders:
> I am not sure about the usefulness of this; specially if it is connected only with msmtp

I think it is useful because there are other programs that just look for a binary called "sendmail" (e.g. `git send-email`). The `sendmail` program can be any program that implements the sendmail interface (which `msmtp` does). Aliasing `sendmail` to other programs allows the user to "plug in" their MTA of choice into that generic interface (this is also how it's done in e.g. Debian with `update-alternatives`).

It is not only connected to msmtp, that is just the mailer that I personally use so it's the only one I updated. I am not sure if there other MTAs available in MacPorts that implement the sendmail interface, but if there are they can also be added to this select group.

And in any case, it doesn't do any harm, since if the select group is set to `none` then it just uses the default macOS sendmail.

ra1nb0w:
I thought a bit and seems a good idea to push this. opensmtpd has a sendmail replacement therefore can be useful to include also that port. I don't find any other up-to-date sendmail replacement on `mail/`.
@ryandesign what do you think? (and sorry to bother you but I don't know anything about select ecosystem).

gpanders:
I added `opensmtpd` to the select group and also changed the select group symlink to be `sbin/sendmail` instead of `bin/sendmail`.

gpanders:
@ryanakca for opensmtpd

ryanakca:
> I think it is useful because there are other programs that just look for a binary called "sendmail" (e.g. `git send-email`). The `sendmail` program can be any program that implements the sendmail interface (which `msmtp` does). Aliasing `sendmail` to other programs allows the user to "plug in" their MTA of choice into that generic interface (this is also how it's done in e.g. Debian with `update-alternatives`).

I maintain OpenSMTPD in Debian. This is *not* what we do. Instead, each MTA provides the "[mail-transport-agent](https://packages.debian.org/en/sid/mail-transport-agent)" [virtual package](https://www.debian.org/doc/debian-policy/ch-binary.html#virtual-packages) and installs `/usr/sbin/sendmail` per [Debian Policy §11.6](https://www.debian.org/doc/debian-policy/ch-customized-programs.html#mail-transport-delivery-and-user-agents).

I don't think this proposal makes sense for several reasons:
1. This sounds like a poor man's implementation of OpenBSD's "[`mailwrapper`](https://man.openbsd.org/mailwrapper.8)" utility. In particular, if you allow for multiple MTAs to be installed, how are you going to ensure that associated utilities `newaliases`, `makemap`, etc., remain consistent with the selected MTA and don't conflict with each other? Are you going to also introduce  selectors for them too?
2. What is the use case for having multiple system-wide MTAs installed? (Debian expressly prohibits multiple MTAs from being installed simultaneously by [requiring conflicts between all MTA packages](https://www.debian.org/doc/debian-policy/ch-customized-programs.html#s-mail-transport-agents).)
3. More importantly: all (?) MTAs already directly install `sendmail`. Will all existing MTA ports have to rename their `sendmail` binary and add a dependency on this selector to provide the `sendmail` interface? The situation is different from, e.g., `python3`, which selects between differently named binaries `python3.8` and `python3.9`.

What you actually want is a subport `msmtp-mta` (à la [msmtp-mta in Debian](https://packages.debian.org/en/sid/msmtp-mta)) that just installs a symlink from `sendmail` to `msmtp`. From what I can tell, Debian's `msmtp-mta` packages uses `update-alternatives` to install the `sendmail` symlink to allow users to switch between different queueing interfaces for msmtp (e.g., vanilla `msmtp` vs `msmtpq` vs `msmtp-enqueue.sh`). It is *not* used to switch between system-wide MTAs from different packages (again, these different packages cannot be installed simultaneously).

gpanders:
Thanks @ryanakca for your feedback!

I double checked my Debian installation and you are right, I was mistaken. `/usr/sbin/sendmail` is simply symlinked to `/usr/bin/msmtp`, it does not use the alternatives system. I also agree that having multiple implementations of sendmail is not useful or valuable.

I will close this PR and just submit a new one that emulates the `msmtp-mta` package, since ultimately that is what I'm trying to achieve.

PR #10543, opened 2021-04-01
py-pytest-mpl: Bugfix build dependency

essandess:
Fixes: https://github.com/macports/macports-ports/pull/10497#issuecomment-811057735

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/python/py-pytest-mpl/Portfile
+++ b/python/py-pytest-mpl/Portfile
@@ -2,11 +2,12 @@
 
 PortSystem          1.0
 PortGroup           python 1.0
-PortGroup           github 1.0
 
-github.setup        matplotlib pytest-mpl 0.12.0 v
-name                py-${name}
+name                py-pytest-mpl
+version             0.12
+revision            1
 categories-append   devel
+
 license             BSD
 platforms           darwin
 supported_archs     noarch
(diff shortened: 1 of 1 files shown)


python/py-pytest-mpl/Portfile
```diff
@@ -11,6 +11,7 @@ license             BSD
 platforms           darwin
 supported_archs     noarch
 maintainers         {aronnax @lpsinger} openmaintainer
+fetch.type          git
```
ryandesign, line 14: Can this be avoided, please?

ryandesign: The [build failure log](https://build.macports.org/builders/ports-10.12_x86_64-builder/builds/142731/steps/install-port/logs/stdio) says:
```
LookupError: setuptools-scm was unable to detect version for '/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_python_py-pytest-mpl/py36-pytest-mpl/work/matplotlib-pytest-mpl-5c60ea6'.

Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
```
So, *don't* fetch from GitHub. Fetch from pypi, which the python portgroup would do by default if you didn't also include the github portgroup. So maybe the fix is not including the github portgroup.

essandess: I’m out of my depth here. The latest force-push that downloads from pypi will fail with:
```
LookupError: setuptools-scm was unable to detect version for '/opt/local/var/macports/build/_opt_local_ports_python_py-pytest-mpl/py39-pytest-mpl/work/matplotlib-pytest-mpl-5c60ea6'.

Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
```

cjones051073: Are tiu sure the tarball being used is the one from pypy ? Do a full clean

`sudo port clean —all py38-pytest-mpl`

then do a sudo `port -d fetch py38-pytest-mpl` And double check the output is using the correct source.

essandess: Thanks, needed to change `master_sites` and `version` to find the download.

PR #10547, opened 2021-04-01
blender: update to version 2.91.0

jasonliu--:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/graphics/blender/Portfile
+++ b/graphics/blender/Portfile
@@ -6,7 +6,7 @@ PortGroup           xcodeversion 1.0
 PortGroup           active_variants 1.1
 
 name                blender
-version             2.90.1
+version             2.91.0
 categories          graphics multimedia
 platforms           darwin
 license             GPL-2+
@@ -26,9 +26,9 @@ long_description    Blender is a ${description}. It contains a set of \
 
 master_sites        https://download.blender.org/source/
 use_xz              yes
-checksums           rmd160  63e2f9de4d48e0d0fac12d1fbc12313dc3d2c750 \
-                    sha256  e78dcc946c823ad1bd89bcc6938fc6b375540ec3eff58649586acf1f6a653d99 \
-                    size    38836408
+checksums           rmd160  536d839a3b054bc6f25c4b160adfc2e12bde27f9 \
+                    sha256  557a4afd09fe086ebcb0256b904896f577fe913683655f4248b881591f356974 \
+                    size    41528920
 
 minimum_xcodeversions-append {15 8.2 16 8.2}
 if {${os.platform} eq "darwin" && ${os.major} < 15} {
@@ -39,13 +39,13 @@ if {${os.platform} eq "darwin" && ${os.major} < 15} {
     }
 }
 
-compiler.cxx_standard 2011
+compiler.cxx_standard 2017
 # We need to blacklist Xcode Clang/LLVM, because Blender needs to
 # use tools such as llvm-config, which are not included in Xcode.
 # (llvm-config is needed because Blender links to libLLVM. libLLVM is
 # also not included in Xcode.)
 # (Reference: https://stackoverflow.com/questions/32959627/where-is-the-system-llvm-config-on-macos)
-compiler.blacklist clang
+compiler.blacklist clang {macports-clang-[5-8].0}
 # Even though we need to blacklist Xcode Clang/LLVM, we still need to
 # make the rest of Xcode available to the build system. If we don't set
 # use_xcode to yes, then the build will fail, because Blender's CMake
@@ -83,7 +83,9 @@ depends_lib-append  port:python${py_ver_nodot} \
                     port:embree \
                     port:oidn \
                     port:tbb \
+                    port:potrace \
                     port:libomp \
+                    port:gmp \
                     port:glew \
                     port:eigen3 \
                     port:lzo2 \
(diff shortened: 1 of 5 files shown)


graphics/blender/Portfile
```diff
@@ -114,6 +116,46 @@ post-patch {
...
+        # * https://www.veripool.org/issues/1231
+        # * https://trac.macports.org/ticket/47191
+        # * https://bugs.llvm.org/show_bug.cgi?id=22951
+        # * https://stackoverflow.com/questions/31686508/why-is-delete-operator-required-for-virtual-destructors
+        # * https://bugs.freedesktop.org/show_bug.cgi?id=54948
+        # * https://github.com/pybind/python_example/issues/48
+        # * https://github.com/llvm-mirror/clang/blob/master/test/SemaCXX/unavailable_aligned_allocation.cpp
+        # * See also the Portfile for port:folly.
+        reinplace -E {s/(-f)(aligned-allocation)/\1no-\2/} \
+            ${worksrcpath}/intern/cycles/kernel/osl/CMakeLists.txt
```
cjones051073, line 158: If the build requires c++17 you should change the requested standard to

compiler.cxx_standard 2017

To indicate this.

it would also be useful to know if you in addition  blacklist older macports clang versions, ie macports-clang-[5-9].0 which then will force the use of the most recent clang 11 version, if this then means the hadk here is no longer required ?

cjones051073 (changes requested):

graphics/blender/Portfile
cjones051073: B.t.w. You should also update the regex

regexp {clang-mp-([0-9]+(?:\.[0-9]+)*)} \
    ${configure.cc} result llvm_version

to work properly with macports clang 10 and newer, i.e. where the major version has double digits and there is no minor version, e.g. macports-clang-11.

jasonliu--: The entire build doesn't rely on C++17... only one specific piece, the Cycles renderer, requires C++17. In fact, there are portions which rely on C++11, and others which need C++14. With so many (nearly independent) moving parts, each with its own needs for C++ standard, it's kind of difficult to use `compiler.cxx_standard` to force a single C++ standard onto the entire source code. Rather, I've been using it as a sort of minimum... Blender uses newer C++ standards in the individual components' CMake files wherever it needs to.

> it would also be useful to know if you in addition blacklist older macports clang versions, ie macports-clang-[5-9].0 which then will force the use of the most recent clang 11 version, if this then means the hack here is no longer required ?

As far as I'm aware, I'm not blacklisting older macports clangs; I'm only blacklisting Xcode Clang. Could you point out which part of my portfile you are referring to?

> B.t.w. You should also update the regex
> ```
> regexp {clang-mp-([0-9]+(?:\.[0-9]+)*)} \
>     ${configure.cc} result llvm_version
> ```
> to work properly with macports clang 10 and newer, i.e. where the major version has double digits and there is no minor version, e.g. macports-clang-11.

I believe that regex should work with clang 10 and newer. The `(?:)*` makes the `.y` portion of version `x.y` optional.

cjones051073: > The entire build doesn't rely on C++17... only one specific piece, the Cycles renderer, requires C++17. In fact, there are portions which rely on C++11, and others which need C++14. With so many (nearly independent) moving parts, each with its own needs for C++ standard, it's kind of difficult to use `compiler.cxx_standard` to force a single C++ standard onto the entire source code. Rather, I've been using it as a sort of minimum... Blender uses newer C++ standards in the individual components' CMake files wherever it needs to.

thats not how it works. The setting does not enforce anything on the build, but indicates to macports what 
Capabilities the compiler it picks needs. so you need to in fact give the most recent required standard, and not the oldest. I.e. if any part of the build requires c++17 thats what you must specify, even if other parts only need something older.

> 
> > it would also be useful to know if you in addition blacklist older macports clang versions, ie macports-clang-[5-9].0 which then will force the use of the most recent clang 11 version, if this then means the hack here is no longer required ?
> 
> As far as I'm aware, I'm not blacklisting older macports clangs; I'm only blacklisting Xcode Clang. Could you point out which part of my portfile you are referring to?

i am suggesting you try adding to the blacklist an entry for macports clang, to enforce a more recent version than the default. I.e. try adding

{macports-clang-[5-9].0} 

to the list, which will black list clang 5 to 9, and thus force use of 11.

> 
> > B.t.w. You should also update the regex
> > ```
> > regexp {clang-mp-([0-9]+(?:\.[0-9]+)*)} \
> >     ${configure.cc} result llvm_version
> > ```
> > 
> > 
> > to work properly with macports clang 10 and newer, i.e. where the major version has double digits and there is no minor version, e.g. macports-clang-11.
> 
> I believe that regex should work with clang 10 and newer. The `(?:)*` makes the `.y` portion of version `x.y` optional.

cjones051073 (approved):

graphics/blender/Portfile
jasonliu--: >> The entire build doesn't rely on C++17... only one specific piece, the Cycles renderer, requires C++17. In fact, there are portions which rely on C++11, and others which need C++14. With so many (nearly independent) moving parts, each with its own needs for C++ standard, it's kind of difficult to use compiler.cxx_standard to force a single C++ standard onto the entire source code. Rather, I've been using it as a sort of minimum... Blender uses newer C++ standards in the individual components' CMake files wherever it needs to.
> thats not how it works. The setting does not enforce anything on the build, but indicates to macports what capabilities the compiler it picks needs. So you need to in fact give the most recent required standard, and not the oldest. I.e. if any part of the build requires c++17 thats what you must specify, even if other parts only need something older.

Sorry to contradict you on this, I hope you believe me when I say that I'm really not trying to be purposely difficult. But I just tested it, and it looks like `compiler.cxx_standard` is indeed enforcing compile flags during the build. Setting `compiler.cxx_standard 2017` is causing the build to add `-std=c++17` to all compile commands during the build phase. When the line is set to `compiler.cxx_standard 2011` as I had it before, the build is adding `-std=c++11` to all compile commands. Regardless, the build is still running to completion when set to C++17, so I've gone ahead and changed it as requested.

> i am suggesting you try adding to the blacklist an entry for macports clang, to enforce a more recent version than the default. I.e. try adding
> 
> `{macports-clang-[5-9].0}`
> 
> to the list, which will black list clang 5 to 9, and thus force use of 11.

I've added macports clang 5-8 to the blacklist. I've left off version 9, because that is the version of clang being used to compile Blender for its official release binaries, and I would like to keep the possibility of performing a macports build that is "as close to" the official release build as possible.

PR #10553, opened 2021-04-02
msmtp: add mta variant

gpanders:
#### Description

Add `msmtp-mta` subport. This will use `msmtp` as the default MTA (i.e. it symlinks `sendmail` to `msmtp` and provides `msmtpd` which acts as a local SMTP server). It is inspired by the [Debian package of the same name](https://packages.debian.org/en/sid/msmtp-mta).

###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H524
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/mail/msmtp/Portfile
+++ b/mail/msmtp/Portfile
@@ -40,3 +40,19 @@ platform macosx {
 }
 
 livecheck.url       https://marlam.de/msmtp/download/
+
+variant mta description {Use msmtp as the system MTA} {
+    conflicts               postfix opensmtpd
+
+    configure.args-append   --with-msmtpd
+
+    startupitem.create      yes
+    startupitem.name        msmtpd
+    startupitem.executable  ${prefix}/bin/msmtpd
+}
+
+post-destroot {
+    if {[variant_isset mta]} {
+        ln -sf ${prefix}/bin/msmtp ${destroot}${prefix}/sbin/sendmail
+    }
+}


ra1nb0w:
for this kind of thing you should use variant not a subport

gpanders:
I used a subport per @ryanakca's [suggestion](https://github.com/macports/macports-ports/pull/10528#issuecomment-812595889) and I agree that a subport seems like a better fit than a variant, since it doesn't change how `msmtp` itself is built, but rather installs additional software entirely (i.e. the `msmtpd` server). This is similar to how e.g. the mysql port also offers a mysql-server subport rather than a +server variant.

That said, I'm not about to die on this hill and I'm happy to change it to whatever consensus we arrive at if other maintainers want to chime in.

ra1nb0w:
The change is made with a configuration option to build something "enchanted" and this is different from mariadb/mysql that  produces two different things. I mean, the msmtp-mta conflicts with msmtp contents? For what I understood yes; without `--with-msmtpd` the port install only `mstmp` but with that option it installs `msmtp` and `msmtpd` plus LaunchDaemons and link as sendmail. Correct?

gpanders:
I changed it to use a `mta` variant (if you can think of a better variant name I will change it).

ra1nb0w:
`post-destroot` should be in the variant since sendmail behave as default MTA; or not?

gpanders:
No, see [this section in the guide](https://guide.macports.org/#development.variants.phase):

> If a variant requires options in addition to those provided by keywords using -append and/or -delete, in other words, any actions that would normally take place within a port installation phase, do not try to do this within the variant declaration. Rather, modify the behavior of any affected phases when the variant is invoked using the variant_isset keyword.

ra1nb0w:
Not too respected (also because it works) but good to know. LGTM. thank you very much.

PR #10558, opened 2021-04-03
clang-*: add clangd to clang_select group

gpanders:
#### Description

Add clangd to clang select group.

I see that clangd was added in #5418 but then reverted by @cjones051073  in #5683. The reversion doesn't provide a reason, so I'm not sure why it was reverted or if clangd is being omitted intentionally for some reason.

###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G8022
Xcode 11.3.1 11C505

###### Verification 
Have you

### DIFF
--- a/lang/llvm-10/files/mp-clang-10
+++ b/lang/llvm-10/files/mp-clang-10
@@ -12,7 +12,7 @@ bin/clang-cl-mp-10
 bin/clang-query-mp-10
 bin/clang-rename-mp-10
 bin/clang-tidy-mp-10
--
+bin/clangd-mp-10
 -
 -
 -
--- a/lang/llvm-11/files/mp-clang-11
+++ b/lang/llvm-11/files/mp-clang-11
@@ -12,7 +12,7 @@ bin/clang-cl-mp-11
 bin/clang-query-mp-11
 bin/clang-rename-mp-11
 bin/clang-tidy-mp-11
--
+bin/clangd-mp-11
 -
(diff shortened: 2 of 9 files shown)


kencu:
Thanks for the thought and the interest! But -- we can't have clangd in the select group as it doesn't build on all the systems that we deploy clang on.  Therefore it will cause errors during select on the systems where clangd doesn't build. That's why it was reverted out.

Now -- if there is invented some (reasonable) way to have different select files for different systems, I guess we can think about that; but at a certain point it sort of becomes too cumbersome to maintain.

besides -- clangd-mp-10 or clangd-mp-11 is not too much to type, I would think?

I'm going to close this so that nobody accidentally commits it, but if some method comes up we can reopen it!

gpanders:
I see. That makes sense. Thanks for clarifying. 

> besides -- clangd-mp-10 or clangd-mp-11 is not too much to type, I would think?

Of course not :) It's more a matter of convenience -- having simple `clangd` work means I don't have to have special cases in configuration files for macOS vs. Linux.

kencu:
are you using macports clangd for anything, by the way? I know it has a role in certain IDEs for on-the-fly source formatting, etc, but I have yet to find a use for it myself IRL.

gpanders:
Yes, I use it as an LSP server in my editor (Vim) and it works very well!

PR #10560, opened 2021-04-03
aom: revert native variant, allow runtime cpu detection

cjones051073:
#### Description

See discussions in https://github.com/macports/macports-ports/commit/6a1f839833a71be9b6c69d7e6ddca7a039d46673 and https://github.com/macports/macports-ports/commit/dafdf1644c207c48f0ca8eab0d255326fe2057f1

I mis-lead the maintainer into suggesting a native variant would be useful. These are useful when ports perform build-the CPU detection and bake into the binaries what it finds based on the build CPU. arm does not do this, and instead uses runtime CPU detection, and as such its perfectly fine, in fact better, to allow this unconditionally as it then allows the best performance on the users machine, based on what that machine supports. This PR reverts the additional of a native PR and goes back to just blacklisting older clangs by default, which is all that was required to fix the avx2 compilation issues.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/multimedia/aom/Portfile
+++ b/multimedia/aom/Portfile
@@ -7,7 +7,7 @@ PortGroup               compiler_blacklist_versions 1.0
 
 name                    aom
 version                 3.0.0
-revision                2
+revision                3
 categories              multimedia
 maintainers             {mcalhoun @MarcusCalhoun-Lopez} openmaintainer
 platforms               darwin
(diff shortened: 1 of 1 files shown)


cjones051073:
closing as maintainer opened https://github.com/macports/macports-ports/pull/10559 simultaneously

PR #8811, opened 2020-10-16
Disable build-time dependency fetching in golang-1.0 portgroup

amake:
#### Description

With the (relatively) new go modules system, go by default will fetch dependencies at build time. This PR disables such dynamic fetching by setting appropriate envars for the build and test phases. Previously this was being done on a port-by-port basis.

There is tooling (go2port) to help generate the appropriate markup to fetch dependencies with the `go.vendors` mechanism in the golang-1.0 portgroup, and this has been used to fix most ports that were downloading at build time (see https://trac.macports.org/ticket/61192). However some of the ports exercise parts of the build system that are harder to support in this fashion, so this PR grandfathers them in to the old behavior of allowing build-time dependency fetching.

I tested the changes in this PR by destrooting every port that uses the golang-1.0 portgroup.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H2  
Xcode 12.0.1 12A7300

###### Verification 
Have you

- [ ] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [ ] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [ ] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [ ] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 
- [ ] checked your Portfile with `port lint`?
- [ ] tried existing tests with `sudo port test`?
- [ ] tried a full install with `sudo port -vst install`?
- [ ] tested basic functionality of all binary files?

PR #10564, opened 2021-04-03
ki-shell: new submission

breun:
#### Description

New port for [Kotlin Language Interactive Shell](https://github.com/Kotlin/kotlin-interactive-shell).

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/java/ki-shell/Portfile
+++ b/java/ki-shell/Portfile
@@ -1,0 +1,64 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem      1.0
+PortGroup       java 1.0
+
+name            ki-shell
+version         0.3.3
+revision        0
+
+categories      java
+platforms       darwin
+maintainers     {breun.nl:nils @breun} openmaintainer
+license         Apache-2
+supported_archs noarch
+
+description     Kotlin Language Interactive Shell (ki)
+
+long_description The shell is an extensible implementation of Kotlin REPL with reach set of features including: \
+                \n\
+                \n* Syntax highlight \
+                \n* Type inference command \
+                \n* Downloading dependencies in runtime using Maven coordinates \
+                \n* List declared symbols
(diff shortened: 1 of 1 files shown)


cjones051073 (commented):
Is building from source not an options ?

breun:
It can be built with Maven, but Maven downloads dependencies at build time and I believe MacPorts doesn't like that.

MarcelBochtler:
Hi @breun, other ports are already build using Maven. E.g. https://github.com/macports/macports-ports/blob/master/devel/protobuf3-java/Portfile

breun:
I know, but MacPorts has also been actively [trying to convert Go ports to build without downloading dependencies at build time](https://github.com/macports/macports-ports/pull/8811). It is unclear to me what MacPorts prefers: no build time fetching of dependencies, or building from source and fetching dependencies at build time? Building from source without fetching dependencies at build time doesn't seem like a feasible option, because that is just not how Java build tools like Maven and Gradle work. They work with a local cache of artifacts and will download any required dependencies on-demand from a repository. By the way, the `spring-boot-cli` port, which is more similar to `ki-shell` than a library like `protobuf3-java`, is also not building from source.

Looking at the `protobuf3-java` Portfile it indeed seems to have a build dependency on `maven3`, but the port is unmaintained and even fails to install on my machine. According to https://ports.macports.org/port/maven3/summary there are currently only 5 ports in MacPorts that have a build time dependency on `maven3`, and I would argue that most of them shouldn't even be in MacPorts, because they are libraries and Java developers generally use their build tool of choice (Maven, Gradle, etc.) to declare the dependency and let the build tool take care of the fetching of that library as needed.

MarcelBochtler:
Yes I agree with your points. As you said it for java it is only feasible to download during build time.
I also don't see the advantage of building from source instead of using the prebuilt binaries.

cjones051073:
Well I guess if downloading the source during the fetch phase, and then building from source is not an option (a number of go and cargo ports seem to manage this), then its a choice of which is the lesser evil. Download during build phase and build from source, or binary download. Personally I favour the first of these, as I think building from source when possible should always be the preference.

PR #10578, opened 2021-04-05
gtk3: fix build on 10.11

elelay:
#### Description

fix 10.11 compatibility break, caused by typos in version macro check

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
```
macOS 10.11.6 15G22010
Xcode 8.2.1 8C1002
```

###### Verification 
Have you

- ~~tried existing tests with `sudo port test`?~~
- ~~tried a full install with `sudo port -vst install`?~~
- [x] tried a full install from source using `sudo port -d install -n gtk3`

### DIFF
--- a/gnome/gtk3/Portfile
+++ b/gnome/gtk3/Portfile
@@ -58,6 +58,9 @@ legacysupport.newest_darwin_requires_legacy 10
 use_autoreconf      yes
 autoreconf.args     -fvi
 
+# OS X 10.11 and earlier does not have NSEventTypeTabletProximity and the version check macro is broken
+# see https://gitlab.gnome.org/GNOME/gtk/-/issues/3592
+patchfiles   patch-gdk_quartz-10.11_compat.diff
 
 # gtk3 +quartz uses instancetype which is not available
 # before approximately Xcode 4.6 (#49391)
--- a/gnome/gtk3/files/patch-gdk_quartz-10.11_compat.diff
+++ b/gnome/gtk3/files/patch-gdk_quartz-10.11_compat.diff
@@ -1,0 +1,22 @@
+--- gdk/quartz/gdkinternal-quartz.h.orig	2021-04-05 15:46:38.000000000 +0200
++++ gdk/quartz/gdkinternal-quartz.h	2021-04-05 15:46:49.000000000 +0200
+@@ -75,7 +75,7 @@
+   GDK_QUARTZ_EVENT_SUBTYPE_EVENTLOOP
+ } GdkQuartzEventSubType;
+ 
+-#if MAC_OS_X_VERSION_MIN_REQUIRED >= 10130
++#if MAC_OS_X_VERSION_MIN_REQUIRED >= 101300
+ #define GDK_QUARTZ_EVENT_TABLET_PROXIMITY NSEventTypeTabletProximity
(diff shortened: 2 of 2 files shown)


kencu:
I'd suggest just apply the patch always (it is always needed) and KISS.

kencu:
Two other little things to ponder for the detail oriented and for upstream to think about...those three defines
```
#define GDK_QUARTZ_EVENT_TABLET_PROXIMITY NSEventTypeTabletProximity
#define GDK_QUARTZ_EVENT_SUBTYPE_TABLET_PROXIMITY NSEventSubtypeTabletProximity
#define GDK_QUARTZ_EVENT_SUBTYPE_TABLET_POINT NSEventSubtypeTabletPoint
```
all are marked by Apple as being in the 10.12 SDK, not actually 10.13.

https://developer.apple.com/documentation/appkit/nseventtype/nseventtypetabletproximity
https://developer.apple.com/documentation/appkit/nseventsubtype/nseventsubtypetabletproximity
https://developer.apple.com/documentation/appkit/nseventsubtype/nseventsubtypetabletpoint

and also, the test should really be on the SDK features (where the defines are), and not the deployment target (as we're not testing for function availability in the system libraries), so I believe the proper test really should be:

```
#if MAC_OS_X_VERSION_MAX_ALLOWED >= 101200
```

because MAC_OS_X_VERSION_MAX_ALLOWED tests the SDK you're building against, whereas MAC_OS_X_VERSION_MIN_REQUIRED tests the deployment target you have set. Usually these are the same, but it is not unusual to build for older systems by setting the deployment target to some older version.

elelay:
Thanks @kencu for the pointers. Indeed, it should be a fix for upstream. I don't have a macOS Sierra install to confirm that NSEventTypeTabletProximity etc. are OK on 10.12.

Regarding KISS I'll follow devans example a few lines above the change, where there is a check for the variant for a similar patch to `gdk/quartz/gdkcursor-quartz.c`.

kencu:
As you wish. The other 10.6 patch is to change code on 10.6 to make it work, tho.

This one you have here is a one-character typo.

gnome/gtk3/Portfile
```diff
@@ -160,6 +160,18 @@ platform darwin {
...
+        if {[variant_isset quartz]} {
+            patchfiles-append   patch-gdk_quartz-10.11_compat.diff
+        }
+    }
+    if {${os.major} <= 15} {
+        # OS X 10.11 and earlier does not have NSEventTypeTabletProximity and the version check macro is broken
+        # see https://gitlab.gnome.org/GNOME/gtk/-/issues/3592
+        if {[variant_isset quartz]} {
+            patchfiles-append   patch-gdk_quartz-10.11_compat.diff
+        }
```
ryandesign, line 174: For OS X 10.8 and earlier, this will add the patchfile twice, which will fail.

Please just apply the patch once unconditionally on all systems and variants. This is simpler. Applying it conditionally on older systems is a good way to ensure that a problem occurs in the future when someone running a newer system updates the port and doesn't realize they have to adjust or remove the patch.

elelay: Yes, you're right. I'll push a fix

elelay:
Patch has been accepted upstream

ra1nb0w:
any update on this?

elelay:
merging, since all review comments have been addressed and the port is openmaintainer

PR #10587, opened 2021-04-06
Fix hdf4 with variants +gcc10 and +szip

MarcusCalhoun-Lopez:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/science/hdf4/Portfile
+++ b/science/hdf4/Portfile
@@ -70,6 +70,8 @@ if {[fortran_variant_isset]} {
     configure.args-delete   --disable-fortran --enable-shared
 }
 
+compilers.allow_arguments_mismatch  yes
+
 if {[variant_isset g95]} {
     configure.fflags-append -fno-second-underscore
 }
--- a/science/hdf4/Portfile
+++ b/science/hdf4/Portfile
@@ -39,7 +39,8 @@ patchfiles          patch-configure.diff \
                     patch-mfhdf-test-tfile.c.diff \
                     patch-mfhdf-test-tncvargetfill.c.diff \
                     patch-mfhdf-test-tsd.c.diff \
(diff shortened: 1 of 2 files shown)


science/hdf4/Portfile
```diff
@@ -69,6 +70,8 @@ if {[fortran_variant_isset]} {
     configure.args-delete   --disable-fortran --enable-shared
 }
 
+compilers.allow_arguments_mismatch  yes
```
cjones051073, line 73: Out of interest what does this option do ?

kencu: https://github.com/macports/macports-ports/blob/73613c0a068534b30ea660189ddd8f9eae46df86/_resources/port1.0/group/compilers-1.0.tcl#L49

MarcusCalhoun-Lopez: The following Fortran code should be an error
```
program main
  print *, my_function(1)
  print *, my_function(1.0)
end program main
```

Newer version of GCC (10+) do not allow this.
With `-fallow-argument-mismatch` or on older version of GCC, it compiles (perhaps with a warning).

Because Fortran passes arguments by reference,  programers sometimes blur the distinction between, e.g., an integer and an array of integers.
There also seem to be clever ways of manipulating pointers/references/arrays when interfacing with C/C++ code.
I do not present to fully understand the details.

When arguments mismatches started becoming an error, `-fallow-argument-mismatch` (https://github.com/macports/macports-ports/commit/20b350bb5eb864e9323f2f79324a2f7f977942d1)
allowed a bridge for ports when a proper solution was not evident.

PR #10618, opened 2021-04-08
tensorflow_macos: Submission

essandess:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/python/tensorflow_macos/Portfile
+++ b/python/tensorflow_macos/Portfile
@@ -1,0 +1,117 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           python 1.0
+
+github.setup        apple tensorflow_macos 0.1 v alpha3
+name                py-${github.project}
+version             ${github.version}.${github.tag_suffix}
+revision            0
+
+platforms           darwin
+categories-append   lang
+license             restrictive/distributable
+maintainers         nomaintainer
+
+description         Mac-optimized TensorFlow and TensorFlow Addons
+
+long_description    Hardware-accelerated TensorFlow and TensorFlow \
+                    Addons for macOS 11.0+. Native hardware \
+                    acceleration is supported on M1 Macs and \
+                    Intel-based Macs through Apple’s ML Compute \
+                    framework.
+
+# https://github.com/apple/tensorflow_macos/blob/master/scripts/download_and_install.sh
+distfiles           ${github.project}-${github.version}${extract.suffix}:tarball \
+                    ${github.project}-${github.version}${github.tag_suffix}${extract.suffix}:releases
+
+master_sites        ${github.homepage}/tarball/${git.branch}:tarball \
+                    github.master_sites ${github.homepage}/releases/download/${git.branch}:releases
+
+checksums           [lindex [split [lindex ${distfiles} 0] :] 0] \
+                    rmd160  975a2ee2508df2fe801b841d3f6f8a0172854292 \
+                    sha256  623237c06941d26f52280535c0139c9d58fc214e0378b4513803a65bfda48ff6 \
+                    size    8095 \
+                    [lindex [split [lindex ${distfiles} 1] :] 0] \
+                    rmd160  812afc151dd87abb657df6d8e66ad8681c91a3bd \
+                    sha256  80807f8b0fc4a98ffa0b7395304fa61239f246a84f1eedbca70a5a1847d4dead \
+                    size    376699291
+
+extract.cmd         ${portutil::autoconf::tar_command}
+extract.pre_args    -xzf
+extract.post_args
+
+if {${os.platform} eq "darwin" && ${os.major} < 20} {
+    pre-fetch {
+        ui_error "TensorFlow with ML Compute acceleration is only available \
+            on macOS 11.0 and later."
+        error {unsupported platform}
+    }
+}
+
+python.versions 38
+
+if {${name} ne ${subport}} {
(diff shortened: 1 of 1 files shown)


essandess:
cc: @cjones051073 

Upstream also supports M1 acceleration. See https://blog.tensorflow.org/2020/11/accelerating-tensorflow-performance-on-mac.html

I'm not sure how to address this in the Portfile. The distribution includes pre-compiled `.whl`'s for both M1 and Intel architectures. The tests I've done on integration with MacPorts Python on Intel have all worked.

python/tensorflow_macos/Portfile
```diff
@@ -0,0 +1,117 @@
...
+                    [lindex [split [lindex ${distfiles} 1] :] 0] \
+                    rmd160  812afc151dd87abb657df6d8e66ad8681c91a3bd \
+                    sha256  80807f8b0fc4a98ffa0b7395304fa61239f246a84f1eedbca70a5a1847d4dead \
+                    size    376699291
+
+extract.cmd         ${portutil::autoconf::tar_command}
+extract.pre_args    -xzf
+extract.post_args
+
+if {${os.platform} eq "darwin" && ${os.major} < 20} {
```
cjones051073, line 45: Please add 

`known_fail yes`

Here to precent the builds on unsupported OSes being attempted on the buildbots

python/tensorflow_macos/Portfile
```diff
@@ -0,0 +1,117 @@
...
+                    ${destroot}${prefix}/bin/tensorboard-${python.branch}
+
+        set github_project_path [glob -nocomplain ${workpath}/${github.author}-${github.project}-*]
+        set docdir ${prefix}/share/doc/${subport}
+        xinstall -d ${destroot}${docdir}
+        xinstall -m 0644 -W ${github_project_path} LICENSE.txt README.md \
+            ${destroot}${docdir}        
+    }
+
+    notes-append \
```
cjones051073, line 111: These notes are not really necessary as the conflicts declaration above does something equivalent when a conflict is detected.

python/tensorflow_macos/Portfile
```diff
@@ -0,0 +1,117 @@
...
+            on macOS 11.0 and later."
+        error {unsupported platform}
+    }
+}
+
+python.versions 38
+
+if {${name} ne ${subport}} {
+    conflicts       py${python.version}-tensorflow \
+                    py${python.version}-tensorflow_addons
```
cjones051073, line 57: You should also declare a conflict with the `tensorflow1` port here, and in addition in both the regular `tensorflow` , `tensorflow_addons` and `tensorflow1` ports you need to added the corresponding conflicts with this new port.

cjones051073 (changes requested):

cjones051073:
> cc: @cjones051073 
> 
> 
> 
> Upstream also supports M1 acceleration. See https://blog.tensorflow.org/2020/11/accelerating-tensorflow-performance-on-mac.html
> 
> 
> 
> I'm not sure how to address this in the Portfile. The distribution includes pre-compiled `.whl`'s for both M1 and Intel architectures. The tests I've done on integration with MacPorts Python on Intel have all worked.
> 
> 

Not really sure exactly what you are asking ? Ince we have Apple Silicon machines back in the buildbots we can just see what that makes of this port. Until then you could ask on the devel mailing e could give it a try.

kencu:
> > cc: @cjones051073
> Not really sure exactly what you are asking ? Once we have Apple Silicon machines back in the buildbots we can just see what that makes of this port. 

AFAICT this port downloads a 376MB tarball and copies the files into destroot, without building anything.  I'm sure the Apple Silicon machines could do that.

The question I guess is whether that, given it is 0.1v-alpha3 version software, is something to have in MP at all, at this stage. But hey, why-the-heck-not, I guess....

If I'm going to use this, I would just go the apple site and install it exactly the way they recommend to install it, myself.

cjones051073:
> > > cc: @cjones051073
> > > Not really sure exactly what you are asking ? Once we have Apple Silicon machines back in the buildbots we can just see what that makes of this port.
> 
> AFAICT this port downloads a 376MB tarball and copies the files into destroot, without building anything. I'm sure the Apple Silicon machines could do that.
> 
> The question I guess is whether that, given it is 0.1v-alpha3 version software, is something to have in MP at all, at this stage. But hey, why-the-heck-not, I guess....
> 
> If I'm going to use this, I would just go the apple site and install it exactly the way they recommend to install it, myself.

Ah right. I missed the fact it wasn’t building from source. Is this not an option ?

cjones051073:
I am though ok with ports sometimes installing prebuilt binaries if building from source is not an easy option. In this case there is I think an advantage to automating it as a port, as then it properly handles installing the wheel into the MacPorts prefix, so can be used agains macports python versions. Users should not do this themselves. Yes, they can use a virtual env or something similar, but offering it as a port seems reasonable. I do though take the point that as this is an early alpha version should we wait for it to stabilise first ?

essandess:
Uh, I tried renaming the branch on this PR to `py-tensorflow_macos` and completely fouled things up.

Does anyone know how to get the commit's for https://github.com/essandess/macports-ports/tree/py-tensorflow_macos back into this PR so we can keep this thread, or should I just start from scratch?

python/tensorflow_macos/Portfile
essandess: Thanks. Please see #10625.

python/tensorflow_macos/Portfile
essandess: Thanks. Please see #10625.

python/tensorflow_macos/Portfile
essandess: Thanks. Please see #10625.

essandess:
Started from scratch. Please go to #10625.

PR #10620, opened 2021-04-08
macfuse: new port

mohd-akram:
#### Description

osxfuse is no longer open-source, doesn't seem to work, and the name of the project has changed to macfuse, hence the new port.

###### Tested on
macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification
Have you

### DIFF
--- a/fuse/macfuse/Portfile
+++ b/fuse/macfuse/Portfile
@@ -1,0 +1,77 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        osxfuse osxfuse 4.1.0 macfuse-
+name                macfuse
+conflicts           osxfuse
+categories          fuse
+platforms           darwin
+supported_archs     x86_64
+license             Restrictive
+maintainers         {outlook.com:mohd.akram @mohd-akram} openmaintainer
+
+description         FUSE for macOS
+
+long_description    FUSE extends macOS by adding support for user space file \
+                    systems.
+
+homepage            https://osxfuse.github.io/
+
+github.tarball_from releases
+distname            ${name}-${version}
+use_dmg             yes
+
+checksums           rmd160  2c5e9741b983e660e34139966838af8a49eb3b5c \
+                    sha256  3cb6a49406fd036c50ef1b4ad717a377f4dcf182811bde172d69f1c289791085 \
+                    size    5747012
+
+if {${os.platform} eq "darwin" && ${os.major} < 16} {
+    known_fail yes
+    pre-fetch {
+        ui_error "${name} is only supported on macOS 10.12 or later."
+        return -code error "unsupported platform version"
+    }
+}
+
+set pkg ${workpath}/pkg/Core.pkg
+
+post-extract {
+    system -W ${worksrcpath} \
+        "pkgutil --expand 'Extras/macFUSE ${version}.pkg' ${workpath}/pkg"
+    system -W ${pkg} "gzip -dc Payload | cpio -id"
+}
+
+use_configure       no
+
+build {
+    system -W ${pkg} "codesign --remove-signature \
+        usr/local/lib/libfuse.2.dylib \
+        Library/Frameworks/macFUSE.framework/Versions/A/macFUSE"
+
+    system -W ${pkg} "install_name_tool -id ${prefix}/lib/libfuse.2.dylib \
+        usr/local/lib/libfuse.2.dylib"
+
+    system -W ${pkg} "install_name_tool \
+        -id ${prefix}/Library/Frameworks/macFUSE.framework/Versions/A/macFUSE \
+        -change /usr/local/lib/libfuse.2.dylib ${prefix}/lib/libfuse.2.dylib \
+        Library/Frameworks/macFUSE.framework/Versions/A/macFUSE"
+}
+
+destroot {
+    copy ${pkg}/Library/Filesystems ${destroot}${prefix}/Library
+    copy {*}[glob ${pkg}/Library/Frameworks/*] ${destroot}${prefix}/Library/Frameworks
+    copy {*}[glob ${pkg}/usr/local/include/*] ${destroot}${prefix}/include
+    copy {*}[glob ${pkg}/usr/local/lib/pkgconfig/*] ${destroot}${prefix}/lib/pkgconfig
+    copy {*}[glob -type f ${pkg}/usr/local/lib/*] ${destroot}${prefix}/lib
+}
+
+set dir /Library/Filesystems/macfuse.fs
+
+notes  "
+    Run the following before using macFUSE:
+
+        sudo ln -sn ${prefix}${dir} ${dir}
+        sudo ${dir}/Contents/Resources/load_macfuse
+"


pmetzger:
I'm not sure it's a great idea to be supporting the developer's whimful closing of the code?

mohd-akram:
I'm not happy about it either, but we already use binary blobs in osxfuse because the kext needs to be signed. In this particular case it wouldn't matter much even if it was open-source.

pmetzger:
One option would be to fork the code and provide a signed binary from an open repository.

herbygillot:
> One option would be to fork the code and provide a signed binary from an open repository.

Fork the code? The last code openly available will not work for the most recent macOS releases....

Most recent seems to be [here](https://github.com/osxfuse/osxfuse/tree/support/osxfuse-3), supporting only up to macOS 10.14.

pmetzger:
> Fork the code? The last code openly available will not work for the most recent macOS releases....

Yes, but the changes required are not dramatic.

herbygillot:
I think starting a whole new fork of MacFuse that is both open source and seeks to be compatible with the most recent versions of macOS is out of scope here.

That being said, I'm actually quite surprised that hasn't already happened.

So I guess a decision needs to be made on whether to have this binary port of MacFUSE or not, and I'm not sure how that decision is to be made.

pmetzger:
We usually avoid packaging binaries...

fuse/macfuse/Portfile
```diff
@@ -0,0 +1,77 @@
...
+    copy {*}[glob ${pkg}/usr/local/lib/pkgconfig/*] ${destroot}${prefix}/lib/pkgconfig
+    copy {*}[glob -type f ${pkg}/usr/local/lib/*] ${destroot}${prefix}/lib
+}
+
+set dir /Library/Filesystems/macfuse.fs
+
+notes  "
+    Run the following before using macFUSE:
+
+        sudo ln -sn ${prefix}${dir} ${dir}
```
ra1nb0w, line 75: you can write outside `destroot` using `destroot.violate_mtree` or there is something else that I ignore?

ra1nb0w (commented):
This package is "essentially" needed and everyone is sad about the opensouce move but the developer has express is idea many many times therefore stop about this point. 
Just one thing: we should thing how to manage the `depends_libs` in fuse/* ports. I mean, changing the name can be a problem. Notice that also this one copy `fuse.pc`. @herbygillot how have you managed that?

pmetzger:
Given that this is a binary now, why do we need to package it? If one wants it, one can just use their binary.

ra1nb0w:
To simplify management and to easily install modules like, sshfs or build your own. This blob without modules is useless.

landonf:
> Given that this is a binary now, why do we need to package it? If one wants it, one can just use their binary.

I agree that the shift to a binary-only distribution model is frustrating, but given that there are FUSE file systems that would be nice to package, it’d be nice if we could declare a dependency on MacFUSE and have things Just Work.

markemer:
I'm half with @pmetzger - I don't know about "we" but we as a oss community should fork osxfuse and fix it - however, I'm also concerned with violating `destroot` is there anyway that we can avoid that?

ra1nb0w:
> I'm half with @pmetzger - I don't know about "we" but we as a oss community should fork osxfuse and fix it 

Ok. find someone that have the time and competence to do it. Then, we can discuss to use the free software fork. Also because with Big Sur you should probably rewrite a lot of the code since kernel extension are not "permitted". And, at the end, we need to distribuite a blob signed by Apple. Just to conclude, I am the first that I don't like binaries (first, for security specially in kernel space).

> however, I'm also concerned with violating `destroot` is there anyway that we can avoid that?

if we need to copy something outside like `/Library/Extensions` we need to use that. This accomplish to things:
- avoid the user to run a not strictly needed command
- clearly remove all files when the user decide to remove the port

At the end of the installation, we can think to display a message in which we say that the port is binary only, closed source.

pmetzger:
There is no need for a binary only closed source port though. You can simply use the upstream and install if you want this. That's why we end up having to put up with these compromises, because MacPorts isn't really built to be an installer for upstream binaries.

ra1nb0w:
I don't like to repeat; but, how do you manage compile/install/uninstall of the modules (the only useful thing of fuse)?
Anyway, @mohd-akram can you check the changes and how do you manage (at the moment) the dependency in the modules?

fuse/macfuse/Portfile
```diff
@@ -0,0 +1,77 @@
...
+use_dmg             yes
+
+checksums           rmd160  2c5e9741b983e660e34139966838af8a49eb3b5c \
+                    sha256  3cb6a49406fd036c50ef1b4ad717a377f4dcf182811bde172d69f1c289791085 \
+                    size    5747012
+
+if {${os.platform} eq "darwin" && ${os.major} < 16} {
+    known_fail yes
+    pre-fetch {
+        ui_error "${name} is only supported on macOS 10.12 or later."
```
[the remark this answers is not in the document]
mohd-akram: Does the osxfuse port even work for older versions of macOS? There seems to be a bunch of issues on the tracker.

fuse/macfuse/Portfile
mohd-akram: Yeah, I saw that. It seemed like overkill for a single symlink, and given that some extra steps are required anyway the user might as well create it.

mohd-akram:
@ra1nb0w Right now the modules have `path:lib/pkgconfig/fuse.pc:osxfuse` in them and so they pick up macfuse.

markemer:
> There is no need for a binary only closed source port though. You can simply use the upstream and install if you want this. That's why we end up having to put up with these compromises, because MacPorts isn't really built to be an installer for upstream binaries.

I'm wondering if the time has come to actually move on changing base to support third party binaries.

fuse/macfuse/Portfile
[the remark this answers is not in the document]
ra1nb0w: I used it on macOS 10.14 and seems to work. Anyway, the new macfuse which version support? and the older one (osxfuse)? because if they are superimposable we can substitute the older osxfuse

fuse/macfuse/Portfile
ra1nb0w: Was just to clear correctly the environment during uninstall without manual intervention. But this can be changed later if there are complaints. Which extra steps the user needs?

ra1nb0w:
> @ra1nb0w Right now the modules have `path:lib/pkgconfig/fuse.pc:osxfuse` in them and so they pick up macfuse.

Probably I was too concise. I mean, if I install sshfs as dependency it installs osxfuse. Generally, I start installing one module not the base fuse. But this question can be useless if the two are superimposable.

fuse/macfuse/Portfile
mohd-akram: The `load_macfuse` command, and they need to allow the kernel extension in System Preferences (they'll get a message when they run `load_macfuse`).

mohd-akram:
Well right now if you install macfuse beforehand it will use that. We should add a message to osxfuse telling users with newer macOS to use macfuse instead.

ra1nb0w:
If macfuse covers all supported macOS version of osxfuse we can use `replaced_by`

ra1nb0w:
macFuse 4.1.0 supports macOS 10.12 or later. 
osxfuse 3.8.3 supports Mac OS X 10.5 to macOS 10.14.
Therefore we need to manage this fact correctly

ra1nb0w:
Any idea on how to manage the dependency?

ra1nb0w:
I can probably merge this and we see in a second time how to manage the dependency; maybe if/else in the module ports.

mascguy:
> macFuse 4.1.0 supports macOS 10.12 or later.
> osxfuse 3.8.3 supports Mac OS X 10.5 to macOS 10.14.
> Therefore we need to manage this fact correctly

I believe 4.0.5 supports MacOS 10.9 through 10.11, so ideally we should select that version for those MacOS releases. And then fallback to 3.8.3 for MacOS 10.5 through 10.8.

ra1nb0w:
This method is not liked too much  by other macports developers because in the long term is hard to manage.
I need to thing about a bit :)

ryandesign:
Having two separate ports, osxfuse and macfuse, for the same software is terrible. It should be one port that installs the right version for the OS. Otherwise it's dependency hell for all the ports that depend on osxfuse/macfuse.

ra1nb0w:
I read both osxfuse and macfuse and they are quite big therefore having a single `if` compromise the reading.  My proposal is to maintain the two ports and add for each fuse module:
```
# use osxfuse up to OS X 10.11
  set fuse_port "osxfuse"
} else {
  set fuse_port "macfuse"
}

depends_lib-append \
  path:lib/pkgconfig/fuse.pc:${fuse_port}
```
Creating a `portgroup` avoid duplication but maybe it is a too small change to justify it.
What do you think @ryandesign?

raimue:
> Having two separate ports, osxfuse and macfuse, for the same software is terrible.

It is definitely not the same software. osxfuse is open source, macfuse is not. Shipping both with the same port name but different licenses causes many edge cases for mirroring, buildbot, etc.

To avoid repeating the conditional in all ports there could be an intermediate dummy port that only declares the right dependency on either osxfuse or macfuse depending on the macOS release. Or put the conditional into a port group.

ra1nb0w:
I don't like too much the idea of dummy port (which name for example?) but you are right about license issue.
Probably the right way to be future proof is to create a port group. I try to implement that  in a new PR.
Thank you @raimue

ryandesign:
> It is definitely not the same software. osxfuse is open source, macfuse is not.

Of course it is the same software.

raimue:
macfuse is a closed-source fork of osxfuse. The code base has diverted and they support different releases of macOS. Dropping support for older versions of macOS implies that macfuse has changed significantly. They have the same base version, but such a fork is no longer the same software project in my opinion.

We used to have macfuse (MacFUSE) and fuse4x, we still have osxfuse and now we have this macfuse (macFUSE) submission. All of these share parts of their code, but that does not mean that they are all the same software.

PR #10625, opened 2021-04-08
py-tensorflow_macos

essandess:
Follow up from: https://github.com/macports/macports-ports/pull/10618#issuecomment-816151194

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/python/py-tensorflow_macos/Portfile
+++ b/python/py-tensorflow_macos/Portfile
@@ -1,0 +1,114 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           python 1.0
+
+github.setup        apple tensorflow_macos 0.1 v alpha3
+name                py-${github.project}
+version             ${github.version}.${github.tag_suffix}
+revision            0
+
+platforms           darwin
+categories-append   lang
+license             restrictive/distributable
+maintainers         nomaintainer
+
+description         Mac-optimized TensorFlow and TensorFlow Addons
+
+long_description    Hardware-accelerated TensorFlow and TensorFlow \
+                    Addons for macOS 11.0+. Native hardware \
+                    acceleration is supported on M1 Macs and \
+                    Intel-based Macs through Apple’s ML Compute \
+                    framework.
+
+# https://github.com/apple/tensorflow_macos/blob/master/scripts/download_and_install.sh
+distfiles           ${github.project}-${github.version}${extract.suffix}:tarball \
+                    ${github.project}-${github.version}${github.tag_suffix}${extract.suffix}:releases
+
+master_sites        ${github.homepage}/tarball/${git.branch}:tarball \
+                    github.master_sites ${github.homepage}/releases/download/${git.branch}:releases
+
+checksums           [lindex [split [lindex ${distfiles} 0] :] 0] \
+                    rmd160  975a2ee2508df2fe801b841d3f6f8a0172854292 \
+                    sha256  623237c06941d26f52280535c0139c9d58fc214e0378b4513803a65bfda48ff6 \
+                    size    8095 \
+                    [lindex [split [lindex ${distfiles} 1] :] 0] \
+                    rmd160  812afc151dd87abb657df6d8e66ad8681c91a3bd \
+                    sha256  80807f8b0fc4a98ffa0b7395304fa61239f246a84f1eedbca70a5a1847d4dead \
+                    size    376699291
+
+extract.cmd         ${portutil::autoconf::tar_command}
+extract.pre_args    -xzf
+extract.post_args
+
+if {${os.platform} eq "darwin" && ${os.major} < 20} {
+    known_fail      yes
+    pre-fetch {
+        ui_error "TensorFlow with ML Compute acceleration is only available \
+            on macOS 11.0 and later."
+        error {unsupported platform}
+    }
+}
+
+python.versions 38
+
+if {${name} ne ${subport}} {
+    conflicts       py${python.version}-tensorflow \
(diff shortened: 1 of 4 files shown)


python/py-tensorflow_macos/Portfile
```diff
@@ -0,0 +1,113 @@
...
+if {${os.platform} eq "darwin" && ${os.major} < 20} {
+    known_fail      yes
+    pre-fetch {
+        ui_error "TensorFlow with ML Compute acceleration is only available \
+            on macOS 11.0 and later."
+        error {unsupported platform}
+    }
+}
+
+python.versions 38
```
cjones051073, line 54: Does this new port not support python 3.9 as well at least ?

python/py-tensorflow_macos/Portfile
```diff
@@ -0,0 +1,113 @@
...
+
+    destroot.cmd    pip-${python.branch}
+    destroot.args   --ignore-installed \
+                    --no-cache-dir \
+                    --no-dependencies \
+                    --root ${destroot}
+    destroot.post_args
+
+    destroot {
+        foreach whl [glob -type f ${workpath}/${github.project}/x86_64/tensorflow*cp${python.version}*.whl] {
```
essandess, line 95: Re: https://github.com/macports/macports-ports/pull/10618#issuecomment-815871065

> Not really sure exactly what you are asking ? Once we have Apple Silicon machines back in the buildbots we can just see what that makes of this port. Until then you could ask on the devel mailing if anyone there with a M1 machine could give it a try.

This line. For M1, the `x86_64` directory must be replaced with the M1 one, and the M1 `.whl`'s will be used.

python/py-tensorflow_macos/Portfile
essandess: Nope. No source code,.No Python 39. Yet. Just pre-compiled binaries for CPython 38.

essandess:
Re: https://github.com/macports/macports-ports/pull/10618#issuecomment-816017055

> I do though take the point that as this is an early alpha version should we wait for it to stabilise first ?

Tradeoff between early adopter issues and the expectation that these performance differences will deliver. Besides, upstream will be less alpha-y by the time this PR gets out of review. 😉 And people have already been waiting a looooooong time to use these frameworks on the Mac, e.g. https://github.com/RadeonOpenCompute/ROCm/issues/262.

The question is whether this will actually by supported and extended to include PyTorch.

Training impact on common models using ML Compute on the Intel-powered 2019 Mac Pro are shown in seconds per batch, with lower numbers indicating faster training time.

python/py-tensorflow_macos/Portfile
cjones051073: Right, so then you should use ${os.arch} to determine the architecture and adjust this accordingly.

cjones051073: Correct. Make that ${build_arch}. See e.g.

https://github.com/macports/macports-ports/blob/master/lang/gcc-devel/Portfile

For an example of a port that does this to do something different on apple silicon machines.

essandess:
> Should this port be considered a possible drop in replacement of the current tensorflow ports ? If so then any port that declares a dependency on the current tensorflow port should be updated to use a path style port dependency, that would allow either to be used to satisfy the dependency.

I think this is a good idea. Personally, I’d want more experience away from Linux GPU cluster before deciding to migrate completely and putting in the cycles to this. Certainly anyone running an M1 wants this version.

python/py-tensorflow_macos/Portfile
essandess: Thanks, done.

essandess:
One comment about the upstream distribution: it also contains specially compiled `.whl`'s for `grpcio`, `numpy`, and `scipy` (see below).

Aside from the necessity of these being compiled packages, I am not aware of any reason these would benefit over what we already do to build these, any use of the GPU, or any reason why they'd be needed. Therefore, to avoid further conflicts with existing ports, I simply did not install these `.whl`'s from upstream and rather included the necessary dependencies.

So far, I haven't seen any issues arising from this initial approach, though perhaps we should ask upstream about it.

```bash
$ tar tfz tensorflow_macos-0.1alpha3.tar.gz 
tensorflow_macos/
tensorflow_macos/.DS_Store
tensorflow_macos/arm64/
tensorflow_macos/arm64/numpy-1.18.5-cp38-cp38-macosx_11_0_arm64.whl
tensorflow_macos/arm64/tensorflow_addons_macos-0.1a3-cp38-cp38-macosx_11_0_arm64.whl
tensorflow_macos/arm64/.DS_Store
tensorflow_macos/arm64/grpcio-1.33.2-cp38-cp38-macosx_11_0_arm64.whl
tensorflow_macos/arm64/h5py-2.10.0-cp38-cp38-macosx_11_0_arm64.whl
tensorflow_macos/arm64/h5py-2.10.0-cp38-cp38-macosx_11_0_arm64.whl.backup
tensorflow_macos/arm64/tensorflow_macos-0.1a3-cp38-cp38-macosx_11_0_arm64.whl
tensorflow_macos/x86_64/
tensorflow_macos/x86_64/.DS_Store
tensorflow_macos/x86_64/numpy-1.18.5-cp38-cp38-macosx_11_0_x86_64.whl
tensorflow_macos/x86_64/scipy-1.5.4-cp38-cp38-macosx_11_0_x86_64.whl
tensorflow_macos/x86_64/h5py-2.10.0-cp38-cp38-macosx_11_0_x86_64.whl
tensorflow_macos/x86_64/tensorflow_addons_macos-0.1a3-cp38-cp38-macosx_11_0_x86_64.whl
tensorflow_macos/x86_64/tensorflow_macos-0.1a3-cp38-cp38-macosx_11_0_x86_64.whl
tensorflow_macos/x86_64/grpcio-1.33.2-cp38-cp38-macosx_11_0_x86_64.whl
tensorflow_macos/install_venv.sh
```

PR #10628, opened 2021-04-08
rb-rubyzip: Moved port from rb-zip and update to 2.3.0

judaew:
Fixed: https://trac.macports.org/ticket/53717

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/ruby/rb-bee/Portfile
+++ b/ruby/rb-bee/Portfile
@@ -2,6 +2,8 @@ PortSystem 1.0
 PortGroup ruby 1.0
 
 ruby.setup		bee 0.5.1 gem {} rubygems
+revision        1
+
 categories-append	devel
 license			Apache
 platforms		darwin freebsd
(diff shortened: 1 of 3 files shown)


reneeotten (changes requested):
I've left a few comments below. I also noticed that `rb-bee` depends on `rb-zip` (do `port echo depends:rb-zip`). So you should make sure that this indeed builds with the new `rb-rubyzip` port and update the dependency to this new port (and increase the `revision`).

One thing to consider though is if it's necessary to rename the port, you could in principle just update the `rb-zip` port to its latest version. But again, I don't use any of these Ruby ports so don't have a strong preference or opinion on what would be best.

judaew:
Done

reneeotten (approved):

PR #10647, opened 2021-04-11
qt5*: fix +universal configure

encounter:
#### Description

Commit https://github.com/macports/macports-base/commit/b00e8efdbc330c7c59b08fb881c069a1b8a8e851 removed `configure_get_universal_cflags` and `configure_get_universal_ldflags`, this replaces usages within the qt5* ports to use `configure_get_universal_archflags` directly.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/aqua/qt5/Portfile
+++ b/aqua/qt5/Portfile
@@ -1035,8 +1035,8 @@ foreach {module module_info} [array get modules] {
             patchfiles-append patch-qmake_flags.diff
             post-patch {
                 if {[variant_exists universal] && [variant_isset universal]} {
-                    set arch_replace_cxx [portconfigure::configure_get_universal_cflags]
-                    set arch_replace_ld  [portconfigure::configure_get_universal_ldflags]
+                    set arch_replace_cxx [portconfigure::configure_get_universal_archflags]
+                    set arch_replace_ld  [portconfigure::configure_get_universal_archflags]
                 } else {
                     set arch_replace_cxx [portconfigure::configure_get_archflags cxx]
                     set arch_replace_ld  [portconfigure::configure_get_ld_archflags]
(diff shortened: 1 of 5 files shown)


encounter:
I just noticed that this still doesn't end up generating universal binaries, more fixes may be needed.

MarcusCalhoun-Lopez:
Thank you for the pull request.
If this is still not quite working, I will change it to a draft.

encounter:
I updated the patch to restore the `QMAKE_APPLE_DEVICE_ARCHS` configuration, which successfully allows building universal binaries.

MarcusCalhoun-Lopez:
Thank you for the update.
Let me try to figure out the situation with `QMAKE_APPLE_DEVICE_ARCHS`.
It make take a little while.
As everyone is no doubt aware, the Qt ports are not always simple to work with.

kencu:
> I updated the patch to restore the `QMAKE_APPLE_DEVICE_ARCHS` configuration, which successfully allows building universal binaries.

There have been changes since I worked on this in December, but I found this would not work then, and disabled that, setting it differently.

As @MarcusCalhoun-Lopez says, be very very careful here. Fix one thing, break two others is the general rule.

kencu:
Also, building universal on an arm64 system and building universal on an Intel system can be quite different, based on arm64 systems being able to run Intel code, but not vice-versa.

MarcusCalhoun-Lopez:
Sorry this has taken a while to properly review.

Replacing `configure_get_universal_cflags` and `configure_get_universal_ldflags` with `configure_get_universal_archflags` is entirely reasonable since, as you say, they no longer exist.

Unfortunately, I cannot test adding `QMAKE_APPLE_DEVICE_ARCHS` since I do not have access to an Apple Silicon machine.
However, the potential danger should be minimal (famous last words), and if you can get it to work, that is great.
There should be no need for a revision increase since this change should only affect the universal build, which is broken anyway.

Thank you for the patch.

PR #10652, opened 2021-04-12
dotnet: new set of ports

tsabirgaliev:
These ports come as a single patch because they are usually used together. I broke them into fine grained parts to make updates easier.

The LTS versions 2.1 and 3.1 can be installed with *Current* version side-by-side. These are the *supported* versions. More info on versioning is [here](https://dotnet.microsoft.com/platform/support/policy).

`dotnet-sdk` depends from `dotnet-runtime` + `aspnet-runtime`. All `*-runtime` ports depend on `dotnet-cli`

Closes: https://trac.macports.org/ticket/58655

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/dotnet/aspnetcore-runtime-2.1/Portfile
+++ b/dotnet/aspnetcore-runtime-2.1/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                aspnetcore-runtime-2.1
+version             2.1.27
+revision            0
+categories          dotnet
+license             MIT
+maintainers         {tsabirgaliev @tsabirgaliev} openmaintainer
+
+description         ASP.NET Core is a cross-platform .NET framework for building modern \
+                    cloud-based web applications on Windows, Mac, or Linux.
+
+long_description    .NET is a free, cross-platform, open source developer platform \
+                    for building many different types of applications. \
+                    \
+                    With .NET, you can use multiple languages, editors, and libraries \
+                    to build for web, mobile, desktop, games, and IoT.
+
+homepage            https://dotnet.microsoft.com/
+platforms           darwin
+supported_archs     x86_64
+distname            aspnetcore-runtime-${version}-osx-x64
+master_sites        https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/${version}/
+
+checksums           rmd160  9c801205795efa5529c2aea823f4f96b83872c5c \
+                    sha256  18c2ea1f5da7a977e46000617b8d80b14ade569e728a5d0468237b124a5d94c8 \
+                    size    58547763
+
+worksrcdir          ${name}-${version}
+extract.mkdir       yes
+
+use_configure       no
+
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.AspNetCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    move ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
+}
--- a/dotnet/aspnetcore-runtime-3.1/Portfile
+++ b/dotnet/aspnetcore-runtime-3.1/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                aspnetcore-runtime-3.1
+version             3.1.14
+revision            0
+categories          dotnet
+license             MIT
+maintainers         {tsabirgaliev @tsabirgaliev} openmaintainer
+
+description         ASP.NET Core is a cross-platform .NET framework for building modern \
+                    cloud-based web applications on Windows, Mac, or Linux.
+
+long_description    .NET is a free, cross-platform, open source developer platform \
+                    for building many different types of applications. \
+                    \
+                    With .NET, you can use multiple languages, editors, and libraries \
+                    to build for web, mobile, desktop, games, and IoT.
+
+homepage            https://dotnet.microsoft.com/
+platforms           darwin
+supported_archs     x86_64
+distname            aspnetcore-runtime-${version}-osx-x64
+master_sites        https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/${version}/
+
+checksums           rmd160  db3325f3e2f81861ceb95c95a33f5122a18a4dc2 \
+                    sha256  0aa95b6a342c2b3c0d5116efb96bf3a76e17a118ef98920432a78b6a42ff1927 \
+                    size    37762429
+
+worksrcdir          ${name}-${version}
+extract.mkdir       yes
+
+use_configure       no
+
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.AspNetCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    move ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
+}
--- a/dotnet/aspnetcore-runtime-devel/Portfile
+++ b/dotnet/aspnetcore-runtime-devel/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                aspnetcore-runtime-devel
+version             5.0.5
+revision            0
+categories          dotnet
+license             MIT
+maintainers         {tsabirgaliev @tsabirgaliev} openmaintainer
+
+description         ASP.NET Core is a cross-platform .NET framework for building modern \
+                    cloud-based web applications on Windows, Mac, or Linux.
+
+long_description    .NET is a free, cross-platform, open source developer platform \
+                    for building many different types of applications. \
+                    \
+                    With .NET, you can use multiple languages, editors, and libraries \
+                    to build for web, mobile, desktop, games, and IoT.
+
+homepage            https://dotnet.microsoft.com/
+platforms           darwin
+supported_archs     x86_64
+distname            aspnetcore-runtime-${version}-osx-x64
+master_sites        https://dotnetcli.azureedge.net/dotnet/aspnetcore/Runtime/${version}/
+
+checksums           rmd160  953b420090625a2c25d205e1cfdc252aa0946dd0 \
+                    sha256  48abb337affc1439f038225bcdff0114f15a57ed9e63b61696d5019d35b39186 \
+                    size    38794603
+
+worksrcdir          ${name}-${version}
+extract.mkdir       yes
+
+use_configure       no
+
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.AspNetCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    move ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
+}
--- a/dotnet/dotnet-cli/Portfile
+++ b/dotnet/dotnet-cli/Portfile
@@ -1,0 +1,54 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                dotnet-cli
+version             5.0.5
+revision            0
+categories          dotnet
+license             MIT
+maintainers         {tsabirgaliev @tsabirgaliev} openmaintainer
+
+description         The .NET command-line interface (CLI) is a cross-platform toolchain \
+                    for developing, building, running, and publishing .NET applications.
+
+long_description    .NET is a free, cross-platform, open source developer platform \
+                    for building many different types of applications. \
+                    \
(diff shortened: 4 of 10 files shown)


devel/dotnet-sdk/Portfile
```diff
@@ -0,0 +1,55 @@
...
+long_description    .NET is a free, cross-platform, open source developer platform \
+                    for building many different types of applications. \
+                    \
+                    With .NET, you can use multiple languages, editors, and libraries \
+                    to build for web, mobile, desktop, games, and IoT.
+
+homepage            https://dotnet.microsoft.com/
+platforms           darwin
+distname            ${name}-${version}-osx-x64
+master_sites        https://download.visualstudio.microsoft.com/download/pr/a7d92db4-5f1f-4478-9caf-6e45c6ac8b09/2363d553132aaed687265c563431673b/
```
judaew, line 22: I advise moving the hash from the link into a separate variable and place it after the version with a comment. I think it will be so more convenient for you to update the port in the future.

```tcl
version 5.0.202
set ver_hash a7d92db4-5f1f-4478-9caf-6e45c6ac8b09/2363d553132aaed687265c563431673b
# Can be found in https://dotnet.microsoft.com/download/dotnet/5.0

...

master_sites https://download.visualstudio.microsoft.com/download/pr/${ver_hash}/
```

devel/dotnet-sdk/Portfile
```diff
@@ -0,0 +1,55 @@
...
+         ${worksrcpath}/host \
+         ${worksrcpath}/packs \
+         ${worksrcpath}/sdk \
+         ${worksrcpath}/shared \
+         ${worksrcpath}/templates \
+         ${target}
+
+    # Symlink gradle into the bin directory
+    ln -s ${dotnet_home}/dotnet ${destroot}${prefix}/bin
+}
```
judaew, line 55: At the end you can add a livecheck to keep track of updates with `port livecheck <portname>`:

```tcl
livecheck.url       https://raw.githubusercontent.com/dotnet/core/main/release-notes/5.0/releases.json
livecheck.regex     ${name}-(\[0-9\.\]+)-osx
```

devel/dotnet-sdk/Portfile
```diff
@@ -0,0 +1,55 @@
...
+
+use_configure       no
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet/${name}
+    set target ${destroot}${dotnet_home}
+
+    # Create the target java directory
```
judaew, line 39: `java directory`, it's typo?

devel/dotnet-sdk/Portfile
tsabirgaliev: @judaew yep, good point

devel/dotnet-sdk/Portfile
tsabirgaliev: @judaew thank you!

tsabirgaliev:
@judaew what do you think about having multiple subports per sdk/runtime version?

dotnet/aspnetcore-runtime-3.1/Portfile
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                aspnetcore-runtime-3.1
```
judaew, line 5: I think the version is better to indicate without a dot, like this in ports ruby30, python39, etc.

dotnet/aspnetcore-runtime-current/Portfile
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                aspnetcore-runtime-current
```
judaew, line 5: Instead of "current", it's better to specify a version, eg 50.

judaew:
Ideally would be to have one commit for each Portfile.

dotnet/aspnetcore-runtime-3.1/Portfile
tsabirgaliev: That was my thought initially, but according to their release cadence, dotnet 10 is coming in 2025, that is not too far away from today

dotnet/aspnetcore-runtime-current/Portfile
tsabirgaliev: here the motivation is to signal that this is a 'rolling' release, that is *not* too stable and should not be used for anything other that experimenting. whenever 6.0 is released, current will switch to next 'odd' major *current* release, that is 7.x. If I name it *-5, it will become obsolete too quickly (3 months after 6.0 release, that is coming this year)

tsabirgaliev:
> Ideally would be to have one commit for each Portfile.

@judaew I'm not even sure that all these have to be separate ports, rather than single port with subports. Unfortunately, I have not enough macports competence to do that for now ))

I can break this into commit per port after reaching consensus on the whole layout

dotnet/aspnetcore-runtime-current/Portfile
tsabirgaliev: May be it is better to also explicitly state that in port description

ryandesign: If it's "not too stable" then the correct suffix in MacPorts would probably be "-devel".

dotnet/aspnetcore-runtime-3.1/Portfile
ryandesign: Removing the dot between version components is an old habit in MacPorts that we may never completely get rid of, but many newer ports are not removing the dot because of the confusion that can result when one or both numbers go to a second digit, for example clang-9.0, bazel-3.7, and even some older ports like perl5.28.

dotnet/aspnetcore-runtime-2.1/Portfile
```diff
@@ -0,0 +1,44 @@
...
+
+long_description    .NET is a free, cross-platform, open source developer platform \
+                    for building many different types of applications. \
+                    \
+                    With .NET, you can use multiple languages, editors, and libraries \
+                    to build for web, mobile, desktop, games, and IoT.
+
+homepage            https://dotnet.microsoft.com/
+platforms           darwin
+distname            aspnetcore-runtime-${version}-osx-x64
```
ryandesign, line 22: Presumably the "x64" in the distname indicates that this only works on x86_64 systems. If so, the port should indicate this by specifying `supported_archs x86_64`.

dotnet/aspnetcore-runtime-2.1/Portfile
```diff
@@ -0,0 +1,45 @@
...
+PortSystem          1.0
+
+name                aspnetcore-runtime-2.1
+version             2.1.27
+revision            0
+categories          dotnet
+license             MIT
+maintainers         {tsabirgaliev @tsabirgaliev} openmaintainer
+
+description         ASP.NET Core is a cross-platform .NET framework for building modern cloud-based web applications on Windows, Mac, or Linux.
```
cjones051073, line 12: Could you add a line break in the description here, just to limit the line length a bit ?

dotnet/aspnetcore-runtime-2.1/Portfile
```diff
@@ -0,0 +1,45 @@
...
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.AspNetCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    copy ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
```
cjones051073, line 44: I guess `move` instead of `copy` would be better here ?

dotnet/aspnetcore-runtime-3.1/Portfile
```diff
@@ -0,0 +1,45 @@
...
+version             3.1.14
+revision            0
+categories          dotnet
+license             MIT
+maintainers         {tsabirgaliev @tsabirgaliev} openmaintainer
+
+description         ASP.NET Core is a cross-platform .NET framework for building modern cloud-based web applications on Windows, Mac, or Linux.
```
cjones051073, line 12: Please add a line break.

dotnet/aspnetcore-runtime-3.1/Portfile
```diff
@@ -0,0 +1,45 @@
...
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.AspNetCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    copy ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
```
cjones051073, line 44: Again probably better to just move here.

dotnet/aspnetcore-runtime-devel/Portfile
```diff
@@ -0,0 +1,45 @@
...
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.AspNetCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    copy ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
```
cjones051073, line 44: Use move

dotnet/dotnet-cli/Portfile
```diff
@@ -0,0 +1,54 @@
...
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set target ${destroot}${dotnet_home}
+
+    # Create the target directory
+    xinstall -d ${target}
+
+    # Copy over the needed elements of our directory tree
+    copy ${worksrcpath}/dotnet \
```
cjones051073, line 46: Use move

dotnet/dotnet-runtime-2.1/Portfile
```diff
@@ -0,0 +1,45 @@
...
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.NETCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    copy ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
```
cjones051073, line 44: Use move

dotnet/dotnet-runtime-3.1/Portfile
```diff
@@ -0,0 +1,45 @@
...
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.NETCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    copy ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
```
cjones051073, line 44: Use move

dotnet/dotnet-runtime-devel/Portfile
```diff
@@ -0,0 +1,45 @@
...
+depends_run         port:dotnet-cli
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+    set runtime_dir /shared/Microsoft.NETCore.App
+    xinstall -d ${destroot}${dotnet_home}${runtime_dir}
+
+    copy ${worksrcpath}${runtime_dir}/${version} ${destroot}${dotnet_home}${runtime_dir}
```
cjones051073, line 44: Use move

dotnet/dotnet-sdk-2.1/Portfile
```diff
@@ -0,0 +1,47 @@
...
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+
+    # Create the target directory
+    xinstall -d ${destroot}${dotnet_home}/sdk
+
+    copy ${worksrcpath}/sdk/${version} ${destroot}${dotnet_home}/sdk
```
cjones051073, line 46: Use move

dotnet/dotnet-sdk-3.1/Portfile
```diff
@@ -0,0 +1,48 @@
...
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+
+    # Create the target directory
+    xinstall -d ${destroot}${dotnet_home}/sdk
+
+    copy ${worksrcpath}/sdk/${version} ${destroot}${dotnet_home}/sdk
```
cjones051073, line 47: Use move

dotnet/dotnet-sdk-devel/Portfile
```diff
@@ -0,0 +1,48 @@
...
+
+build {}
+
+destroot {
+    set dotnet_home ${prefix}/share/dotnet
+
+    # Create the target directory
+    xinstall -d ${destroot}${dotnet_home}/sdk
+
+    copy ${worksrcpath}/sdk/${version} ${destroot}${dotnet_home}/sdk
```
cjones051073, line 47: Use move

cjones051073 (changes requested):

dotnet/aspnetcore-runtime-2.1/Portfile
tsabirgaliev: why? I don't see many uses of `move worksrcpath -> destroot` in other Portfiles

cjones051073: It just makes sense no ? If ${worksrcpath} is going to be wiped after installation, why perform the additional I/O of a copy compared to a move. Just because other ports use copy doesn't mean it wouldn't make sense there either in most cases.

For sure there are plenty that do use move ...

```
Oberon ~/Projects/MacPorts/ports > git grep move | grep destroot | grep work
<snip>
science/geant4/Portfile:                move ${workpath}/${data.name}${data.version} ${destroot}${geant.datadir}
science/libad9361-iio/Portfile:    move ${destroot}${prefix}/lib/ad9361.framework ${destroot}${frameworks_dir}
science/mpi-doc/Portfile:    move ${worksrcpath}/man/man1 ${destroot}${prefix}/share/man/
science/mpi-doc/Portfile:    move ${worksrcpath}/man/man3 ${destroot}${prefix}/share/man/
<snip>
```

mascguy:
@tsabirgaliev Tair, if you can create separate commits for each port, we'll be able to merge this PR.

tsabirgaliev:
@mascguy 
> create separate commits for each port

done

mascguy:
Beautiful! Merging.

PR #10678, opened 2021-04-15
prest: New port

judaew:
Version: 1.0.8

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/prest/Portfile
+++ b/devel/prest/Portfile
@@ -1,0 +1,229 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/prest/prest 1.0.8 v
+revision            0
+
+categories          devel databases
+license             MIT
+maintainers         {outlook.de:judaew @judaew} openmaintainer
+
+description         pREST (PostgreSQL REST) is REST API for any Postgres database.
+long_description    \
+    pREST (PostgreSQL REST), simplify and accelerate development, instant, \
+    realtime, high-performance on any Postgres application, existing or new
+homepage            https://prestd.com
+
+checksums           ${distname}${extract.suffix} \
+                        rmd160  448a195892366ae2fd5a07888a9c3bb285ffee4f \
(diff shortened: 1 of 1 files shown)


judaew:
Can you help to explain what the error is that causes the checks to fail?

devel/prest/Portfile
```diff
@@ -0,0 +1,227 @@
...
+    pREST (PostgreSQL REST), simplify and accelerate development, instant, \
+    realtime, high-performance on any Postgres application, existing or new
+homepage            https://prestd.com
+
+checksums           ${distname}${extract.suffix} \
+                        rmd160  448a195892366ae2fd5a07888a9c3bb285ffee4f \
+                        sha256  0751ceb27bf9e32cbc67a6e01f657b9532b9eac6aa6198f1929da2c56a3316c1 \
+                        size    97783
+
+build.args          "-o prestd cmd/prestd/main.go"
```
herbygillot, line 22: You can drop the double quotes, and you don't have to specify the specific Go file as the build target.

The following should work:
```
build.args            -o prestd ./cmd/prestd
```

herbygillot:
> Can you help to explain what the error is that causes the checks to fail?

https://dev.azure.com/macports/macports-ports/_build/results?buildId=13619&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=08703039-49d0-5b78-7691-5b9414c62c1d&l=1397

Looks like a vendored Go dependency whose Github repo URL is different than its module name.  This is a problem one runs into quite frequently when using `go2port`.

The error states that it can't find `jtolds-gls`, likely caused by this vendored dependency:  **github.com/jtolds/gls**

If you visit that URL, you will actually see the URL resolve to: _github.com/**jtolio**/gls_  (_jtolio_ instead of _jtolds_)

So you need to correct the name of the vendored dependency (to **github.com/jtolio/gls**), and you may have to adjust the checksum after doing so.  You may encounter this for several other vendored dependencies.

devel/prest/Portfile
judaew: Thanks 🌟

judaew:
Oh, it's sly! I tried to fix the checksums before, but I didn't know to need another name of the dependency. Thanks so much.

herbygillot:
Np.  Let's also ensure a `revision` field.

judaew:
Done

herbygillot:
Thanks @judaew

PR #10692, opened 2021-04-16
materialx: update to version 1.38.0

jasonliu--:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/graphics/materialx/Portfile
+++ b/graphics/materialx/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           cmake 1.1
 
-github.setup        materialx MaterialX 1.37.4 v
+github.setup        materialx MaterialX 1.38.0 v
 name                materialx
 categories          graphics
 platforms           darwin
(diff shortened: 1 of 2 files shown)


graphics/materialx/Portfile
```diff
@@ -79,6 +79,12 @@ post-destroot {
     if {[file exists ${destroot}${prefix}/README.md]} {
         move ${destroot}${prefix}/README.md $mtlx_share_path
     }
+
+    fs-traverse f $mtlx_share_path/resources {
+        if {[file isfile $f] && [regexp {CMakeLists.txt} $f]} {
```
ryandesign, line 84: Seems like just a string equality comparison of the filename would have been simpler and more accurate?
```suggestion
        if {[file isfile $f] && [file tail $f] eq "CMakeLists.txt"} {
```

jasonliu--: Yes, it would be simpler. I didn't know you could do that.

ryandesign: Yeah `file tail` gets used a lot for this kind of thing. Tcl has lots of other `file` commands that you may find useful from time to time. See https://www.tcl.tk/man/tcl8.5/TclCmd/file.htm

PR #10726, opened 2021-04-19
PULL_REQUEST_TEMPLATE: add machine hardware name

jrjsmrtn:
#### Description

Update "Tested On " section in GitHub PR Template to include hardware machine name.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,7 +11,7 @@
 
 ###### Tested on
 <!-- Triple-click and copy the next line and paste it into your shell. It will copy your OS and Xcode version to the clipboard. Paste it here replacing this section.
-printf "%s\n" "macOS $(sw_vers -productVersion) $(sw_vers -buildVersion)" "$(xcodebuild -version|awk 'NR==1{x=$0}END{print x" "$NF}')"|tee /dev/tty|pbcopy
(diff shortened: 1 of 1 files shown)


raimue:
Technically speaking, `uname -m` is the architecture of the kernel. A long time ago, on the first Intel 64-bit generations, the macOS kernel was still 32-bit, while the userland was already 64-bit. For reference, there is a nice summary in mesonbuild/meson#6187. However, I think `uname -m` is good enough for now and I am not aware of any other (short) command to get this.

As a side note for the future, what about using the output of `port platform` for this? Ideally we should extend it to also include the Xcode version and even better it could show the default `build_arch`, which is actually what we would like to know. Then filling this field would could become much simpler.

jrjsmrtn:
> Le 20 avr. 2021 à 20:58, Rainer Müller ***@***.***> a écrit :
> 
> Technically speaking, uname -m is the architecture of the kernel.
> 
I took it from the `uname` man page:

	-m print the machine hardware name.

> As a side note for the future, what about using the output of port platform for this? Ideally we should extend it to also include the Xcode version and even better it could show the default build_arch, which is actually what we would like to know. Then filling this field would could become much simpler.
> 
Sounds good ! :-)

Another idea: add the `supported_archs` field to the `port info` subcommand (#62715 <https://trac.macports.org/ticket/62715>)

G.

PR #10737, opened 2021-04-20
igraph: add variants

szhorvat:
* choice for BLAS/LAPACK
 * TLS variant
 * remove obsolete livecheck

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G8022 on x86_64
Xcode 11.3.1 11C504

###### Verification 
Have you

### DIFF
--- a/math/igraph/Portfile
+++ b/math/igraph/Portfile
@@ -3,9 +3,10 @@
 PortSystem          1.0
 PortGroup           cmake 1.1
 PortGroup           github 1.0
+PortGroup           linear_algebra 1.0
 
 github.setup        igraph igraph 0.9.2
-revision            0
+revision            1
 github.tarball_from releases
 
 categories          math science devel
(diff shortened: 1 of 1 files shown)


math/igraph/Portfile
```diff
@@ -30,36 +31,41 @@ checksums           rmd160  433e8cb78287600b6d8b2d9173faa2e8ea5d4f42 \
...
+                            -DIGRAPH_USE_INTERNAL_LAPACK=OFF
+
+    # Link to chosen BLAS/LAPACK
+    configure.args-append   ${cmake_linalglib}
+}
+
+variant tls description {Build thread-safe version} {
+    compiler.thread_local_storage   yes
+    configure.args-append           -DIGRAPH_ENABLE_TLS=ON
+}
```
mf2k, line 65: Why is this a variant? Why not always enable this? No dependencies are added.

szhorvat: @mf2k I would like to have a consistent build on all versions of macOS, with the same features. I don't know if it will build correctly on older macOS with TLS. I think it won't on 10.6.

I believe most users of igraph don't need TLS. Also, thread-safety is not mature or well tested in igraph. Finally, it depends on the thread safety of dependencies as well. E.g. GLPK is only built to be thread safe on macOS 10.11 and later.  I kept it off by default for these reasons.

Would it be possible to keep tls as a variant, and have it enabled by default only for recent macOS versions? Is this possible with MacPorts? I know it is possible to compile with TLS only on some versions, but having an explicit variant, as it would make it absolutely clear to users what features are enabled in the installed build.

szhorvat: I force-pushed a version which enables it unconditionally, but I would prefer not to go with that option (and break the port on old macOS) if possible.

mf2k: I see. You are the maintainer. If you need the variant to support older versions of macOs, that is fine.

mf2k: Make the variant default though because most people are on macOS beyond 10.6.

mf2k: Actually you can make it conditionally default for macOS > 10.6.

szhorvat: Done, please have another look Frank.

mf2k (approved):

PR #10740, opened 2021-04-20
cargo_fetch PG, rust, cargo: Add support for using legacysupport PG 

cjones051073:
https://trac.macports.org/ticket/62656

Support older systems requiring legacy support 

    - Patch bootstrapping binaries to use legacysupport dylib for missing libSystem calls
    - Wrap underlying compiler commands used by rust/cargo to enforce use of legacy support flags

Fixes things perfectly for macOS10.9. rust, cargo install fine and was able to build a number of cargo/rust based ports.

Work on older systems is WIP, as currently failing for other unknown reasons.

@kencu FYI

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/lang/rust/Portfile
+++ b/lang/rust/Portfile
@@ -2,9 +2,15 @@
 
 PortSystem          1.0
 
+PortGroup           compiler_blacklist_versions 1.0
+PortGroup           legacysupport 1.1
+
+legacysupport.use_static                    yes
+legacysupport.newest_darwin_requires_legacy 15
+
 name                rust
 version             1.51.0
-revision            1
+revision            2
 categories          lang devel
 platforms           darwin
 supported_archs     x86_64 arm64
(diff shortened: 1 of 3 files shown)


herbygillot:
I've gotta be honest, I'm still pretty wary of doing any `legacysupport` modifications to anything Rust-related.

lang/rust/Portfile
```diff
@@ -2,6 +2,11 @@
...
 PortSystem          1.0
 
+# Needed to patch the bootstrap binaries and for the final build
+PortGroup                     legacysupport 1.1
+legacysupport.use_static                    yes
+legacysupport.newest_darwin_requires_legacy 13
+
 name                rust
 version             1.51.0
 revision            1
```
herbygillot, line 12: Shouldn't Rust's revision be bumped for all these changes?

cjones051073:
> I've gotta be honest, I'm still pretty wary of doing any `legacysupport` modifications to anything Rust-related.

I understand, there was problems last time. I know why there was problems last time this was tried, and this time I avoid them. For one, last time the use of legacysupport was not limited to 10.9 or older, so it was also used on 10.10 and 10.11. This time it is currently limited to 10.9 and older, so will only affect platforms rust currently does not build on anyway. 

The other reason it shouldn’t be a problem this time is i use the static link option, which avoids the issue with the internal llvm build, and also wrap the commands to make completely sure the right flags are used.

lang/rust/Portfile
cjones051073: No. Nothing is changed at all for 10.10 and new, and for 10.9 and older rust doesn’t currently build.

cjones051073:
Tested now extensively on a number of systems. Fixes build of rust itself on 10.9, and also on 10.{10-11} a number of port builds that currently fail due to missing features in libSystem, like clock_gettime.

PR #10745, opened 2021-04-20
git-interactive-rebase-tool: Update to 2.1.0

judaew:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91 on x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/git-interactive-rebase-tool/Portfile
+++ b/devel/git-interactive-rebase-tool/Portfile
@@ -4,11 +4,11 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           cargo  1.0
 
-github.setup        MitMaro git-interactive-rebase-tool 2.0.0
+github.setup        MitMaro git-interactive-rebase-tool 2.1.0
+revision            0
 
 categories          devel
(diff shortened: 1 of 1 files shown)


judaew:
The port cannot be built correctly because is no architecture in the CARGO_BUILD_TARGET var at build time, e. g.:

```bash
CARGO_BUILD_TARGET='-apple-darwin'
# -----------------^^^
```

I don't know why this is happening, but it looks like a bug in PortGroups Cargo.

herbygillot:
My guess is that it's `supported_archs noarch`... I have no idea why that's even there.

This is a `cargo` project, meaning Rust code that gets compiled down to a a specific processor architecture.  So this is either going to be x86_64 _or_ arm64.  `noarch` only makes sense for code that doesn't get compiled and run via interpreter usually, like Python, Perl or Javascript.

I would recommend removing that `supported_archs` line and seeing if that changes anything.

judaew:
Thanks, all is good now.

PR #10753, opened 2021-04-21
Add JSON-enabled variant to `mtr`

alanbrent:
#### Description

This change creates a `+json` variant of `mtr` that enables JSON output via the `--json` flag. If upstream would prefer to enable this by default, I'd be happy to refactor it to do so.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91 on x86_64
Xcode 12.4 12D4e

###### Verification

Have you

### DIFF
--- a/net/mtr/Portfile
+++ b/net/mtr/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                mtr
 version             0.94
-revision            0
+revision            1
 categories          net
 platforms           darwin
 license             GPL-2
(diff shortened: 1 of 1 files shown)


herbygillot:
I think it shouldn't hurt to just include it by default.

You will also want to increase the revision since the binary will change as a result of this:
```
revision            1
```

alanbrent:
Thanks for that feedback, makes total sense. This is my first time editing a Portfile :)

Fixed.

herbygillot:
> Thanks for that feedback, makes total sense. This is my first time editing a Portfile :)
> 
> Fixed.

No worries. However I'm still seeing the commit showing the variant and no change to `revision`.

alanbrent:
Alright we should be good now. I missed your point about enabling JSON by default.

net/mtr/Portfile
```diff
@@ -24,14 +24,15 @@ long_description    mtr combines the functionality of the 'traceroute' \
 homepage            http://www.bitwizard.nl/mtr/
 
 depends_build       port:pkgconfig
-depends_lib         port:ncurses
+depends_lib         port:ncurses port:jansson
```
herbygillot, line 27: Let's add a line break + continuation here to have one dep per line:
```
depends_lib         port:ncurses \
                    port:jansson
```

herbygillot:
OK thanks. One minor nitpick and we're good to go.

alanbrent:
All set @herbygillot

herbygillot:
Thanks @alanbrent. Merged.

PR #10763, opened 2021-04-22
netdata-dashboard: new subport

gpanders:
#### Description

Netdata split out the web dashboard from the core repo, so we need to manually pull that down during installation.

The netdata agent (https://github.com/netdata/netdata) includes an older version of the dashboard bundled with it, but this is being kept just for backwards compatibility. New installations need to include the dashboard from https://github.com/netdata/dashboard (see [docs](https://community.netdata.cloud/t/i-have-an-old-version-of-the-netdata-agent-dashboard/1207) and the [discussion with Netdata](https://github.com/netdata/netdata/pull/11013#issuecomment-824987164)).

I welcome any criticism or suggestions from more experienced maintainers on the best way to manage two separate upstream repos. If the method I used here is not optimal I'm more than happy to update it.

###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H524 on x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/sysutils/netdata/Portfile
+++ b/sysutils/netdata/Portfile
@@ -5,7 +5,7 @@ PortGroup               github 1.0
 PortGroup               legacysupport 1.1
 
 github.setup            netdata netdata 1.29.3 v
-revision                0
+revision                1
 
 github.tarball_from     releases
 distname                ${name}-v${version}
@@ -23,9 +23,10 @@ long_description        Netdata's distributed, real-time monitoring Agent \
                         containers, cloud deployments, and edge/IoT devices, and \
                         is perfectly safe to install on your systems mid-incident \
                         without any preparation.
-                        
-compiler.thread_local_storage   yes
- 
+
+compiler.thread_local_storage \
+                        yes
+
 depends_build-append    port:autoconf \
                         port:automake \
                         port:pkgconfig
@@ -53,9 +54,6 @@ configure.args          --disable-dependency-tracking \
                         --enable-dbengine \
                         --with-user=netdata
 
-startupitem.create      yes
-startupitem.executable  ${prefix}/sbin/netdata -D
-
 set netdata_user        netdata
 set netdata_group       netdata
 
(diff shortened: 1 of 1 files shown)


sysutils/netdata/Portfile
```diff
@@ -5,11 +5,19 @@ PortGroup               github 1.0
...
-revision                0
+revision                1
 
 github.tarball_from     releases
 distname                ${name}-v${version}
 
+# Netdata ships the dashboard (netdata/dashboard) separately from the agent,
+# so we need to download and extract that as well.
+set netdata_dashboard_version \
+                        2.16.0
```
cjones051073, line 16: Wouldn’t it make more sense to make a sub-port for the dashboard ? Having this ‘hidden’ second version in the port is not ideal, and making it a subport will make it more transparent.

gpanders: They are not separate programs: the dashboard and the agent are two halves of a whole and neither is very useful without the other. So anyone installing the `netdata` port needs to also get the dashboard. The fact that we have to do this extra step is strictly a consequence of the fact that Netdata (the organization) splits up the source code for each part into separate source repos. Creating a subport (or even a variant) for the dashboard is counter-productive as it basically forces the user to go through an extra step to get a version of the software that works as they would expect.

cjones051073: No, it will not cause the user to have to do an extra step, as you can just make the primary port depend on the subport, so installing the main port will trigger bith to be installed.

cjones051073: But fair enough, as maintainer its your choice, its just personally I dislike it when versions like this are hidden, as you cannot check for updates via livecheck etc.

gpanders: Ok that is fair, but I still don't see the value in doing that. That is like offering two options to the user: "do you want to install the up-to-date version, or an out-of-date version?" What is the benefit of even making that an option?

Maybe I didn't explain the situation very well, but the files from `netdata/dashboard` are just static assets that go into the `web/` directory. An older (legacy) version of those files already exists in the default installation from when the dashboard was bundled in `netdata/netdata`. So if we *don't* install these files, the user still gets a dashboard, just an out-of-date one.

From [I have an old version of the Netdata Agent dashboard](https://community.netdata.cloud/t/i-have-an-old-version-of-the-netdata-agent-dashboard/1207):
> At some point last year, we decided to split the Netdata Agent between the core components and the dashboard. The dashboard was moved to a different GitHub repository named netdata/dashboard.
>
>The Netdata Agent repository continues to have embedded an older version of the dashboard, so even if no code is downloaded from the netdata/dashboard repo, the agent is still usable.
>
>Some maintainers of the Netdata Agent are unaware of this change since it’s not obvious as the Agent works either way!
>
>We are working on letting maintainers know of the situation and update their packages.

If we added a subport, then there would be two choices in MacPorts: `netdata` and `netdata-dashboard`. But *both* of these would actually provide the web dashboard, only one of them (the one in `netdata`) would be old and out-of-date. Even if we make `netdata-dashboard` the default, there are still now two `netdata` entries in the port index, and inevitably someone will ask what the difference is between the two, to which the answer will be: "one of them is updated and one of them isn't".

Pulling in the assets from `netdata/dashboard` doesn't add any additional functionality (which is what a subport implies), it just ensures that everything is up-to-date.

Apologies for the long-winded response, hopefully that provides some clarification.

cjones051073: I was not suggesting you give the user the option of which to install, I agree that would. Ot be useful. My suggestion was to make the sub port a mandatory dependency of the main port, and to notminstall the out of dat assets from the main port.

let me ask another question, do these dashboard assets get updated independently of the main code ? If so, then having them in a subport would allow that to be updated independently of the main port.

gpanders: > I was not suggesting you give the user the option of which to install, I agree that would not be useful. My suggestion was to make the sub port a mandatory dependency of the main port, and to not install the out of date assets from the main port.
> 
> let me ask another question, do these dashboard assets get updated independently of the main code ? If so, then having them in a subport would allow that to be updated independently of the main port.

Ah okay, I misunderstood your original suggestion and I think what you're actually suggesting is becoming clear to me now, and I agree it makes sense, particularly because the answer to your question is "yes".

I'll push up an updated commit momentarily.

gpanders: @cjones051073 Take a look at the changes and let me know if this is what you had in mind or if you have any feedback.

sysutils/netdata/Portfile
```diff
@@ -67,25 +65,66 @@ set netdata_web_dir     ${prefix}/share/${name}/web
...
+        touch ${destroot}${netdata_conf_dir}/.opt-out-from-anonymous-statistics
+
+        # Delete any files provided by netdata-dashboard to avoid conflicts. We
+        # can't just delete the whole directory because this directory contains
+        # some files that are not present in netdata/dashboard. This will
+        # hopefully change someday, see
+        # https://github.com/netdata/dashboard/issues/83
+        foreach f [glob -dir ${netdata_web_dir} * **/*] {
+            delete ${destroot}${f}
+        }
```
gpanders, line 100: @cjones051073 In particular, take a look at this block. This is a bit of a tricky situation, as `netdata/netdata` provides some of the static web files that are *not* present in `netdata/dashboard`, so in order to prevent conflicts upon installation we have to delete the files provided by `netdata/dashboard` while preserving the ones from `netdata/netdata`. If you know of a better way to do this I'm all ears.

sysutils/netdata/Portfile
cjones051073: LGTM, thanks for taking the time to make the changes.

PR #10774, opened 2021-04-23
terminus-font: use python 3.9

exg:


mf2k:
You will need to increase the `revision` by 1 so people who have it installed get this change.

PR #10785, opened 2021-04-25
cdo: update dependencies to use proj7

thiagoveloso:
#### Description
cdo: update dependencies to use proj7

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 11.0.1
Xcode 12.4

###### Verification 
Have you

### DIFF
--- a/science/cdo/Portfile
+++ b/science/cdo/Portfile
@@ -5,7 +5,7 @@ PortGroup                   mpi 1.0
 
 name                        cdo
 version                     1.9.10
-revision                    0
+revision                    1
 platforms                   darwin
 maintainers                 {takeshi @tenomoto} openmaintainer
 license                     GPL-2
@@ -35,7 +35,7 @@ depends_lib                 port:netcdf \
                             port:hdf5 \
                             port:curl \
                             port:udunits2 \
-                            port:proj \
+                            port:proj7 \
                             port:fftw-3
 
 patchfiles                  patch-skip-nc4-test.diff
@@ -43,7 +43,7 @@ patchfiles                  patch-skip-nc4-test.diff
 configure.args              --with-netcdf=${prefix} \
                             --disable-dependency-tracking \
                             --disable-openmp \
-                            --with-proj=${prefix}/lib/proj5 \
+                            --with-proj=${prefix}/lib/proj7 \
                             --with-hdf5=${prefix} \
                             --with-udunits2=${prefix} \
                             --with-fftw3 \


science/cdo/Portfile
```diff
@@ -35,15 +35,15 @@ depends_lib                 port:netcdf \
                             port:hdf5 \
                             port:curl \
                             port:udunits2 \
-                            port:proj \
+                            port:proj7 \
                             port:fftw-3
```
herbygillot, line 39: Since you are changing the dependencies for `cdo`, the port will require a rebuild, which normally gets done with the `cdo`'s version changes.

Since we are changing the dependencies while the version stays the same, you will need to increment the revision.

Please edit this commit and add:
```
revision     1
```

herbygillot (changes requested):

herbygillot:
Also, the commit description should follow the following standard:
```
<port name>: <description>
```

So can you change the first line of your commit description to something like:
```
cdo: update dependencies to use proj7
```

thiagoveloso:
Thanks for the feedback @herbygillot. However, I am not sure how to edit a commit. Is that fine if I submit another commit to include the changes you requested?

cjones051073:
> Thanks for the feedback @herbygillot. However, I am not sure how to edit a commit. Is that fine if I submit another commit to include the changes you requested?

No, thats not ok, as we also want a single commit. You need to amend the current commit in your local checkout and then force push he change back to github.

herbygillot:
> Thanks for the feedback @herbygillot. However, I am not sure how to edit a commit. Is that fine if I submit another commit to include the changes you requested?

Usually MacPorts prefers to see each change to a port as a single commit, so we'd prefer _not_ submit additional commits.

You can change your current commit by _amending_ it.

First go back to your local branch and make your changes.

Then when you're ready, you:

1. Stage them as you normally would:
  ```
  $ git add Portfile
  ```
 2. Then instead of committing the normal way (which would add a new commit), we will use __--amend__  instead.  Amend will add the changes you added in the previous step to the most recent existing commit (your current commit):
   ```
    $ git commit --amend
   ```
 3. As a result of this, the branch on your local machine and your remote branch in Github diverge.  Use force push to overwrite your changes on Github with your amended branch:
   ```
    $ git push --force origin master:master
   ```

thiagoveloso:
> > Thanks for the feedback @herbygillot. However, I am not sure how to edit a commit. Is that fine if I submit another commit to include the changes you requested?
> 
> Usually MacPorts prefers to see each change to a port as a single commit, so we'd prefer _not_ to have additional commits.
> 
> You can change your current commit by _amending_ it.
> 
> First go back to your local branch and make your changes.
> 
> Then when you're ready, you:
> 
> 1. Stage them as you normally would:
> 
> ```
> $ git add Portfile
> ```
> 
> 1. Then instead of committing the normal way (which would add a new commit), we will use **--amend**  instead.  Amend will add the changes you added in the previous step to the most recent existing commit (your current commit):
> 
> ```
>  $ git commit --amend
> ```
> 
> 1. As a result of this, the branch on your local machine and your remote branch in Github diverge.  Use force push to overwrite your changes on Github with your amended branch:
> 
> ```
>  $ git push --force origin master:master
> ```

Ok thanks a lot for the hints, it was easier that I thought. I think I got all the changes right - everything should be ready to go now.

herbygillot (approved):

herbygillot:
Looks good.  Thank you @thiagoveloso.

PR #10788, opened 2021-04-25
kitty: Fix build on macOS 10.12

judaew:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/aqua/kitty/Portfile
+++ b/aqua/kitty/Portfile
@@ -33,7 +33,8 @@ checksums           rmd160  f14a923892e26886f8079b3aaa39bcb362b21699 \
 use_xz              yes
 
 # Changes is already in master, remove the patch in next version
-patchfiles-append   patch-turn-off-werror.diff
+patchfiles-append   patch-turn-off-werror.diff \
+                    patch-fix-compilation-on-macos-10.12.diff
 
 patchfiles-append   patch-kitty-no-deprecated-declarations.diff
 
--- a/aqua/kitty/files/patch-fix-compilation-on-macos-10.12.diff
+++ b/aqua/kitty/files/patch-fix-compilation-on-macos-10.12.diff
@@ -1,0 +1,19 @@
+From: Luflosi <luflosi@luflosi.de>
+Date: Wed, 21 Apr 2021 16:23:42 +0200
+Subject: [PATCH] Fix compilation on macOS 10.12
+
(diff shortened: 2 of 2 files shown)


cjones051073:
Do you know what header these defines are in ? It looks like this is something we should handle via the legacysupport PG @kencu FYI

judaew:
This is a patch from the master branch, but not included in the current version.

https://github.com/kovidgoyal/kitty/commit/6606f5163605247df4b97d21f75ca7d3d07323c2

kencu:
Jason @jasonliu--  has been working on a very complete set of these:

<https://github.com/macports/macports-ports/blob/master/graphics/blender/files/old_appkit_compat/AppKit/AppKit.h>

with the idea that if enough ports use them, we could include them in legacysupport.

The issue we have at present in just dumping them into legacysupport now is that these changes are so extensive it's likely to break as many builds as it fixes, so it has to be able to be toggled appropriately.

kencu:
for example, what happens if you want the legacysupport getline and such fixes, but the AppKit.h redefines break the build on your project? Have to be able to handle things like that without too much incomprehensible Portgroup and #define spaghetti making it collapse.

jasonliu--:
Indeed, @kencu is very much correct. It looks like as time goes on, we are discovering more instances of specific pieces of AppKit that Apple has deprecated/obseleted, but are being used by code authors. I don't yet have a good answer for how to structure the legacysupport for AppKit.h that won't result in the incomprehensible Portgroup and #define spaghetti, but every time another instance is brought to my attention, it does help to give me fuller picture so that hopefully I can eventually add the AppKit fixes to legacysupport in a reasonable way.

For the time being, implementing the #defines in the individual project is the correct course of action. I have added this thread to my notes that I am using to track the AppKit deprecations.

aqua/kitty/Portfile
```diff
@@ -6,7 +6,7 @@ PortGroup           python 1.0
 
 github.setup        kovidgoyal kitty 0.20.1 v
 github.tarball_from releases
-revision            2
+revision            3
```
ryandesign, line 9: If you're only fixing a build failure, then you should not increase the revision.

PR #10793, opened 2021-04-25
Create Janet ecosystem

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/janet-1.0.tcl
+++ b/_resources/port1.0/group/janet-1.0.tcl
@@ -1,0 +1,84 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+# Usage:
+# PortGroup     janet 1.0
+#
+# This PortGroup allows projects to be built with jpm
+#     (see https://janet-lang.org/docs/jpm.html).
+#
+
+# search in ${prefix}/lib/janet/bin and create link in ${prefix}/bin
+options janet.link_bin
+default janet.link_bin          yes
+
+# program to create and maintain library archives
+options janet.compiler.ar
+default janet.compiler.ar       {[janetpg::get_default_ar]}
+
+namespace eval janetpg {
+}
+
+proc janetpg::get_default_ar {} {
+    # no doubt that at some point, for some systems, the ar from cctools will have to be used
+    return /usr/bin/ar
+}
+
+# The build command jpm does not respect CFLAGS/CXXFLAGS.
+# You can set the compilers, but no spaces seem to be allowed.
+# Hence, we cannot simply append the CFLAGS/CXXFLAGS to the compiler name.
+# As a workaround, create custom compilers with CFLAGS/CXXFLAGS baked in.
+pre-configure {
+    # create directory
+    xinstall -d -m 0755 ${worksrcpath}/macports_bin
+    # create empty files with the correct permissions
+    foreach bin {cc cxx ar} {
+        close [open ${worksrcpath}/macports_bin/${bin} w 0755]
+    }
+}
+default configure.pre_args      {}
+default configure.cmd           {
+    /bin/echo \\\\#!/bin/bash >> ${worksrcpath}/macports_bin/cc  && echo exec \\\$CC  \\\$CFLAGS   \\\\\"\\\\$@\\\\\" >> ${worksrcpath}/macports_bin/cc  &&
+    /bin/echo \\\\#!/bin/bash >> ${worksrcpath}/macports_bin/cxx && echo exec \\\$CXX \\\$CXXFLAGS \\\\\"\\\\$@\\\\\" >> ${worksrcpath}/macports_bin/cxx &&
+    /bin/echo \\\\#!/bin/bash >> ${worksrcpath}/macports_bin/ar  && echo exec ${janet.compiler.ar} \\\\\"\\\\$@\\\\\" >> ${worksrcpath}/macports_bin/ar
+}
+default configure.post_args     {}
+
+# Build with jpm.
+default build.cmd               {${prefix}/bin/jpm}
+default build.args              {--verbose
+                                 --offline
+                                 --compiler=${worksrcpath}/macports_bin/cc
+                                 --cpp-compiler=${worksrcpath}/macports_bin/cxx
+                                 --archiver=${worksrcpath}/macports_bin/ar
+                                 build
+                                }
+default build.target            {}
+
+# Install with jpm.
+default destroot.cmd            {${prefix}/bin/jpm}
+default destroot.args           {--verbose
+                                 --offline
+                                 install
+                                }
+default destroot.target         {}
+default destroot.destdir        {}
+
+pre-destroot {
+    xinstall -d -m 0755         ${destroot}${prefix}/lib/janet
+}
+
+post-destroot {
+    if {${janet.link_bin}} {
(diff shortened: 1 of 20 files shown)


MarcusCalhoun-Lopez:
@tsujp since your are the maintainer of janet, does any of this look right?

_resources/port1.0/group/janet-1.0.tcl
```diff
@@ -0,0 +1,92 @@
...
+        close ${fl}
+    }
+}
+post-configure {
+    foreach bin {cc cxx ar} {
+        set fl [open "${worksrcpath}/macports_bin/${bin}" a 0755]
+        puts ${fl} { "$@"}
+        close ${fl}
+    }
+}
```
cjones051073, line 45: Not sure I see any advantage to creating the wrapper scripts in three different phases... 

Wouldn’t it be simpler to just create the entire file in one go in the main configure phase ?

MarcusCalhoun-Lopez: I agree splitting the wrapper script generation into `pre-configure`, `configure`, and `post-configure` is less than ideal.
The problem is that `configure { ... }` means `portconfigure::configure_main` from port [portconfigure.tcl](https://github.com/macports/macports-base/blob/a0172d6655263e9de47e05dd33da0651c1b93ced/src/port1.0/portconfigure.tcl#L1640) does not run, and so none of the `append_to_environment_value configure CFLAGS/CXXFLAGS` commands are called.
I was hoping to just use $CFLAGS add defined by the MacPorts base.

cjones051073: I was not suggesting overriding the entire configure phase but just create the entire wrap script with a single echo command. Line breaks can be passed as \n etc.

MarcusCalhoun-Lopez: Thank you for the clarification.
It would certainly require less code.
I will try to make the change tomorrow.

cjones051073: Actually... you also don’t need to evaluate the environment variables when you write them to file, as you should actually write $CC into the file, so that the variable is then evaluated each time the script is  ran during the build phase.

MarcusCalhoun-Lopez: I did not think `$CC` (and `$CXX`) were set by the MacPorts base in the build phase.
However, I could be wrong.

janet/janet-bundler/Portfile
```diff
@@ -0,0 +1,24 @@
...
+maintainers         nomaintainer
+platforms           darwin
+description         A very simple css and js bundler for janet
+long_description    {*}${description}
+
+checksums           rmd160  fd7ea7ec29852b9cb16e778d9aeae4ffb5ae9377 \
+                    sha256  aeea6442314562f6ee9136d16c17266064bf49b2af95851d271070242ac79b62 \
+                    size    1861
+
+depends_lib         port:janet-path \
```
cjones051073, line 23: Can you always use -append when adding dependencies. It avoids potential future bugs if a portgroup in the future adds dependencies itself, as without -append you would inadvertently remove them.

cjones051073 (changes requested):

_resources/port1.0/group/janet-1.0.tcl
cjones051073: If not you could append them yourself to the build env.

MarcusCalhoun-Lopez: In that case, it seems just easier to write `$CC` and `$CXX` directly to the wrapper file and be done with it.
I cannot image a case where `$CC/$CXX` are different in the configure and build phases.

cjones051073: Yes, but my point is it does not then matter if they are actually set or not when you create the scripts, it only matters that they are set when they are used. So you could create these scripts in any phase, e.g. post-extract or whatever.

cjones051073: Cresting the script in post-extract would also have the advantage you would have to play tricks with echo etc to get them into a single ‘configure’ command, which is a bit ugly. You could just use the standard tcl puts.

MarcusCalhoun-Lopez: I may be missing some important part of your concern.
Please forgive me if that is the case.

It seems to me that the wrapper scripts can *only* be generated during the configure phase since that is the *only* time `$CFLAGS/$CXXFLAGS` are fully set by the MacPorts base.

Were it not for `$CFLAGS/$CXXFLAGS`, then I would agree that there would be much easier ways of generating the wrapper scripts.
Since `$CFLAGS/$CXXFLAGS` must be set during the configure phase, we might as well set `$CC/$CXX` as well and not worry about setting any more environmental variables in the build phase.

janet/janet-bundler/Portfile
MarcusCalhoun-Lopez: I agree that `-append` is safer and causes no harm.
I will look into making the change.
Ideally, port groups should be designed so this is not necessary.

_resources/port1.0/group/janet-1.0.tcl
cjones051073: Its not I have a huge concern, I am sure it works. It just looked a little overly complicated having the file creation split into three parts.

janet/janet-bundler/Portfile
cjones051073: Its not possible to avoid portgroups sometimes having to add dependencies. E.g. the legacy support PG has to decide if or not a port needs a dependency on the port or not. There are other examples I am sure.

MarcusCalhoun-Lopez: Requiring `-append` may indeed sometimes be necessary.
I will make the change to `-append` just to be safe.
My point is only that dependencies should be added by `port::register_callback` in port groups whenever possible (sometimes it is not).

tsujp:
Would we not leave `jpm` to manage janet packages?

Tatsh:
I like the idea of this as long as packages within the Janet ecosystem can have command line utilities that are useful for system tasks, work for those not wanting to deal with yet another package manager, and do not have insane dependency trees (like NPM does for most things). Otherwise, we would just have users use Pip to manage user-installed Python packages.

Is there a killer app in the Janet ecosystem? For Haskell, that's things like Shellcheck and Pandoc. For Python, it's almost anything that has a useful CLI with it, like (in my case) [PyGitUp](https://pypi.org/project/git-up/).

MarcusCalhoun-Lopez:
@tsujp it is true that `jpm` could handle all the installations.
However, global installation of packages into the MacPorts prefix outside of the MacPorts infrastructure is not recommended (see, e.g., [npm](https://github.com/macports/macports-ports/blob/9ee3dd323c9f98e7dba7cf3e26dea9833e52c01c/devel/npm7/Portfile#L86)).
This patch allows MacPorts to install the janet packages, which brings all the advantages MacPorts has to offer.

MarcusCalhoun-Lopez:
@Tatsh I am not aware of any "killer app" in the janet ecosystem.
However, I am new to the language.

cjones051073:
I think its fine to have this available.

cjones051073 (approved):

cjones051073:
FYI - I fixed the build of janet on older systems with https://github.com/macports/macports-ports/commit/daa97db0f396766735063fbd279dd17a73ec34c1 - with that it builds all the way down to 10.6.

PR #10812, opened 2021-04-27
melonds: new port

Tatsh:
#### Description

Nintendo DS emulator.

[Homepage](http://melonds.kuribo64.net/)
[Repository](https://github.com/Arisotura/melonDS)

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91 x86_64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/emulators/melonds/Portfile
+++ b/emulators/melonds/Portfile
@@ -1,0 +1,39 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           github 1.0
+PortGroup           qt5 1.0
+
+github.setup        Arisotura melonDS 0.9.2
+revision            0
+
+name                melonds
+platforms           darwin
+license             GPL-3
+maintainers         {gmail.com:audvare @Tatsh} openmaintainer
+categories          emulators games
+description         Nintendo DS emulator.
+long_description    {*}${description}
+
+checksums           rmd160  87d6585506c8abdae6bf7417708e1b30578b1927 \
+                    sha256  d20ce23725e63be5c60d42c87ed504015aa57ddf57c5161302c98e1837e9acae \
+                    size    1331032
+
+depends_lib-append  port:gettext \
+                    path:lib/pkgconfig/glib-2.0.pc:glib2 \
+                    port:libarchive \
+                    port:libepoxy \
+                    port:libiconv \
+                    port:libsdl2 \
+                    port:libslirp
+
+destroot {
+    move ${workpath}/build/melonDS.app ${destroot}${applications_dir}
+}
+
+if {${os.platform} eq "darwin" && ${os.major} < 18} {
+    post-install {
+        ui_warn "${name} @${version} is only supported on macOS 10.14 or later and may not work properly on your OS version."
+    }
+}


szhorvat:
MacPorts novice here. Just wondering if it isn't better to use the qt5 PortGroup and indicate `qt5.min_version` to get better compatibility with older macOS that isn't supported by the latest Qt.

Tatsh:
> MacPorts novice here. Just wondering if it isn't better to use the qt5 PortGroup and indicate `qt5.min_version` to get better compatibility with older macOS that isn't supported by the latest Qt.

I'd be okay with this but I do not know what is the minimum Qt version melonDS requires. Also in the `CMakeLists.txt` they are setting minimum macOS version to 10.14. Not sure if this will make a difference in the MacPorts CMake toolchain.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           github 1.0
+
+github.setup        Arisotura melonDS 0.9.2
+
```
herbygillot, line 8: Can you add a revision field?

```
revision            0
```

emulators/melonds/Portfile
```diff
@@ -0,0 +1,39 @@
...
+name                melonds
+platforms           darwin
+license             GPL-3
+maintainers         {gmail.com:audvare @Tatsh}
```
herbygillot, line 12: Please also add `openmaintainer` to your maintainers section.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,39 @@
...
+categories          emulators games
+description         Nintendo DS emulator.
+long_description    ${description}
```
herbygillot, line 15: This should be written with array expansion:
```
```

herbygillot:
> > MacPorts novice here. Just wondering if it isn't better to use the qt5 PortGroup and indicate `qt5.min_version` to get better compatibility with older macOS that isn't supported by the latest Qt.
> 
> I'd be okay with this but I do not know what is the minimum Qt version melonDS requires. Also in the `CMakeLists.txt` they are setting minimum macOS version to 10.14. Not sure if this will make a difference in the MacPorts CMake toolchain.

You can use the Qt5 portgroup and still have your own minimum OS version check.  I have no preferences either way though, so others should chime in.

Tatsh:
> > > MacPorts novice here. Just wondering if it isn't better to use the qt5 PortGroup and indicate `qt5.min_version` to get better compatibility with older macOS that isn't supported by the latest Qt.
> > 
> > 
> > I'd be okay with this but I do not know what is the minimum Qt version melonDS requires. Also in the `CMakeLists.txt` they are setting minimum macOS version to 10.14. Not sure if this will make a difference in the MacPorts CMake toolchain.
> 
> You can use the Qt5 portgroup and still have your own minimum OS version check. I have no preferences either way though, so others should chime in.

I replaced the OS check with `qt5.min_version 5.12` as this is the earliest version that macOS 10.14 accepts. This may allow users on earlier versions of macOS to at least build the emulator. Not sure if it will work properly at runtime, so I put in a warning. Upstream probably will look into issues on earlier OS versions as long as they are not definitely caused by having an older macOS.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,42 @@
...
+                    size    1331032
+
+depends_lib         port:gettext \
+                    port:glib2 \
+                    port:libarchive \
+                    port:libepoxy \
+                    port:libiconv \
+                    port:libsdl2 \
+                    port:libslirp \
+                    port:qt5-qtbase
```
herbygillot, line 30: If you are using the `qt5` portgroup, then you no longer need to add qtbase as an explicit dependency. You should be able to remove this.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,41 @@
...
+maintainers         {gmail.com:audvare @Tatsh} openmaintainer
+categories          emulators games
+description         Nintendo DS emulator.
+long_description    {*}${description}
+
+checksums           rmd160  87d6585506c8abdae6bf7417708e1b30578b1927 \
+                    sha256  d20ce23725e63be5c60d42c87ed504015aa57ddf57c5161302c98e1837e9acae \
+                    size    1331032
+
+depends_lib         port:gettext \
```
cjones051073, line 23: Please use `depends_lib-append` here, to avoid potentially wiping out previous set dependencies.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,41 @@
...
+
+qt5.min_version     5.12
+
+destroot {
+    copy ${workpath}/build/melonDS.app ${destroot}${applications_dir}
+}
+
+if {${os.major} < 18} {
+    post-install {
+        ui_warn "${name} @${version} is only supported on macOS 10.14 or later and may not work properly on your OS version."
```
cjones051073, line 39: Are you saying it doesn’t build at all, or builds but the results might be unstable, or ??

cjones051073 (changes requested):

cjones051073:
> > > MacPorts novice here. Just wondering if it isn't better to use the qt5 PortGroup and indicate `qt5.min_version` to get better compatibility with older macOS that isn't supported by the latest Qt.
> > 
> > 
> > I'd be okay with this but I do not know what is the minimum Qt version melonDS requires. Also in the `CMakeLists.txt` they are setting minimum macOS version to 10.14. Not sure if this will make a difference in the MacPorts CMake toolchain.
> 
> You can use the Qt5 portgroup and still have your own minimum OS version check. I have no preferences either way though, so others should chime in.

A qt check is not directly equivalent to an os check. 

If you don’t know what qt version is required, lets leave it out and see what happens in the buildbots. It can always be added later on.

Regarding supported OS its very common that within macports we can build projects on more OS versions that upstream expects, due to various tools with have in a our toolbox, like newer compilers for each OS, a legacy support package that adds features missing in the OS, and general dependencies that aren’t default in the OS. So again, I would not put too much into upstreams claims and lets see what the build bots make of them and go from there.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,41 @@
...
+                    port:libarchive \
+                    port:libepoxy \
+                    port:libiconv \
+                    port:libsdl2 \
+                    port:libslirp
+
+qt5.min_version     5.12
+
+destroot {
+    copy ${workpath}/build/melonDS.app ${destroot}${applications_dir}
```
cjones051073, line 34: As The workpath is removed after the build, just use move here instead of copy.

emulators/melonds/Portfile
Tatsh: If it does build, all bets are off because upstream only wants 10.14 or later. That's why I put this in `post-install`.

cjones051073: well ok, lets se what the reasons are in the buildbots then.

emulators/melonds/Portfile
ryandesign: The workpath is not removed after the build phase, but yes it is removed after the port is installed.

It is pretty normal for destroot phases to copy files into the destroot. If the app is extremely large then moving it could be better, otherwise copying it is perfectly fine.

emulators/melonds/Portfile
```diff
@@ -0,0 +1,39 @@
...
+categories          emulators games
+description         Nintendo DS emulator.
+long_description    {*}${description}
+
+checksums           rmd160  87d6585506c8abdae6bf7417708e1b30578b1927 \
+                    sha256  d20ce23725e63be5c60d42c87ed504015aa57ddf57c5161302c98e1837e9acae \
+                    size    1331032
+
+depends_lib-append  port:gettext \
+                    port:glib2 \
```
ryandesign, line 24: Allow glib2-devel to satisfy the dependency:
```suggestion
                    path:lib/pkgconfig/glib-2.0.pc:glib2 \
```

herbygillot (approved):
lgtm

emulators/melonds/Portfile
```diff
@@ -0,0 +1,39 @@
...
+                    port:libepoxy \
+                    port:libiconv \
+                    port:libsdl2 \
+                    port:libslirp
+
+destroot {
+    move ${workpath}/build/melonDS.app ${destroot}${applications_dir}
+}
+
+if {${os.major} < 18} {
```
cjones051073, line 35:
```suggestion
if {${os.platform} eq “darwin” && ${os.major} < 18} {
```

cjones051073 (changes requested):

emulators/melonds/Portfile
Tatsh: Smart quotes in this suggestion tricked me.

cjones051073 (approved):

PR #10816, opened 2021-04-27
hwloc: update to 2.4.1; add new variants opencl and cuda

mascguy:
#### Description

* Update hwloc to 2.4.1
* Explicitly disable OpenCL and CUDA support by default, to avoid issues with certain MacOS releases (like Big Sur).
* Add new variants `opencl` and `cuda`, to allow users to re-enable that functionality if desired.
* Add port note mentioning the change, along with the new variants.

###### Type(s)
- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

###### Verification 
Have you

### DIFF
--- a/devel/hwloc/Portfile
+++ b/devel/hwloc/Portfile
@@ -5,7 +5,7 @@ PortGroup           active_variants 1.1
 
 name                hwloc
 set branch          2.4
-version             ${branch}.0
+version             ${branch}.1
 epoch               1
 categories          devel
 platforms           darwin
@@ -23,16 +23,27 @@ long_description \
 homepage            https://www.open-mpi.org/projects/hwloc/
 master_sites        https://www.open-mpi.org/software/hwloc/v${branch}/downloads/
 
-checksums           rmd160  863d1f1f78b20f8d3d370be8b85aa1c636be9233 \
-                    sha256  30404065dc1d6872b0181269d0bb2424fbbc6e3b0a80491aa373109554006544 \
-                    size    6965149
+checksums           rmd160  246a33becaaf623aec8d6fcdd6efd5864c248d72 \
+                    sha256  4267fe1193a8989f3ab7563a7499e047e77e33fed8f4dec16822a7aebcf78459 \
+                    size    6956102
 
 depends_build       port:pkgconfig
 depends_lib-append  port:libxml2
-configure.args      --without-x --disable-cairo
+configure.args      --without-x \
+                    --disable-cairo \
+                    --disable-cuda \
+                    --disable-opencl
 
 livecheck.type      regex
 livecheck.regex     "${name} v(\[0-9.\]+) "
+
+variant cuda description {Enable CUDA support} {
+    configure.args-replace  --disable-cuda --enable-cuda
+}
+
+variant opencl description {Enable OpenCL support} {
+    configure.args-replace  --disable-opencl --enable-opencl
+}
 
 variant gui description {Add graphical output capability} {
     depends_lib-append      path:bin/openssl:openssl \
@@ -48,3 +59,8 @@ variant gui description {Add graphical output capability} {
 }
 
 default_variants    +gui
+
+notes {
+  OpenCL and CUDA support is now disabled by default. They may be enabled by new\
+  variants opencl and cuda, respectively.
+}


mascguy:
@i0ntempest any thoughts/concern with this change?

ryandesign:
We don't use variants whose names begin with `no_` anymore. Use positively-named variants instead, like `cuda` and `opencl` and turn them on by default with `default_variants +cuda +opencl`.

mascguy:
Well, here's the problem: The default behavior of hwloc - when no configure arguments for OpenCL and CUDA are specified - is to automatically enable those features _if supported by the platform_. If the features are not supported, then the build succeeds, with the functionality disabled.

However, when they are explicitly enabled, hwloc _requires_ that the support is there. And if not, the build fails.

CUDA, in particular, isn't available for most users, so we wouldn't want to enable that by default. OpenCL should be there in principle for most MacOS releases, though it isn't for MacOS 10.5. And it's deprecated, so it's likely to disappear at some point.

The easier route would be to simply leave them both disabled by default, though that changes the behavior of hwloc. Whether that has any tangible user impact is another story, but still worth mentioning.

So... thoughts?

ryandesign:
> Well, here's the problem: The default behavior of hwloc - when no configure arguments for OpenCL and CUDA are specified - is to automatically enable those features _if supported by the platform_. If the features are not supported, then the build succeeds, with the functionality disabled.

Well this sounds all kinds of problematic. What is the decision based on: what is "the platform"? Does it depend on the specific CPU model? If so, then we have a problem if we build a binary on a server with one CPU (like our Xserves with Intel Xeon processors) and distribute it to a user with a Mac with a different processor (like an Intel Core 2 Duo). The solution to that would be for the port to clear `archive_sites` so that users do not get binaries from us and always build from source on their own computers.

What is your motivation for wanting to add these variants?

mascguy:
> > Well, here's the problem: The default behavior of hwloc - when no configure arguments for OpenCL and CUDA are specified - is to automatically enable those features _if supported by the platform_. If the features are not supported, then the build succeeds, with the functionality disabled.
> 
> Well this sounds all kinds of problematic. What is the decision based on: what is "the platform"? Does it depend on the specific CPU model? If so, then we have a problem if we build a binary on a server with one CPU (like our Xserves with Intel Xeon processors) and distribute it to a user with a Mac with a different processor (like an Intel Core 2 Duo). The solution to that would be for the port to clear `archive_sites` so that users do not get binaries from us and always build from source on their own computers.
> 
> What is your motivation for wanting to add these variants?

The platform support relates to what header files exist, rather than hardware like CPU types. So I don't foresee a need to stop using binaries, otherwise users would have likely encountered issues long ago.

As for the goal: To be able to explicitly disable OpenCL, and/or CUDA, if the user wishes. Otherwise, retain the default behavior.

Does that make sense?

ryandesign:
> As for the goal: To be able to explicitly disable OpenCL, and/or CUDA, if the user wishes. Otherwise, retain the default behavior.

I got that, but why would the user wish to do that?

mascguy:
> > As for the goal: To be able to explicitly disable OpenCL, and/or CUDA, if the user wishes. Otherwise, retain the default behavior.
> 
> I got that, but why would the user wish to do that?

One of our Big Sur users is having issues with OpenMPI, caused by use of OpenCL. And the latter is being pulled in by hwloc.

Here's the ticket:

https://trac.macports.org/ticket/62070

ryandesign:
Ok, mentioning that ticket URL in the commit message would be good.

So then my recommendation remains to make the port use `--disable-opencl` and `--disable-cuda` in the main `configure.args`, rename the variants to `cuda` and `opencl` and make them change the `--disable-*` flag to an `--enable-*` flag, and enable the variants with `default_variants` on systems where the build system would have chosen to enable that functionality. If you don't know which OS versions those are, you could enable them on all OS versions and then watch the buildbot for failures and make adjustments as needed. Alternately, if this functionality is not generally useful and other ports don't depend on it being present, you could leave them off by default on all systems.

i0ntempest:
Will merge after you've done what Ryan said.

mascguy:
> Ok, mentioning that ticket URL in the commit message would be good.
> 
> So then my recommendation remains to make the port use `--disable-opencl` and `--disable-cuda` in the main `configure.args`, rename the variants to `cuda` and `opencl` and make them change the `--disable-*` flag to an `--enable-*` flag, and enable the variants with `default_variants` on systems where the build system would have chosen to enable that functionality. If you don't know which OS versions those are, you could enable them on all OS versions and then watch the buildbot for failures and make adjustments as needed. Alternately, if this functionality is not generally useful and other ports don't depend on it being present, you could leave them off by default on all systems.

Ryan, sounds good. The safest bet is to leave the variants disabled by default.

Added port notes mentioning the change, along with the two new variants.

mascguy:
All changes made per discussions, and CI jobs have completed successfully. This PR is ready to merge.

i0ntempest:
Thank you, looks fine to me.

mascguy:
Great news, this change solved the problem for our Big Sur user in ticket: [62070 - mpif90-openmpi-gcc10 fails to link FORTRAN program](https://trac.macports.org/ticket/62070).

Thanks guys!

PR #10817, opened 2021-04-27
retroarch: new port

Tatsh:
#### Description

RetroArch expects to be able to write downloaded cores and other 'system' data into the app bundle and this is not a normal behaviour in any other macOS app especially now with app bundle integrity checks. Because of this and because MacPorts installs with files certain permissions, this version has a patch applied that sets default directories to be in `~/Library/Application Support/RetroArch`.

This will allow downloading cores on first launch without needing extra configuration, and without writing into `~/Documents` (a fallback path).

The alternative to downloading cores is to have ports for each core. This should be encouraged when cores are stable, but they do not have to be in this package or even reference it as a dependency.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.2.3 20D91 x86_64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/emulators/retroarch/Portfile
+++ b/emulators/retroarch/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           xcode 1.0
+
+github.setup        libretro RetroArch 1.9.1 v
+revision            0
+
+name                retroarch
+platforms           darwin
+license             GPL-3
+maintainers         {gmail.com:audvare @Tatsh} openmaintainer
+categories          emulators games
+description         Frontend for the libretro API.
+long_description    {*}${description}
+
+checksums           rmd160  2c4f65be5cb1be55da9f242af9721c5098fe01dc \
+                    sha256  e1fb48a0c9f8ce4ca31699c9611727e3b04b127d7c7eb2de171809279681ddf3 \
+                    size    39878651
+
+# See https://github.com/libretro/RetroArch/issues/8641
+patchfiles          patch-${name}-library-dirs.diff
+
+xcode.project       ${worksrcpath}/pkg/apple/RetroArch.xcodeproj
+xcode.target        RetroArch
+xcode.destroot.settings \
+    CODE_SIGN_IDENTITY=- \
+    CODE_SIGN_STYLE=Manual \
+    ENABLE_HARDENED_RUNTIME=NO \
+    GCC_TREAT_WARNINGS_AS_ERRORS=NO
+
+# Compile and install in one step, for now. See https://trac.macports.org/ticket/57137
+build {}
+
+if {${os.platform} eq "darwin" && ${os.major} < 10} {
+    known_fail yes
+    pre-fetch {
+        ui_error "${name} @${version} requires Mac OS X 10.6 or later."
+        return -code error "incompatible Mac OS X version"
+    }
+}
+
+notes {
+    On first run, go to "Online Updater" and choose each item beginning with "Update" (e.g. "Update Core Info Files").
+}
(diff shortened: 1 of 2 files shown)


emulators/retroarch/Portfile
```diff
@@ -0,0 +1,45 @@
...
+    GCC_TREAT_WARNINGS_AS_ERRORS=NO
+
+# Compile and install in one step, for now. See https://trac.macports.org/ticket/57137
+build {}
+
+if {${os.major} < 10} {
+    known_fail yes
+    pre-fetch {
+        ui_error "${name} @${version} requires OS X 10.6 or later."
+        return -code error "incompatible OS X version"
```
ryandesign, line 39: 10.7 and earlier are "Mac OS X" not "OS X".
```suggestion
        ui_error "${name} @${version} requires Mac OS X 10.6 or later."
        return -code error "incompatible Mac OS X version"
```

emulators/retroarch/Portfile
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+post-install {
+    ui_msg "On first run, go to 'Online Updater' and choose each item beginning with 'Update' (e.g. 'Update Core Info Files')."
+}
```
ryandesign, line 45: Any messages you want to display to the user after installation should usually go in `notes`.
```suggestion
notes {
    On first run, go to "Online Updater" and choose each item beginning with "Update" (e.g. "Update Core Info Files").
}
```

emulators/retroarch/files/patch-retroarch-library-dirs.diff
```diff
@@ -0,0 +1,174 @@
...
++   SETTING_ARRAY("log_dir", settings->paths.log_dir, true, "~/Library/Application Support/RetroArch/logs", true);
+ 
+    *size = count;
+ 
+@@ -3099,10 +3099,10 @@
+       size_t tmp = 0;
+       if (config_get_size_t(conf, size_settings[i].ident, &tmp))
+          *size_settings[i].ptr = tmp ;
+-      /* Special case for rewind_buffer_size - need to convert 
++      /* Special case for rewind_buffer_size - need to convert
```
ryandesign, line 131: This and all subsequent changes seem to be just whitespace changes and could be removed since they aren't relevant to the issue being fixed.

ryandesign (commented):
> this version has a patch applied that sets default directories to be in `~/Library/Application Support/RetroArch`.

That sounds like a much better place to put those files than into the app bundle itself. Is there an upstream bug report about this problem? Have you suggested this change to them?

Tatsh:
> > this version has a patch applied that sets default directories to be in `~/Library/Application Support/RetroArch`.
> 
> That sounds like a much better place to put those files than into the app bundle itself. Is there an upstream bug report about this problem? Have you suggested this change to them?

I found this issue: https://github.com/libretro/RetroArch/issues/8641

The user can fix this in the settings UI manually without this patch. Which may be their argument. Or that `~/Documents` can be used, and is indeed sometimes used instead (I do not fully know the behaviour).

I'm providing a little convenience for first time users of RetroArch. Plus, I really dislike RetroArch putting stuff in `~/Documents` even if it's in a subdirectory.

I know that with downloads of assets like the cheat files, if the storage path is not writeable, then the download succeeds but does not extract/install the content of the download, and fails silently. It does not tell the user to use a writeable path or how to fix it.

emulators/retroarch/Portfile
```diff
@@ -0,0 +1,46 @@
...
+github.setup        libretro RetroArch 1.9.1 v
+revision            0
+
+name                retroarch
+platforms           darwin
+license             GPL-3
+maintainers         {gmail.com:audvare @Tatsh} openmaintainer
+categories          emulators games
+description         Frontend for the libretro API.
+long_description    ${description}
```
cjones051073, line 16:
```suggestion
long_description    {*}${description}
```

emulators/retroarch/Portfile
```diff
@@ -0,0 +1,46 @@
...
+xcode.destroot.settings \
+    CODE_SIGN_IDENTITY=- \
+    CODE_SIGN_STYLE=Manual \
+    ENABLE_HARDENED_RUNTIME=NO \
+    GCC_TREAT_WARNINGS_AS_ERRORS=NO
+
+# Compile and install in one step, for now. See https://trac.macports.org/ticket/57137
+build {}
+
+if {${os.major} < 10} {
```
cjones051073, line 36:
```suggestion
if {${os.platform} eq “darwin” && ${os.major} < 10} {
```

cjones051073 (changes requested):

cjones051073 (approved):

PR #10828, opened 2021-04-28
raylib: Update to 3.7.0

judaew:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.3 20E232 x86_64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/multimedia/raylib/Portfile
+++ b/multimedia/raylib/Portfile
@@ -4,7 +4,8 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           cmake 1.1
 
-github.setup        raysan5 raylib 3.5.0
+github.setup        raysan5 raylib 3.7.0
+github.tarball_from archive
 revision            0
 
 homepage            https://www.raylib.com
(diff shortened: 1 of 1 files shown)


herbygillot:
Thanks @judaew.  I was planning on changing the tarball source to archive over time for many of these ports.

Can you make that change?
```
github.tarball_from archive
```

You'll need to recalculate the checksum as the tarball will be slightly different.
You may need to do a full clean so that MacPorts isn't reusing the old tarball:

```
$ sudo port clean -f --all raylib
```

judaew:
Done, thanks.

What is the meaning of the `archive` source for `github_tarball`? Is this the auto-generated `git archive` from the tag or?..

herbygillot:
> Done, thanks.
> 
> What is the meaning of the `archive` source for `github_tarball`? Is this the auto-generated `git archive` from the tag or?..

So for every Github repository, there are several different methods of downloading a branch or tag.

You can:
1. Download an asset from the repo's Releases section
2. You can download the branch or tag directly (using an "archive" link)
3. You can download the branch or tag using the "tarball" URL, which is the older method that Github intends to get rid of at some point


The Github portgroup is using method 3 by default.  You can see the various methods here: https://github.com/macports/macports-ports/blob/436249aefabd252950cc0537555b8985cca74ee7/_resources/port1.0/group/github-1.0.tcl#L26

So since Github wants to remove the tarball style URL, then over time we will want to move towards using either the `release` or `archive` style URLs.

PR #10832, opened 2021-04-28
py-ansible: Further performance tweak to patch

eborisch:
This further improves (building on 63bac4a947 from @tsabirgaliev) the time required to patch py*-ansible; from 200s to 14s on my 2019 MBP with Apex One anti-virus software. (I'm guessing the AV software may have something to do with it.)

This uses:
 * a single excecution of find
 * multiple (from find) executions of egrep as needed, but as few as possible (multiple files per execution)
 * a single execution of xargs
 * (via xargs) multiple executions of sed as needed, again with multiple files each time / as few executions as possible, and with both replace steps in each execution.

### DIFF
--- a/python/py-ansible/Portfile
+++ b/python/py-ansible/Portfile
@@ -37,12 +37,14 @@ python.versions 27 35 36 37 38 39
 
 if {${name} ne ${subport}} {
     patch {
-        foreach f [ exec find ${worksrcpath} -type f \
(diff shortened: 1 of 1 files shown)


eborisch:
Link to previous PR #10765.

judaew:
Thank you so much, you've made it easier to maintain dependent ports from `py-ansible`

PR #10852, opened 2021-04-30
openmpi: update to 4.1.1; add new variants heterogeneous and mpi1; enable use of published binaries for all compiler-specific subports

mascguy:
#### Description

* Update `openmpi` to 4.1.1. In addition to general bug fixes, upstream also fixed Big Sur ARM builds.
* Add new variants `heterogeneous` and `mpi1`
* Enable use of published binaries for all compiler-specific subports, to match mpich
* Add lib dependency for zlib, and corresponding configure flag
* Ensure target compilers are added to depends_lib

Fixes Included:
* [62795 - openmpi: declare lib dependency on target compiler](https://trac.macports.org/ticket/62795)
* [62788 - openmpi: opportunistically uses zlib; add formal dependency and configure flag](https://trac.macports.org/ticket/62788)
* [62783 - openmpi: validate that buildbot binaries are hardware-independent](https://trac.macports.org/ticket/62783)
* [62214 - openmpi-default +gccdevel fails to build on aarch64 with "error: address argument to atomic operation must be a pointer to integer or pointer"](https://trac.macports.org/ticket/62214)
* [62070 - mpif90-openmpi-gcc10 fails to link FORTRAN program](https://trac.macports.org/ticket/62070)
* [59547 - openmpi-gcc9 & enable-mpi1-compatibility](https://trac.macports.org/ticket/59547)

Reviewers/Maintainers:
* Given the number of subports that need to be built, there's little chance that the CI jobs will complete successfully. (Though I won't complain if they do!)
* The use of published binaries has been thoroughly researched, and tested. See ticket: [62783 - openmpi: validate that buildbot binaries are hardware-independent](https://trac.macports.org/ticket/62783)

###### Type(s)
- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

###### Verification 
Have you

### DIFF
--- a/science/openmpi/Portfile
+++ b/science/openmpi/Portfile
@@ -7,9 +7,23 @@ PortGroup           compilers 1.0
 PortGroup           muniversal 1.0
 PortGroup           legacysupport 1.0
 
+#===============================================================================
+#
+# *** IMPORTANT NOTE ***
+#
+# When making logic changes to this port, PLEASE review port 'mpich' to see
+# if the same changes should be applied. While the subports and variants aren't
+# exactly the same between the two - and things like configure arguments
+# certainly differ, as they're different code bases - much of the core logic
+# is very similar. (And often identical.)
+#
+# Please help us avoid divergent MPI ports, which cause serious migraines.
+#
+#===============================================================================
+
 name                openmpi
-version             4.1.0
-revision            1
+version             4.1.1
+revision            0
 set branch          [join [lrange [split ${version} .] 0 1] .]
 categories          science parallel net
 platforms           darwin
@@ -29,9 +43,9 @@ homepage            https://www.open-mpi.org/
 set subdir          ompi/v${branch}/downloads/
 master_sites        https://www.open-mpi.org/software/${subdir}
 
-checksums           rmd160  15e3b5429e74f0f39a46ca0cc35de92b68ea21d0 \
-                    sha256  73866fb77090819b6a8c85cb8539638d37d6877455825b74e289d647a39fd5b5 \
-                    size    10022171
+checksums           rmd160  60fff94c4c7a4bfaaa11285f35d76265516ed8b0 \
+                    sha256  e24f7a778bd11a71ad0c14587a7f5b00e68a71aa5623e2157bafee3d44c07cda \
+                    size    10052770
 
 use_bzip2           yes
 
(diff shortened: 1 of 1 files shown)


science/openmpi/Portfile
```diff
@@ -185,9 +185,11 @@ if {${cname} in ${clist_unsupported}} {
         set cname                   mp
     }
 
-    # Force buildbot to skip this since openmpi depends on hwloc and that would
-    # be different for a user's machine
-    archive_sites
+    if {[lsearch -exact {mp llvm clang} ${cname}] != -1} {
```
mascguy, line 188: Reviewers: Note that we don't presently have an openmpi subport named `openmpi-llvm`. However, to be consistent with mpich, I'm using the same logic when determining which buildbot binaries are not to be used.

cjones051073: If this line need to be kept in sync with that in another port please add a clear comment here *and* in that other port detailing this.

mascguy: Chris, technically the line doesn't have to be kept in-sync with mpich. I'm simply using the same logic for consistency.

I'm willing to add a comment though, as part of a separate commit. But I'd prefer to wait until I'm granted commit access.

Can we merge this as-is for now?

mascguy:
Of note, this OpenMPI update also fixes builds for Big Sur ARM.

mascguy:
Added fix for [62788 - openmpi: opportunistically uses zlib; add formal dependency and configure flag](https://trac.macports.org/ticket/62788)

Otherwise, this PR is ready to merge.

mascguy:
Two out of three CI jobs completed successfully, which is better than expected.

Can this be merged?

science/openmpi/Portfile
mascguy: Chris, I didn't do a very good job of explaining this before, so let me elaborate further.

In short, there is a _significant_ amount of logic that is duplicated between ports `openmpi` and `mpich`. Some of it was refactored to the `mpi` portgroup long before my arrival, but there's still plenty of duplication between the two. And there are just enough differences - such as the target compilers supported - making logic sharing more difficult.

Nonetheless, there is a long-term opportunity to share more logic, and simplify both ports, via the `mpi` portgroup. I'll likely take a look at some point during the coming months, unless someone beats me to it. And I'd certainly welcome anyone who wishes to collaborate on it.

Regardless, here's the point: If we start adding comments to logic duplicated between the two, that would relate to far more than just this block. And we'd be littering the two ports with comments, which simply isn't practical.

Instead, it makes more sense to point out the fact that the two share quite a bit of logic, and emphasize the need to keep both in-sync where it makes sense. And indeed, my latest commit includes just such a comment block, on line 10 above.

I'll be adding an identical comment block to the top of 'mpich' as part of a separate PR, as some of fixes here now need to be applied there too. But given that the CI jobs have a great chance of timeout for one of these ports alone, including both in one PR is pointless.

Does all of that make sense?

mascguy:
Added fix for [62795 - openmpi: declare lib dependency on target compiler](https://trac.macports.org/ticket/62795)

mascguy:
@cjones051073/anyone, please review, and let me know if you have specific concerns. It would be great to merge this within the next day or so, for the Big Sur ARM fixes alone.

science/openmpi/Portfile
mascguy: PR for `mpich`:

[10879 - mpich: ensure target compilers are added to depends_lib](https://github.com/macports/macports-ports/pull/10879)

mascguy: Proposed new portgroup, allowing us to share logic between `openmpi` and `mpich`, tracked by ticket:

[62807 - openmpi/mpich: refactor duplicated logic, and simplify ports, via new portgroup](https://trac.macports.org/ticket/62807)

PR #10854, opened 2021-04-30
openjdk: update to latest versions

breun:
#### Description

Update to the latest versions of AdoptOpenJDK OpenJ9 (8u292, 11.0.11, 16.0.1) and Azul Zulu (16.0.1).

###### Tested on

macOS 11.3 20E232 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/java/openjdk/Portfile
+++ b/java/openjdk/Portfile
@@ -13,16 +13,6 @@ supported_archs  x86_64
 # Latest Long Term Support (LTS) major version
 version          11
 
-if {${subport} eq "openjdk"} {
-    # The openjdk port is not installable, but recommends users to install the latest Long Term Support (LTS) subport
-    PortGroup    obsolete 1.0
-    replaced_by  openjdk${version}
-}
-
-# Dummy default values for the obsoleted openjdk port
-set build        0
-set openj9_version 0
-
 set long_description_adoptopenjdk_intro \
    "OpenJDK build provided by AdoptOpenJDK, built from a fully open-source set of build scripts and infrastructure."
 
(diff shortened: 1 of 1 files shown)


java/openjdk/Portfile
```diff
@@ -441,6 +415,12 @@ if {${subport} in [list openjdk12 openjdk13 openjdk14 openjdk15]} {
     PortGroup    obsolete 1.0
     # Can replace with openjdk17 when it is added and openjdk16 is obsoleted
     replaced_by  openjdk16
+} elseif {${subport} eq "openjdk8-openj9-large-heap"} {
+    PortGroup    obsolete 1.0
+    replaced_by  openjdk8-openj9
+} elseif {${subport} eq "openjdk11-openj9-large-heap"} {
+    PortGroup    obsolete 1.0
+    replaced_by  openjdk11-openj9
```
cjones051073, line 423: Wouldn’t it be better to include these above when the subport is first declared ? I am not sure I see the advantage to separating them out down here ? I also wonder if doing so wouldn’t help with some of the nitpick warnings?

either way, can you also please leave an empty line after each portgroup line.

cjones051073 (changes requested):

java/openjdk/Portfile
breun: I've refactored the Portfile to only have a single declaration of `PortGroup obsolete 1.0` and I've inlined all the `replaced_by` declarations.

`port lint` reports 0 errors, and `port lint --nitpick` only warns about lines that 'seem to hardcode the version number', but we want a different version per 'subport' in this case.

cjones051073 (approved):

PR #10895, opened 2021-05-05
gcc11: new port

wyuenho:
#### Description

New port: gcc11, mostly copied from gcc10 and gcc-devel, also update libgcc to 4.0 and reference libgcc11 on macOS >= 11.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.3.1
Xcode 12.5

###### Verification 
Have you

### DIFF
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -1,0 +1,379 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem                                       1.0
+PortGroup           select                       1.0
+PortGroup           compiler_blacklist_versions  1.0
+PortGroup           active_variants              1.1
+PortGroup           conflicts_build              1.0
+PortGroup           cltversion                   1.0
+
+epoch               6
+name                gcc11
+
+homepage            https://gcc.gnu.org/
+
+platforms           darwin
+categories          lang
+maintainers         nomaintainer
+# an exception in the license allows dependents to not be GPL
+license             {GPL-3+ Permissive}
+description         The GNU compiler collection
+long_description    The GNU compiler collection, including front ends for \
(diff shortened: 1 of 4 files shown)


cjones051073:
Thanks for this suggestion.

I actually already have a branch for gcc11 ready to go at

https://github.com/cjones051073/macports-ports/tree/gcc11

There are a number of other changes you need to make, which you would not be aware of that makes adding a new GCC a bit more involved.  The bigger issue though is there is no gcc11 ARM support at this time, and that will need addressing before we can update as currently gcc10 which currently provides the primary libgcc runtime works on both macOS11 Intel and Arm. I am keeping my eye out for a update for this and will push my branch when it is ready. So I suggest closing this for now..

wyuenho:
Ok. BTW, is there a way to separate libgccjit into its own port, and add a variant to gcc1[01] to depend on it? This is all I care about really so I can change emacs-devel's nativecompile variant to just depend on libgccjit.

cjones051073:
@wyuenho please see https://github.com/macports/macports-ports/commit/8759dbd93fb71e89a59cd58be4ca2d854db24535

Home-brew produced an arm back port to gcc11 which the above uses.

cjones051073:
Has a patch been submitted upstream for this ? Presumably its a bug it doesn't match the macOS convention of using .dylib ?

wyuenho:
Doesn't look like it's been reported. The patch is just a reworked patch from the gcc10 port, I presumed whoever made that patch reported it, but it seems I was wrong.

wyuenho:
It now requires an account on gcc's bugzilla to report a bug and I have to email somebody to request one. It's annoying, do you have a gcc account to report this?

cjones051073:
sorry no I don't, for similar reasons I guess...

jwakely:
> It's annoying

Yes, sorry. But it's not as annoying or difficult as dealing with thousands and thousands of spam bugs per day, which is why we had to restrict account creation to be a manual process.

wyuenho:
Bug submitted: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100613

PR #10896, opened 2021-05-05
igraph: update to 0.9.3

szhorvat:
* update to 0.9.3
 * add openmp variant (new feature in 0.9.3)

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G9028 x86_64
Xcode 11.3.1 11C504

###### Verification 
Have you

### DIFF
--- a/math/igraph/Portfile
+++ b/math/igraph/Portfile
@@ -5,8 +5,8 @@ PortGroup           cmake 1.1
 PortGroup           github 1.0
 PortGroup           linear_algebra 1.0
 
-github.setup        igraph igraph 0.9.2
-revision            1
+github.setup        igraph igraph 0.9.3
+revision            0
 github.tarball_from releases
 
 categories          math science devel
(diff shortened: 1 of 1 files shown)


szhorvat:
Getting OpenMP to work took a lot of time and trouble because of this problem:

https://github.com/macports/macports-base/commit/7c91604891fa0d071b8d598490c4dc2edb8e0031

It is still unclear to me what the proper workaround is. For now, I simply copied the solution from the `darktable` Portfile.

In short, MacPorts sets `CC_PRINT_OPTION`, and that somehow causes CMake not to link to the OpenMP library correctly. Please document this clearly somewhere, including the preferred workaround. For example here: https://trac.macports.org/wiki/CompilerSelection

CC @MarcusCalhoun-Lopez

szhorvat:
One of the problems with the workaround I used is that it will actually add `-fopenmp` to the compilation flags unconditionally. A proper solution would allow igraph to add this flag only where it is needed instead of forcing it everywhere.

Is there any way to forcibly remove environment variables? I tried `configure.env-remove CC_PRINT_OPTIONS`, but there seems to be no `configure.env-remove`.

szhorvat:
Can someone explain why a simple

```
compiler.log_verbose_output no
```

does not work?

szhorvat:
@nerdling I finally understood what is happening. `compiler.log_verbose_output` will only be available in MacPorts 2.7. Until then it is fine to use `-fopenmp` with all source files. This PR is good to go as it is now.

kencu:
We had previously had lots of trouble with trying to enable openmp generally by default, eg:

https://trac.macports.org/ticket/57009

And there was moderate discussion and a sort-of plan that we would just avoid it until openmp was supported by Xcode clangs, and then all the various bugs would shake out it, and we would come up with a proper plan for it.

Although I suspect it works in various carefully-checked situations.

PR #10909, opened 2021-05-06
pyinfra: new port

herbygillot:
#### Description

New port for the **[pyinfra](https://pyinfra.com/)** configuration management tool.

###### Tested on

macOS 10.15.7 19H1030 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/sysutils/pyinfra/Portfile
+++ b/sysutils/pyinfra/Portfile
@@ -1,0 +1,50 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           python 1.0
+
+name                pyinfra
+version             1.4
+revision            0
+
+homepage            https://pyinfra.com
+
+description         pyinfra automates infrastructure super fast at massive \
+                    scale.
+
+long_description    {*}${description}  It can be used for ad-hoc command \
+                    execution, service deployment, configuration management \
+                    and more. Core design features include super fast \
+                    execution over thousands of hosts with predictable \
+                    performance, agentless execution against \
+                    SSH/Docker/subprocess/winrm hosts, extendable with any \
+                    Python package as configured & written in standard \
+                    Python, and integrated with Docker, Vagrant/Mech & \
+                    Ansible out of the box.
+
+platforms           darwin
+license             MIT
+categories          sysutils python
+maintainers         {gmail.com:herby.gillot @herbygillot} \
+                    openmaintainer
+supported_archs     noarch
+
+checksums           rmd160  3c428307414bc86d06126ddb7aeb5fbbba007cdd \
+                    sha256  3cad5ca81b4d842091f39a6e24ea8117e9c2fcb1a4abfcfb254bf1617d87dca0 \
+                    size    166876
+
+python.default_version      39
+
+depends_lib-append  port:py${python.version}-setuptools
+
+depends_run-append  port:py${python.version}-click          \
+                    port:py${python.version}-colorama       \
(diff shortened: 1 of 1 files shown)


sysutils/pyinfra/Portfile
```diff
@@ -0,0 +1,52 @@
...
+
+checksums           rmd160  3c428307414bc86d06126ddb7aeb5fbbba007cdd \
+                    sha256  3cad5ca81b4d842091f39a6e24ea8117e9c2fcb1a4abfcfb254bf1617d87dca0 \
+                    size    166876
+
+python.default_version \
+                    39
+
+depends_build-append \
+                    port:py${python.version}-setuptools
```
harens, line 40: pyinfra contains [entry_points](https://github.com/Fizzadar/pyinfra/blob/92cd818ce72a6a1bec030932e2f11f3e9732bb77/setup.py#L107-L110), which means setuptools will be required at both build and runtime (hence it should be a lib dependency). See https://github.com/macports/macports-ports/pull/6314#issuecomment-585490282

```
entry_points={
            'console_scripts': (
                'pyinfra=pyinfra_cli.__main__:execute_pyinfra',
            ),
```

sysutils/pyinfra/Portfile
```diff
@@ -0,0 +1,52 @@
...
+
+depends_lib-append  port:py${python.version}-click          \
+                    port:py${python.version}-colorama       \
```
harens, line 43: This is a Windows-only dependency in this scenario.

```
'colorama<1',  # Windows color support for click
```

Just to note, the [py-click port](https://github.com/macports/macports-ports/blob/d275a24edbefe7f361f2aa6eb8f59648092058c0/python/py-click/Portfile) doesn't list colorama as a dependency for macOS, whilst it's always required on Windows (see https://github.com/pallets/click/commit/d7fac6e16099005c642f4dcb71d84a091a787073). It also shouldn't be required here.

sysutils/pyinfra/Portfile
```diff
@@ -0,0 +1,52 @@
...
+                    sha256  3cad5ca81b4d842091f39a6e24ea8117e9c2fcb1a4abfcfb254bf1617d87dca0 \
+                    size    166876
+
+python.default_version \
+                    39
+
+depends_build-append \
+                    port:py${python.version}-setuptools
+
+depends_lib-append  port:py${python.version}-click          \
```
harens, line 42: I've always wondered why by convention we list `install_requires` dependencies as library deps even though they're not required at build time. I normally leave them as runtime deps in my ports, which is what the [Gentoo docs](https://wiki.gentoo.org/wiki/Project:Python/Dependencies#Finding_dependencies) also suggests (unless there are tests present).

Out of curiosity, is there any particular reason why we do this? cc @reneeotten.

harens (commented):
Thanks as always @herbygillot, this looks like an interesting port 👀.

A few points from me below.

sysutils/pyinfra/Portfile
herbygillot: Good point; change made

sysutils/pyinfra/Portfile
herbygillot: This is true - changed to `depends_run`

sysutils/pyinfra/Portfile
herbygillot: Since `colorama` appears to be an unequivocal install requirement for `pyinfra` according to its `setup.py`, I'll keep this as a dependency for now.

harens: That's a fair point, since they've only noted it as a comment.

harens (approved):
LGTM. Thanks @herbygillot.

herbygillot:
> LGTM. Thanks @herbygillot.

Thank you @harens

PR #10933, opened 2021-05-08
libsndfile: update to 1.0.31

kode54:
* Moved to new home page on GitHub
    * Move build system to CMake
    * Try to replace original features

#### Description

Update libsndfile to its new home on GitHub, and update to the latest release.
This also attempts to retain some of the original feature parity, but I cannot
guarantee at this point that it still builds on 10.4, like one of the past patches
supported.

Unlike the current release, this one actually supports enabling libogg,
libvorbis, libFLAC, and the new libopus and speex support. Note, that at
this time, both libopus and speex will enable if detected, so I don't know
what to do about that, other than trying to force the CMake script to
disable those unless specifically requested.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.3.1 20E241 arm64
Xcode 12.5 12E262

###### Verification 
Have you


audio/libsndfile/Portfile
```diff
@@ -18,36 +19,30 @@ long_description    \
 	a simple, elegant and easy to use API, \
 	and on the fly format conversion.
 license             LGPL-2.1+
-homepage            http://www.mega-nerd.com/libsndfile/
-master_sites        ${homepage}files/
-
+homepage            https://libsndfile.github.io/libsndfile/
```
cjones051073, line 22: This line shouldn’t be required as the github PG should set this correctly.

cjones051073 (changes requested):

audio/libsndfile/Portfile
kode54: I'll get rid of that, then.

szhorvat: @cjones051073 Would the github PG set it to https://github.com/libsndfile/libsndfile which is just the repo? This was the actual homepage, not the repo.

cjones051073 (approved):

PR #10954, opened 2021-05-09
saml2aws: New port

jamesog:
#### Description

saml2aws is a CLI tool which enables you to login and retrieve AWS temporary credentials using a SAML IDP.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.3 20E232 x86_64

###### Verification 
Have you

### DIFF
--- a/security/saml2aws/Portfile
+++ b/security/saml2aws/Portfile
@@ -1,0 +1,318 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/Versent/saml2aws 2.30.0 v
+revision            0
+categories          security
+platforms           darwin
+license             MIT
+maintainers         {netinertia.co.uk:jamesog @jamesog} \
+                    openmaintainer
+
+description         Log in to AWS using a SAML IDP
+long_description    CLI tool which enables you to login and retrieve AWS \
+                    temporary credentials using a SAML IDP.
+
+homepage            https://github.com/Versent/saml2aws
+
+checksums           ${distname}${extract.suffix} \
+                        rmd160  0cdd193229f49d1a0585f6fb71442baf41a9f7a3 \
+                        sha256  34de318eb3a725a392433d19da3991e30f67304b8db6340d061795661f703d76 \
+                        size    214434
+
+go.vendors          gopkg.in/yaml.v3 \
+                        lock    496545a6307b \
+                        rmd160  16a43936d8ae6243895e23465132977d3a1193c2 \
+                        sha256  333e78b3b9cb73b3572d62f692d32426a8554b86c93025ea032f779395869e84 \
+                        size    90145 \
+                    gopkg.in/yaml.v2 \
+                        lock    v2.4.0 \
+                        rmd160  66e9feb7944b3804efa63155ed9b618717b8955c \
+                        sha256  72812077e7f20278003de6ab0d85053d89131d64c443f39115a022114fd032b6 \
+                        size    73231 \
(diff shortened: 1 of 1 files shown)


herbygillot:
Although we have plenty of ports that distribute pre-built binaries just like this current and your previous submissions are doing, MacPorts' preference is for ports to build their software from source.

For Go software specifically, a Portfile can:

1. allow Go to pull dependencies on its own and build as it normally does...
_or_
2. ...explicitly specify each dependency in the Portfile so that it's MacPorts that fetches everything, and Go needs only build with all the source and dependencies already in place

MacPorts' philosophy prefers method 2.  When each required Go module is specified in the Portfile, MacPorts can mirror each module, allowing MacPorts to build the software even if the upstream repos disappear for both the target project iself, or its dependencies.

Examples of method 1:
- https://github.com/macports/macports-ports/blob/master/devel/argocd/Portfile (calls `make` [via `build.cmd`], which uses the project's Makefile to build it from source)
- https://github.com/macports/macports-ports/blob/master/devel/grpcurl/Portfile

Examples of method 2 (note the dependencies specified in the `go.vendors` section):
- https://github.com/macports/macports-ports/blob/master/security/cfssl/Portfile
- https://github.com/macports/macports-ports/blob/master/devel/delve/Portfile

The only issue is that it can be a hassle putting together the `go.vendors` list, but the [`go2port`](https://ports.macports.org/port/go2port/summary) tool can help a lot here.

jamesog:
Thanks for the feedback @herbygillot, I really appreciate it. I'd taken the example of the [`gh`](https://github.com/macports/macports-ports/blob/master/devel/gh/Portfile) port when I did the first port submission.

I'll work on converting this and the flyctl port over to source builds. Thanks for providing the examples.

herbygillot:
Much appreciated.  You can identify most Go ports by any port using the `golang` portgroup.  If you have `ripgrep` installed, you can just run `rg 'PortGroup\s+golang' -l` at the root of your macports-ports source tree.

security/saml2aws/Portfile
```diff
@@ -0,0 +1,319 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/Versent/saml2aws 2.30.0 v
+revision            0
+categories          security
+platforms           darwin
+supported_archs     arm64 x86_64
```
herbygillot, line 10: You can skip `supported_archs` - the `golang` portgroup should do the right thing per architecture.

herbygillot (commented):
Looking good @jamesog.  Just one comment, but other than that, this looks good to go.

security/saml2aws/Portfile
jamesog: Thanks, I've removed this.

jamesog:
One thing I noticed since switching to the `go.vendors` method is `port lint` complains a lot now - is that down to the way `go.vendors` works compared to the usual `checksums`?

herbygillot:
> One thing I noticed since switching to the `go.vendors` method is `port lint` complains a lot now - is that down to the way `go.vendors` works compared to the usual `checksums`?

That's an unfortunate known issue - certain portgroup specific facilities, like `go.vendors` or `cargo.crates` (for the Rust `cargo` portgroup) are not yet understood by `port lint`.  The portgroups are like "macros" that expose to port maintainers language or framework -specific options that help when defining ports for particular languages or frameworks, but then underneath the portgroup, Portfile-native options are may be getting set in a non-standard way. As a result, the final Portfile options that `port lint` sees will not look right (to it).

herbygillot (approved):

jamesog:
That makes a lot of sense. Thanks so much for your help, Herby!

herbygillot:
No worries, thanks for your contribution 👍🏾

PR #10958, opened 2021-05-09
sqlitebrowser: make qttools a build dependency

szhorvat:
#### Description

This makes qttools a build dependency instead of a library dependency, so this port wouldn't force the installation of all qt dev tools.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G9028 x86_64
Xcode 11.3.1 11C504

###### Verification 
Have you

### DIFF
--- a/databases/sqlitebrowser/Portfile
+++ b/databases/sqlitebrowser/Portfile
@@ -70,7 +70,9 @@ platform darwin {
     use_parallel_build  no
     depends_lib-append  port:sqlite3
     qt5.depends_component \
-                        qtmacextras qttools
+                        qtmacextras
+    qt5.depends_build_component \
+                        qttools
 
     destroot {
             copy "${worksrcpath}/src/DB Browser for SQLite.app" ${destroot}${applications_dir}


szhorvat:
The `linux` section could have the same improvement, but I did not touch it as I cannot test it, and I didn't understand why it even included `qtmacextras`.

@i0ntempest It would be nice if you could resolve the license conflicts so as to make binaries distributable for this port. It's the same thing that affects every single Qt-based port: the `OpenSSLException` blocks all GPL'd dependents. This port takes an unpleasantly long time to build, probably because of `use_parallel_build  no`.

i0ntempest:
Thanks. The app executable links against `QtMacExtras` framework, you can see by running `otool -L` on it.

szhorvat:
I was referring to the _Linux_ section, which also includes qtmacextras

https://github.com/macports/macports-ports/blob/master/databases/sqlitebrowser/Portfile#L93

I cannot test this on Linux, so I did not apply the same improvement to that part of the portfile.

i0ntempest:
Oh okay, the Linux block actually was used on macOS until I thought cmake sucks (for this particular project and on macOS only) and switched it to use qmake, and moved the original code inside Linux block. Most likely it won't hurt to remove `qtmacextras` (also the `qttools` change) but I also can't test it.

PR #10985, opened 2021-05-11
exim: attempt to fix build on 10.11 and earlier

chrstphrchvz:
#### Description
See build failures: https://ports.macports.org/port/exim/builds

###### ~~Tested on~~

Untested.

###### Verification 
Have you

### DIFF
--- a/mail/exim/Portfile
+++ b/mail/exim/Portfile
@@ -2,6 +2,11 @@
 
 PortSystem              1.0
 
+# clock_gettime(), openat()
+PortGroup               legacysupport 1.1
+
+legacysupport.newest_darwin_requires_legacy 15
+
 name                    exim
 version                 4.94.2
 revision                0


mail/exim/Portfile
```diff
@@ -2,6 +2,10 @@
 
 PortSystem              1.0
 
+# clock_gettime(), openat()
+PortGroup               legacysupport 1.1
+legacysupport.newest_darwin_requires_legacy 15
```
cjones051073, line 7: Please add an empty line after the port group to silence the lint warnings

cjones051073 (changes requested):

cjones051073 (approved):

chrstphrchvz:
Fails because the `-lMacportsLegacySupport` flag isn’t passed to the linker.

I notice that the `clock_gettime()` usage seems optional, so maybe the check being used (`#ifdef _POSIX_MONOTONIC_CLOCK`) does not work properly.

cjones051073:
no, the check is fine as it just tests for the declaration in the headers, which will be found. The issue is the build system is not properly passing the linker flags, and if you do one and not the other you have problems...

chrstphrchvz:
> no, the check is fine as it just tests for the declaration in the headers, which will be found.

I meant without legacy support—that the check is not useful on macOS because its headers define `_POSIX_MONOTONIC_CLOCK` even for macOS 10.11 and earlier which lack `clock_gettime`, i.e. an upstream bug (if they intend to support older macOS).

chrstphrchvz:
> i.e. an upstream bug (if they intend to support older macOS).

…unless one argues this is a macOS bug (POSIX noncompliance)

PR #10992, opened 2021-05-12
llvm-11: use system python for set-xcode-analyzer

szhorvat:
* Fixes: https://trac.macports.org/ticket/62308

#### Description

This fixes https://trac.macports.org/ticket/62308. I tested on macOS 10.14.6 by manually reverting the `set-xcode-analyzer` back to use `/usr/bin/python` (I did not use the modified portfile to rebuild the port). I use Xcode 11.3.1 and I confirm that the `set-xcode-analyzer` script works correctly after this script, and that my Xcode version can successfully use MacPorts's clang-11 for static analysis.

The same fix can probably be applied to other versions of the Clang port as well, but I only looked at version 11. I would rather leave this change to the maintainers, or at least wait until this PR is approved before putting in effort into additional changes.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/lang/llvm-11/Portfile
+++ b/lang/llvm-11/Portfile
@@ -25,7 +25,7 @@ checksums               rmd160  f566b4b75c8f30418f19069a9a84864ead766401 \
 
 name                    llvm-${llvm_version}
 revision                0
-subport                 clang-${llvm_version} { revision 2 }
+subport                 clang-${llvm_version} { revision 3 }
 subport                 flang-${llvm_version} { revision 0 }
 subport                 lldb-${llvm_version} { revision 0 }
 set suffix              mp-${llvm_version}
@@ -523,9 +523,6 @@ if {${subport} eq "llvm-${llvm_version}"} {
             # pythonfullpath is set above, depending on presence of system python2.7
             reinplace "s|/usr/bin/env python|${pythonfullpath}|g" \
                  ${worksrcpath}/tools/clang/tools/scan-view/bin/scan-view
-
-            reinplace "s|/usr/bin/python|${pythonfullpath}|g" \
-                ${worksrcpath}/tools/clang/tools/scan-build/bin/set-xcode-analyzer
         }
     }
 


kencu:
scan view is still fine being set with the python3 in pythonpath? 

I am not sure if all darwin systems have /usr/bin/python set to what we need. I'll have to check them all. Certainly on some systems, /usr/bin/python is too old to use. Perhaps we just have to completely disable the analyzer below some cutoff.

And if /usr/bin/python is 2.7 as it may or may not be, then I'm fairly sure that llvm needs a python 3 to build, so we'll have to make sure that we have that set.

Let me look into this in more detail. In the meantime, you have fixed your person version it appears, and TBH you are the only one who has brought this up in all these years, so you're OK.

I don't want to revbump the entire LLVM ecosystem for just this -- too minor to make 100,000 machines work for hours for this one little change. I may pair it with some other update someday.

Because PRs have a habit of bring committed I will close this to prevent that, but thanks for your thoughts and noticing something was funny about the analyzer.

szhorvat:
> scan view is still fine being set with the python3 in pythonpath?

I used it as it is currently (i.e. with MacPorts Python 3.9), and it worked perfectly fine.

> you have fixed your person version it appears

Actually, I edited the `set-xcode-analyzer` script manually and reverted it to `/usr/bin/python`. I would have to do this after every clang update (and I'd have to run `set-xcode-analyzer` after every Xcode update). It would be nice to make this work out of the box if possible.

> I don't want to revbump the entire LLVM ecosystem for just this

Perhaps it's possible to make the fix without the revbump? After all, personally I'm set until the next clang update anyway.

> Because PRs have a habit of bring committed I will close this to prevent that, but thanks for your thoughts and noticing something was funny about the analyzer.

Certainly, that makes sense. Thanks for looking into this!

PR #10993, opened 2021-05-12
mpich: Include gcc11 cases as needed.

eborisch:
Fixes https://trac.macports.org/ticket/62866

Maintainer update.

### DIFF
--- a/_resources/port1.0/group/compilers-1.0.tcl
+++ b/_resources/port1.0/group/compilers-1.0.tcl
@@ -785,7 +785,7 @@ proc compilers::add_fortran_legacy_support {} {
         } else {
             set fortran_compiler    [fortran_variant_name]
         }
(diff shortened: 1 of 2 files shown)


mascguy (approved):

mascguy:
I suppose we'll need to do this for `openmpi` too...

eborisch:
> I suppose we'll need to do this for `openmpi` too...

I don't see any switches in it currently, so perhaps it is already handled?

mascguy:
> > I suppose we'll need to do this for `openmpi` too...
> 
> I don't see any (similar) switches in it currently, so perhaps it is already handled?

Sorry for the confusion, I was thinking in terms of the addition of a gcc11 subport. But that's not what this PR is about, so never mind...  :-)

PR #10996, opened 2021-05-13
llvm/clang/lldb-12: new port

kencu:
This is a WIP of a moderate refactoring and simplification of the llvm / clang / lldb installation.

All patches and compatibility code for older systems currently disabled and will be reviewed and used as needed.

Most variants currently removed. Will add back the needed variants prior to committing.


cjones051073:
Thanks @kencu . If you want pull requests to not be merged, the GitHub language for this is to mark them as draft. I’ve just done that here.

kencu:
yes indeedy, Even after all these years, I was unaware that github had chosen that particular label for special treatment. Thanks. (Somebody should write up a cheat sheet for github -- oh wait, I did :> I guess I better add that bit.)

kencu:
as far as I can tell, these builds succeeded.

cjones051073:
Closed ?

kencu:
I have the patches sorted out, so ready to work on this again.

lang/llvm-12/Portfile
```diff
@@ -0,0 +1,306 @@
...
+        -DCLANG_INCLUDE_TESTS=OFF          \
+        -DCLANG_ENABLE_STATIC_ANALYZER=OFF \
+        -DCLANG_ENABLE_ARCMT=OFF           \
+        -DDARWIN_PREFER_PUBLIC_SDK=ON      \
+        -DLLVM_BUILD_RUNTIME=ON            \
+        -DLIBCXX_ENABLE_SHARED=ON          \
+        -DLIBCXX_INSTALL_LIBRARY=ON
+
+    # clang project requires libcxx project
+    # else the built 1st stage compiler tries to link against libstdc++ and finds an old version
```
kencu, line 159: this is relevant only when using BOOTSTRAP clang, of course, which I want to do next

lang/llvm-12/Portfile
```diff
@@ -0,0 +1,306 @@
...
+    # flang project requires mlir project
+    configure.args-append   \
+        -DLLVM_ENABLE_PROJECTS="clang\;clang-tools-extra\;compiler-rt\;libcxx\;libcxxabi\;flang\;mlir\;lld\;openmp\;libunwind"
```
kencu, line 162: I'll take openmp out of here, as we use an external build of openmp and hack the clang code to find it instead of this one anyway

lang/llvm-12/Portfile
```diff
@@ -0,0 +1,306 @@
...
+variant tests description {Enable building of test code} {
+    depends_test-append     port:py39-psutil
+    test.run                yes
+    configure.args-append   -DLLVM_INCLUDE_TESTS=ON
+    test.target             check
+
+# this should work for lldb, but doesn't as gmock is not being found
+#   if {${subport} eq "lldb-${llvm_version}"} {
+#       test.target    check-lldb
+#   }
```
kencu, line 204: I'm not 100% sure how I'm going to fix this lldb testing issue. We use llvmtree/lldb as the source root only for the lldb build because lldb has no good unique install targets and otherwise we would have to build all of llvm/clang/compiler-rt to build lldb. I guess perhaps that is just what I will do, even though it's a bit time consuming to do it that way. Probably < 5 users even use the lldb port on MacPorts, I would imagine...I'd be curious to know --- hah! -- it's only 4 (probably all me :> ) <https://ports.macports.org/port/lldb-11/stats?days=30&days_ago=0>

lang/llvm-12/Portfile
```diff
@@ -0,0 +1,424 @@
...
+
+if {${os.platform} eq "darwin" && ${os.major} < 14} {
+    patchfiles-append \
+        0016-10.10-and-less-compiler-rt-find-macosx-sdk.patch \
+}
+
+if {${os.platform} eq "darwin" && ${os.major} < 11} {
+    patchfiles-append \
+        0017-10.6-and-less-libcxx-work-around-no-reexported_symbo.patch \
+        0018-10.6-and-less-clang-default-to-libc-on-all-darwin-sy.patch \
```
kencu, line 108: whoopsy -- this one should be at least 10.8 and less, as we need to catch the 10.7 and 10.8 systems too.

lang/llvm-12/files/0005-clang-support-macports-libstdcxx.patch
```diff
@@ -0,0 +1,202 @@
...
+@@ -643,6 +643,8 @@ HexagonToolChain::GetCXXStdlibType(const ArgList &Args) const {
+       return ToolChain::CST_Libstdcxx;
+   }
+   StringRef Value = A->getValue();
++  if (Value == "macports-libstdc++")
++    return ToolChain::CST_MacPortsLibstdcxx;
+   if (Value != "libstdc++" && Value != "libc++")
+     getDriver().Diag(diag::err_drv_invalid_stdlib_name) << A->getAsString(Args);
+ 
+diff --git a/clang/lib/Frontend/CompilerInvocation.cpp.rej b/clang/lib/Frontend/CompilerInvocation.cpp.rej
```
kencu, line 160: yoops. this bit snuck in and is not needed. Causes no harm, but does no good. I'll get rid of this.

PR #11096, opened 2021-05-21
Mysql8 is GA, update socket paths

haraldgroven:
#### Description

Mysql 51, 55, 56 no longer officially supported. Php should inform users on path to sockets for newer versions (i.e. mysql 57 and 80)

###### Type(s)

- [ ] enhancement

### DIFF
--- a/lang/php/Portfile
+++ b/lang/php/Portfile
@@ -1343,9 +1343,8 @@ mysql.default_socket, mysqli.default_socket and pdo_mysql.default_socket\
 to the path to your MySQL server's socket file.
 
 For mysql5, use ${prefix}/var/run/mysql5/mysqld.sock
-For mysql51, use ${prefix}/var/run/mysql51/mysqld.sock
-For mysql55, use ${prefix}/var/run/mysql55/mysqld.sock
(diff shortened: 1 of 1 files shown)


lang/php/Portfile
```diff
@@ -1343,9 +1343,8 @@ mysql.default_socket, mysqli.default_socket and pdo_mysql.default_socket\
 to the path to your MySQL server's socket file.
 
 For mysql5, use ${prefix}/var/run/mysql5/mysqld.sock
```
chrstphrchvz, line 1345: The mention of `mysql5` can likely be removed as well: `mysql5` and `mysql51` are both for MySQL 5.1, but nobody ever removed one of the ports in favor of the other.
```suggestion
```

chrstphrchvz:
Also, as of 2020-10-19 upstream MySQL no longer supports version 5.7 on macOS; they only support MySQL 8 on macOS. See https://www.mysql.com/support/eol-notice.html

ryandesign:
Note that the php port intentionally lets the user use older versions of php or of its database dependencies even if those versions are not supported by the developers anymore. Upstream support is not of particular concern to me in this port; allowing the user to do what they want to do is more important to me.

PR #11114, opened 2021-05-22
fuse: new PortGroup

ra1nb0w:
#### Description

This PortGroup chooses the right FUSE library to use.
It is the follow-up of https://github.com/macports/macports-ports/pull/10620

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G9028 x86_64
Xcode 10.3 10G8

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/fuse-1.0.tcl
+++ b/_resources/port1.0/group/fuse-1.0.tcl
@@ -1,0 +1,35 @@
+# -*- coding: utf-8; mode: tcl; c-basic-offset: 4; indent-tabs-mode: nil; tab-width: 4; truncate-lines: t -*- vim:fenc=utf-8:et:sw=4:ts=4:sts=4
+#
+# This PortGroup chooses the right FUSE library to use.
+#
+# Usage:
+#
+# PortGroup           fuse 1.0
+#
+
+# use osxfuse up to OS X 10.11
+if {${os.platform} eq "darwin" && ${os.major} < 16} {
+
+    set fuse_port "osxfuse"
+    set fuse_path "lib/pkgconfig/fuse.pc"
+
+    configure.cflags-append -I${prefix}/include/osxfuse -I${prefix}/include/osxfuse/fuse
+    configure.cppflags-append -I${prefix}/include/osxfuse -I${prefix}/include/osxfuse/fuse
+
+} else {
+
+    set fuse_port "macfuse"
+    set fuse_path "lib/pkgconfig/fuse.pc"
+
+    configure.cflags-append -I${prefix}/include/fuse
+    configure.cppflags-append -I${prefix}/include/fuse
+
+}
+
+depends_lib-append \
+    path:${fuse_path}:${fuse_port}
+
+depends_build-append \
+    port:pkgconfig
+
+universal_variant no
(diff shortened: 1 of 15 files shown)


_resources/port1.0/group/fuse-1.0.tcl
```diff
@@ -0,0 +1,35 @@
...
+#
+# This PortGroup chooses the right FUSE library to use.
+#
+# Usage:
+#
+# PortGroup           fuse 1.0
+#
+
+# use osxfuse up to OS X 10.11
+if {${os.platform} eq "darwin" && ${os.major} < 16} {
```
mascguy, line 11: Since we now support a cask-style binary, my recommendation is that we select the optimal one for each macOS release. This is per the discussion we had for [PR 10620](https://github.com/macports/macports-ports/pull/10620):

> IMHO, if we're going to utilize the binary for this port, then we should select the "best" version for a given MacOS release. And it's only three cases - 3.11.2 for 10.5 through 10.8, 4.0.5 for 10.9 through 10.11, and the latest (currently 4.1.0) for 10.12+.

Since all binary releases of macFUSE are newer than the old one we still have [prior to the project switching to a closed-source model] - and there are likely numerous bug fixes, which would benefit all macOS releases, even those as old as 10.5 - my suggestion is that we auto-select the binary macFUSE version.

Then add support for a variant, allowing users to select the legacy open-source version, on macOS releases that support it.

mascguy, line 11: Since we now support a cask-style binary, my recommendation is that we select the optimal one for each macOS release. This is per the discussion we had for [PR 10620](https://github.com/macports/macports-ports/pull/10620):

> IMHO, if we're going to utilize the binary for this port, then we should select the "best" version for a given MacOS release. And it's only three cases - 3.11.2 for 10.5 through 10.8, 4.0.5 for 10.9 through 10.11, and the latest (currently 4.1.0) for 10.12+.

Since all binary releases of macFUSE are newer than the old one we still have [prior to the project switching to a closed-source model] - and there are likely numerous bug fixes, which would benefit all macOS releases, even those as old as 10.5 - my suggestion is that we auto-select the binary macFUSE version.

Then add support for a variant, allowing users to select the legacy open-source version, on macOS releases that support it.

mascguy:
I definitely like the idea of using a portgroup for FUSE-related ports. Great stuff!

_resources/port1.0/group/fuse-1.0.tcl
ra1nb0w: I don't catch what do you mean with "auto-select the optimal". My idea was to use opensource whenever possible and then use binary/blob. Anyway, the choose between 4.0.5 or 4.1.0 must be done at macfuse port level and not in the portgroup. You are right about fixes/improvement but I personally prefer, as I said, to use the opensource one before everything else.

Variant: no. If you want to install the binary or the opensource you can still install it manually before installing the fuse module.

ra1nb0w: can you answer?

mascguy: > I don't catch what do you mean with "auto-select the optimal". My idea was to use opensource whenever possible and then use binary/blob. Anyway, the choose between 4.0.5 or 4.1.0 must be done at macfuse port level and not in the portgroup. You are right about fixes/improvement but I personally prefer, as I said, to use the opensource one before everything else.

Ah yes, you're right: the auto-selection [for binary version] should occur in port `macfuse`.

> Variant: no. If you want to install the binary or the opensource you can still install it manually before installing the fuse module.

Fair enough.

adfernandes:
LGTM! Thank you!

ra1nb0w:
Ok. Merge and we will see if problems arise.

PR #11143, opened 2021-05-26
swift-format: new port

goranmoomin:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.4 20F71 arm64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/lang/swift-format/Portfile
+++ b/lang/swift-format/Portfile
@@ -1,0 +1,129 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+name                swift-format
+categories          lang
+platforms           darwin
+license             MIT
+maintainers         @goranmoomin openmaintainer
+supported_archs     x86_64 arm64
+
+description         Formatting technology for Swift source code
+long_description    swift-format provides the formatting technology for \
+                    SourceKit-LSP and the building blocks for doing \
+                    code formatting transformations.
+
+homepage            https://github.com/apple/swift-format
+
+set swiftversions       {55 54 53 52}
+set latest_version      0.50500.0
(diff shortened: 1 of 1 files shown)


lang/swift-format/Portfile
```diff
@@ -0,0 +1,58 @@
...
+supported_archs     x86_64 arm64
+
+# Latest version gets no suffix
+github.setup        apple swift-format 0.50400.0
+set requiredXcodeVersion 12.5
+checksums           rmd160  beb5b6df42f722293097d7eed88ba320023b6bda \
+                    sha256  d8f128c1fb9a4c0f9d0bbebc4dbe6455dcb6026d7b279abbead4e1cb9d73b4de \
+                    size    227662
+
+subport ${name}54 {}
```
ryandesign, line 19: Looks like the swift-format54 subport does the same thing as the main swift-format port? We don't want to have multiple ports that do the same thing. How about making the swift-format54 subport install that version, and make the main swift-format port be a stub that does nothing but install a readme file and declare a runtime dependency on swift-format54? We use this pattern elsewhere. See for example how the php port has subports and the main port just declares a dependency on php80.

goranmoomin: @ryandesign This was my first portfile I've ever written – I'm reading the php port file, but it's too complex for me to grok how to do it.

Can you give me some pointers on how to do make the main port a stub and make it depend to a specific subport? Are there some portfiles that are less complex than the php port?

pmetzger: We have a manual! The macports guide is actually pretty good! https://guide.macports.org/

goranmoomin: I did read the Portfile reference part of the guide, but my experience was that it didn't have all the information that I needed?

E.g. I couldn't find much information on how subports work – the Portfile I currently have was written based on other Portfiles in the repository.

goranmoomin: @ryandesign I've just updated the main port to be a stub one that just depends on swift-format54. I'm not sure if this is an idiomatic Portfile – could you please take a look at it?

PR #11151, opened 2021-05-26
ncbi_tools: fix compilation and update to 20120620

weitzner:
ncbi_tools: update to version 20120620

* update to version 20120620
* add new patch file to fix compilation issues
* remove outdated patch files
* update Portfile to copy the correct binaries and data files
* fixing remaining warnings in Portfile

Closes: https://trac.macports.org/ticket/19513
Closes: https://trac.macports.org/ticket/62513

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.3
Xcode 12.5

###### Verification 
Have you


weitzner:
Remaining issues: attempts at local testing leads to 
```
:notice:fetch --->  Attempting to fetch ncbi.tar.gz from ftp://ftp.ncbi.nlm.nih.gov/toolbox/ncbi_tools/old/20120620/ncbi.tar.gz
:debug:fetch Fetching distfile failed: Server denied you to change to the given directory
```
however,
```
curl ftp://ftp.ncbi.nlm.nih.gov/toolbox/ncbi_tools/old/20120620/ncbi.tar.gz --output ncbi.tar.gz
```
executes correctly in my terminal. I am a bit perplexed at this point, and would hope that the distfile will be mirrored and this problem will go away.

science/ncbi_tools/Portfile
```diff
@@ -1,52 +1,79 @@
...
 description     blast is a set of tools for doing nucleotide and protein searches
 maintainers     nomaintainer
+license         public-domain
 categories      science
 platforms       darwin
 homepage        https://www.ncbi.nlm.nih.gov/blast/
 fetch.use_epsv  no
-master_sites    ftp://ftp.ncbi.nlm.nih.gov/toolbox/ncbi_tools/old/20080302
+master_sites    ftp://ftp.ncbi.nlm.nih.gov/toolbox/ncbi_tools/old/${version}
 distfiles       ncbi.tar.gz
```
amake, line 13: Usually this would be `distname   ncbi` and then the `.tar.gz` comes automatically from `extract.suffix`

science/ncbi_tools/Portfile
```diff
@@ -1,52 +1,79 @@
...
-checksums       md5 f8de9d7264aed7de1d87d2185df2e07c
+checksums       md5 e84826b8d390ca1ec6d0dfcaaf4cb6aa \
```
herbygillot, line 14: The `md5` checksum is no longer needed.  You can drop that one, and keep `sha256`, `rmd160` and `size`.

mascguy:
Brian, thank you so much for contributing this PR! I've approved the remaining CI workflows, to provide a more comprehensive check.

If you can go ahead and make the minor changes suggested by Aaron and Herby, we'll merge it.

science/ncbi_tools/Portfile
jmroot: But it's fine to keep md5 if it's what upstream publishes. (Yes, of course they should use a better hash function.)

jmroot:
Out of curiosity, what's the reason for not using version 20170106?

mascguy:
Brian, I pushed another commit, with proposed changes/cleanup.

Note that we wouldn't normally expect you to make most of these, as you're not the maintainer. But since we're changing this port anyway, now would be a good time for cleanup, etc.

science/ncbi_tools/Portfile
mascguy: Fixed in the 2nd "proposed cleanup" commit

science/ncbi_tools/Portfile
mascguy: Fixed in the 2nd "proposed cleanup" commit

mascguy:
One key change I made: Formally declaring 'tcsh' as a build dependency, and utilizing that to run the build script. This will help ensure consistent results, as we're not dependent on the system's `csh` shell.

mascguy:
Fellow reviewers: While the CI jobs say they've passed, if you drill into them, you'll see that dependency installation is actually failing. So we have two issues on the CI side that needs to be looked into:
* The jobs shouldn't show a pass status for dependency install failures.
* Dependency archive fetches appear to be failing, I believe.
CC: @ryandesign

Brian, you're fine though; I've verified that your port builds locally.

mascguy:
Just kicked of a rerun of all CI jobs. Hopefully the dependency fetches won't fail this time.

mascguy:
I'm not too worried about the CI jobs, as we have two successful builds: local 10.13, and CI Big Sur.

Great work on this PR Brian, and we appreciate the contribution! Merging.

mascguy:
Ticket submitted for CI fetch failure issue:

[62978 - github: CI jobs: dependency fetch failures don't fail workflow](https://trac.macports.org/ticket/62978)

mascguy:
> Out of curiosity, what's the reason for not using version 20170106?

Brian, any thoughts on Josh's question?

weitzner:
Thank you for hopping in here, reviewing, cleaning this up and getting it across the finish line, @amake, @herbygillot, @jmroot and @mascguy!

> Out of curiosity, what's the reason for not using version 20170106?

The later releases seem to be incomplete (perhaps only the deltas as the c-based tools were replaced with the modern c++ implementations?), while the 20120620 release is the most-recent that includes all of the executables expected (for my uses, `blastpgp` is the one I care most about). 

The newer tools are generally considered improvements, but some behaviors differ between them (e.g. the new tools write binary checkpoint files, while the old versions write ascii-formatted checkpoint files that other tools expect), which creates dependencies on these old versions in some workflows.

science/ncbi_tools/Portfile
```diff
@@ -56,9 +110,12 @@ destroot {
     foreach manpage ${manpages} {
         xinstall -m 444 ${worksrcpath}/ncbi/doc/man/${manpage} ${destroot}${prefix}/share/man/man1/
     }
+    xinstall -m 755 -d  ${destroot}${prefix}/data
+    foreach datafile ${datafiles} {
+        xinstall -m 444 ${worksrcpath}/ncbi/data/${datafile} ${destroot}${prefix}/data/
+    }
```
ryandesign, line 116: Note that you don't need a `foreach` loop here. It could be just:
```
xinstall -m 444 -W ${worksrcpath}/ncbi/data/ {*}${datafiles} ${destroot}${prefix}/data/
```

science/ncbi_tools/Portfile
```diff
@@ -1,51 +1,105 @@
...
 platforms       darwin
+license         public-domain
+maintainers     nomaintainer
+
+version         20120620
+revision        0
+
+description     Blast is a set of tools for doing nucleotide and protein searches
+long_description \
+                "${description}"
```
ryandesign, line 16: Note that you don't need quotation marks here.

mascguy: That was me, old habits die hard...

science/ncbi_tools/Portfile
weitzner: Good to know! I was replicating the style of the other file operations the were already in the portfile -- your suggestion is cleaner for sure.

PR #11166, opened 2021-05-28
Gpilab updates

eborisch:
Maintainer updates.

### DIFF
--- a/python/py-gpilab-core/Portfile
+++ b/python/py-gpilab-core/Portfile
@@ -8,8 +8,7 @@ github.setup            gpilab core-nodes 2.2.0 v
 name                    py-gpilab-core
 # Bump to rebuild with latest py-gpilab-framework
 revision                0
-python.versions         36 37 38
-python.default_version  37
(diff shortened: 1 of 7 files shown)


python/py-gpilab-core/Portfile
```diff
@@ -8,8 +8,8 @@ github.setup            gpilab core-nodes 2.2.0 v
 name                    py-gpilab-core
 # Bump to rebuild with latest py-gpilab-framework
 revision                0
-python.versions         36 37 38
-python.default_version  37
+python.versions         36 37 38 39
+python.default_version  38
```
cjones051073, line 12: Any reason to not make 3.9 the default at this time ?

eborisch: Yes, there is a (binary only) "Philips Data Reader" that they make available; they don't have a py39 version of it  available yet.

eborisch: I do need to drop the 36 flavors, however, with h5py dropping 36 support...

PR #11174, opened 2021-05-29
apbs: Fix build on Big Sur

jwhowarth:
apbs: Fix build on Big Sur in finding Eigen headers and add patch to fix missing forward declaration of BEMparm_copy

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.4
Xcode 12.5

###### Verification 
Have you

### DIFF
--- a/science/apbs/Portfile
+++ b/science/apbs/Portfile
@@ -26,7 +26,7 @@ github.setup            Electrostatics apbs-pdb2pqr d4e78c62e6a07f92c26924318d83
 version                 1.4.1
 homepage                http://www.poissonboltzmann.org
 worksrcdir              ${worksrcdir}/apbs
-revision                5
+revision                6
 
 checksums               rmd160  5c4d583e12deb3fbc2b5a8031882311cbfd22c7e \
                         sha256  f98ce6a51d8f813e1b4fa626c054ddbf7a985403ca30f890733cb1abf2bd6e05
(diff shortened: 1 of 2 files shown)


cjones051073:
I see you have used the master branch in your fork for this PR. This is bad practise, as if you make any other changes to that branch they will then also appear here. You should instead create a new feature branch for each PR you wish to submit, based off a clean and up to date master.

science/apbs/Portfile
```diff
@@ -41,12 +41,13 @@ use_parallel_build no
 configure.optflags      -O3 -ffast-math
 configure.args-append   -DENABLE_OPENMP:BOOL=OFF \
                         -DENABLE_BEM=OFF \
+                        -DCMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING="-I${worksrcpath}/include -O2 -g -DNDEBUG -std=c++98" \
```
cjones051073, line 44: You use -O2 optimisation here, but the line above is selecting -O3 instead. Which should the build use ?

jwhowarth: I set the CXX_FLAGS_RELWITHDEBINFO flags to those currently be set in the build. Note that the current cmake build does a poor job of transferring includes down to the compile lines and using CMAKE_CXX_FLAGS_RELWITHDEBINFO was the only way I could find to get the include for the headers passed. I would rather not mess further with the build and instead move on to porting the newer 3.0.0 release later which should have much improved cmake support.

cjones051073: If you don’t want the build to use `-O3 -ffast-math ` then why set these flags in `configure.optflags ` ?

jwhowarth: I didn't add that.

[https://github.com/macports/macports-ports/commit/da7b18b678144793d38a81c19c8cd71a3220f040#diff-feac28b79ae2f0c4429d43f0539579d5326e3bd39778bb27ce639016d53e2e64](url)

jwhowarth: Feel free to change the -O2 to -O3 -ffast-math when you commit.

cjones051073: I am not going to change it. For one its not possible easy the github ui to make changes like that as you merge. You are also the maintainer so its not unreasonable to ask you why the port has these inconsistent settings.

jwhowarth: Since pymol is unlikely to support ABPS tools again, feel free to switch this package to unmaintained.

cjones051073: If you wish to make that change you should submit a MR to do so. It is no one else's job to do the work for you.

PR #11178, opened 2021-05-30
math/r: build with implicit dependencies, remove dependency on openmp

adfernandes:
math/r: build with implicit dependencies, remove dependency on openmp

Closes: https://trac.macports.org/ticket/62626

### DIFF
--- a/math/R/Portfile
+++ b/math/R/Portfile
@@ -103,6 +103,8 @@ configure.args              --disable-openmp \
                             --without-x \
                             --with-included-gettext
 
+configure.cflags-append -Wno-implicit-function-declaration
+
 if {${os.platform} eq "darwin" && ${os.major} < 13} {
     configure.args-append --disable-openmp
 }
@@ -198,7 +200,7 @@ variant java description {enable Java} {
     configure.args-replace --disable-java --enable-java
 }
 
-default_variants +cairo +recommended +openmp
+default_variants +cairo +recommended
 
 if {![variant_isset quartz]} {
     default_variants-append +x11


i0ntempest:
I believe you can't just disable that error with a flag because it could cause weird behaviors on ASi Macs. Apple added it for a reason. Also this has nothing to do with openmp.

adfernandes:
Hey, @i0ntempest, what weird behavior are you referring to? C has always supported implicit function declarations. They're a cause of lots of headaches, that's true, but warning on them should not change code generation in any way. What part of the ABI would be affected?

The connection to `openmp` is that by using the `+openmp` variant, you now depend on the old `clang-9.0` toolchain which only works because that old toolchain does _not_ have the `-Werror,-Wimplicit-function-declaration` setting. (Which, if that is needed for ASi Macs, should be just as problematic as removing it!)

So the only reason this port works at all is because the automatic `openmp` variant has masked the whole implicit-function problem, but at the cost of pulling in the OpenMP port(s) and the (now very old) `clang-9.0` and `llvm-9.0` ports.

How would you suggest solving the problem? Especially in light of the fact that Fortran has always had, and depends on, implicit function declaration?

i0ntempest:
R 4.0.5 with and without +openmp both builds fine on my system and looking at the logs it is indeed using the system provided clang (for the one without openmp). Also R is now 4.1.0 so maybe updating it will solve the problem.
I'm not skilled enough in C to provide a complete explanation, @ryandesign could you please help by explaining a bit? Is there a way to use a newer MP clang when `compiler.openmp_version` is set?

i0ntempest:
If I blacklist `{macports-clang-[0-9].*}` in openmp variant it will now pull MP clang 11 instead, not sure that's the correct thing to do though.

i0ntempest:
Update pushed, please try the new version and report back if you still have the problem.

cjones051073:
Closing this, as we definitely do not want ports adding the `-Wno-implicit-function-declaration` flag. The correct fix is indeed to not use implicit function declarations which are not allow c99 and newer, and definitely cannot work on arm macs.

adfernandes:
Okay, so a couple of things. 

First, building with `+accelerate+cairo+gcc11-openmp+recommended+tcltk+x11` does **not** work and gives the same error as [62626](https://trac.macports.org/ticket/62626).

Second, that's _twice_ that claims have been made that "we don't want to enable implicit function declaration", but with no evidence nor justification. **Yes**, it's generally a bad idea (the increase in safety is the reason the default was changed in the C standard), but a _lot_ of old code needs it, in this case the `f2c`-ed Fortran that's part of R.

So, @i0ntempest and @cjones051073, how do you suggest we build the port with the `gcc11` variant?

adfernandes:
> Update pushed, please try the new version and report back if you still have the problem.

Same problem.

cjones051073:
> Okay, so a couple of things.
> 
> First, building with `+accelerate+cairo+gcc11-openmp+recommended+tcltk+x11` does **not** work and gives the same error as [62626](https://trac.macports.org/ticket/62626).
> 
> Second, that's _twice_ that claims have been made that "we don't want to enable implicit function declaration", but with no evidence nor justification. **Yes**, it's generally a bad idea (the increase in safety is the reason the default was changed in the C standard), but a _lot_ of old code needs it, in this case the `f2c`-ed Fortran that's part of R.

the justification is because on macOS11 arm64 implicit declarations are not possible. This is largely why Apple decided to make this an error by default in the latest Xcode, to catch bad odd code relies on it. yes, we know a lot of code relies on it, but that does not make it right. Adding the flag might fixe things for intel but will not help on arm systems, so going forward is not a solution. 
 
> 
> (There's some background on the change [here](https://github.com/astropy/astropy/issues/10770#issuecomment-698939733).)
> 
> So, @i0ntempest and @cjones051073, how do you suggest we build the port with the `gcc11` variant?

cjones051073:
If you want to read up more on why this fix is no good on arm machines you need to read up on how the two architectures handle variadic functions. In short, on arm the compiler has to see the declaration of a function before calling it, in order to know how to align parameters etc. See eg.

https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code

adfernandes:
Ah! Variadic functions! _Sound of me screaming quietly into a pillow..._ Always a source of problems!

Okay, thanks - that explains it.

I'm still curious why using the `openmp` variant supposedly works. The Fortran stuff simply _doesn't have_ function prototypes. So... what's going on here, I wonder...

ryandesign:
> @ryandesign could you please help by explaining a bit?

I explained at great length here: https://lists.macports.org/pipermail/macports-dev/2020-November/042648.html

PR #11039, opened 2021-05-17
crossgcc-1.0: Updated compiler blacklist gcc11

Gcenx:
Conditionally append compiler blacklist for gcc11+ only

Added the missing compiler_blacklist_versions PortGroup without this the compiler just default to /usr/bin/clang (tested on OS X 10.9)

revision bumped i686-w64-mingw32-gcc & x86_64-w64-mingw32-gcc

Closes https://trac.macports.org/ticket/62901

#### Description

This was tested on an OS X 10.9 VM, prior versions of mingw-bootstrap etc were removed prior to testing to ensure this worked correctly.
```
--->  Fetching distfiles for i686-w64-mingw32-gcc-bootstrap
--->  Verifying checksums for i686-w64-mingw32-gcc-bootstrap
--->  Extracting i686-w64-mingw32-gcc-bootstrap
--->  Applying patches to i686-w64-mingw32-gcc-bootstrap
--->  Configuring i686-w64-mingw32-gcc-bootstrap
--->  Building i686-w64-mingw32-gcc-bootstrap
--->  Staging i686-w64-mingw32-gcc-bootstrap into destroot
Warning: i686-w64-mingw32-gcc-bootstrap installs files outside the common directory structure.
--->  Installing i686-w64-mingw32-gcc-bootstrap @11.1.0_0
--->  Activating i686-w64-mingw32-gcc-bootstrap @11.1.0_0
--->  Cleaning i686-w64-mingw32-gcc-bootstrap
```

###### Type(s)

- [x] bugfix
- [x] enhancement

###### Tested on

macOS 10.9.5 13F1911 x86_64
Xcode Command Line Tools installed via `xcode-select —install`
```
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
```

###### Verification 

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 
- [x] checked your Portfile with `port lint`?
- [x] tried existing tests with `sudo port test`?
- [x] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?

PR #11179, opened 2021-05-30
mingw-*: Fix everything and update to v9.0.0

Gcenx:
#### Description

- crossgcc Portgroup fix compiling gcc-11 on 10.10 and lower (compiler blacklist from gcc11)
- Mingw-ctr Fix build (don't build in parallel)
- Mingw-gcc Build on ARM64/Apple Silicon (patch from gcc-11.1.0-arm branch)
- Mingw-gcc Change patch names (fix lint warnings)
- Mingw-w64 Updated to v9.0.0

###### Type(s)

- [x] enhancement

###### Tested on

macOS 11.3 20E232 x86_64
Xcode 12.5 12E262

###### Verification 
Have you


Gcenx:
Still requires https://github.com/macports/macports-ports/pull/11039 to fix mingw-gcc

Gcenx:
Combined all the needed changes into this single PR.

cjones051073:
@mojca are you ok with the changes here ?

Gcenx:
@cjones051073 I know @mojca was in favor of the crossgcc changes and mentioned fixing the lint issues within the mingw-gcc Ports so they could built automatically via the buildbots see https://github.com/macports/macports-ports/pull/11039#issuecomment-842281873

I’d added a new patch into mingw-gcc to allow building on Apple silicon, the same patch is already used by brew.

The mingw-crt fix is also applied to mingw-w64 for brew as well as proton-ge

So if @mojca is in favor of the additional two fixes.

_resources/port1.0/group/crossgcc-1.0.tcl
```diff
@@ -261,6 +262,18 @@ proc crossgcc.setup {target version} {
...
+        if { ${version} >= 11.0 } {
+            # https://trac.macports.org/ticket/29067
+            # https://trac.macports.org/ticket/29104
+            # https://trac.macports.org/ticket/47996
+            # https://trac.macports.org/ticket/58493
+            compiler.blacklist-append {clang < 800} gcc-4.0 *gcc-4.2 {llvm-gcc-4.2 < 2336.1} {macports-clang-3.[4-7]}
+
+            # https://build.macports.org/builders/ports-10.13_x86_64-builder/builds/105513/steps/install-port/logs/stdio
+            # c++/v1/functional:1408:2: error: no member named 'fancy_abort' in namespace 'std::__1'; did you mean simply 'fancy_abort'?
+            compiler.blacklist-append {clang < 1000}
```
mojca, line 274: Would you be willing to open a new trac ticket for this?

There is something strikingly similar among the old tickets (listing further references):
https://trac.macports.org/ticket/59908

cross/i686-w64-mingw32-gcc/Portfile
```diff
@@ -22,10 +22,13 @@ depends_lib-append  port:${mingw_target}-headers
...
 
-patchfiles-append   notparallel-install-headers.patch
+patchfiles-append   patch-notparallel-install-headers.diff
 
 # Fixes libgomp compile
-patchfiles-append   0020-libgomp-Don-t-hard-code-MS-printf-attributes.patch
+patchfiles-append   patch-libgomp-Don-t-hard-code-MS-printf-attributes.diff
+
+# Fix ARM64
+patchfiles-append   patch-ARM64.diff
```
mojca, line 31: @kencu: do you have any better idea about how to solve this for all cross-compilers at a single place?

mojca (commented):
@kencu: would you be willing to take a look at these changes?

It generally looks reasonable to me. I would add a bit more details to the commit message, I didn't test on different systems, but other than that I guess that this can be merged.

I wonder if anyone has a better idea for cross-compilers on arm, but we don't really need to wait for a better idea in order to merge the patches.

I would suggest to include the list of changes into the commit message rather than just listing them on the pull request.

Thanks a lot for trying to figure all of this out.

_resources/port1.0/group/crossgcc-1.0.tcl
Gcenx: The compile blacklist information was taken directly from the gcc11 port, so the same trac ticket of open would be relevant for this.

Gcenx:
This was tested on macOS Mojave and a 10.9 VM, also built on Apple Silicon.

For compile arm(64) PE binaries I believe we’d require mingw-clang but would be a PITA to make into a coherent Portfile.

cross/i686-w64-mingw32-gcc/Portfile
Gcenx: This new patch taken from homebrew it’s used in one of there gcc cross-compilers and I’d then reused it for updating there mingw-gcc Formula.

kencu: @mojca -- not a simple way. As we imagine this will be disappearing again with some upcoming update, it may be the most clear to just carry it in each place, and then delete again.

@Gcenx -- thanks for your ongoing efforts to help us & @mojca  keep all these (tricky) cross compilers in a working state!

_resources/port1.0/group/crossgcc-1.0.tcl
Gcenx: Added a note that this section was taken from gcc11 Port.

cross/i686-w64-mingw32-gcc/Portfile
Gcenx: Changed the note clearing up the origin of the ARM64 patch.

_resources/port1.0/group/crossgcc-1.0.tcl
Gcenx: @kencu was there a ticket associated with the listed error?

kencu: Not that I know of; looks like a buildbot discovery, fixed by a compiler bump, with no further digging in on the details at the time.

Gcenx: @kencu makes sense I just figured you’d know more about it since compilers seem to be your thing. But I’m guessing since it’s resolved by a compiler bump it didn’t qualify for a ticket.

@mojca I’d added a note explaining the section was taken from the gcc11 Port.

kencu:
It looks to me like this is about as close as we can expect to get to a finished PR, and no objections were raised to the overall process. I can't test this -- perhaps I should learn how to do so -- but I think at this point we put this through and if there are still any issues to be fixed, we deal with those.

Gcenx:
@kencu looks like I’d missed something and it wasn’t caught by lint.

The arm64 patch name is lower case but in the Portfile it’s uppercase.

I’ll opened a PR after work unless someone else is able to get to it first.

kencu:
And it passed all the CI tests .... that is wacky.

mojca:
macOS is case insensitive by default, so it will work on almost every machine by default.
That probably includes the environment used by GitHub Actions.

Buildbot workers run on a case sensitive filesystem on purpose: primarily to catch issues like this (before users running a case sensitive OS get hit), but also in order to avoid ending up with multiple files (that only differ in casing) being merged into a single file in the binary tarballs.

There's not much we can do about GitHub Actions other than setting up our own CI.

Gcenx:
@mojca that makes sense, yeah without hosting a custom runner for the CI tests there isn’t much that can be done.

Gcenx:
@mojca mingw-gcc ports have built bar 10.9 but that buildbot is listed as offline anyway.

In guessing the build of mingw-w64 related Ports need to now be manually triggered, or are these just waiting to be built. (Hard to check on mobile)

mojca:
I don't believe we need to trigger anything manually, but:
* 10.5, 10.6, 10.9, 10.10, 10.13 are offline
* 10.11, 10.12, 10.14 aren't yet done with commits earlier than the one for mingw
* the build on arm has passed
* the build on 10.7 almost passed, only [widl](https://build.macports.org/builders/ports-10.7_x86_64-builder/builds/58885/) failed

Gcenx:
widl might fail for anything below 10.10 depending on that’s called I’d forgotten about the current wine requirements here sigh… (widl is from wine), so maybe need to block that unless it’s some easy modifications to get it working.

Last time I’d attempted to use legacysupport for wine it failed so not sure if that’s an option but maybe it could work for this line binary.

Gcenx:
@mojca we need to build `widl` with a compiler that supports `__builtin_ms_va_list` I forced macports-clang-9.0 and it now built in my 10.9VM

PR #11180, opened 2021-05-30
fluidsynth: update to 2.1.9

szhorvat:
#### Description

 * update to 2.1.9
 * enable tests
 * make a default soundfont available

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G9028 x86_64
Xcode 11.3.1 11C504

###### Verification 
Have you


szhorvat:
Maintainers, note:

 - Latest version is 2.2.1, but it breaks API and breaks some dependent ports. Therefore, this update is to 2.1.9 which is a post-2.2.0 maintenance release.
 - Check the patch that I needed to adapt
 - When trying to play a MIDI file, I get:

        fluidsynth: error: fluid_sfloader_load(): Failed to open '/opt/local/share/soundfonts/default.sf2': File does not exist.
        fluidsynth: error: Unable to open file '/opt/local/share/soundfonts/default.sf2'
        fluidsynth: error: Failed to load SoundFont "/opt/local/share/soundfonts/default.sf2"

   To make it work, I need to manually specify the location of a soundfont (e.g. one from the port `fluid-soundfont`). I don't know if this is expected.

RJVB:
Thanks, I'll have a look!

On Sunday May 30 2021 03:13:39 Szabolcs Horvát wrote:
>Maintainers, note:
>
> - Latest version is 2.2.1, but it breaks API and breaks some dependent ports.

Breaks as in they don't build with the latest version?

>   To make it work, I need to manually specify the location of a soundfont (e.g. one from the port `fluid-soundfont`). I don't know if this is expected.

No. The current version of the port has a runtime dependency on port:generaluser-soundfont (a great all round soundfont) and this font is declared as the default at configure time via `-DDEFAULT_SOUNDFONT=${prefix}/share/sounds/sf2/GeneralUser_GS_v1.471.sf2` .
If that no longer works you'll need to figure out why.

szhorvat:
> Breaks as in they don't build with the latest version?

Yes, precisely. It's not that they simply need to be re-built, but they won't build. A version-specific fix is possible though, see e.g. here: https://github.com/rncbc/qsynth/commit/61aec88fd52bb47943ab236927f61c916858fc5a

szhorvat:
> No. The current version of the port has a runtime dependency on port:generaluser-soundfont (a great all round soundfont) and this font is declared as the default at configure time via `-DDEFAULT_SOUNDFONT=${prefix}/share/sounds/sf2/GeneralUser_GS_v1.471.sf2` .

Thanks for the hint. I can set up this solution, but the current version of the port does not actually have this dependency, and does not set the default soundfont.  What am I missing?

szhorvat:
@RJVB I force-pushed a change adding `generaluser-soundfont` as the default soundfont, and verified that it works. However, this was _not_ set up in the current version.

I notice that the filename of the soundfont filename hard-codes its version. Is this not going to be an issue when the soundfont gets updated?

szhorvat:
@RJVB It looks like your personal copy differs from the original MacPorts version: https://github.com/RJVB/macstrop/blob/master/multimedia/fluidsynth/Portfile Perhaps you want to take over this PR and add your improvements?

szhorvat:
@RJVB Do you have any more concerns about the current version of this PR? If not, can you please indicate approval so the PR can move ahead?

RJVB:
On Thursday June 03 2021 04:19:10 Szabolcs Horvát wrote:
>@RJVB Do you have any more concerns about the current version of this PR? If not, can you please indicate approval so the PR can move ahead?

I honestly have no idea; I was hoping to take a close look tomorrow. (I know I've spent time on another PR but that was, well, just venting opinions ;) )

RJVB:
OK, this looks good now. I haven't tried building it but I trust you did (and maybe I should too since you suppressed the port:gmake build.dep which may have been there for older systems).

Thanks for adding the soundfont thing and sorry for the confusion; I thought I had seen it in the official version of the port but I must have looked at one that I modified (and possibly sent to Mojca for assessment).

One thing to consider: fluidsynth has support for libomp and AFAICT will pull it in as a serendipious dependency if it happens to be installed. I presume there must be some advantage to it so in my own copy of the port (as an additional depends_lib and with this patch:

```
> cat `port dir/fluidsynth`/files/patch-force-omp.diff
--- orig.CMakeLists.txt 2019-05-24 20:07:55.000000000 +0200
+++ CMakeLists.txt      2019-05-24 20:08:17.000000000 +0200
@@ -596,9 +596,9 @@
 unset ( HAVE_OPENMP CACHE )
 find_package ( OpenMP QUIET )
 if ( OpenMP_FOUND OR OpenMP_C_FOUND )
-    message(STATUS "Found OpenMP ${OpenMP_C_SPEC_DATE}")
+    message(STATUS "Found OpenMP ${OpenMP_C_SPEC_DATE} CFLAGS=${OpenMP_C_FLAGS} LDFLAGS=${OpenMP_EXE_LINKER_FLAGS}")
     # require at least OMP 4.0
-    if ( ( NOT OpenMP_C_SPEC_DATE LESS "201307" ) OR NOT ( OpenMP_C_VERSION VERSION_LESS "4.0" ) )
+    #if ( ( NOT OpenMP_C_SPEC_DATE LESS "201307" ) OR NOT ( OpenMP_C_VERSION VERSION_LESS "4.0" ) )
         set ( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}" )
         set ( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}" )
         
@@ -606,7 +606,7 @@
         #  set ( CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS}" )
         #  set ( LIBFLUID_LIBS "${OpenMP_C_LIBRARIES};${LIBFLUID_LIBS}" )
         set ( HAVE_OPENMP 1 )
-    endif()
+    #endif()
 endif()
 
 # manipulate some variables to setup a proper test env
```

szhorvat:
I think for OpenMP to work it is necessary to use a MacPorts-provided compiler.

To ensure a consistent build we could:

 1. Disable OpenMP, whether it's supported by the compiler or not
 2. Add an OpenMP variant which requires a MacPorts-provided compiler.

What is your preference? I would go with 1, at least for now.

RJVB:
1 would be fine with me too (but port:libomp could mention this, rather than its somewhat cryptic notes which suggest that it should work with any clang v3.8 and up).

kencu:
don't forget about 

```compiler.openmp_version```

https://trac.macports.org/wiki/CompilerSelection#Parallelism

**Comment by szhorvat** [comment]
I tried to add an `openmp` variant, but I can't get it to work:

Asking for `compiler.openmp_version 4.0` results in clang-9 being used. According to the docs, Clang 9.0 supports OpenMP 4.0. However, fluidsynth says that it found OpenMP version 201107, which is supposed to be 3.something not 4.0. 4.0 should be 201307.

I tried to blacklist Clang 9.0, but couldn't figure out how to do it. Adding `compiler.blacklist-append   {clang < 1100}` to the variant or toplevel and `PortGroup compiler_blacklist_versions 1.0` to the toplevel doesn't help: it still uses Clang 9.0.

I give up.

**Comment by szhorvat** [comment]
I simply added `-DCMAKE_DISABLE_FIND_PACKAGE_OpenMP=ON` to make sure that OpenMP stays turned off, no matter what compiler is used.

The PR is good to go @RJVB .

**Comment by kencu** [comment]
To blacklist a MacPorts clang, you would do this:
{{{
compiler.blacklist-append macports-clang-9.0
}}}

I believe that except for very special ports, openmp generally will continue to be rather problematic for MacPorts until Xcode supports it, if it ever does.

Once Xcode clang supports openmp, I think then it will spread quickly.

**Comment by szhorvat** [comment]
My gut feeling is that OpenMP will not bring many advantages with this port. This is a guess only though.

Parallelization usually has overhead. The same task takes more total CPU time to complete with parallelization than without. Depending on the quality of parallelization, it may be _much_ more. If fluidsynth is used for real-time playback, and a single core is sufficient for this, then parallelization will only have drawbacks. If a single core is not sufficient, then I guess it might help just to make real-time playback possible ... but that actually happen?  If someone wants to render a long MIDI file to wave, I guess parallelization will reduce the rendering time. But how long does it take to render a single file anyway? If the need for performance comes from having to render many files, it can be done with several fluidsynth processes run in parallel—that's definitely more efficient than OpenMP.

In short, let's not bother until someone asks for the feature.

> Once Xcode clang supports openmp

I am told that it _does_ support it, it just lacks `libomp`. Installing `libomp` into a standard location with Homebrew is supposed to make it work. I never personally tried this.

**Comment by kencu** [comment]
If that is all it takes, we have libomp available with our libomp port. Last time I tried it, it did not work, but that was a while ago.

**Comment by RJVB** [comment]
On Tuesday June 08 2021 08:30:04 Szabolcs Horvát wrote:
I don't have commit permissions so someone else will have to take care of that. I have no further objections.

FWIW, my build log says that OpenMP 201107 was found and OpenMP 4.0 support enabled. That's on 10.9.5 with port:clang-5.0 and port:libomp+top_level @6.0.0 . I appreciate you trying to support it.

**Comment by RJVB** [comment]
On Tuesday June 08 2021 10:35:45 Szabolcs Horvát wrote:

>Parallelization usually has overhead. The same task takes more total CPU time to complete with parallelization than without. Depending on the quality of parallelization, it may be _much_ more.

I haven't looked at the code, but on Linux I've seen fluidsynth use 4 threads (on a quadcore CPU) during playback. For playback there could be 1 thread that runs in RT mode and handles the actual sound generation, 1 thread is undoubtedly for the "console" because everyone must have noticed that fluidsynth presents a prompt while playback happens in the background and the other 2 threads could cut the MIDI score in chunks and each render their part. Or maybe the process is inherently a multi-stage one. What I do assume is that the authors wouldn't go through the trouble of maintaining the support if there no significant gain to be had.

**Comment by szhorvat** [comment]
I see:

```
-- Found OpenMP 201107
...
Developer nerds info:
  Samples type:          double
  Multithread rendering: yes
  OpenMP 4.0:            no
```

Without OpenMP, I get:

```
Developer nerds info:
  Samples type:          double
  Multithread rendering: yes
  OpenMP 4.0:            no
```

(I.e. the same.)

----

If I add the horrible looking `compiler.blacklist-append macports-clang-6.0 macports-clang-7.0 macports-clang-8.0 macports-clang-9.0` to blacklist all older Clang versions it would be willing to use, it ends up using Clang 12.0 instead. With that I get:

```
-- Found OpenMP 201811
...
Developer nerds info:
  Samples type:          double
  Multithread rendering: yes
  OpenMP 4.0:            yes
```

**Comment by szhorvat** [comment]
Notice that it seems to have multithread support even without OpenMP.

**Comment by RJVB** [comment]
On Tuesday June 08 2021 12:50:20 Szabolcs Horvát wrote:
>I see:
...
>Without OpenMP, I get:
...
>(I.e. the same.)

Maybe that's why I have the patch I showed?

>If I add the horrible looking `compiler.blacklist-append macports-clang-6.0 macports-clang-7.0 macports-clang-8.0 macports-clang-9.0` to blacklist all older Clang versions it would be willing to use, it ends up using Clang 12.0 instead. With that I get:

Define older... (9.0 is very new in my book ;)). With MacPorts's compiler selection algorithm, you'll get to use the latest (supported) compiler that MacPorts has (for your system). I've always found that annoying, so a number of my ports use a trick that will pick the latest *installed* clang version instead of the latest available (which would still be what a user gets who has no clang ports installed). I can share it if it's of interest to you.

> Notice that it seems to have multithread support even without OpenMP.

Yeah. I delved into the source (still 2.0.5 in this case) and found that only the fluid_rvoice_mixer module uses OMP in constructs like 

```
    base_src = fluid_align_ptr(src->left_buf, FLUID_DEFAULT_ALIGNMENT);
    base_dst = fluid_align_ptr(dst->left_buf, FLUID_DEFAULT_ALIGNMENT);

    for(i = 0; i < minbuf; i++)
    {
        #pragma omp simd aligned(base_dst,base_src:FLUID_DEFAULT_ALIGNMENT)

        for(j = 0; j < scount; j++)
        {
            int dsp_i = i * FLUID_MIXER_MAX_BUFFERS_DEFAULT * FLUID_BUFSIZE + j;
            base_dst[dsp_i] += base_src[dsp_i];
        }
    }
```

To me that looks more like the compiler is forced to parallelise vector operations using SIMD instructions - but that shouldn't require a library...

So maybe it could be just as useful to provide a +cputuned variant instead of +openmp, which would build using `-Ofast -march=native -flto`.

**Comment by kencu** [comment]
> If I add the horrible looking compiler.blacklist-append macports-clang-6.0 macports-clang-7.0 macports-clang-8.0 macports-clang-9.0 to blacklist all older Clang versions it would be willing to use, it ends up using Clang 12.0 instead. With that I get:

Try this instead:
```
compiler.blacklist-append {macports-clang-[5-9].0}

kencu:
BTW -- the ```compiler.openmp_version``` is exactly designed to make it so you need to know nothing about which compilers support which openmp standards, so if you find that is not working correctly, that is a ticket to open that needs fixing, because it is supposed to be working :>

szhorvat:
> BTW -- the `compiler.openmp_version` is exactly designed to make it so you need to know nothing about which compilers support which openmp standards, so if you find that is not working correctly, that is a ticket to open that needs fixing, because it is supposed to be working :>

That's clear, but I am not familiar enough with OpenMP to be able to decide what is broken here. Is it MacPorts's compiler selection? The Clang 9.0 docs _do_ claim full OpenMP 4.5 support. fluidsynth claims to require OpenMP 4.0, yet it's not happy with Clang 9.0 for some reason. Is fluidsynth's OpenMP version detection broken? It seems to use both a "spec date" and a "version". What is the relationship between the two?

https://github.com/FluidSynth/fluidsynth/blob/master/CMakeLists.txt#L704

I would suggest merging this PR as is and leaving OpenMP support (which the previous version of the port didn't have either) or "cputune" support for later, if there is a desire to enable it.

RJVB:
>https://github.com/FluidSynth/fluidsynth/blob/master/CMakeLists.txt#L704

I disable that check in my patch, causing OpenMP to be used unconditionally if found. Have you tried that? If the build succeeds that way it means the CMake logic is simply broken (it sure seems brittle enough!).

kencu:
So, if you want your openmp variant, you can have it I believe. You just have to add this to the variant block:
```
compiler.blacklist-append {macports-clang-[5-9].0}
```

RJVB:
>compiler.blacklist-append {macports-clang-[5-9].0}

Should I repeat that it seems to work just fine with clang 5 for me?

kencu:
On a standard MacPorts install? That would be unexpected, for it to report an newer ```INFO:OpenMP-date[201107]``` on clang-5.0 than on clang-9.0 ,but WTH, software authors do medical marijuana sometimes.

What ```INFO:OpenMP-date[201107]``` does clang-5.0 report itself as being?

szhorvat:
Once again, I would like to suggest to merge this PR as-is for now (to get the update), and sort out the OpenMP situation in a followup. 

There are two things to sort out:

 - OpenMP in fluidsynth specifically. It seems like removing the CMake check makes it at least _compile_ even with Clang 5.0, according to @RJVB, but it would be nice to understand what's going on with the check, and why it doesn't work as expected. Is there an upstream bug?
 - Does MacPorts's compiler selection work properly when requesting a given OpenMP version? (@kencu )

szhorvat:
@kencu Regarding OpenMP versions, this is what I found:

The `_OPENMP` macro is defined to `yyyymm` which corresponds to the date of the OpenMP spec of the chosen version.

In Clang 9.0, I get `201107`, i.e. July 2011, which is version 3.1. https://www.openmp.org/specifications/

However, [the Clang 9.0 documentation](https://releases.llvm.org/9.0.0/tools/clang/docs/OpenMPSupport.html) claims full support for OpenMP 4.5 and partial support for OpenMP 5.0. Why does it indicate version 3.1 then? I do not know, but I found the not-very-well-documented `-fopenmp-version` compiler flag:

https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fopenmp-version

If I compile my test program as `clang-mp-9.0 -fopenmp-version=40 -fopenmp`, then `_OPENMP` is defined to `201307`, which is indeed version 4.0. With `-fopenmp-version=45` I get `201511`, which is indeed version 4.5.

Remaining questions:

 - What does `-fopenmp-version` actually do? If not specified, can OpenMP 4.0 features be used with Clang 9.0 at all?
 - Does switching to a later version merely enable newer features, or does it also change behaviour?
 - Does fluidsynth _really_ require OpenMP 4.0?
 - When specifying `compiler.openmp_version 4.0` in MacPorts, should `-fopenmp-version=40` be automatically added @kencu ?

I don't have the answers to these.

szhorvat:
@RJVB I just added an OpenMP variant, fixing the problem not with a patch but with `-fopenmp-version=40`. Do you want to use this? If yes, do you want to rename it to something like `cputune`?

RJVB:
>On a standard MacPorts install?

I don't see what MacPorts has to do with that given that the port:fluidsynth version I refer to (and which you know how to find ;)) doesn't use any of MacPorts's `compiler.foo` magic.

>What ```INFO:OpenMP-date[201107]``` does clang-5.0 report itself as being?

How do I get the answer to that question?

Either way, as said before, my patch causes any version of the OpenMP package (libomp?) to be accepted. AFAIC that's acceptable because we control libomp (and can control the compiler version to be used, too).

I compared the assembly code with and without the `-fopenmp=libomp` compiler argument that is added by the optional openmp support. With it, I see references to `llvm.mem.parallel_loop_access` and a bunch of other instructions that aren't in the no-openmp object code (which is about 3% smaller).

I do agree that we should just get the upgrade in. If anyone has suggestions how to do a cost/benefit analysis of the openmp support I could try to handle that - that should give an idea if there's any interest in supporting it in the port.

RJVB:
> @RJVB I just added an OpenMP variant, fixing the problem not with a patch but with `-fopenmp-version=40`. Do you want to use this? If yes, do you want to rename it to something like `cputune`?

No, `cputune` would be something else, or at least so I assume. It would involve adding `-march=native` which will introduce a form of CPU dependency. I would assume (but based on a hunch only) that that would be much less the case with OpenMP.

szhorvat:
> EDIT: does this mean libomp isn't required (used)?

The way the port is implemented now, `libomp` is a library dependency. I believe `compiler.openmp_version` does this automatically. But according to the fluidsynth comments, fluidsynth does not actually require linking to `libomp`.

kencu:
> On a standard MacPorts install?
> I don't see what MacPorts has to do with that given that the port:fluidsynth version I refer to (and which you know how to find ;)) doesn't use any of MacPorts's `compiler.foo` magic.

I wasn't clear. I was thinking of a "standard macports install" of clang-5.0. It turns out with all this exploring we're doing we've learned there is quite a bit of fuzz on this here openmp peach :>

RJVB:
On Thursday June 10 2021 16:21:10 Ken wrote:

>I wasn't clear. I was thinking of a "standard macports install" of clang-5.0. It turns out with all this exploring we're doing we've learned there is quite a bit of fuzz on this here openmp peach :>

Ah, well, my clang-5.0 install on Mac has a few tweaks to make it smaller and possibly just a tad faster but I'm not meddling with OpenMP support. And my Linux clang-5.0 install is absolutely stock ("official" builds from llvm.org).

szhorvat:
Are we happy wit the current state of this PR and can it be merged?

@RJVB These are the lines I added since you last approved it:

https://github.com/macports/macports-ports/pull/11180/files#diff-a9ccd1f97d21ef9fa094a33af657ebc87bdc91ca43d7ad6fe5a0cb9f596c82dbR79-R84

Let's either merge as-is, or if there are concerns about the `openmp` variant, I'll remove it, let's merge the PR without, and I can open a new PR specifically for OpenMP support.

kencu:
> @kencu Regarding OpenMP versions, this is what I found:
> 
> The `_OPENMP` macro is defined to `yyyymm` which corresponds to the date of the OpenMP spec of the chosen version.
> 
> In Clang 9.0, I get `201107`, i.e. July 2011, which is version 3.1. https://www.openmp.org/specifications/
> 
> However, [the Clang 9.0 documentation](https://releases.llvm.org/9.0.0/tools/clang/docs/OpenMPSupport.html) claims full support for OpenMP 4.5 and partial support for OpenMP 5.0. Why does it indicate version 3.1 then? I do not know, but I found the not-very-well-documented `-fopenmp-version` compiler flag:
> 
> https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fopenmp-version
> 
> If I compile my test program as `clang-mp-9.0 -fopenmp-version=40 -fopenmp`, then `_OPENMP` is defined to `201307`, which is indeed version 4.0. With `-fopenmp-version=45` I get `201511`, which is indeed version 4.5.
> 
> Remaining questions:
> 
> * What does `-fopenmp-version` actually do? If not specified, can OpenMP 4.0 features be used with Clang 9.0 at all?
> * Does switching to a later version merely enable newer features, or does it also change behaviour?
> * Does fluidsynth _really_ require OpenMP 4.0?
> * When specifying `compiler.openmp_version 4.0` in MacPorts, should `-fopenmp-version=40` be automatically added @kencu ?
> 
> I don't have the answers to these.

It is indeed quite common for clang to hold certain default settings (like the c++ standard) at certain levels for compatibility, even if the software can support newer levels. clang only defaulted to c++17 as of about clang-8.0 or clang-9.0, I forget which. Up to then, it defaulted to c++03.

So no doubt clang-9.0 is capable or a newer openmp than it defaults to. You are quite right, we probably do need to add the fopenmp-version flag to set the openmp version we want.

PR #11196, opened 2021-06-01
kerberos5: update python version to 3.9

ferdy:
###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on

macOS 11.3.1 20E241
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/net/kerberos5/Portfile
+++ b/net/kerberos5/Portfile
@@ -32,7 +32,7 @@ master_sites                ${homepage}dist/krb5/${branch}/
 distname                    krb5-${version}
 
 depends_build               port:pkgconfig \
-                            port:python38
+                            port:python39
 
 depends_lib                 port:gettext \
                             port:libcomerr \
@@ -59,7 +59,7 @@ use_autoreconf              yes
 # remove ${prefix}/lib from configure.ldflags to allow linking against its own libs first
 # adding ${worksrcpath}/lib is not necessary and pollutes krb5-config --libs and pkg-config files
 configure.ldflags-delete    -L${prefix}/lib
-configure.python            ${prefix}/bin/python3.8
+configure.python            ${prefix}/bin/python3.9
 configure.args              --mandir=${prefix}/share/man \
                             --with-system-et \
                             --without-system-db \


ryandesign:
@ferdy The switch to from python38 to python39 in this PR seems reasonable. Python is only a build dependency, so in what way does this change the installed files? If, as I assume, it does not, then increasing the revision is not warranted.

I don't understand why there are 106 other commits in this PR or know what effect merging this PR with those extra commits would have.

The other comments in this PR appear to be unrelated to it. Presumably they relate to the other commits.

ferdy:
> @ferdy The switch to from python38 to python39 in this PR seems reasonable. Python is only a build dependency, so in what way does this change the installed files? If, as I assume, it does not, then increasing the revision is not warranted.

i see your point, of course you are right.
i have wrongly given for granted you are always supposed to increase the revision. my fault.
 
> I don't understand why there are 106 other commits in this PR or know what effect merging this PR with those extra commits would have.
> 
> The other comments in this PR appear to be unrelated to it. Presumably they relate to the other commits.

actually I do not understand either. surely I created a custerfck messing with git.

right after i opened the pr. i received the message notifying the test failed.
i checked the log and i noticed the it failed to fetch the bzip2 archive, thus i thought it was a temporary error and i pushed again to have the test restarted.

can i just close this pr and open a brand new (i am aware i should amend this, but i do not want to do any further damage).

please let me know,

ryandesign:
> i checked the log and i noticed the it failed to fetch the bzip2 archive, thus i thought it was a temporary error and i pushed again to have the test restarted.

Sorry about that, the archive fetch failures in CI were my fault. They shouldn't happen anymore if your PR is from an up to date master.

> can i just close this pr and open a brand new (i am aware i should amend this, but i do not want to do any further damage).

I think I've rescued the PR with a rebase and force push, and I removed the revision increase. Let's make sure the CI builds pass now and then we can merge it.

ferdy:
sorry for the annoyance and (as always) thank you very much indeed!

jmroot:
CI passed so I went ahead and merged. This also resolves [#63040](https://trac.macports.org/ticket/63040).

PR #11230, opened 2021-06-03
mkvtoolnix: use the new boost pg

i0ntempest:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.5 20G5033c x86_64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/multimedia/mkvtoolnix/Portfile
+++ b/multimedia/mkvtoolnix/Portfile
@@ -2,6 +2,7 @@
 
 PortSystem          1.0
 PortGroup           muniversal 1.0
+PortGroup           boost 1.0
 
 # The developer does not accept macOS-specific bug reports, but does
 # accept pull requests.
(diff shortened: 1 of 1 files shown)


ryandesign (changes requested):
I don't know about this portgroup. If this is how it is used, then I leave it up to you to make the right changes.

Changing boost to a build dependency is not the right change for all of the versions of mkvtoolnix that this port offers. mkvtoolnix 41.0.0 and 53.0.0 link with boost libraries. I see that mkvtoolnix 57.0.0 does not link with them anymore, so I do not know whether that means boost is then needed only at build time or whether it means that the use of boost has been replaced with newer standard C++ features and that boost is no longer needed.

i0ntempest:
Yes, from their change log of version 54: "MKVToolNix is now using the C++17 library feature "file system library"
  instead of Boost's "file system" and "system" libraries." Maybe it doesn't even need boost to build any more, I'll try.

i0ntempest:
Okay it still needs some boost headers. I've moved `boost.depends_type build` into v57 so other versions pull boost as a lib dep.

PR #11286, opened 2021-06-09
doas doesn't install vidoas

anahata0108:
We don't install a vidoas command, so instead tell the user to run a command that has a better chance of working.

### DIFF
--- a/sysutils/doas/Portfile
+++ b/sysutils/doas/Portfile
@@ -72,7 +72,7 @@ post-activate {
 notes "
 To complete the installation, run:
 
-\t${prefix}/bin/doas ${prefix}/sbin/vidoas
+\t${prefix}/bin/doas $EDITOR ${prefix}/etc/doas.conf
 
 Edit as necessary. (See the doas.conf manpage for additional information.)
 "


cjones051073 (changes requested):
Please can you change your commit message to follow our guidelines, i.e.

portname: description

herbygillot:
Also, what happens if `$EDITOR` isn't set?

cjones051073:
Does this port build vidoas ? If so surely a better fix is to also install it ? Is the fact it isn’t just an oversight ?

anahata0108:
I suppose the existence of this issue depends on whether the port is supposed to install vidoas. Is anyone able to answer that question? danchr has already been pinged, so I don't want to ping them again.

cjones051073:
It seems the port does install `vidoas` so this MR is not required.

```
Oberon ~/Projects/MacPorts/ports > port contents doas
Port doas contains:
  /opt/local/bin/doas
  /opt/local/bin/vidoas
  /opt/local/share/doc/doas/doas.conf.dist
  /opt/local/share/doc/doas/doas.conf.sample
  /opt/local/share/man/man1/doas.1.gz
  /opt/local/share/man/man5/doas.conf.5.gz
  /opt/local/share/man/man8/vidoas.8.gz
```

anahata0108:
I don't think this is the correct solution. The port installs vidoas to $PREFIX/bin, whereas the message references $PREFIX/sbin. It seems appropriate to change where vidoas is installed, as it's a sysadmin tool. Thoughts?

cjones051073:
Good point, missed the message referred to sbin.  https://github.com/macports/macports-ports/commit/e410e5e5def47e0073330055cfd9711a12d8dba7

PR #11311, opened 2021-06-11
logisim-evolution: set fallback dep to openjdk16

harens:
#### Description

Following from the recent update https://github.com/macports/macports-ports/commit/b90614c7aea403a02dc9b328ef27f4765263d586.

[10.12_x86_64-builder log](https://build.macports.org/builders/ports-10.12_x86_64-builder/builds/149069/steps/install-port/logs/stdio):

```
> java.io.IOException: Cannot run program "/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home/bin/jpackage": error=2, No such file or directory
```

`jpackage` is available in openjdk14+, so this PR sets the portfile to use openjdk 16, since 14 and 15 are obsolete. Not using LTS openjdk 8 or 11 however isn't good practice. 

What I didn't realise was that trace mode wouldn't pick up on this error since openjdk installs outside of the standard prefix.

My thoughts were to set to openjdk 16 now, and set it to LTS 17 when it comes out? CC @chrstphrchvz.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/cad/logisim-evolution/Portfile
+++ b/cad/logisim-evolution/Portfile
@@ -5,7 +5,7 @@ PortGroup           java 1.0
 PortGroup           github 1.0
 
 github.setup        reds-heig logisim-evolution 3.5.0 v
-revision            0
+revision            1
 
 categories          cad education java
 platforms           darwin
@@ -22,8 +22,10 @@ checksums           rmd160  87b24defcabd6ab81333bc1c52c76928a64bdab1 \
                     sha256  8b52f1ce57205f68c396d12637ccb1d6955f631987c2e6b4247acce31116cbe4 \
                     size    40050282
 
-java.version        9+
-java.fallback       openjdk11
+# non-LTS Java required for jpackage
+# Set to LTS Java 17 when released
+java.version        14+
+java.fallback       openjdk16
 
 use_configure       no
 


cad/logisim-evolution/Portfile
```diff
@@ -23,7 +23,7 @@ checksums           rmd160  87b24defcabd6ab81333bc1c52c76928a64bdab1 \
                     size    40050282
 
 java.version        9+
```
chrstphrchvz, line 25: Does this need to be updated as well?

```suggestion
java.version        14+
```

chrstphrchvz (commented):
Thanks for the CC. This usage should be fine; I would suggest adding a comment by `java.fallback` explaining the use of non-LTS Java and that it intends to migrate to and then probably remain on LTS Java when Java 17 is released.

I had neglected the possibility of something requiring functionality only provided in Java versions newer than the latest LTS version. I will consider reflecting that in macports/macports-guide#40.

I remember discussing with @breun (`openjdk*` maintainer) that the non-LTS Java ports could potentially be obsoleted immediately or soon after a new version is added to MacPorts, which is easiest to do when there are as few dependents as possible. So that is what may eventually require some coordination.

cad/logisim-evolution/Portfile
harens: That's a good spot. Thanks @chrstphrchvz.

harens:
Thanks for reviewing @chrstphrchvz. I've made your suggested changes (as well as bumping the revision).

> I remember discussing with @breun (openjdk* maintainer) that the non-LTS Java ports could potentially be obsoleted immediately or soon after a new version is added to MacPorts, which is easiest to do when there are as few dependents as possible. So that is what may eventually require some coordination.

That's a good point. This port should be fine since I can update it quickly when required, but it could become problematic in the future if more ports require these non-LTS features. Fortunately, so far this hasn't been the case.

Assuming the tests pass, I'll commit this fix shortly. Thank you again for your help @chrstphrchvz!

PR #11260, opened 2021-06-07
curl: upstream does not support --with-darwinssl switch, use --with-secure-transport

seabadger:
…; also, --with-ssl is really --with-openssl; --without-ssl means not SSL support at all

#### Description

Fix curl +darwinssl variant. 
The '--with-darwinssl' configure script switch has not worked for a while.
Update port file to use '--with-secure-transport' instead.

(without the fix, curl +darwinssl builds without ssl support at all, since --with-darwinssl was being silently ignored by curl's configure script; upstream curl was recently updated in master (not yet released) to fail the build  as of https://github.com/curl/curl/pull/7200/commits/5731d5a9be4fcf89e356293a28b4289cf65bd0ae).

I also updated the '--with-ssl' switch to the synonymous but more explicit --with-openssl switch.
The reason for this is that '--with-ssl' and '--without-ssl' are not symmetric: the former enables openssl, the latter  disables ALL ssl backends. This should make the +ssl variant clearer. Functionally nothing changes.

###### Type(s)
- [x ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.4 20F71 x86_64
Xcode 12.3 12C33
###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] checked your Portfile with `port lint`?
- [x] tried existing tests with `sudo port test`?
- [x] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?

PR #11361, opened 2021-06-13
curl: fix +darwinssl variant 

seabadger:
#### Description

Fix curl +darwinssl variant.
The '--with-darwinssl' configure script switch has not worked for a while.
Update port file to use '--with-secure-transport' instead.

(without the fix, curl +darwinssl builds without ssl support at all, since --with-darwinssl was being silently ignored by curl's configure script; upstream curl was recently updated in master (not yet released) to fail the build as of curl/curl@5731d5a).

I also updated the '--with-ssl' switch to the synonymous but more explicit --with-openssl switch.
The reason for this is that '--with-ssl' and '--without-ssl' are not symmetric: the former enables openssl, the latter disables ALL ssl backends. This should make the +ssl variant clearer. Functionally nothing changes.

###### Type(s)
- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.4 20F71 x86_64
Xcode 12.3 12C33

###### Verification
Have you

### DIFF
--- a/net/curl/Portfile
+++ b/net/curl/Portfile
@@ -60,7 +60,7 @@ if {${name} eq ${subport}} {
                                 --without-nghttp2 \
                                 --without-nss \
                                 --without-ssl \
-                                --without-darwinssl \
+                                --without-secure-transport \
                                 --disable-ares \
                                 --disable-ldap \
                                 --disable-ldaps \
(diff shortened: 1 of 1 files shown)


seabadger:
Please note that this PR has the same change as (the now closed) https://github.com/macports/macports-ports/pull/11260 but without messed up git history.

PR #11402, opened 2021-06-18
webkit2-gtk-devel: add enhancements to WebKitGTK+

jmercouris:
#### Description
This fixes a few things that help webkit2-gtk-devel more suitable for development.

- [ ] enhancement
- [ ] security fix

@pmetzger 
@kencu 

I no longer have a macOS machine handy, so I can't readily test this.

### DIFF
--- a/www/webkit2-gtk-devel/Portfile
+++ b/www/webkit2-gtk-devel/Portfile
@@ -91,6 +91,16 @@ patchfiles-append    patch-enable-plugin-architecture-unix.diff
 # runs without this patch.
 patchfiles-append    patch-bundle-link-webcore.diff
 
+# allow the webkit executables typically located in libexec to be
+# dynamically located, this is useful for app bundling
+patchfiles-append    process-executable-path.diff
+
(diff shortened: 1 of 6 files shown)


pmetzger:
Just for @dbevans and @kencu's benefit, this is apparently a prerequisite for packaging the Nyxt browser for MacPorts (see https://nyxt.atlas.engineer/ )

kencu:
Tricky to test this properly indeed. One successful build, and one failed build that might have simply timed out. The changes look reasonable. Nit point is that we'll need a revbump to get it to happen, but that is trivial for me to add.

I suppose I should check it out and test this on at least one system.... when I originally did these webkit patches I ran builds on every system from 10.6 i386 up, but I'm not in a position to invest that kind of time any longer.

I'll run a build, and if it doesn't fail or crash on use, we'll push this with a revbump.

jmercouris:
Thanks @kencu , let me know how it works for you.

kencu:
10.14 worked OK.

PR #11428, opened 2021-06-21
joshuto: New port

judaew:
Version 0.9.0

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.4 20F71 x86_64
Xcode 12.5 12E262

###### Verification 
Have you


sysutils/joshuto/Portfile
```diff
@@ -0,0 +1,133 @@
...
+    wasi                          0.9.0+wasi-snapshot-preview1  cccddf32554fecc6acb585f82a32a72e28b48f8c4c1883ddfeeeaa96f7d8e519 \
+    wasi                          0.10.2+wasi-snapshot-preview1  fd6fbd9a79829dd1ad0cc20627bf1ed606756a7f77edff7b66b7064f9cb327c6 \
+    which                            4.1.0  b55551e42cbdf2ce2bedd2203d0cc08dba002c27510f86dab6d0ce304cba3dfe \
+    whoami                           0.9.0  7884773ab69074615cb8f8425d0e53f11710786158704fca70f53e71b0e05504 \
+    winapi                           0.3.9  5c839a674fcd7a98952e593242ea400abe93992746761e38641405d28b00f419 \
+    winapi-i686-pc-windows-gnu       0.4.0  ac3b87c63620426dd9b991e5ce0329eff545bccbbb34f3be09ff6fb6ab51b7b6 \
+    winapi-x86_64-pc-windows-gnu     0.4.0  712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f \
+    xdg                              2.2.0  d089681aa106a86fade1b0128fb5daf07d5867a509ab036d99988dec80429a57
+
+build.pre_args      --release -v -j${build.jobs}
```
herbygillot, line 124: Hey @judaew - you provide `cargo.crates` above, but at the same time you set `build.pre_args` without the `--frozen` flag... 

`--frozen` is required to make `cargo` not attempt to download deps from the Internet and only use what's provided by `cargo.crates`.  Is there any reason why you felt you needed to override `build.pre_args`?

judaew: Hmm, if I use `--frozen` (default PG settings), then the port is not built. Does this mean it cannot be built without Internet access or am I doing something wrong?

```
:debug:build system:  cd "/opt/local/var/macports/build/.../joshuto/work/joshuto-0.9.0" && /opt/local/bin/cargo build --release --frozen -v -j8 
:info:build error: the lock file /opt/local/var/macports/build/.../kamiyaa-joshuto-65f7aaf/Cargo.lock needs to be updated but --frozen was passed to prevent this
:info:build If you want to try to generate the lock file without accessing the network, use the --offline flag.
:info:build Command failed:  cd "/opt/local/var/macports/build/.../joshuto/work/joshuto-0.9.0" && /opt/local/bin/cargo build --release --frozen -v -j8 
:info:build Exit code: 101
```

herbygillot: This means that the `Cargo.lock` file that you generated your `cargo.crates` from is not matching the original `Cargo.toml`.  As a result of this it wants to connect to the Internet to fetch what it thinks is missing.

You can leave as is, but if you want to have everything work using the default `--frozen` configuration, then you will need to regenerate `Cargo.lock` and then regenerate `cargo.crates` from that.

You would then need to include a patch in the Portfile that patches `Cargo.lock` when the port is extracted.

To regenerate the `Cargo.lock`, you would need to do something like:
1. `sudo port extract joshuto`
2. `cd $(port work joshuto)/joshuto...`
3. `sudo -u macports HOME=../.home cargo regenerate-lockfile`

Now with that new `Cargo.lock` file, you can create the patch file and create a correct `cargo.crates` section.

judaew: Thanks so much. Earlier with other ports, I also had a similar problem, but now I will know how to solve it.

herbygillot:
This looks good @judaew.  In situations like this, you can also open an issue with the upstream project requesting them to keep their `Cargo.lock` up-to-date whenever they release, so that things work without you having to do these extra steps.

PR #11432, opened 2021-06-21
duckdb: enable parquet and full text search extensions

fs111:
#### Description

adds a variant to duckdb that enables support for Apache Parquet. 

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance
macOS 10.15.7 19H2 x86_64

###### Verification 
Have you

### DIFF
--- a/databases/duckdb/Portfile
+++ b/databases/duckdb/Portfile
@@ -6,7 +6,7 @@ PortGroup           cmake 1.1
 PortGroup           legacysupport 1.0
 
 github.setup        cwida duckdb 0.2.7 v
-revision            0
+revision            1
 
 homepage            https://www.duckdb.org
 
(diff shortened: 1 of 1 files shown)


herbygillot:
This doesn't seem to require any extra deps - if that's so, we should just build it in by default. No need to have it as a separate variant.

fs111:
@herbygillot yes, they use the in-tree parquet library for that. I thought that since the upstream project is building binaries w/o parquet by default this should be a variant, but we can def. enable it by default. LMK if you want me to update the  PR

herbygillot:
> @herbygillot yes, they use the in-tree parquet library for that. I thought that since the upstream project is building binaries w/o parquet by default this should be a variant, but we can def. enable it by default. LMK if you want me to update the PR

OK, since it's in-tree, that's great - it means we can just turn it on "for free".  Let's enable it, as well as other features that could make sense to have by default and update this commit.

Since doing this changes the contents of the current DuckDB version, you will need to bump the `revision`.

fs111:
@herbygillot I have updated the PR. We now enable Parquet and Full Text Search by default

herbygillot:
Awesome @fs111, thank you and merged.

PR #11457, opened 2021-06-25
go-tools: update to 0.1.4

enckse:
#### Description

- fix building using go 1.16 (set `GO111MODULE=off` if not using `go.mod`)
- update version to commit cd1d0887 (required new dep, removed duplicate dep...)
- replaced errant tabs on 2 lines

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.4 20F71 arm64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/devel/go-tools/Portfile
+++ b/devel/go-tools/Portfile
@@ -1,71 +1,32 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           github 1.0
+PortGroup           golang 1.0
 
-github.setup        golang tools 89232532d2d57ce53cc88eccb5d060e92f86729b
-version             20200630-[string range ${github.version} 0 7]
-epoch               6
+go.setup            github.com/golang/tools 0.1.4 v
+epoch               7
+revision            0
+
+
 name                go-tools
 categories          devel
 platforms           darwin freebsd linux
 license             BSD
-maintainers         {ciserlohn @ci42}
+maintainers         {ciserlohn @ci42} {@enckse voidedtech.com:enckse} openmaintainer
 description         Various packages and tools that support the Go programming language.
 long_description    $description
 
-checksums           rmd160  97410382ccbe3efce08e96756c63db33eb8b1703 \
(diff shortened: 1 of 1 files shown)


devel/go-tools/Portfile
```diff
@@ -3,8 +3,8 @@
 PortSystem          1.0
 PortGroup           github 1.0
```
herbygillot, line 5: This port could probably benefit from using the `golang` PortGroup.

It knows how to compile Go software in a platform-portable way across older versions of macOS, among other benefits.

By default we want to use it for any port for software written in Go.

We'd start off here with:
```
```

devel/go-tools/Portfile
```diff
@@ -14,11 +14,11 @@ maintainers         {ciserlohn @ci42}
...
-checksums           rmd160  97410382ccbe3efce08e96756c63db33eb8b1703 \
-                    sha256  c451b5af8a10b06624d5ae92c33aa1b30beaf380e195d1373b333fde1553c15a \
-                    size    2484284
-		    
-depends_build	    port:go
+checksums           rmd160  861f7f72627b4449752dded9ab7c1a376435e841 \
+                    sha256  b26d3b51fbcbabc7ff8fd2cfdb152421753dfeb8f7e38be9041c0bddfa1b8acd \
+                    size    2785629
+
+depends_build       port:go
```
herbygillot, line 21: You can remove the `depends_build       port:go` if you are using the `golang` portgroup.

enckse:
@herbygillot 

Yep, I didn't want to rewrite it entirely but I saw that it needed to be modernized (but didn't want to step on protocol). I've done another pass, please see if you agree with it now

I was mostly uncertain about what to do for version (I understand the idea of epoch/versioning) because go-tools does have a version (nowadays) so we should probably track to that instead of date+commits

herbygillot:
> @herbygillot
> 
> Yep, I didn't want to rewrite it entirely but I saw that it needed to be modernized (but didn't want to step on protocol). I've done another pass, please see if you agree with it now
> 
> I was mostly uncertain about what to do for version (I understand the idea of epoch/versioning) because go-tools does have a version (nowadays) so we should probably track to that instead of date+commits

@enckse your changes look great, and 100% that we should use the version number since that's available now.

The Portfile is a lot more straightforward now - my only comment might be we could probably remove `livecheck none`.

enckse:
`livecheck.type` removed

herbygillot:
Looks good and merged @enckse - thanks.

PR #11466, opened 2021-06-27
isync: add +bdb variant

enckse:
#### Description

Adds the `+bdb` variant which can be used to enable db53 dependency which is required for `mdconvert` (converting mailbox format - less likely to be used executable)

This does make db53 an _optional_ dependency which, for isync on arm64, allows install that is blocked by https://trac.macports.org/ticket/62216

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.4 20F71 arm64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/mail/isync/Portfile
+++ b/mail/isync/Portfile
@@ -4,7 +4,7 @@ PortSystem              1.0
 
 name                    isync
 version                 1.4.2
-revision                0
+revision                1
 categories              mail
 platforms               darwin
 license                 GPL-2
@@ -28,7 +28,6 @@ depends_build           port:perl5 \
                         port:pkgconfig
 
 depends_lib             port:cyrus-sasl2 \
-                        port:db53 \
                         path:lib/libssl.dylib:openssl \
                         port:zlib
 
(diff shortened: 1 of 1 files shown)


mail/isync/Portfile
```diff
@@ -60,8 +59,11 @@ post-destroot {
     xinstall -m 644 -W ${worksrcpath} mbsync.plist ${destroot}${examples_dir}/${startupitem.plist}
 }
 
-configure.cppflags-append   -I${prefix}/include/db53
-configure.ldflags-append    -L${prefix}/lib/db53
+variant db5 description {Enable mdconvert compilation with Berkeley DB datatabase} {
```
exg, line 62: Please name the variant `bdb` (so that we can keep the name as is when we bump the dependency) and change the description to `Enable support for the alternative UID storage scheme` (see the `mbsync` `AltMap` store setting).

exg (changes requested):

mail/isync/Portfile
enckse: updated

mail/isync/Portfile
```diff
@@ -60,8 +59,11 @@ post-destroot {
     xinstall -m 644 -W ${worksrcpath} mbsync.plist ${destroot}${examples_dir}/${startupitem.plist}
 }
 
-configure.cppflags-append   -I${prefix}/include/db53
-configure.ldflags-append    -L${prefix}/lib/db53
+variant bdb description {Enable support for the alternative UID storage scheme} {
```
exg, line 62: The variant should be enabled by default (with `default_variants`).

exg (changes requested):

mail/isync/Portfile
enckse: done

exg (approved):

PR #11496, opened 2021-07-01
perl5: address problematic baked-in values

jmroot:
Use generic `/usr/bin/cc` for compiler since the specific one used to
build perl may not be available at runtime. Remove flags specifying SDK
and deployment target since different ones may be specified in the
environment at runtime, and, similar to the compiler, the SDK that was
used at build time may not be available at runtime.

Fixes: https://trac.macports.org/ticket/59786
Fixes: https://trac.macports.org/ticket/62440

#### Description

There is of course still more work to be done in collaboration with upstream, but this should deal with the main problems that we are commonly seeing.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

### DIFF
--- a/lang/perl5/Portfile
+++ b/lang/perl5/Portfile
@@ -24,15 +24,15 @@ master_sites        https://www.cpan.org/src/5.0/
 # - sha256
 # - size
 set perl5.versions_info {
-    5.16 3 9 f25fdd72449156a7cbe989e8bd339fdba1afabc0  bb7bc735e6813b177dcfccd480defcde7eddefa173b5967eac11babd1bfa98e8  13724906
-    5.18 4 8 d97181a98f7acc80125b0d2a182a6a2cd7542ceb  1fb4d27b75cd244e849f253320260efe1750641aaff4a18ce0d67556ff1b96a5  14059430
-    5.20 3 7 499846a1c92e00dd357cb782bc14787b8cd47051  1b40068166c242e34a536836286e70b78410602a80615143301e52aa2901493b  13743405
-    5.22 4 5 54fdbcbf249134dc7d82b693417900286201b5e7  8b3122046d1186598082d0e6da53193b045e85e3505e7d37ee0bdd0bdb539b71  13745983
-    5.24 4 4 8d6b67fc6d58334b2fdbfa9d6d7456265dca1f4e  e34ff38c54857f431f37403b757267c9998152bf46b5c750b462f62461279b10  14125130
-    5.26 3 4 84ed404407c198ca2b8194c374c7914d941b6f49  9ff35a613213f29ab53975141af6825ae7d4408895538cac0922e47ab92a1477  14539342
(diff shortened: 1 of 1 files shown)


lang/perl5/Portfile
```diff
@@ -209,6 +209,17 @@ foreach {perl5.v perl5.subversion perl5.revision perl5.rmd160 perl5.sha256 perl5
         post-build {
             reinplace -E {s/-arch [a-z0-9_]+//g} \
                 ${worksrcpath}/lib/Config_heavy.pl
+            reinplace -E {s/-mmacosx-version-min=[0-9.]+//g} \
+                ${worksrcpath}/lib/Config_heavy.pl
+            reinplace "s|${configure.cc}|/usr/bin/cc|g" \
```
mojca, line 214: Will that work on < 10.9 where we switched to `clang-mp` as the default compiler? How will the rest of the flags cope with that different compiler? I mean: perl comes up with tons of flags depending on various checks it does when it's being built with the latest clang, but then those flags and feature checks will be applied to `gcc-4.2` when building the modules like `p5-tk`? (Maybe we come up with a proper patch for `cc` soon anyway.)

Maybe we could introduce `${prefix}/bin/cc-mp-default` at some point ;) ;) ;)

@kencu

lang/perl5/Portfile
```diff
@@ -209,6 +209,17 @@ foreach {perl5.v perl5.subversion perl5.revision perl5.rmd160 perl5.sha256 perl5
...
+                ${worksrcpath}/lib/Config.pm \
+                ${worksrcpath}/lib/Config_heavy.pl \
+                ${worksrcpath}/config.h
+            reinplace -E {s|-isysroot[[:space:]]*[^[:space:]]+\.sdk||g} \
+                ${worksrcpath}/lib/Config_heavy.pl \
+                ${worksrcpath}/config.h
+            reinplace -E {s|-Wl,-syslibroot,[^[:space:]]+\.sdk||g} \
+                ${worksrcpath}/lib/Config_heavy.pl
```
mojca, line 222: Awesome, thank you very much.
I came up with a `sed` expression to do exactly that, but I had some troubles translating it into Tcl ;)

This still leaves one trace of a header file for time in `Config_heavy.pl` with full path hardcoded in, but I have no clue how to get rid of that one.

mojca (approved):
Thanks a lot. I'm somewhat sceptic about the patch for `cc`, but other than fixing the perl build system, I don't have any significantly better idea.

I'll add #11440 once this gets merged.

lang/perl5/Portfile
jmroot: We still want to set `CC` explicitly when building perl modules in MacPorts. This just fixes certain build systems that like to check if the baked-in compiler works. (And it can't get any more broken than trying to use a nonexistent compiler.)

lang/perl5/Portfile
jmroot: Yes, that's some of the further work I alluded to.

PR #11504, opened 2021-07-02
elftoolchain: Fix implicit func decls; other fixes

ryandesign:
#### Description

Fix implicit declaration of function htole32 which caused build failure
on Xcode 12 and later.

Closes: https://trac.macports.org/ticket/62419

Remove unneeded additions of defines of htobe32 in archive.c and write.c
because it's already defined in _elftc.h. Use makefile portgroup to
simplify using the right compiler and flags. Copy build.args to
destroot.args programmatically to avoid code duplication. Don't need to
set destroot.target since it's already set by base. Remove unneeded
addition of CFLAGS and LDFLAGS in elftoolchain.prog.mk since the
makefile portgroup already sets them. Patch make-toolchain-version to no
longer use svnversion or git describe since we're not using a working
copy or clone and since those utilities might not be installed. Create
relative symlinks rather than absolute. Fix erroneously formatted
warning in elftoolchain.tex.mk by changing echo -n to printf. Specify
full sourceforge path to avoid redirects and fetch failures. Add size to
checksums. Add modeline.

###### Type(s)

- [X] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on
macOS 10.13.6 17G14042 x86_64
Xcode 9.4.1 9F2000

###### Verification 
Have you

### DIFF
--- a/devel/elftoolchain/Portfile
+++ b/devel/elftoolchain/Portfile
@@ -1,7 +1,11 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
 PortSystem      1.0
+PortGroup       makefile 1.0
 
 name            elftoolchain
 version         0.7.1
+revision        1
 categories      devel
 platforms       darwin
 license         BSD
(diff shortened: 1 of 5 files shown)


ryandesign:
It fails to build on macOS 11 due to an unrelated problem:
```
  /opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_devel_elftoolchain/elftoolchain/work/elftoolchain-0.7.1/libelftc/make-toolchain-version: line 74: svnversion: command not found
  ERROR: cannot determine a revision number.
```
But I'll push a fix for that to this PR in a moment.

raimue:
Looks good to me! Thank you for the extensive work on the port.

PR #2587, opened 2018-09-16
github portgroup: add archive download source

herbygillot:
A non-trivial amount of Github repositories generate **archive** URLs for their releases.

This is an issue when trying to use the `github` portgroup because there is no automatic redirection happening between the "modern" **releases** URL and the **archive** one.

This PR adds a new option for the `github.tarball_from` parameter, `archives`, that allows port authors to properly download from repositories like these when using this portgroup.

Related Trac tickets:
- https://trac.macports.org/ticket/47895
- https://trac.macports.org/ticket/40518

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G2307
Xcode 9.4.1 9F2000

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL?

- [x] tried a full install with `sudo port -vst install`?

PR #11505, opened 2021-07-02
sslscan: update to 2.0.10

herbygillot:
###### Tested on

macOS 10.15.7 19H1217 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/net/sslscan/Portfile
+++ b/net/sslscan/Portfile
@@ -3,7 +3,7 @@
 PortSystem 1.0
 PortGroup github 1.0
 
-github.setup    rbsec sslscan 2.0.9
+github.setup    rbsec sslscan 2.0.10
 revision        0
 categories      net
 maintainers     {raimue @raimue} \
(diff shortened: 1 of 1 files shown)


net/sslscan/Portfile
```diff
@@ -35,3 +35,5 @@ build.args      PREFIX="${prefix}" \
                 LDFLAGS="[get_canonical_archflags ld] -L${prefix}/lib"
 
 destroot.args   PREFIX="${prefix}"
+
+github.tarball_from archive
```
raimue, line 39: Hm, what is the reason for this change? The default 'tarball' seems fine to me?

herbygillot: The `tarball` URL is actually legacy, and though Github still supports it, it's considered deprecated.  So as projects are updated, we convert them to use `archive` or `release` URLs instead of the default.  Then perhaps at some point, the `github` portgroup's `tarball_from`option can be made to default to `archive`.

More discussion here:
- https://github.com/macports/macports-ports/pull/2587
- https://trac.macports.org/ticket/40518

raimue: This is still the default in the port group and it works for many ports.

If we want to get rid of `tarball`, I would rather change it in the port group. If there are concerns that it could break ports, then make a github-1.1 port group with the changed default and update the ports. But setting this individually in each port kind of goes against the purpose of a port group in my opinion.

herbygillot: Alright, removed.

raimue:
Thank you! 👍

herbygillot:
Thanks @raimue

PR #11519, opened 2021-07-05
autorestic: new port

manojkarthick:
#### Description

Autorestic is a wrapper around restic for making backing up to multiple locations easier. This port cannot be built without downloading deps during the build phase. Ref: https://trac.macports.org/ticket/61192

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.4 20F71 x86_64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/sysutils/autorestic/Portfile
+++ b/sysutils/autorestic/Portfile
@@ -1,0 +1,38 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/cupcakearmy/autorestic 1.1.1 v
+revision            0
+
+categories          sysutils
+maintainers         {gmail.com: smanojkarthick @manojkarthick} \
+                    openmaintainer
+license             Apache-2
+
+description         High level CLI utility for restic
+
+long_description    Autorestic is a wrapper around the amazing restic. \
+                    While being amazing the restic cli can be a bit overwhelming \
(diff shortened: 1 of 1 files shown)


sysutils/autorestic/Portfile
```diff
@@ -0,0 +1,39 @@
...
+                        sha256  19779b2582dd871f2a0f6707dc7054f0a96b283d5ef6031bc1d6d7625160f5c2 \
+                        size    186465
+
+# FIXME: This port currently can't be built without allowing go mod to fetch
+# dependencies during the build phase. See
+# https://trac.macports.org/ticket/61192
+build.env-delete    GOPROXY=off GO111MODULE=off
+
+depends_build-append \
+                    port:restic
```
herbygillot, line 35: Looks good @manojkarthick, but if `autorestic` is a wrapper around `restic`, then it seems like this should be `depends_run` or `depends_lib` instead of `depends_build`.

`depends_build`: dep needed only at build time.  The port will install without requiring this dependency be present.

`depends_run`: dep not needed at build time, but needed when the port is activated/installed.

`depends_lib`: combination of both `build` and `run`.  (ie., a binary may need a dep, like a library, to be present both when it is being built *and* when it is being run).

manojkarthick: Thank you @herbygillot ! - It should be `depends_run`, I've updated it.

herbygillot: OK great.  You should also be aware that the `-append` suffix works for all the depend types I listed. You should probably use the `-append` style by default every time as a precaution, as without it, you could be resetting what might have already been set by a portgroup or other method.  This looks fine though, so merging.

manojkarthick: I wasn't aware of that, thank you for that info!

PR #11540, opened 2021-07-09
fish: update to 3.3.1

szhorvat:
#### Description

Update fish to 3.3.1.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G9216 x86_64
Xcode 11.3.1 11C504

###### Verification 
Have you

### DIFF
--- a/shells/fish/Portfile
+++ b/shells/fish/Portfile
@@ -5,11 +5,11 @@ PortGroup               github 1.0
 PortGroup               cmake 1.1
 PortGroup               legacysupport 1.0
 
-github.setup            fish-shell fish-shell 3.3.0
+github.setup            fish-shell fish-shell 3.3.1
 revision                0
-checksums               rmd160  8bdb6c6951ae5b0ab09628e9bd17959eca7d837e \
-                        sha256  a4215e4cab2a5b101b0b8843720bda3c7eb98e8a14dca0950b8ef17e94282faa \
-                        size    3489124
+checksums               rmd160  9ec5b3ee590427a9e7d7b4691da13970385ded0f \
+                        sha256  b5b4ee1a5269762cbbe993a4bd6507e675e4100ce9bbe84214a5eeb2b19fae89 \
+                        size    3489152
 use_xz                  yes
 
 name                    fish


szhorvat:
The `sigint` test, which was disabled by @tsabirgaliev recently, still fails.

The `type` test, which failed in 3.3.0, still fails. I did not disable it though.

tsabirgaliev:
@szhorvat what about tmux-complete.fish test? does it fail on your machine too?

tsabirgaliev:
@szhorvat BTW type test doesn't fail on my machine (Macos 11.4)

szhorvat:
> @szhorvat what about tmux-complete.fish test? does it fail on your machine too?

Yes, it does. `tmux-prompt` fails too. I did not notice this because I did not have `tmux` installed, so the test was skipped.

> @szhorvat BTW type test doesn't fail on my machine (Macos 11.4)

I am on 10.14. After installing the `realpath` port, now the test passes. Before it failed with:

```
  The CHECK on line 41 wants:
    function realpath --description {{.+}}

  which failed to match line stdout:12:
    function realpath --wraps=grealpath --description \'print the resolved path [command grealpath]\'
```

I think this is harmless.

szhorvat:
BTW regarding the disabled test: would it not be better to note the failure in the portfile instead of disabling the test?

tsabirgaliev:
@szhorvat 
1. I have fixes for tmux tests, not sure how add them to current PR. What do you think?
2. May be add realpath as test dep?

szhorvat:
@tsabirgaliev If you want to open a new PR where you do the update and make the fixes, feel free to go ahead. I will then close this PR. I am a rather naive fish user. I enjoy its user-friendliness but I don't know how to use its advanced features.

In fact I would prefer that you do that, so I'll close this right away. If you don't want to, let me know and I'll re-open this PR.

tsabirgaliev:
> BTW regarding the disabled test: would it not be better to note the failure in the portfile instead of disabling the test?

@szhorvat I don't really know :) didn't find a better way other than disabling.

Where exactly you would note that in the Portfile?

PR #11587, opened 2021-07-13
ConvertAll: New portfile submission. 

khorton:
ConvertAll: New portfile submission.  First submission by Kevin Horton

I have never created a Portfile before, and am a git newbie, so make no assumptions

###### Type(s)

- [x] enhancement

###### Tested on
macOS 11.4 20F71 x86_64
Xcode 12.5.1 12E507

###### Verification 
Have you


math/convertall/Portfile
```diff
@@ -0,0 +1,45 @@
...
+PortGroup           github 1.0
+PortGroup           python 1.0
+PortGroup           qt5 1.0
+
+github.setup        doug-101 ConvertAll 0.8.0 v
+github.tarball_from releases
+
+python.versions     34 35 36 37 38 39
+python.default_version     38
+
```
kencu, line 13: we'll want to default 39 now

math/convertall/Portfile
```diff
@@ -0,0 +1,45 @@
...
+qt5.min_version     5.4
+
+name                convertall
+version             0.8.0
+license             GPL-2+
+categories          math science
+platforms           darwin
+maintainers         {@khorton kilohotel.com:kevin01}
```
kencu, line 21: if you add ```openmaintainer``` you allow others to make minor changes to things. We usually do that unless it's a critical port of some kind.

math/convertall/Portfile
```diff
@@ -0,0 +1,45 @@
...
+
+depends_lib-append  port:py${python.version}-pyqt5
+
+use_configure       no
+build.cmd           ${python.bin} install.py
+build.args          -p ${prefix} \
+                    -b ${destroot}
+
+destroot {}
+test {}
```
kencu, line 40: you don't actually need to disable this phase -- if you haven't specifically enabled a test phase, then there is none by default.

math/convertall/Portfile
```diff
@@ -0,0 +1,45 @@
...
+                    size    281055
+
+depends_lib-append  port:py${python.version}-pyqt5
+
+use_configure       no
+build.cmd           ${python.bin} install.py
+build.args          -p ${prefix} \
+                    -b ${destroot}
+
+destroot {}
```
kencu, line 39: this is kind of odd. No destroot phase? How does it install anything ?

kencu: oh, perhaps it installs during the build.cmd phase? I see it runs install.py there.

math/convertall/Portfile
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
```
kencu, line 2: Nice looking job for your first Portfile! Thanks for taking the leap.

math/convertall/Portfile
khorton: I don't see an easy way to split the install.py actions into build and destroot phases, so I propose to leave the portfile as-is

math/convertall/Portfile
khorton: Thanks - I hope to slowly add more Portfiles, as I stumble across tools I use that aren't in MacPorts, or see one that has no maintainer

math/convertall/Portfile
khorton: Removed line that disabled the test phase

kencu:
Regarding squashing your commits -- you did the first thing right by working on a branch.

So on your branch, in your git repo, you do a ```git log``` and see how many commits you have made so far after you branched off master. Let's say it's 6 commits.

Then you do this ```git rebase -i HEAD~6```.

A list of all your commits comes up. You can do various things here, but for this process, you just want to squash them all into the first commit. Mark them with an 's'. The one commit you want to keep, you mark with a `p`.  Then when you save that, git starts condensing them all into one commit.

Once it's done, you have one commit with all your bits and pieces in it. Then "force push" that into this PR to make it one commit.

Google is your friend -- there are hundreds of walkthroughs on how to do this. Forget about the Git application -- it's just not useful IMHO for anything like this.

Try it on something disposable before you do it on this PR, so you get the idea.

Look back at the commit history a bit to see what a commit message is supposed to look like.

math/convertall/Portfile
```diff
@@ -0,0 +1,44 @@
...
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           python 1.0
+PortGroup           qt5 1.0
+
+github.setup        doug-101 ConvertAll 0.8.0 v
+github.tarball_from releases
+
+python.versions     34 35 36 37 38 39
+python.default_version     39
```
harens, line 12:
```suggestion
```
This should default to the latest version automatically.

math/convertall/Portfile
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           python 1.0
+PortGroup           qt5 1.0
+
+github.setup        doug-101 ConvertAll 0.8.0 v
```
harens, line 8:
```suggestion
github.setup        doug-101 convertall 0.8.0 v
```

This then allows us to remove the `name` attribute.

math/convertall/Portfile
```diff
@@ -0,0 +1,44 @@
...
+github.tarball_from releases
+
+python.versions     34 35 36 37 38 39
+python.default_version     39
+
+qt5.min_version     5.4
+
+name                convertall
```
harens, line 16:
```suggestion
```

math/convertall/Portfile
```diff
@@ -0,0 +1,44 @@
...
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           python 1.0
+PortGroup           qt5 1.0
+
+github.setup        doug-101 ConvertAll 0.8.0 v
+github.tarball_from releases
+
+python.versions     34 35 36 37 38 39
```
harens, line 11:
```suggestion
python.versions     38 39
```

For now, I'd stick with 38 and 39. New versions can then be added on later.

See https://github.com/macports/macports-ports/pull/8620#issuecomment-703742167 for more info.

math/convertall/Portfile
```diff
@@ -0,0 +1,44 @@
...
+use_configure       no
+build.cmd           ${python.bin} install.py
+build.args          -p ${prefix} \
+                    -b ${destroot}
+
+destroot {}
+
+post-patch {
+    reinplace "s|/usr/local|${prefix}|g" ${worksrcpath}/install.py
+    reinplace "s|buildRoot = '/'|buildRoot = '${workpath}/destroot'|g" ${worksrcpath}/install.py
```
harens, line 43: I think this would make more sense as a [patch](https://guide.macports.org/#development.patches.source), since you're only replacing two constants.

Let me know if you need any help with this, and I'd be happy to assist :)

harens (commented):
Thank you @khorton for your first contribution! 🎉 You've done a great job here. I've left a few minor comments below.

Regarding squashing commits, @kencu has put some really great advice. Since you're only adding one port, whoever merges this can very easily [squash the commits for you](https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits), so I wouldn't worry too much about it if you're not able to do it.

Thanks again for adding this port @khorton!

math/convertall/Portfile
harens: I'm willing to agree with @kencu here. I'm not sure how yet, but we definitely need a destroot stage here. Otherwise things start to breakdown. 

As an example, I'm not sure how you would install buildbot binaries of this port, and surely the destroot dir wouldn't even exist during the build stage?

Maybe it's something to raise upstream, since they have a very unconventional install method. Alternatively, we could try to bypass the script entirely and write our own build stage based on the script.

khorton: Would it be OK to move the install.py actions into the destroot phase, and declare an empty build phase?  

The docs show an example of an empty build phase in [section 4.3.4](https://guide.macports.org/#development.examples.eliminate), so hopefully this doesn't break the buildbot.

ConvertAll seems to not be actively updated now.  The last commit was over a year ago.  His build system apparently works correctly on Linux, so I doubt he would be interested in revising it to fit MacPorts requirements. 

Writing a custom build phase is a bit more than I was hoping to tackle for my first Portfile

math/convertall/Portfile
khorton: I've reviewed the MacPorts docs related to patches, and looked at a few existing ports, but I must admit I'm not yet understanding the advantage of a patch over using a post-patch reinplace here.  If I understand correctly, I shouldn't hard code /opt/local into a patch.  The patch would need to replace the existing /usr/local with some placeholder, then I would need a post-patch reinplace to replace that placeholder with ${prefix}.  Or, is MacPorts smart enough to replace a hard coded /opt/local in a patch with the correct PREFIX?

I must be missing something here.  Please educate me.

harens: > If I understand correctly, I shouldn't hard code /opt/local into a patch. The patch would need to replace the existing /usr/local with some placeholder, then I would need a post-patch reinplace to replace that placeholder with ${prefix}.

Sorry about that @khorton, you're completely right. I didn't think it through properly, so that mistake is on me :)

khorton:
It isn't clear to me whose court the ball is in.  Is there something I need to do to move this forward, or is everyone else busy and I simply need to be patient.  Thanks - Kevin

harens:
> It isn't clear to me whose court the ball is in. Is there something I need to do to move this forward, or is everyone else busy and I simply need to be patient? Thanks - Kevin

I think this port looks nearly good to go, the only thing left is the build script. My first two thoughts were to either patch it/rewrite the build section, but as you mention that's not ideal.

Alternatively, maybe we can set the build script's destroot path to a new folder in `worksrcpath`, and during the destroot stage, just copy everything across from that folder?

CC @reneeotten who might have a better idea.

Either way, thank you for chasing me up on this @khorton and for all your work on this portfile. You've done an amazing job. 👍

math/convertall/Portfile
```diff
@@ -0,0 +1,42 @@
...
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           python 1.0
+PortGroup           qt5 1.0
+
+github.setup        doug-101 convertall 0.8.0 v
+github.tarball_from releases
+
+python.versions     38 39
```
harens, line 11:
```suggestion
python.default_version 39
```

Just remembered that since this is an application, it will always default to python39. `python.versions` is normally used for python modules that can be imported in different versions of python.

See https://guide.macports.org/#reference.portgroup.python.keywords

khorton:
I'll be on the road for 8 days starting on July 17th.  I'll attack the build script after I return, unless someone suggests a better approach by then.

math/convertall/Portfile
```diff
@@ -0,0 +1,42 @@
...
+PortGroup           qt5 1.0
+
+github.setup        doug-101 convertall 0.8.0 v
+github.tarball_from releases
+
+python.default_version 39
+
+qt5.min_version     5.4
+
+version             0.8.0
```
ryandesign, line 15: You can remove this version line; the github.setup line sets the version for you.

math/convertall/Portfile
ryandesign: As long as the files to be installed end up in the destroot directory by the time that the destroot phase is over, that's fine. Typically a build would happen in the build phase and the built files would be copied into the destroot directory in the destroot phase. But if this build system does not easily accommodate splitting those tasks it is acceptable for both tasks to occur in a single phase (either build or destroot) and for the other one to be made a no-op. The [mongodb Portfile](https://github.com/macports/macports-ports/blob/d4bd2a5edc85e9aeb307eb7f8e9d58a5a2f582fc/databases/mongodb/Portfile#L136-L138) is an example of a port that already does that. My feeling today is that in this case having both build and destroot tasks take place in the build phase is preferable, since users usually expect the build phase to take longer than the destroot phase. (MacPorts also adds a `-j` flag to enable parallel building in the build phase but not in the destroot phase, though that is not relevant to this port since its build command is not `make`.) In the mongodb port, the build installs some things to the wrong places within the destroot directory, so I use a very brief destroot phase to clean that up.

math/convertall/Portfile
ryandesign: > If I understand correctly, I shouldn't hard code /opt/local into a patch. The patch would need to replace the existing /usr/local with some placeholder, then I would need a post-patch reinplace to replace that placeholder with ${prefix}.

Yes, that is a pattern you will see used by many ports.

P.S: Use `${destroot}` here, not `${workpath}/destroot`.

```suggestion
    reinplace "s|/usr/local|${prefix}|g" ${worksrcpath}/install.py
    reinplace "s|buildRoot = '/'|buildRoot = '${destroot}'|g" ${worksrcpath}/install.py
```

ryandesign:
Looks like the reason why the `-p` and `-b` flags you were sending to install.py were not having any effect is because the python portgroup by default sets `build.target` to `build[python_get_defaults jobs_arg]` which install.py wasn't expecting. I guess it stopped processing arguments after that. The solution is to clear `build.target`. Then the post-patch block can go away.

You refer to the repository in the github.setup line as "convertall" but the repository name is actually "ConvertAll". Although GitHub currently allows case-mismatched access, it's best to specify case correctly. That would change the port name capitalization to "ConvertAll" as well. We could either leave it that way or specify `name convertall` to keep the port name lowercase.

When we change the repo name, the distname won't be right anymore, since the distname should indeed be lowercase, so that'll need to be fixed.

The distfile extracts to a directory called just "ConvertAll" without a version number. Although MacPorts will handle that automatically now by creating a symlink for you when worksrcdir is wrong, I prefer to set worksrcdir correctly. It seems cleaner to me to have just the extracted directory in the work dir, rather than both an extracted directory and a symlink.

I mentioned in an earlier comment that, since install.py cannot separate build from install, I thought doing everything in the build phase was best. I've changed my mind. Because install.py doesn't really build anything—all it does is install things—it would be best to do it in the destroot phase.

The fact that nothing actually gets built, files just get copied, was a clue to the likelihood that this port does not install any architecture-specific files. That being the case, this needs to be indicated via the line `supported_archs noarch`.

We're also getting into the habit of always adding the `revision` line, even when it is 0, so I'll add that too.

Since you mentioned you'll be away, I hope you don't mind if I make the various corrections myself.

PR #11598, opened 2021-07-14
libgcrypt: Support for Tiger

evanmiller:
Closes https://trac.macports.org/ticket/62431

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/devel/libgcrypt/Portfile
+++ b/devel/libgcrypt/Portfile
@@ -34,6 +34,10 @@ depends_lib     port:libgpg-error
 patchfiles      patch-configure.diff \
                 patch-random-rndlinux.c.diff
 
+platform darwin 8 {
+    patchfiles-append patch-tiger-random-rndlinux.c.diff
+}
+
 configure.args  --disable-asm
 
 # Build fix for compilers that default to c99 (clang)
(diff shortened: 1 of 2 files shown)


kencu:
I have to push my Availabilty.h fix for Tiger, or we'll be doing 1000 of these...

kencu:
OK, it's merged.

https://github.com/macports/macports-legacy-support/commit/221ec6c352b144a06c28a91e27e82773f8d7a6c1

I will make up a new update to legacy-support-devel that copies these two headers into the right place for use for Tiger only, and we'll see if that helps this.

Rather than patching ports individually, we should be able to just add
```
PortGroup legacysupport 1.1
```
like always, and (if we are at all lucky) that will be the end of this particular Tiger headache.

kencu:
OK, done. 

https://github.com/macports/macports-ports/commit/b4bd782caf77c8c32017905d5e4de057d534fc84

please, if you would Evan, install ```legacy-support-devel``` instead of legacy-support and help me kick the tires of this new addition. I believe it has a high probability of being all good with no bad, but not much in life works out that way.

This update to legacy-support-devel does fix the build of this port, libgcrypt, without your added patch here, and ideally will save us all lots of troubles.

My goal in general here is to make the MacOSX10.4u.sdk, with the additions of legacy-support, look as close as it can to the SDK for 10.12 to software that want to build against it, so that software that can build against the 10.12 SDK will build all the way back to the 10.4u SDK without any modifications or even knowing they are doing so.

PR #11603, opened 2021-07-15
Update Portfile

asvitkine:
#### Description

Fix openal-soft build on Mac OS X 10.6.

###### Type(s)

- [X] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 10K549 i386
Xcode 3.2.3 10M2262

###### Verification 
Have you

### DIFF
--- a/audio/openal-soft/Portfile
+++ b/audio/openal-soft/Portfile
@@ -59,6 +59,12 @@ configure.args-append   -DALSOFT_EXAMPLES=OFF \
                         -DALSOFT_REQUIRE_SDL2=OFF \
                         -DALSOFT_BACKEND_SDL2=OFF
 
+if {${os.platform} eq "darwin" && ${os.major} < 11} {
+  configure.cxx_stdlib macports-libstdc++
+  configure.args-append -DCMAKE_CXX_FLAGS="-D__STDC_FORMAT_MACROS -stdlib=macports-libstdc++" \
+                        -DCMAKE_SHARED_LINKER_FLAGS="-stdlib=macports-libstdc++"
+}
+
 variant gui description {Enable alsoft-config Qt GUI} {
     configure.args-replace  -DALSOFT_NO_CONFIG_UTIL=ON -DALSOFT_NO_CONFIG_UTIL=OFF
 }


kencu:
This approach is not likely to be the right one. Forcing the gcc libstdc++ is not going to be a good plan in the end I believe, even if it does build.

Why don't we open up a ticket for what is going wrong when building it against libc++, and see if we can fix it properly?

asvitkine:
> This approach is not likely to be the right one. Forcing the gcc libstdc++ is not going to be a good plan in the end I believe, even if it does build.
> 
> Why don't we open up a ticket for what is going wrong when building it against libc++, and see if we can fix it properly?

The gcc libstdc++ isn't the relevant bit, it's the macports one - which provides thread-local-storage support for C++, which is missing on 10.6 otherwise.

The build fails with:

```
:info:build Undefined symbols for architecture x86_64:
:info:build   "___cxa_thread_atexit", referenced from:
:info:build       (anonymous namespace)::ThreadCtx::~ThreadCtx() in alc.cpp.o
:info:build       ALCcontext::deinit() in alc.cpp.o
:info:build       GetContextRef() in alc.cpp.o
:info:build       _alcGetCurrentContext in alc.cpp.o
:info:build       _alcGetThreadContext in alc.cpp.o
:info:build       _alcMakeContextCurrent in alc.cpp.o
:info:build       _alcSetThreadContext in alc.cpp.o
:info:build       ...
```

My understanding is that's the whole purpose of `macports-libstdc++` per:
https://lists.macports.org/pipermail/macports-users/2018-February/044716.html

The secondary -D__STDC_FORMAT_MACROS thing is required because otherwise we get a PRIu64 undefined error when building with `macports-libstdc++`.

Let me know if you prefer that I open a ticket instead - I was going by the contributor instructions which suggests that a github PR is the easiest way to get something fixed.

By the way, the openal-soft macports status page shows that it indeed fails on 10.6. As a meta question, would it make sense for there to automatically be a ticket for every failing build that's already known to macports?

kencu:
Ah, I know this error.

I have fixed the libcxx for 10.4, 10.5, and 10.6 to have full thread_local_storage support, but the problem here is that at present, we cannot build the full, thread_local_storage enabled libcxx in one go from bare system roots. I am working on an approach to allow that, but for now, do this:

1. install clang-9.0 (you have it already).
2. now rebuild libcxx with the full TLS capability:
```sudo port -v -n upgrade --force --enforce variants libcxx +emulated_tls```
3. now try installing openal-soft with a fully capable libcxx and you should be good to go.

kencu:
Thread_local_storage support once upon a time was the underlying reason for macports-libstdc++, but we realized that approach was overall doomed because all software on macOS is geared towards building with clang and against libc++. So I came up with a way to add TLS to libc++ on < 10.7.

It works better, and very nicely, on 10.5 and 10.6. It also working on 10.4 Intel as well, but it requires a step that I have not as yet added to MacPorts (as there is a stepping-stone needed that I am in the process of working on:

https://lists.macports.org/pipermail/macports-dev/2021-July/043633.html

asvitkine:
Thanks for the clarification, glad to hear it's being worked on.

Does it make sense to still merge this but add a TODO to clean this up once the above work materializes? I'm trying your suggested workaround right now.

kencu:
No, merging this would be a disaster. macports-libstdc++ and libc++ do not play nice together.

I'll close this. If you have any further trouble, please go ahead and open a ticket and we'll work through it.

We already have an open ticket to make the full libxx automatic.

Years ago I volunteered to install a full-capable premade binary on 10.4, 10,5, and 10.6 but that was not allowed, so we have this problem. But I finally have a workable plan to fix it, I believe, as per that email message i referenced.

asvitkine:
> No, merging this would be a disaster. macports-libstdc++ and libc++ do not play nice together.

Oh is it because other ports that use this library might use libc++ and we don't want the two to mix? Is it only a problem if the library exposes a C++ API - or just having the two std libs in the same process is enough to cause issues?

In any case, your suggested workaround seems to have worked. Thanks!

kencu:
Yep, you got it 100% right there.

PR #11625, opened 2021-07-19
libjpeg-turbo: Support non-Altivec PowerPC

evanmiller:
The CMake file supplied by libjpeg-turbo 2.1.0 incorrectly detects SIMD support on G3 processors. As a result, dynamic libraries are produced which will not load on the host machine. This patch forces SIMD to be disabled on G3 processors.

Closes: https://trac.macports.org/ticket/63259

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/graphics/libjpeg-turbo/Portfile
+++ b/graphics/libjpeg-turbo/Portfile
@@ -38,6 +38,7 @@ checksums           sha1    6bf63c869105d341011cd4915816de888338231a \
                     sha256  bef89803e506f27715c5627b1e3219c95b80fc31465d4452de2a909d382e4444 \
                     size    2255497
 
+configure.args      -DWITH_JPEG8=ON
 if {![info exists universal_possible]} {
     set universal_possible [expr {${os.universal_supported} && [llength ${configure.universal_archs}] >= 2}]
 }
@@ -50,8 +51,12 @@ if {${universal_possible} && [variant_isset universal]} {
 } elseif {${configure.build_arch} in {i386 x86_64}} {
     depends_build-append port:nasm
     configure.env       ASM_NASM=${prefix}/bin/nasm
+} elseif {${configure.build_arch} eq "ppc"} {
+    if {[catch {sysctl hw.vectorunit} result] || $result == 0} {
+        configure.args-append   -DWITH_SIMD=OFF
+        archive_sites
+    }
 }
-configure.args      -DWITH_JPEG8=1
 
 variant java description "Add Java support" {
     PortGroup           java 1.0


kencu:
We wouldn’t want to turn off Altivec for the 99% of users who have it, to support one person on a G3.

Instead, I mentioned there are Portfiles in the repo that test for Altivec using sysctl now. We want to do what they do, and only turn it off if the user doesn’t have it.

kencu:
There is a good example how to do this properly in the ffmpeg Portfile.

kencu:
something like this:
```
platform powerpc {
    # absence of altivec is not automatically detected
    if {[catch {sysctl hw.vectorunit} result] || $result == 0} {
        configure.args-append   -DWITH_SIMD=OFF
    }
}
```

I don't know about the other part of your PR here yet.

evanmiller:
@kencu Thanks for the feedback. Do G4s have a vector unit? My understanding was that Altivec is only available on `ppc64` (i.e. G5 processors).

kencu:
the G4/74xx series was the first to have Altivec, I believe. For these older systems we like to squeak all the work out of the cpu cycles we can, ergo the test-before-disable is desirable.

Thanks!

evanmiller:
@kencu Please see the latest commit. I can squash and force-push if needed.

kencu:
that looks great. I'm not near my PowerPC systems at present. As that has been in ffmpeg forever, I assume it works. Does that indeed do the trick for your G3? By any hope, do you have a G4 or G5 to show it does the right thing there too?

If you don't have a G4 or G5 then that is fine, we can merge and push this and I can tweak it later if needed.

evanmiller:
@kencu It configures correctly on my G3. I do not have a G4/G5 to test with.

kencu:
We’ll wait the required interval for MacPorts maintainer timeout and then commit if no further input. Thanks!

ryandesign:
> ```
> platform powerpc {
>     # absence of altivec is not automatically detected
>     if {[catch {sysctl hw.vectorunit} result] || $result == 0} {
>         configure.args-append   -DWITH_SIMD=OFF
>     }
> }
> ```

This means the port builds differently depending on the user's computer, which we don't usually want to do in MacPorts. (The ffmpeg port that this method was copied from is wrong in this regard too.) We don't currently have binaries for Tiger, but if we did, and if our Tiger PowerPC build machine had a G4 or better, then the binary we shipped would be built for G4 and would not work on a G3.

Usually we like to express build differences as variants. For example we might define a variant called "g3" that turns off the G3-incompatible SIMD optimizations (with optimizations on being the default if the variant is not selected). Or we could invert it and define a variant called "g4" that turns them on (and make the default that optimizations are off unless the variant is selected). Whichever we decide, code like the above could be used to decide whether to enable or disable the variant by default. If we use a variant name "g3" or "g4" which is PowerPC-specific we would have to decide how to handle non-PowerPC systems. Defining the variants only on PowerPC systems is one possibility, though the buildbot mirroring script doesn't like ports that define variants conditionally. A more platform-neutral variant name might be "simd"; it could be turned on by default on all systems except G3s.

It would be terribly nifty if MacPorts could just build it twice—once for arch ppc without SIMD optimizations for G3s and once for arch ppc7400 with SIMD for G4s and G5s—and ship it as a universal binary. MacPorts does have support for universal building but it was never designed for building for two different 32-bit PowerPC architectures and there is probably not a great deal of interest in adding such support at this late date.

ryandesign:
There's also the matter of universal builds. `platform powerpc` only applies when building *on* a PowerPC. What happens when you build universal for i386 / ppc on a PowerPC Mac vs. what happens when you build universal for i386 / ppc on an Intel Mac?

What you probably meant is for `-DWITH_SIMD=OFF` to be used when building *for* PowerPC (regardless what system you're building *on*). So you need to add that flag only for the ppc part of the build. Doing that requires using the muniversal portgroup; fortunately, this port already does.

kencu:
that's what Evan had at first and I chased him away from it.

I would not like to cripple most systems for the extremely rare G3 users.

As you say there is no ppc Tiger buildbot. Leopard requires a G4. So this is not an issue really, but to future-proof it you someday put up a ppc Tiger buildbot we can just disable archive sites so the user has to build it.

ppc needs every Altivec enhancement it can get!

evanmiller:
"The little G3 that could" has raised a more complicated issue than I expected! I'm treating it as a recreational coding machine, so I'd be open to working on a universal ppc + ppc7400 build in the future.

If you want to take the variant approach for now, I'd suggest a name along the lines of `simd`, `nosimd`, or `scalar`. The `WITH_SIMD=OFF` build option in libjpeg-turbo can be applied to any architecture, not just PowerPC.

evanmiller:
Here's a quick implementation of a `no_simd` variant in a few ports, removing the problematic run-time CPU checks: https://github.com/evanmiller/macports-ports/commit/633979a9bcd87dde25d9e9e4bfa3afd5f1d229e2

My thinking is that we happy few G3 users can just add `+no_simd` to our variants.conf and avoid fighting with the G4/G5 gang in the future.

mojca:
Using negative variants has been considered an antipattern for years now. The idea is to have `simd` and have it turned on by default where supported.

kencu:
The way you would do this is to have runtime Altivec detection still, but have that detection enable the variant that turns on Altivec.

graphics/libjpeg-turbo/Portfile
```diff
@@ -46,8 +47,11 @@ if {${universal_possible} && [variant_isset universal]} {
 } elseif {${configure.build_arch} in {i386 x86_64}} {
     depends_build-append port:nasm
     configure.env       ASM_NASM=${prefix}/bin/nasm
+} elseif {${configure.build_arch} eq "ppc"} {
+    if {[catch {sysctl hw.vectorunit} result] || $result == 0} {
```
jmroot, line 51: This isn't correct if binary distribution can occur (the sysctl result on the build machine could easily be different to that on the host machine), but of course we don't build binaries for ppc. Someone else could though.

graphics/libjpeg-turbo/Portfile
```diff
@@ -46,8 +47,11 @@ if {${universal_possible} && [variant_isset universal]} {
...
+        configure.args-append   -DWITH_SIMD=OFF
```
jmroot, line 52: Also needs to go in `merger_configure_args(ppc)` in the universal case.

kencu:
the presented solutions noted above are all OK...disable the (nonexistant) buildbot by setting archive_sites to nothing, or toggle a variant (perhaps named +ppcg3 to be a positive rather than a negative thing).

Just please don't disable Altivec across the bosrd for the 99% of users who have it.

evanmiller:
Here is a solution following @mojca’s comment, adding a `simd` variant and disabling by default on G3:

https://github.com/evanmiller/macports-ports/commit/95745c3636e6d09caddba03281d632d371c4e768

If it looks OK to the parties concerned, I will open a separate pull request.

kencu:
the test you have put in to override the simd detection built in to libjpeg-turbo is set to affect every system of every kind now, not just the powerpc systems.

I'm not sure every processor with simd instructions (which I think is all of them except a G3 ppc) passes that sysctlt test, and for the future, who knows?

So whatever is done to support G3 powerpc has to be set up to leave everything else alone.

By the way, any idea why the existing Altivec test in the simd CMakeLists.txt is apparently passing for you, even though you don't have Altivec? They are going to some lengths to do this right already, and would be likely interested in fixing this if it is broken somehow.

I continue to fell that the better approach here,  is to use runtime detection (best by the build itself, but in the Portfile if the build detection is broken) but disable the ( nonexistant) buildbot by clearing archive_sites instead of these variants, which add confusion.

evanmiller:
@kencu You can see how libjpeg-turbo tests for Altivec support here:

https://github.com/libjpeg-turbo/libjpeg-turbo/blob/e6e952d5d5db6c0f9c505854995988ad13b30ec8/simd/CMakeLists.txt#L410-L419

If it can compile Altivec code using the `-maltivec` flag, the test passes. However, this compilation succeeds on G3 machines, even when setting other flags such as `-mcpu=G3`. So technically it appears to be a bug in GCC.

Since getting a bug fixed in GCC will take a while to make its way to MacPorts, I can modify the sysctl test to run only on PowerPC. I will note that `hw.vectorunit = 0` on the Apple M1 machine on which I'm typing – so you're right, the test shouldn't be relied upon on other platforms.

graphics/libjpeg-turbo/Portfile
evanmiller: Can you clarify this comment or point me to an example?

kencu:
i looked at that cmake test. It should fail on your G3. Is is somehow passing? If it is, that should be fixed to fail, if you don't have Altivec, as that is the point of it (altho no doubt nobody has tested this path in a decade or more).

Bug in gcc? Have to see...

A G3 can probably "compile" code that needs Altivec, just not run it...their test might need tweaking.

Many ports disable archive_sites to force a build on the native machine. This continues to be my preference here instead of variants, to get past the (nonexistant) buildbot issue.

evanmiller:
Yes, the Altivec test passes on a G3. GCC emits ppc7400 code when `-maltivec` is set, regardless of other flags.

If the libjpeg-turbo test attempts to run the code, rather than just compile, it will break cross-compilation.

kencu:
we don't care about cross-compilation here...we want to build what the build machine can run.

kencu:
Would you have a moment to change:
```
https://github.com/libjpeg-turbo/libjpeg-turbo/blob/e6e952d5d5db6c0f9c505854995988ad13b30ec8/simd/CMakeLists.txt#L412
``` 
from  ```check_c_source_compiles``` to ```check_c_source_runs```?

That should fix it to properly do the test, that can be sent upstream, and we can keep that patch in the Portfile until they do.

kencu:
If someone is still worried about the possibility of a buildbot someday existing, we can do this:
```
platform darwin powerpc {
    # clear archive sites so this is built on the user's system
    archive_sites
}
```

evanmiller:
@kencu I'm afraid the `check_c_source_runs` change cannot be sent upstream. It will break the build for anyone trying to cross-compile libjpeg-turbo for PowerPC (e.g. Intel -> PowerPC builds). We're not the only cross-compilers!

I guess it's debatable, but the fact that `gcc -mcpu=G3 -maltivec` produces code that does not actually run on a G3 seems like a problem to me. I would expect an error about conflicting flags.

I will wait for a consensus on the correct overall strategy before making additional commits. For anyone interested, there is a similar issue with `-maltivec` incorrectly enabled on clang-3.4 that I have reported here: https://trac.macports.org/ticket/63304

kencu:
> @kencu I'm afraid the `check_c_source_runs` change cannot be sent upstream. It will break the build for anyone trying to cross-compile libjpeg-turbo for PowerPC (e.g. Intel -> PowerPC builds). We're not the only cross-compilers!

Sure it can. I will send it. We are interested here in the code running on the build machine. Not cross compiling.

> I guess it's debatable, but the fact that `gcc -mcpu=G3 -maltivec` produces code that does not actually run on a G3 seems like a problem to me. I would expect an error about conflicting flags.

GCC will have been all through this, and they obviously decided maltivec wins. Or try them reverse order -- maybe last flag wins.


> I will wait for a consensus on the correct overall strategy before making additional commits. For anyone interested, there is a similar issue with `-maltivec` incorrectly enabled on clang-3.4 that I have reported here: https://trac.macports.org/ticket/63304

There will never be a consensus on an ancient processor for a system that almost nobody uses for a potential issue with a buildbot that will never exist.

evanmiller:
> Sure it can. I will send it. We are interested here in the code running on the build machine. Not cross compiling.

I think we are misunderstanding each other. Other consumers of libjpeg-turbo may want to cross-compile. I'm not talking about MacPorts folks. If someone downloads libjpeg-turbo with the proposed change onto an Intel machine and tries to cross-compile for PPC/Altivec, `check_c_source_compiles` will succeed, but `check_c_source_runs` will fail. So they will not get the SIMD/Altivec build that they desire. That's why I'm saying `check_c_source_runs` cannot be sent upstream.

evanmiller:
How difficult / insane would it be to define a separate `ppc7400` (i.e. Altivec) architecture in MacPorts? My thinking is that instead of reinventing the wheel with run-time detection and/or variants just use the existing architecture mechanism to select the correct build settings and provide better error messages on unsupported hardware. G4s would prefer this architecture to `ppc` the same way G5s currently prefer `ppc64`. Would such an approach be at all feasible?

kencu:
that would mean dozens or hundreds of ports and many portgroups would need to be altered, including the whole muniversal PG setup.

It is very disappointing this rabbit hole of variants was ever opened or mentioned. Just do the runtime detection that other ports do, clear archive_sites at the same time to cover the unlikely-to-ever-exist Tiger PPC buildbot, and lets move on.

You're two weeks in on this 10-second fix, with no end in sight and very little point to any of it.

kencu:
I'm thinking we use this, very similar to what we do in ffmpeg:
```
configure.args      -DWITH_JPEG8=ON

if {![info exists universal_possible]} {
    set universal_possible [expr {${os.universal_supported} && [llength ${configure.universal_archs}] >= 2}]
}
if {${universal_possible} && [variant_isset universal]} {
    if {"x86_64" in ${configure.universal_archs} || "i386" in ${configure.universal_archs}} {
        depends_build-append port:nasm
    }
    set merger_configure_env(i386)  ASM_NASM=${prefix}/bin/nasm
    set merger_configure_env(x86_64)    ASM_NASM=${prefix}/bin/nasm
} elseif {${configure.build_arch} in {i386 x86_64}} {
    depends_build-append port:nasm
    configure.env       ASM_NASM=${prefix}/bin/nasm
}

platform powerpc {
    # absence of altivec is not automatically detected
    if {[catch {sysctl hw.vectorunit} result] || $result == 0} {
        configure.args-append -DWITH_SIMD=OFF

        # as this machine is a G3, force a local build so we don't download a buildbot-built
        # version possibly built on a G4+ or cross-compiled from an Intel system
        archive_sites
    }
}
```

evanmiller:
@mascguy OK, opened #12597

mascguy:
> > I'm not an expert, but wouldn't it be possible to adapt the [init_simd](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/simd/powerpc/jsimd.c#L112) function to detect the presence of Altivec on PPC Macs at runtime, rather than at build time? In particular, [SDL detects Altivec at runtime](https://github.com/libsdl-org/SDL/blob/main/src/cpuinfo/SDL_cpuinfo.c#L320), so I'd assume it would be simple to do for libjpeg-turbo as well.
> 
> @ccawley2011 The issue is one of instruction sets. G3 machines cannot load a dylib that self-reports as `ppc7400`. I believe you'd need to have a separate Altivec dylib that is conditionally loaded at run-time after the initial detection takes place, or else some PPC assembly trickery. This may require significant changes to libjpeg-turbo.

Okay, this PR has been sitting here long enough. And these changes have been in-play for `libjpeg-turbo-devel` for a few weeks now, without issue.

@jmroot Apart from some minor disagreement related to precisely how we detect Altivec - and frankly, that shouldn't hold this up any longer - are you OK with merging?

mascguy:
(Evan, my force-pushes were simply to rebase against master. Otherwise, nothing was touched/changed.)

mascguy:
Evan, are there any other refinements you'd like to make, or are we ready to merge?

evanmiller:
@mascguy I'm fine with it as it stands. Since it will need a rev-bump to get the change out to my fellow G3 user(s)(?), it would be great to roll out the 2.1.1 release whenever it is deemed ready.

mascguy:
> @mascguy I'm fine with it as it stands. Since it will need a rev-bump to get the change out to my fellow G3 user(s)(?), it would be great to roll out the 2.1.1 release whenever it is deemed ready.

I'll update the Devel port to 2.1.1 ASAP.

On a semi-related note, it looks like upstream recently committed a change related to AltiVec detection:

https://github.com/libjpeg-turbo/libjpeg-turbo/pull/552

It was added just 12 days ago, so it's not released yet. But worth taking a look at, if you haven't seen it!

mascguy:
PR created for 2.1.1 update:

https://github.com/macports/macports-ports/pull/12910

PR #11641, opened 2021-07-21
gcc7-bootstrap: new port

kencu:
this port is to be part of a solution to bootstrapping
older systems. It is as self-contained as I can make it,
relying on only system roots to build on all systems,
and using no other ports in MacPorts to build or run.

This makes it immune to changes in Portfiles over time.

This port will be used to build gcc11 directly, so lets us
get to libgcc11 on older systems without bootstrapping
headaches. libgcc11 requires a c++11 compiler to build.

This port will also be used to build a version of clang,
perhaps clang-3.8 or clang-4.0, directly against libstdc++,
bypassing libc++. This "bootstrap clang" will then be used
to build a fully-capable libc++ to be installed, and also can
be used then to build any "clang dependency" ports that are
needed, without worrying about the path to a capable clang.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you


kencu:
This is not yet ready to merge. I put this here for any thoughts, unseen pitfalls, or improvements needed. It builds on many systems, but there are a few left to check. 

The other parts of the plan, that use this, will come later but can be more fully described if needed.

kencu:
I think this is getting close to a final version. I would like to see how we would go about specifying it in a Portfile to be used during a build, ideally with "configure.compiler" if I can find a way to do that.

kencu:
This is coming along nicely. I'll push it when it's ready.

PR #11668, opened 2021-07-23
npm-1.0: new portgroup

harens:
#### Description

As a first step in creating a javascript portgroup, this npm portgroup aims to make openly available which node versions each npm/yarn depends on.

This follows from the discussion on the [mailing list](https://lists.macports.org/pipermail/macports-dev/2021-July/043636.html). As part of this, I've also tried to refactor the npm ports by creating an `npm.setup` procedure.

I've tested this against npm6 and npm7. If errors appear in the CI for the other ports, I'll aim to correct them.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.1 20C69 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/npm-1.0.tcl
+++ b/_resources/port1.0/group/npm-1.0.tcl
@@ -1,0 +1,127 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+#
+# This PortGroup is mainly used for npm ports (e.g. npm7). It helps setting up conflicts/dependencies/etc.
+#
+# Usage:
+#
+#     PortGroup            npm 1.0
+#     npm.setup            version
+#
+# It also makes publically accessible which node version each NPM/yarn depends on.
+# This can be accessed via [node_version NPM_VERSION]
+# e.g. npm7: [node_version 7] returns 16 for nodejs16
+#      yarn: [node_version yarn]
+
+
+# Set the minimum and maximum versions of npm
+# Required for livecheck and for setting conflicts
+# Also used by javascript portgroup
+options npm.maximum_version npm.minimum_version
+default npm.maximum_version 7
+default npm.minimum_version 3
+
+
+# This publically makes accessible which node version goes with which npm/yarn.
+# N.B. the npm.major of yarn is "yarn"
+proc node_version {npm.major} {
+    global os.major
+
+    # The platform is already darwin, so no need to double check
+    if {${npm.major} <= 5 || ${os.major} < 13} {
+        return 8
+    } else {
+        return 16
+    }
+}
+
+use_configure       no
+
+# Set all npm-specific attributes in npm.setup
+# This is only used by npm-based ports
+options npm.major npm.version
+proc npm.setup {npm_version} {
+    global prefix destroot distname conflicts
+    global npm.major npm.version npm.maximum_version npm.minimum_version
+    global PortInfo
+
+    npm.version         ${npm_version}
+
+    # This returns the first number of the npm version
+    # e.g. 7 for npm7
+    npm.major           [lindex [split ${npm.version} .] 0]
+
+    if {!([info exists PortInfo(name)] && (${PortInfo(name)} ne "npm${npm.major}"))} {
+        name            npm${npm.major}
+    }
+
+    categories          devel
+    platforms           darwin
+    license             MIT
+
+    supported_archs     noarch
+
+    homepage            https://www.npmjs.com/
+
+    description         node package manager
+    long_description    npm is a package manager for node. \
+                        You can use it to install and publish your node programs. \
+                        It manages dependencies and does other cool stuff.
+
+    worksrcdir          "package"
+
+    version             ${npm.version}
+
+    master_sites        https://registry.npmjs.org/npm/-/
+    distname            npm-${npm.version}
+    extract.suffix      .tgz
+
+    # Add all the conflicts
+    for {set index ${npm.minimum_version}} { $index <= ${npm.maximum_version} } { incr index } {
+        if {${index} ne ${npm.major}} {
+            lappend conflicts npm${index}
+        }
+    }
+
+    depends_lib         port:nodejs[node_version ${npm.major}]
+    
+    destroot.destdir    --prefix=${destroot}${prefix}
+    
+    # As of npm5 installing from a directory will result in a symlink being used. 
+    # Change the install to use the built in pack functionality instead.
+    if {${npm.major} >= 5} {
+        build {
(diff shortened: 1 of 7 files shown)


harens:
npm3, npm4:
```
npm ERR! cb.apply is not a function
```

npm5 - should probably set default to nodejs8

```
npm WARN npm npm does not support Node.js v16.5.0
npm WARN npm You should probably upgrade to a newer version of node as we
npm WARN npm can't make any promises that npm will work with this version.
npm WARN npm Supported releases of Node.js are the latest release of 4, 6, 7, 8, 9, 10.
npm WARN npm You can find the latest version at https://nodejs.org/
npm ERR! path /opt/local/var/macports/build/_Users_runner_work_1_s_ports_devel_npm5/npm5/work/package/npm-5.10.0.tgz
npm ERR! code ENOENT
npm ERR! errno -2
npm ERR! syscall stat
npm ERR! enoent ENOENT: no such file or directory, stat '/opt/local/var/macports/build/_Users_runner_work_1_s_ports_devel_npm5/npm5/work/package/npm-5.10.0.tgz'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```

npm6, npm7: Same error as above (without node warning)

```
npm ERR! enoent ENOENT: no such file or directory, stat '/opt/local/var/macports/build/_Users_runner_work_1_s_ports_devel_npm6/npm6/work/package/npm-6.14.13.tgz'
npm ERR! enoent This is related to npm not being able to find a file.
```

yarn was fine.

_resources/port1.0/group/npm-1.0.tcl
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    post-destroot {
+        set completions_path ${destroot}${prefix}/share/bash-completion/completions/
+        xinstall -d ${completions_path}
+        xinstall -m 644 ${worksrcpath}/lib/utils/completion.sh ${completions_path}/npm
+    }
+
+    notes "
+    It is not recommended to install packages globally. But if you do so\
```
ryandesign, line 112:
```suggestion
    It is not recommended to install packages globally, but if you do so\
```

_resources/port1.0/group/npm-1.0.tcl
```diff
@@ -0,0 +1,127 @@
...
+
+# Set all npm-specific attributes in npm.setup
+# This is only used by npm-based ports
+options npm.major npm.version
+proc npm.setup {npm_version} {
+    global prefix destroot distname conflicts
+    global npm.major npm.version npm.maximum_version npm.minimum_version
+    global PortInfo
+
+    npm.version         ${npm_version}
```
ryandesign, line 47: `.setup` procedures aren't great. The existing portgroups where we have them should really be redesigned to eliminate them. Consider how the python-1.0 and php-1.1 portgroups work without needing setup procedures.

How about instead of calling `npm.setup`, make each npm* port set `npm.version`, and make the portgroup do its setup process when `npm.version` is set?

ryandesign (commented):
It looks like this is a very special-purpose portgroup, intended only for use by npm* ports. Do the existing npm* portfiles differ from one another in their structure so greatly that a separate portgroup is needed for this common code? Would it not be more ideal to have a single npm portfile with subports for each npm* version and this portgroup code would then be in that portfile instead? Consider how the perl5 and php portfiles offer subports of multiple versions from a single portfile.

_resources/port1.0/group/npm-1.0.tcl
```diff
@@ -0,0 +1,127 @@
...
+# Required for livecheck and for setting conflicts
+# Also used by javascript portgroup
+options npm.maximum_version npm.minimum_version
+default npm.maximum_version 7
+default npm.minimum_version 3
+
+
+# This publically makes accessible which node version goes with which npm/yarn.
+# N.B. the npm.major of yarn is "yarn"
+proc node_version {npm.major} {
```
ryandesign, line 26: If a portgroup defines a procedure, its name should be preceded by the portgroup name and a delimiter to avoid potential conflicts with other portgroups or procedures that portfiles might define themselves. If the procedure is public (is intended for use by the portfiles or other portgroups that include the portgroup), then we typically use the delimiter `.` (`npm.node_version`). If the procedure is private (intended for use within this portgroup only), then we typically use a namespace with the name of the portgroup (which you have to declare separately) and then the delimiter is `::` (`npm::node_version`).

_resources/port1.0/group/npm-1.0.tcl
ryandesign: In fact I'm not sure why a separate `npm.version` variable is needed. There's already the `version` variable. Why not use that?

_resources/port1.0/group/npm-1.0.tcl
```diff
@@ -0,0 +1,127 @@
...
+    global os.major
+
+    # The platform is already darwin, so no need to double check
+    if {${npm.major} <= 5 || ${os.major} < 13} {
```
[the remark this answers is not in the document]
harens: > It makes no difference if you have already set platforms darwin since MacPorts does not check the value of platforms.

Ah ok, I didn't realise that. Thank you.

_resources/port1.0/group/npm-1.0.tcl
harens: > .setup procedures aren't great. The existing portgroups where we have them should really be redesigned to eliminate them. Consider how the python-1.0 and php-1.1 portgroups work without needing setup procedures.

I see. The unpublished javascript port that I've been working on makes use of a `js.setup` structure, so I might have to rethink that.

> How about instead of calling npm.setup, make each npm* port set npm.version, and make the portgroup do its setup process when npm.version is set?

That's a great idea.

> In fact I'm not sure why a separate npm.version variable is needed. There's already the version variable. Why not use that?

I maybe should've documented that better. I wanted the `npm.setup` to have an attribute called `npm.version`, which sets the version number. This is similar to how `github.setup` sets the version number for you.

However, as you mentioned, there are better approaches that I should consider.

harens:
Thank you very much for your detailed review @ryandesign. The fact that you even spotted some foolish typos of mine really shows how much effort you put into it :)

I greatly appreciate your time and your work, so thanks for reviewing.

> It looks like this is a very special-purpose portgroup, intended only for use by npm* ports.

The original plan was just to make available which node versions each npm port depends on. I then thought about trying to refactor them.

> Do the existing npm* portfiles differ from one another in their structure so greatly that a separate portgroup is needed for this common code? Would it not be more ideal to have a single npm portfile with subports for each npm* version and this portgroup code would then be in that portfile instead? Consider how the perl5 and php portfiles offer subports of multiple versions from a single portfile.

That is a good idea. I didn't realise that perl5/php portfiles are based on a different portfile.

This PR has raised a few things I haven't considered properly, particularly the fix for the previous build failure which throws a lot of the JS PG logic out the window. Given this, I think I need to rethink how I approach this issue, so I'll close the PR for now.

Thanks @ryandesign for your help.

PR #11677, opened 2021-07-24
repo: Update to version 2.16.3

MarcelBochtler:
#### Description

Change fetch.type to git as the versioned script is no longer available
from storage.googleapis.com/git-repo-downloads.

###### Tested on

macOS 11.4 20F71 x86_64
Xcode 12.5.1 12E507

###### Verification 
Have you

### DIFF
--- a/python/repo/Portfile
+++ b/python/repo/Portfile
@@ -1,10 +1,12 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           github 1.0
 
 name                repo
 epoch               20160202
-version             2.15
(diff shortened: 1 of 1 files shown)


python/repo/Portfile
```diff
@@ -3,8 +3,8 @@
 PortSystem          1.0
 
 name                repo
-epoch               20160202
```
MarcelBochtler, line 6: not sure about removing it. Is it required?

ryandesign: A port's epoch can never be decreased (or removed, which decreases it to 0). https://guide.macports.org/#reference.keywords

ryandesign:
> Change fetch.type to git as the versioned script is no longer available
> from storage.googleapis.com/git-repo-downloads.

`fetch.type git` should be used only if there is no distfile alternative available, but there is: the project's homepage mentions a GitHub mirror at https://github.com/GerritCodeReview/git-repo. The GitHub portgroup makes it easy for you to access generated GitHub tarballs. (Also use `github.tarball_from archive`)

PR #11689, opened 2021-07-25
nmap: remove subversion variant

grimreaper:
Problem

the `nmap-update` command was removed in 7.90 and never had fully
functional infrastructure. Further, it causes a default dependency on
subversion.

Solution

Remove `subversion` variant

See Also

https://nmap.org/changelog.html
> Removed nmap-update. This program was intended to provide a way to
> update data files and NSE scripts, but the infrastructure was never
> fielded. It depended on Subversion version control and would have
> required maintaining separate versions of NSE scripts for compatibility.

Note that the upstream build system does not appear to have fully
removed references so `--without-subversion` still is required.

Tested on

macOS 10.14.6 18G9216 x86_64
Xcode 11.3.1 11C504

### DIFF
--- a/net/nmap/Portfile
+++ b/net/nmap/Portfile
@@ -2,6 +2,7 @@ PortSystem 1.0
 
 name		nmap
 version		7.91
+revision	1
 categories	net
 maintainers	darkart.com:opendarwin.org {geeklair.net:dluke @danielluke}
 description	Port scanning utility for large networks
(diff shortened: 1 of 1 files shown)


danielluke:
Is there an upstream bug to fix their configure script so --without-subversion is no longer required? If not, we should suggest that also. Otherwise this looks good - thanks!

grimreaper:
> Is there an upstream bug to fix their configure script so --without-subversion is no longer required? If not, we should suggest that also. Otherwise this looks good - thanks!

I can't find one. I'll add it to my TODO list to write up a report or patch.

PR #11690, opened 2021-07-25
gdb-apple: PowerPC support

evanmiller:
#### Description

Enable building on PowerPC (10.4+) with the supplied patches. PPC64 debugging is not supported.

To circumvent a hang witnessed on Tiger/PPC, replace a call to `pthread_join` with `pthread_detach`. It's possible that this change introduces a thread leak on 10.4/10.5, but the program appears to function normally.

Tiger's native GDB has Apple revision 696, so this port will make 1000+ commits available on that platform.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/devel/gdb-apple/Portfile
+++ b/devel/gdb-apple/Portfile
@@ -28,7 +28,7 @@ can run on most popular UNIX and Microsoft Windows variants.
 
 homepage        http://opensource.apple.com/source/gdb
 platforms       darwin
-supported_archs x86_64 i386
+supported_archs x86_64 i386 ppc
 
 # xm.h can be created out of order
 use_parallel_build no
@@ -43,7 +43,21 @@ depends_build   port:gettext port:zlib port:flex port:texinfo
 
 depends_lib     port:libiconv port:ncurses port:sqlite3
 
-patchfiles	patch-macosx-nat.c.diff
+patchfiles      patch-macosx-nat.c.diff
+
+if {${os.platform} eq "darwin" && ${os.major} < 10} {
+    patchfiles-append \
+                patch-macosx-tdep.c.diff \
+                patch-macosx-nat-threads.c.diff \
+                patch-gdb-availability.diff
+
+    if {${os.major} == 8} {
+        depends_build-append port:libmacho-headers
+        patchfiles-append \
+                patch-macosx-nat-info.c.diff \
+                patch-include-libiberty.h.diff
+    }
+}
 
 worksrcdir	${worksrcdir}/src
 
@@ -55,7 +69,7 @@ post-patch {
         ${worksrcpath}/bfd/configure.host
 
     # https://trac.macports.org/ticket/63153
-     reinplace "s|\$(SDKROOT)/usr/bin/mig|/usr/bin/mig|" \
+    reinplace "s|\$(SDKROOT)/usr/bin/mig|/usr/bin/mig|" \
         ${worksrcpath}/gdb/config/i386/macosx.mh \
         ${worksrcpath}/gdb/config/powerpc/macosx.mh
 }
@@ -76,8 +90,10 @@ destroot.args \
     LEXLIB="${prefix}/lib/libfl.a"
 
 post-destroot {
-    system "chgrp procmod ${destroot}${prefix}/bin/*-apple"
-    system "chmod g+s ${destroot}${prefix}/bin/*-apple"
+    if { ${os.arch} ne "powerpc" } {
+        system "chgrp procmod ${destroot}${prefix}/bin/*-apple"
+        system "chmod g+s ${destroot}${prefix}/bin/*-apple"
+    }
 
     foreach info [glob -tails -directory ${destroot}${prefix}/share/info g*] {
         move ${destroot}${prefix}/share/info/${info} ${destroot}${prefix}/share/info/apple-${info}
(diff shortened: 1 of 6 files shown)


cooljeanius:
Ah, good to see you're addressing [63153](https://trac.macports.org/ticket/63153) with this, too!

evanmiller:
Based on what I've gathered from other PRs, these patches should be rewritten to be applied unconditionally (i.e. move the OS checks out of the Portfile and into the C code).

kencu:
I thought I would let you know about the gdb  7.68 versions that Cameron built for Tiger PPC, that are kept here:

https://sourceforge.net/projects/tenfourfox/files/tools/

These are the ones I've been using lately on Tiger PPC. I'm not sure if they are newer than the ones built by this gdb-apple port, as I have not used this one.

evanmiller:
@kencu Very cool – I was not aware of that fork.

evanmiller:
I've moved some (but not all) of the platform branching into the C files.... After #11724 is merged, and if things look good to everyone, I will squash the commits.

cooljeanius:
I'm trying to merge these patches into my own fork here: https://github.com/cooljeanius/apple-gdb-1824

evanmiller:
@cooljeanius The patches are currently broken – I tried to be clever and edit the old `configure` script to define `HAVE_SPAWN_H`, `HAVE_EXECINFO_H`, etc but it doesn't quite work correctly.

evanmiller:
@kencu - I downloaded the TenFourFox GDB that you linked. It appears to be based on GDB 6.3.50 with Apple revision 768 (i.e. not GNU version 7.68). So while Cameron may have added features, he was forking off a significantly older version than what is used in this port (Apple revision 1824).

CI is building again so the patches at least are ready for review / comment.

kencu:
hey, that is good to know, thanks

evanmiller:
Rebased and ready for testing.

kencu:
I'm looking this over. It's building away on TigerPPC now.

I think to sell this comfortably into MacPorts, we would first need to completely leave unchanged the existing gdb port. Your edits may or may not be right. This port was set up by Jeremy, so we know it was right at that point. What changes might be needed in the main port for Intel archs is going to be a larger discussion that will possibly go on forever.

For the powerpc archs though, we can do pretty much whatever we want, as it never worked there and anything added is a bonus. So there, what we say goes. So let's restrict any changes for now to the powerpc part of the port.

Then we just need to show these changes work on Tiger PPC and Leopard PPC and we are good to push.

kencu:
unless of course some of the changes need to apply to Tiger i386 I guess, and in that case, we have to restrict them to there, or a common Tiger block as you have already started.

The idea though is that when this commit gets looked over, people can see it did nothing to Intel archs 10.5 and above, so there is nothing to worry about.

kencu:
looks like we need:
```
depends_build-append port:libmacho-headers
```
otherwise the build failed on Tiger.

with that added it builds though, Tiger PPC. Will kick the tires over the next couple of days

evanmiller:
@kencu Well, my strategy in a previous PR was to confine my patches to PPC/Tiger. Then I was told it's better to have a patch that applies cleanly on all platforms (so that CI / BB will catch out of date patches) and do the branching inside the patch. This PR is a mixture of those two strategies. I'm happy to tilt it in either direction. Since this is a legacy code release from Apple the "out of date patches" problem may not apply.

Almost all the patches are blocking out functions that are not available on a given platform, i.e. changes to get something to compile where the compilation failed previously. So the only risk they carry is that I screwed up the configure logic. Still it involves fairly low-priority functionality, e.g. getting the name of a thread. The one exception to the foregoing is `patch-macosx-nat-threads.c.diff`. Here I fixed a hang on Tiger/PPC when ending a debugging session and restricted the patch to Tiger. It may or may not be useful on 10.5 or on Intel, but the side effects should be negligible. (`pthread_detach()` is basically an asynchronous version of `pthread_join()`, and this applies to a secondary communication thread set up by GDB.)

Will add the missing dependency and force-push.

kencu:
As I think you well understand, the strategy of applying the patches always holds if every patch is itself blocked by a system blocker so it is only actuated on the new systems. ```MAC_OS_X_VERSION_MIN_REQUIRED``` etc. We do prefer that (because the patches don't get accidentally broken during updated).

This is not a simple source code patch. 

Here you are changing configure logic, etc -- you can't macro that out -- it applies to every build it appears. You can't really test it on all the systems that would need to be test either. That is why this is different, and if you learn one thing hanging around MacPorts for a while, it is that some little patch that somebody thought harmless and worked fine on their local system can very easily screw up four other systems in some way that the patcher / committer never considered possible.

kencu:
Actually, skip the test every system project. We can't possibly do that properly with something like gdb. It's too hard to properly test. We will just apply the patches to broken systems only. After all, it's not like apple-gdb is ever going to be updated again anyway, it's locked forever as is, so the "apply always" patch strategy is meaningless here.

evanmiller:
I think the easiest approach then is to apply my new "universal" patches only to <10.6:

* patch-macosx-tdep.c.diff
* patch-macosx-nat-threads.c.diff
* patch-gdb-availability.diff

I think that should resolve any fears of breaking 10.6+.

I mis-wrote earlier and the "hang" fix currently is applied to all platforms, not just Tiger. So we can restrict that to Tiger/Leopard.

evanmiller:
Patches are now applied conditionally and provide more explanation.

kencu:
thanks!

PR #11706, opened 2021-07-26
fluidsynth: Fix build on Tiger/Leopard

evanmiller:
#### Description

A number of Core Audio symbols were renamed in 10.6, with the old names dropped in macOS 11.0.

See: https://trac.macports.org/ticket/63230

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/multimedia/fluidsynth/Portfile
+++ b/multimedia/fluidsynth/Portfile
@@ -42,6 +42,10 @@ depends_lib         port:flac \
 
 depends_run-append  port:generaluser-soundfont
 
+# Remove with fluidsynth 2.2.3 or later
+# https://trac.macports.org/ticket/63230
+patchfiles-append   patch-pre-snow-leopard.diff
+
 # https://trac.macports.org/ticket/36962
 platform darwin 8 {
     patchfiles-append \
--- a/multimedia/fluidsynth/files/patch-pre-snow-leopard.diff
+++ b/multimedia/fluidsynth/files/patch-pre-snow-leopard.diff
@@ -1,0 +1,49 @@
+--- src/drivers/fluid_coreaudio.c.orig
++++ src/drivers/fluid_coreaudio.c
+@@ -187,14 +187,22 @@ new_fluid_core_audio_driver2(fluid_settings_t *settings, fluid_audio_func_t func
+     dev->data = data;
+ 
+     // Open the default output unit
++#if MAC_OS_X_VERSION_MIN_REQUIRED < 1060
++    ComponentDescription desc;
++#else
+     AudioComponentDescription desc;
++#endif
+     desc.componentType = kAudioUnitType_Output;
+     desc.componentSubType = kAudioUnitSubType_HALOutput; //kAudioUnitSubType_DefaultOutput;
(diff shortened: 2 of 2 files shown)


mojca:
Thank you very much for the patches.

While this looks ok as it is in principle (I don't have any plans to test compatibility with Leopard and earlier, I just believe you that it works), these kind of patches are very likely to stop working as software gets updated, since nobody working on a modern machine would ever notice the breakage. For this reason we usually slightly prefer patches that get applied unconditionally on all platforms, but only affect those where the patch is needed.

My request nr. 1 would be to rewrite the patch with one of the macros checking whether the target version is smaller than X.

My request nr. 2 would be to try to submit that patch upstream (https://github.com/FluidSynth/fluidsynth/pulls, CC-ing @ringoz) and link to the upstream report. If upstream is willing to accept the patch, we really win in the long run. If not, that's still fine, we'll keep maintaining the patch ourselves.

I leave this to @kencu. (If you decide that it's ok to merge it as it is, go forward, I'm fine with the decision.)
I don't know what @RJVB says (he's a more active maintainer of this port than I am), but last thing I remember he was usually testing on >= 10.9.

mojca:
Also, this port is waiting for an update. In case anyone is willing to test the new version, that would be welcome.

evanmiller:
I suspect the most sustainable solution would be to add the newer symbols to Legacy Support, so that other projects will not break in the same way. However, this would involve a bit more work and testing.

RJVB:
I agree with Mojca that the proper way would be to use the reversible patch to add conditional code that gets built only when needed - or only when supported in this case.

Submitting a patch upstream would be the ideal of course but let's be realistic: they probably dropped support for older OS versions for a reason. Best to check their attitude to this kind of patch before filing a full PR or whatever they use!

As usual I'm even more swamped with real-life issues during holidays than outside so I can't make any promises as to when I can test this, but I'll try.

R.

evanmiller:
I have opened a PR upstream to accommodate both `Component` and `AudioComponent`: https://github.com/FluidSynth/fluidsynth/pull/947

They sound amenable to it. I will leave this PR open since it could be several weeks before that patch makes its way into an official release.

kencu:
Evan, if we could have your upstream patch here we could commit that right away.

mojca:
I totally agree. Let's just take the patch that was submitted upstream, use it unconditionally, and merge the PR.

mojca:
Upstream merged the changes. @evanmiller: thank you very much.

RJVB:
+++

And thanks!

evanmiller:
Rebased with the upstream patch applied unconditionally.

kencu:
Thanks!

RJVB:
On Monday July 26 2021 11:41:47 Mojca Miklavec wrote:

>Also, this port is waiting for an update. In case anyone is willing to test the new version, that would be welcome.

Would you be willing to open a ticket for that, one which lists all the current dependents (and the ones known to break with 2.2+, as far as feasible)?

PR #11736, opened 2021-07-30
cmake: restore no-default-rpath usage on Tiger

kencu:
cmake is gradually removing support for Tiger
at present, only the default adding of an -rpath
to all library and executables is troublesome however

this reverts:
https://gitlab.kitware.com/cmake/cmake/-/commit/4aed96e2309e64571aabfda263b3ca4c4f62d619

closes: https://trac.macports.org/ticket/63261

this needs someone to actually test it on Tiger to be 100% certain that it does the needed changes, but it looks to be correct.

What changes newer and future versions of cmake might make that make supporting Tiger difficult remain to be seen.

### DIFF
--- a/devel/cmake/Portfile
+++ b/devel/cmake/Portfile
@@ -47,7 +47,7 @@ if {${subport} eq ${name}} {
     checksums       rmd160  f0f13ae525878d7929dcc95f322d224e7b47443d \
                     sha256  54dd3e0dcd64e27ff8442071aacc4f5a9605896b5701de46d76f22d686238eba \
                     size    7264123
-    revision        1
+    revision        2
 
     gitlab.livecheck.regex {([0-9.]+)}
     compiler.cxx_standard 2011
@@ -78,7 +78,8 @@ if {${subport} eq ${name}} {
         patch-fix-system-prefix-path.diff \
         patch-cmake-leopard-tiger.diff \
         patch-fix-clock_gettime-test.diff \
-        patch-qt5gui.diff
+        patch-qt5gui.diff \
+        patch-restore-no-default-rpath-added-on-Tiger.diff
 
     depends_lib-append \
         port:curl \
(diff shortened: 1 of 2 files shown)


kencu:
@ryandesign , @evanmiller -- please feel free to test as you both have noted this issue so far.

evanmiller:
Do we know whether `DARWIN_MAJOR_VERSION` refers to the host OS version or target OS version (i.e. `-mmacosx-version-min=`)? I'm wondering if the logic ought to be modified to test the target version (to facilitate 10.5 -> 10.4 compiles).

mojca:
I decided to be bold and opened an upstream request:
* https://gitlab.kitware.com/cmake/cmake/-/issues/22499

mojca:
Oh, I'm sorry. I see that you already reported it upstream. I wasn't reading all the references close enough.

kencu:
darwin major version will relate to the host os. 

Let's for now restore what we had, which is an easy step that gets everyone working again. 

Building on 10.5 for deployment on 10.4 will still work (the LC_RPATH load command will be ignored by Tiger once it is in there), but this is overall not well supported in MP anyway due to the many assumptions in Portfiles and open-source software that they will run on the building system. We can work on specific issues as we cross them, if needed. 

Can you confirm for Michael you built this and it works on Tiger  now? Then we can push I think.

evanmiller:
@kencu OK, restoring the old behavior for MP makes sense. I have not built or tested on Tiger.

michaelld (approved):
LGTM ; I haven't tested, but since it impacts just Tiger / 10.4, I think this is fine either way. I will note from the noted CMake commit, that it specifically notes not just deprecation of OSX 10.4 or older, not removing support ... if this is the only real change that adds support back, then I think it's fine that we include it in MP & I don't see why they don't do the same upstream ...
```
Since CMake 3.19, we no longer support macOS SDKs older than 10.5,
which corresponds to Xcode 3. Supporting older Xcode versions for
device platforms is also not realistic. We therefore expect the -rpath
linker option should always be supported now.
```

michaelld:
@evanmiller Can you confirm that this change allows current CMake to build and run on Tiger / 10.4?

evanmiller:
Confirmed fixed on 10.4.11.

PR #11755, opened 2021-07-31
dylibbundler: update to 1.0.0

Lord-Kamina:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G9216 x86_64
Xcode 11.3 11C29

###### Verification 
Have you

Just one note, I added the `configure.cxx_standard` and it did nothing, not sure if it's because the port doesn't use configure or what, but that's the reason I *also* added the `-std=c++11` flag like that in the `CXXFLAGS`.

### DIFF
--- a/devel/dylibbundler/Portfile
+++ b/devel/dylibbundler/Portfile
@@ -3,11 +3,11 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        auriamg macdylibbundler 0.4.5 {} -release
+github.setup        auriamg macdylibbundler 1.0.0 {}
 revision            0
-checksums           rmd160  e87ffd25d5405e78bf02cb3c57795b0d5502bfec \
-                    sha256  cd41e45115371721e0aa94e70c457134acf49f6d5f6d359b5bae060fd876d887 \
-                    size    11114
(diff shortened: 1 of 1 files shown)


ryandesign:
Thanks for letting me know version 1.0.0 was available.

> Just one note, I added the `configure.cxx_standard` and it did nothing,

`compiler.cxx_standard` doesn't do nothing: it asks MacPorts to select a compiler that supports that language standard.

> that's the reason I _also_ added the `-std=c++11` flag like that in the `CXXFLAGS`.

The Makefile already adds this flag, we were just overriding the Makefile. I've fixed it so that we no longer override the Makefile for this flag.

Lord-Kamina:
> Thanks for letting me know version 1.0.0 was available.
> 
> > Just one note, I added the `configure.cxx_standard` and it did nothing,
> 
> `compiler.cxx_standard` doesn't do nothing: it asks MacPorts to select a compiler that supports that language standard.
> 
> > that's the reason I _also_ added the `-std=c++11` flag like that in the `CXXFLAGS`.
> 
> The Makefile already adds this flag, we were just overriding the Makefile. I've fixed it so that we no longer override the Makefile for this flag.

Ah, I see.

PR #11780, opened 2021-08-03
catgirl: use libretls for libtls dependency by default

causal-agent:
catgirl can build with libtls provided by either libretls when
openssl is installed, or by libressl when libressl is installed.
Since openssl is the default, use libretls to provide the missing
library. Also use a path: dependency since we always want to build
against a MacPorts libtls.

While here, add license_noconflict for both openssl and libressl,
since catgirl's license includes additional permission for
OpenSSL/LibreSSL.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H1323 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/libretls/Portfile
+++ b/devel/libretls/Portfile
@@ -25,7 +25,10 @@ checksums           rmd160  7a6bd3ecf5ae0385ad175a4cb97b02e3571a69c3 \
 depends_lib         port:openssl
 
 configure.args      --with-openssl=${prefix}
+
+if {[vercmp [macports_version] 2.6.99] >= 0} {
 configure.checks.implicit_function_declaration.whitelist-append getpagesize
+}
 
 livecheck.type      regex
 livecheck.url       ${master_sites}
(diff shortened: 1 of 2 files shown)


ryanakca:
I don't have my macbook handy to test, but OK by me so long as it continues to build against libressl.

Also, please feel free to add yourself as a maintainer.

causal-agent:
Seems I should add `if {[vercmp [macports_version] 2.6.99] >= 0}` before the `configure.checks` line that was suggested to add in libretls' Portfile to fix this build.

reneeotten (changes requested):
@casual-agent @ryanakca since you're changing the dependency here you'll want to make sure that this is picked up by all users. Since the actual `versions` did not change, MacPorts will not consider this change as an update unless you increase the `revision` field. If you can please take care of that, I think this PR is all good to go!

causal-agent:
@reneeotten Ah right, fixed. Thanks

reneeotten (approved):

PR #11139, opened 2021-05-25
mp4v2: Update to 4.1.6 from TechSmith fork

mkuron:
#### Description

[mp4v2](https://code.google.com/archive/p/mp4v2/) hasn't been updated in almost a decade. @TechSmith continues to maintain a [fork](https://github.com/TechSmith/mp4v2). https://trac.macports.org/ticket/53467 has, for a long time, requested that ports hosted on Google Code (which was made read-only in 2012) should be updated, including mp4v2, and suggests precisely this fork as the replacement repository. I have also added a patch from @sandreas, who has a separate [fork](https://github.com/sandreas/mp4v2) that is also (somewhat) maintained. This specific patch is currently the main difference to the @TechSmith repo, it's useful on its own and would be required if we ever add a port for https://github.com/sandreas/m4b-tool.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.4 20F71 arm64
Xcode 12.5 12E262

###### Verification
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 
- [x] checked your Portfile with `port lint`?
- [x] tried existing tests with `sudo port test`?
- [x] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?

PR #11781, opened 2021-08-03
mp4v2: Patch out C++ code introduced by TechSmith fork

mkuron:
Closes https://trac.macports.org/ticket/63297

#### Description

Since we switched from the unmaintained original version to the TechSmith fork in #11139, dependents that used a C compiler (as opposed to a C++ compiler) would fail to build. This problem was reported in https://trac.macports.org/ticket/63297 and further discussed in #11139 by @ryandesign and @sandreas. The issue has also been reported upstream as https://github.com/TechSmith/mp4v2/issues/47.

I initially intended to completely revert the change that TechSmith made, but in order to remain compatible with possible future dependents that make use of TechSmith's new API I decided to go with a variadic macro to emulate the behavior in plain C.

cmus builds fine after applying my patch.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.5.1 20G80 arm64
Xcode 12.5.1 12E507

###### Verification
Have you


ryandesign:
Thanks for attempting to fix this.

I don't know if this fix is correct. It would be nice if you could submit it to TechSmith as a pull request and have them agree that it is the correct solution and merge that PR, but I don't know if we can expect them to respond in that way or indeed at all.

Since this changes installed files, the port's revision would need to increase.

mkuron:
> I don't know if this fix is correct.

I have done a few tests and think it's correct. It restores API compatibility with the old 2.0 version, though not ABI compatibility. Does that mean I should increase the revision on all dependent ports as well so they get rebuilt?

> I don't know if we can expect them to respond in that way or indeed at all.

Indeed they don't seem to be currently reacting to anything going on on their GitHub. So support clearly isn't better than with the old and unsupported Google Code repository, but at least TechSmith are taking care of security fixes and adapting to new macOS versions...

> Since this changes installed files, the port's revision would need to increase.

Will do that once I know whether the dependent ports also need to have their revisions increased.

sandreas:
> I have done a few tests and think it's correct.
Although this might be the case, I just wanted to ask as a non-expert: Wouldn't it be easier to just replace all `nullptr` with `NULL`?

> Indeed they don't seem to be currently reacting to anything going on on their GitHub
I made the same experience. If you would like to, I would really appreciate a submit a PR to my fork https://github.com/sandreas/mp4v2. Although I'm not a professional C developer, I'll try to keep an eye on the fixes, updates and PRs submitted to TechSmith and reintegrate as many of them as possible (as well as some of my own extensions like `sorttitle`, `purchasedate`, etc.).

mkuron:
> Wouldn't it be easier to just replace all nullptr with NULL?

C doesn‘t have default arguments. But you can emulate them via variadic macros like I did.

> If you would like to, I would really appreciate a submit a PR to my fork

I guess we could consider switching to your fork, but then we would need version numbers. But if your fork essentially follows TechSmith plus your own extension (which we do already include as a patch), we wouldn’t gain much.

ryandesign:
> I guess we could consider switching to your fork, but then we would need version numbers. But if your fork essentially follows TechSmith plus your own extension (which we do already include as a patch), we wouldn’t gain much.

Ideally, MacPorts does not want to be in the business of carrying around large patches forever. If changes like those by sandreas are useful, they should be submitted to the upstream developers. If his changes are based on TechSmith's fork, then I guess TechSmith is upstream. That seems preferable to me, since they've put in some development effort already and continue to release new versions, as recently as June 30.

However, if TechSmith does not accept these changes because they are only publishing their fork out of obligation and do not intend accept outside contributions or fix compatibility bugs then we should rethink selecting them as the basis for this port. If @sandreas would like his fork to be considered upstream, then he should advertise this fact in the readme, add a project description, enable the GitHub issue tracker, create tags and releases, and perhaps coordinate with TechSmith to have them mention his fork in their readme as the preferable one for public use.

Perhaps we should survey what other distributions are doing. It would be confusing for users and developers alike if some distributions are using the TechSmith fork and some are using the sandreas fork and they diverge in their feature set and version numbers. One place to start such a survey could be https://repology.org/project/libmp4v2/versions.

mkuron:
> I would love to do so, BUT I think I'm not qualified, since I'm not a C developer and probably I would not have enough time

So we probably wouldn't gain much by switching to your fork. At least you respond on GitHub, @sandreas, which is a giant advantage over @TechSmith. mp4v2 is such a nice tool and library and it is highly unfortunate that the original developer abandoned it. If we want to keep it in MacPorts (and not drop it like Ubuntu did), we will need to use a fork that is at least somewhat maintained.

Right now I think the best solution would be to stick with the @TechSmith fork and ship this relatively simple patch to remove the C++ `nullptr` with MacPorts. If @TechSmith breaks mp4v2 compatibility in worse ways in future versions, we should re-evaluate which fork to use. Reverting to 2.0.0 or 4.1.3 is not an option as I wrote at https://trac.macports.org/ticket/63297#comment:10.

sandreas:
> So we probably wouldn't gain much by switching to your fork. 

> Right now I think the best solution would be to stick with the @TechSmith fork and ship this relatively simple patch to remove the C++ nullptr with MacPorts. We should also try to upstream the patch, but that will take longer with @TechSmith not being responsitve, and we don't want to keep the cmus ports etc. broken for longer. 

I'm totally with you on that. Best solution for now in my opinion - thanks for sharing your opinion and working things out like this. If I can do something to help, feel free to ask.

mkuron:
I incremented the revision on all mp4v2 dependents so they get rebuilt. As discussed above, ABI compatibility was already broken with the upgrade from 2.0.0 to 4.1.6, and the present pull request at least restores API compatibility. Unfortunately, three of the dependents fail to build for entirely unrelated (non-mp4v2) reasons, so I cannot get a green status check here.

PR #11800, opened 2021-08-05
tippecanoe: blacklist on ports-10.7

sikmir:
https://build.macports.org/builders/ports-10.7_x86_64-watcher/builds/20638

cc @kencu @herbygillot

### DIFF
--- a/gis/tippecanoe/Portfile
+++ b/gis/tippecanoe/Portfile
@@ -4,6 +4,7 @@ PortSystem              1.0
 PortGroup               github 1.0
 PortGroup               makefile 1.0
 PortGroup               legacysupport 1.0
+PortGroup               compiler_blacklist_versions 1.0
 
 github.setup            mapbox tippecanoe 1.36.0
 revision                0
@@ -22,3 +23,5 @@ depends_lib-append      port:sqlite3 \
                         port:zlib
 
 destroot.args-append    PREFIX=${destroot}${prefix}
+
+compiler.blacklist-append {clang < 500}


ryandesign:
It is a good idea for a commit message to say *why* something is being done, not just *what* is being done. *What* is being done can be seen already by looking at the commit.

In the PR you only linked to output from a buildbot run which will be automatically deleted before long so for posterity I will record the errors from the log here:

```
In file included from geometry.cpp:14:
./mapbox/geometry/multi_polygon.hpp:17:27: error: no member named 'container_type' in 'std::__1::vector<mapbox::geometry::polygon<long long, std::vector>, std::__1::allocator<mapbox::geometry::polygon<long long, std::vector> > >'
    using container_type::container_type;
          ~~~~~~~~~~~~~~~~^
geometry.cpp:196:26: note: in instantiation of template class 'mapbox::geometry::multi_polygon<long long, std::vector>' requested here
        for (size_t i = 0; i < t.size(); i++) {
                                ^
In file included from geometry.cpp:14:
In file included from ./mapbox/geometry/multi_polygon.hpp:4:
./mapbox/geometry/polygon.hpp:27:27: error: no member named 'container_type' in 'std::__1::vector<mapbox::geometry::linear_ring<long long, std::vector>, std::__1::allocator<mapbox::geometry::linear_ring<long long, std::vector> > >'
    using container_type::container_type;
          ~~~~~~~~~~~~~~~~^
geometry.cpp:197:30: note: in instantiation of template class 'mapbox::geometry::polygon<long long, std::vector>' requested here
                for (size_t j = 0; j < t[i].size(); j++) {
                                           ^
In file included from geometry.cpp:14:
In file included from ./mapbox/geometry/multi_polygon.hpp:4:
./mapbox/geometry/polygon.hpp:18:27: error: no member named 'container_type' in 'std::__1::vector<mapbox::geometry::point<long long>, std::__1::allocator<mapbox::geometry::point<long long> > >'
    using container_type::container_type;
          ~~~~~~~~~~~~~~~~^
geometry.cpp:200:34: note: in instantiation of template class 'mapbox::geometry::linear_ring<long long, std::vector>' requested here
                        for (size_t k = 0; k < t[i][j].size(); k++) {
                                                      ^
/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_gis_tippecanoe/tippecanoe/work/compwrap/cxx/usr/bin/clang++ -MMD  -I/usr/local/include -I. -g -Wall -Wshadow -Wsign-compare -Wextra -Wunreachable-code -Wuninitialized -Wshadow -O3 -DNDEBUG -Os -arch x86_64 -stdlib=libc++ -std=c++11 -c -o main.o main.cpp
tile.cpp:1774:33: error: variable length array of non-POD element type 'std::atomic<long long>'
                std::atomic<long long> geompos[child_shards];
                                              ^
main.cpp:413:34: error: variable length array of non-POD element type 'std::atomic<long long>'
        std::atomic<long long> layer_seq[CPUS];
                                        ^
main.cpp:694:36: error: variable length array of non-POD element type 'std::atomic<long long>'
        std::atomic<long long> sub_geompos[splits];
                                          ^
main.cpp:1370:36: error: variable length array of non-POD element type 'std::atomic<long long>'
                        std::atomic<long long> layer_seq[CPUS];
                                                        ^
main.cpp:1426:36: error: variable length array of non-POD element type 'std::atomic<long long>'
                        std::atomic<long long> layer_seq[CPUS];
                                                        ^
```

I am not familiar with these errors so I don't know whether they indicate a requirement for a particular version of the C++ standard or something else. However the project's readme says:

> Tippecanoe now requires features from the 2011 C++ standard

MacPorts already knows (see https://github.com/macports/macports-base/blob/master/src/port1.0/portconfigure.tcl) that clang < 500 does not support C++11. Therefore the better fix would have been to simply convey to MacPorts that this project requires C++11 and let it select a suitable compiler:

```
compiler.cxx_standard 2011
```

PR #11814, opened 2021-08-08
ipfs: update to 0.9.1

herbygillot:
###### Tested on

macOS 10.15.7 19H1323 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/net/ipfs/Portfile
+++ b/net/ipfs/Portfile
@@ -3,25 +3,36 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/ipfs/go-ipfs 0.8.0 v
+go.setup            github.com/ipfs/go-ipfs 0.9.1 v
 name                ipfs
-categories          net
-
-checksums           rmd160  96a3cc49da44a98068571e71e8c728b08279e144 \
-                    sha256  60a090ace201de65043f3fcf22ea447ec4abd4c334e71e0185fb6f1e45f56aa9 \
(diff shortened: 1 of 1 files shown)


sirn:
Is it possible to change the source distfile to `https://dist.ipfs.io/go-ipfs/v0.9.0/go-ipfs-source.tar.gz`? It was recommended by upstream (https://github.com/ipfs/go-ipfs/pull/8220) to use this vendored tarball instead of the one in GitHub.

sirn (approved):

sirn:
Looks good to me. Thank you!

herbygillot:
Np.  I think we can improve the launchd .plist handling by making `ipfs` run as a non-user-specific daemon, with config living at say `/opt/local/etc/ipfs`, instead of in `/Users/<whoever>/.ipfs`.  In this way, we can have MacPorts automatically install the launchd configuration, and a user need only `sudo port [load/unload] ipfs` to start its service in launchd, instead of having to follow the manual steps listed in the notes.

sirn (approved):

herbygillot:
OK, I merged the current changes - but let me know your thoughts around what I mentioned previously

sirn:
Definitely, I think we could do that. It appears that we can use `IPFS_PATH` environment variable to set IPFS path, but this path also contains other data other than config (cache, etc.) so I think something other than `etc` (var?) might be a better fit?

PR #11883, opened 2021-08-15
lean: new port

kakuhen:
#### Description
[leanprover-community/lean](https://github.com/leanprover-community/lean ) is the canonical upstream for the Lean theorem prover. Lean 3.x was discontinued by the official Lean developers in order to work on Lean 4.x, which does not have a stable release yet (as it is still in development). The official developers [defer to the community-maintained fork](https://github.com/leanprover/lean/blob/master/README.md) for anything involving Lean 3. Moreover, mathlib and related communities centered around Lean use the community-maintained fork. Lastly, other package repositories (such as FreeBSD ports) use the community branch's releases for math/lean.

For these reasons, I named the port lean and chose the community branch.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H1323 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/math/lean/Portfile
+++ b/math/lean/Portfile
@@ -1,0 +1,31 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           github 1.0
+
+github.setup        leanprover-community lean 3.32.1 v
+revision            0
+
+categories          math
+license             Apache-2
+maintainers         {@kakuhen} openmaintainer
+platforms           darwin
+supported_archs     x86_64 arm64
+description         The Lean theorem prover.
+long_description    A fork of Lean 3 maintained and updated by the Lean \
+    community. Lean is a functional programming language that makes it easy to \
+    write correct and maintainable code. You can also use Lean as an \
+    interactive theorem prover.
+
+depends_lib         port:gmp
+
+checksums \
+    rmd160    6b1b8a1134c40ab772e8a3795ae0b29e13a24a22 \
+    sha256    50c49a426024fa35ee2547f7e82803deecb81fbdc187734da9b0fb56ff54e99f \
+    size      1873303
+
+cmake.build_type    Release
+cmake.generator     Ninja
+
+set worksrcdir ${worksrcdir}/src


kencu:
nice job!

You won't need this line

depends_build       port:cmake port:gmp

because the cmake port is specified in the cmake PortGroup already (a little sugar), and gmp is also specified as a lib dep later on, and lib deps are automatically also build deps.

But the rest looks great, and once we confirm the build, should be good to go.

math/lean/Portfile
```diff
@@ -0,0 +1,33 @@
...
+maintainers         {@kakuhen} openmaintainer
+platforms           darwin
+supported_archs     x86_64 arm64
+description         The Lean theorem prover.
+long_description    A fork of Lean 3 maintained and updated by the Lean \
+    community. Lean is a functional programming language that makes it easy to \
+    write correct and maintainable code. You can also use Lean as an \
+    interactive theorem prover.
+
+depends_build       port:cmake port:gmp
```
herbygillot, line 21: A port specified in `depends_lib` does not have to be specified again in `depends_build`.

If a library is needed **only** at runtime, then it would get specified in `depends_run`.

Something in `depends_lib` is understood to be needed both at build *and* run -time.

herbygillot: Ah, just saw @kencu already commented on this 😅

pmetzger:
Very glad to see this packaged! No comments to make beyond the ones already made.

math/lean/Portfile
kakuhen: Both comments gave me good information, for this port and any future ports I may submit. Thanks to you two for the feedback. If there is anything else, please let me know.

math/lean/Portfile
```diff
@@ -0,0 +1,32 @@
...
+checksums \
+    rmd160    6b1b8a1134c40ab772e8a3795ae0b29e13a24a22 \
+    sha256    50c49a426024fa35ee2547f7e82803deecb81fbdc187734da9b0fb56ff54e99f \
+    size      1873303
+
+cmake.build_type    Release
+cmake.generator     Ninja
+
+configure.post_args \
+    ${workpath}/${distname}/src
```
kencu, line 32: I missed this last thing first time through. Although this does work, there is a more "MacPortsy" way to do this, but setting the variable ```${worksrcdir}```. The default ```${worksrcdir}``` is ```${distname}```, so what we do is:
```
${worksrcdir} ${worksrcdir}/src
```

Now i always have to try this to be sure I have it right. Some times you need this construct:
```
set ${worksrcdir} ${worksrcdir}/src
```
and some times you just need this:
```
${worksrcdir} ${worksrcdir}/src
```
and I forget just which one is when.

Then ```${worksrcpath}``` is contructed from ```${workpath}/${worksrcdir}``` and that is passed to cmake by the cmake PortGroup in the post-args.  

i admit, this is confusing for new users, and it might just be easier to do what you did:
```
configure.post_args  ${workpath}/${distname}/src
```
which seems a lot clearer and easier to remember. But that is not MacPorts' way, and we do it this other way in all ports.

kakuhen: The reason I used the configure argument is becuase the cmake port group makes a folder called "build," separate from the actual folder that holds the extracted source, and the CMake sources are in a "src" folder inside the extracted folder. I will see what I can do right now to apply your advice.

kencu: You're 100% right about that. I think you'll find that one of these works. Now that I read it the 3rd time, I think it might be this:
```
worksrcdir ${worksrcdir}/src
```

or perhaps
```
set worksrcdir ${worksrcdir}/src
```

kakuhen: The latter option worked on my machine, and the binaries are still functional. I pushed the changes just now.

kencu:
Yep, that looks really tight now. If this builds through on the CI, I'll push it. thanks for your patience!~

kencu:
looks good. I ran it locally. I have a couple of tweaks to add later.

PR #11897, opened 2021-08-16
stack: Move architecture for stack-built ports to haskell_stack PG

essandess:
###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.5.2 20G95 x86_64
Xcode 12.5.1 12E507

**and**

macOS 11.5.2 20G95 arm64
Xcode 12.5.1 12E507

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/haskell_stack-1.0.tcl
+++ b/_resources/port1.0/group/haskell_stack-1.0.tcl
@@ -70,6 +70,9 @@ post-extract {
     xinstall -m 0755 -d "[option haskell_stack.stack_root]"
 }
 
+# stack builds x86_64 binaries
+supported_archs     x86_64
+
 # libHSbase shipped with GHC links against system libiconv, which provides the
 # 'iconv' symbol, but not the 'libiconv' symbol. Because the compilation
 # process statically links libHSbase.a, we must have /usr/lib in the library
--- a/lang/stack/Portfile
+++ b/lang/stack/Portfile
@@ -28,8 +28,6 @@ homepage            https://haskellstack.org
 variant prebuilt \
 	    description {Do not bootstrap stack; install the pre-built binary.} {}
 
-supported_archs     x86_64
-
 distname            ${name}-${github.version}-osx-x86_64
 
 worksrcdir          ${name}-${github.version}


kencu:
still not completely clear to me why we would need to depends_skip_archcheck if we are specifying x86_64 for everything. Something is funny.

But I don't have an M1 machine handy to work through it, so I'm all talk no action.

essandess:
> still not completely clear to me why we would need to depends_skip_archcheck if we are specifying x86_64 for everything. Something is funny.

Without `depends_skip_archcheck-append stack`, trying to install any stack-built ports on Apple silicon throws this error:
> ```
> Cannot install pandoc for the arch 'arm64' because
> It’s dependency stack is only installed for the arch 'x86_64'
> and does not have a universal variant.
> Unable to execute port: architecture mismatch
> ```

kencu:
Yeah, that is still the part that is hard to understand.

IF the Portfile (or properly included PortGroup) says that the ```supported_archs``` are ```x86_64```, then how in God's Green Earth is base trying to build it as ```arm64``` on **any** system?

essandess:
As I understand the install logic, it’s trying to install an x86_64 binary that will run on an arm64 M1 because macOS uses Rosetta. Native arm64 binaries are preferred, but x86_64 binaries work just fine.

Until stack/cabal/ghc can produce arm64 binaries, we must install its x86_64 ones.

kencu:
Let me see if I can get access to an M1 machine to see if/why supported_archs seems to be utterly non-functional for you here.

kencu:
> Cannot install pandoc for the arch 'arm64' because
> It’s dependency stack is only installed for the arch 'x86_64'
> and does not have a universal variant.
> Unable to execute port: architecture mismatch

So I now have access to an M1 once again, thanks to @michaelld , who is always very generous with his time and resources. Thanks, Michael.

I ran ```sudo port -v install pandoc``` and I saw this this same error.

I added ```supported_archs x86_64``` to the pandoc Portfile, and as it should, it gave no error, and has started building away. There is a fair amount to build, so it will take a while, but it would seem that base is indeed properly working, and there was no need for me to put ```depends_skip-archcheck stack``` into the Portfile, at least with pandoc.

I will report back later, which it is finished, and then try shellcheck.

kencu:
pandoc installed properly as an x86_64 binary on the M1 machine, and works.

pandoc and stack are the only two ports that have installed as x86_64 to date:
```
% port -v installed | grep x86
  pandoc @2.14.1_0 (active) requested_variants='' platform='darwin 20' archs='x86_64' date='2021-08-21T16:13:14-0400'
  stack @2.7.3_0 (active) requested_variants='' platform='darwin 20' archs='x86_64' date='2021-08-21T15:49:29-0400'
```

kencu:
OK, this time I put 
```
supported_archs x86_64
```
into the ```haskell_stack-1.0.tcl``` PortGroup, and then tried your shellcheck PR.

Without the above addition, I got the usual error:
```
Cannot install shellcheck for the arch 'arm64' because
It’s dependency stack is only installed for the arch 'x86_64'
and does not have a universal variant.
Unable to execute port: architecture mismatch
```

after I added the supported_archs line to the haskell_stack-1.0.tcl PortGroup, everything worked just as expected, there were no  issues with the build, and shellcheck is building away. I will report back when done.

So -- something is not quite working right on your M1 system, Steve, but it is working perfectly as expected here.

kencu:
confirmed. shellcheck also now properly installed as an x86_64 binary on the M1:
```
% port -v installed | grep x86_64
  pandoc @2.14.1_0 (active) requested_variants='' platform='darwin 20' archs='x86_64' date='2021-08-21T16:13:14-0400'
  shellcheck @0.7.2_0 (active) requested_variants='' platform='darwin 20' archs='x86_64' date='2021-08-21T17:07:18-0400'
  stack @2.7.3_0 (active) requested_variants='' platform='darwin 20' archs='x86_64' date='2021-08-21T15:49:29-0400'
```

just to reiterate, the only change is to add:
```
supported_archs x86_64
```
to either the Portfile (pandoc tested) or the haskell_stack PortGroup (shellcheck tested).

This looks right and proper and is the way base is designed to work.

essandess:
@kencu Thank you for checking this. I'm scratching my head about what could be the issue on my M! box because it's a brand new box with a brand new MacPorts install.

I'll defer to your simpler approach and hope that it just works on my end. I suggest modifying this PR to remove the `depends_skip_archcheck-append stack` lines. Agreed?

kencu:
so it would seem. I admit to being puzzled by the differences between your system and what I saw, but Josh says that the right way is to spec the supported_archs and things should work.

If there are Heisenbergs in the wild, we'll find out quickly I suppose!

kencu:
Well, this looks right. But just to be 100% sure, let me actually use this exact PR on the M1 and make shamma-sure.

kencu:
Yeah - far as I can tell, this is working as it should.

PR #11948, opened 2021-08-21
harfbuzz/pango: update to latest releases

mascguy:
#### Description

* harfbuzz: update to 2.9.0
* pango: update to 1.48.9
* py-pygtk: patch for latest pango

###### Type(s)
- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.12.6 16G2136
Xcode 9.2 9C40b

macOS 10.13.6 17G14019
Xcode 10.1 10B61

macOS 10.14.6 18G103
Xcode 11.3.1 11C505

macOS 10.15.6 19G2021
Xcode 12.0 12A7209

macOS 10.15.7 19H1217
Xcode 12.0 12A7209

###### Verification 
Have you


evanmiller:
Attempting to build pango with your Portfile on a Tiger box I get a configure error:

```
Run-time dependency xrender found: YES 0.9.10
Run-time dependency appleframeworks found: NO (tried framework)

meson.build:373:2: ERROR: Dependency "appleframeworks" not found, tried framework
```

It's using macports-gcc-7. Any ideas?

python/py-pygtk/Portfile
```diff
@@ -1,11 +1,12 @@
 PortSystem      1.0
 PortGroup       python 1.0
+PortGroup       compilers 1.0
```
kencu, line 3: I don't see that you are using this added PortGroup for anything

x11/pango/Portfile
```diff
@@ -2,32 +2,25 @@
 
 PortSystem              1.0
 PortGroup               active_variants 1.1
-PortGroup               gobject_introspection 1.0
+PortGroup               meson 1.0
 
 # Please keep the pango and pango-devel ports as similar as possible.
```
ryandesign, line 7: Don't forget to regard this comment. Usually when a new stable version was out I would update pango-devel to a new version first, verify that it built on the buildbot on all OS versions, fix any problems I noticed, wait awhile to see if any bug reports came in about it, and then later update pango.

x11/pango/Portfile
```diff
@@ -2,19 +2,19 @@
 
 PortSystem              1.0
 PortGroup               active_variants 1.1
-PortGroup               gobject_introspection 1.0
+PortGroup               meson 1.0
```
Gcenx, line 5: As you moved to meson and removed the gobject_introspection you need to add that as a build dependency, or are you leaving that being added via harfbuzz

x11/pango/Portfile
```diff
@@ -45,13 +45,14 @@ depends_lib \
...
     port:harfbuzz
 
-patchfiles-append       fix-pango_layout_get_pixel_extents-crash.patch
+configure.args-append   -Dxft=disabled \
+                        -Dintrospection=enabled
 
-configure.args          --enable-static \
-                        --disable-silent-rules \
-                        --without-xft
+configure.cflags-append \
```
Gcenx, line 51: Maybe add a not about why this is required, since this isn’t needed for newer versions of glib

x11/pango/Portfile
```diff
@@ -91,10 +90,14 @@ if {[variant_isset x11]} {
     require_active_variants path:lib/pkgconfig/cairo.pc:cairo x11
 }
 
+pre-configure {
```
Gcenx, line 93: Maybe add a note about why this is required, since this isn’t needed for newer versions of glib

x11/pango/Portfile
mascguy: Updated `pango-devel` via the following commit. (Per the commit message, this was done without a PR, to start validating these changes ASAP.)

https://github.com/macports/macports-ports/commit/51aea14775709205251c6004641f5d3996ed22a0

x11/pango/Portfile
mascguy: @i0ntempest Zhenfu, any thoughts?

mascguy:
> Attempting to build pango with your Portfile on a Tiger box I get a configure error:
> 
> ```
> Run-time dependency xrender found: YES 0.9.10
> Run-time dependency appleframeworks found: NO (tried framework)
> 
> meson.build:373:2: ERROR: Dependency "appleframeworks" not found, tried framework
> ```
> 
> It's using macports-gcc-7. Any ideas?

Hey Evan, I'll happily try to help, once the overall changes are complete. (Still looking at a few things, based on the outstanding comments.)

In the interim, is there a lower-level Meson log with detail?

evanmiller:
@mascguy There is a lower level Meson log, but it contains no other information about “appleframeworks” besides what I posted. I can try a verbose mode if you have any pointers.

mascguy:
> @mascguy There is a lower level Meson log, but it contains no other information about “appleframeworks” besides what I posted. I can try a verbose mode if you have any pointers.

This is the first time I've dealt with Meson before. So until I'm a bit more familiar with it, I won't be much help.

Gcenx:
```
if host_system == 'darwin'
  has_core_text = cc.links('''#include <Carbon/Carbon.h>
                              int main (void) {
                                CTGetCoreTextVersion ();
                                return 0;
                              }''',
                           name: 'CoreText availability',
                           dependencies: dependency('appleframeworks', modules: 'ApplicationServices'))
  if has_core_text
    pango_conf.set('HAVE_CORE_TEXT', 1)
  endif

  pango_deps += dependency('appleframeworks', modules: [ 'CoreFoundation', 'ApplicationServices' ])
endif
```

So I’d guess it’s not finding ApplicationServices

x11/pango/Portfile
i0ntempest: This might be a mistake that I forgot to remove, it was used to force meson to not build glib as a subproject. Since we now have an updated glib this should be able to be removed without problems.

Gcenx: @i0ntempest Isn't that also the case for `-Wno-error,-Wimplicit-fallthrough` as I’ve not needed that for my updated version of pango.

x11/pango/Portfile
jmroot: Given the reasons for the gobject_introspection portgroup's existence, I think you might find that you need to add meson support to it rather than stop using it.

Gcenx: I’d think it would be the other way around, adding gobject_introspection support into the meson portgroup since most protects that moved to meson need it.

There’s some Ports that use the meson portgroup and add in the needed functions to handle gobject_introspection usage.

mascguy:
Update: We finally have a working `pango-devel` - based on this PR - after creating a corresponding Devel version of `harfbuzz`. (It seems that these two are more intertwined than first expected.)

So that's a sigh of relief.

However, we still need to test and validate ports dependent on those. And that will require updating some ports to use a path-style dependency on `harfbuzz`, to support using the Devel version of the latter. (Whether those changes are committed to master, vs. simply testing locally with them, is still TBD.)

x11/pango/Portfile
mascguy: Sounds like further thought is needed, as both may be necessary. We'll figure it out longer-term, once we've finally gotten all of the umpteen Gimp dependencies - including these two - working and committed to master.

mascguy:
FYI: Since Pango 1.49.0 was just released, I updated `pango-devel` to that. And it built on all of the buildbots the first time. Woo Hoo!

Also updated `pango` to 1.49.0, and pushed to this PR's branch.

x11/pango/Portfile
mascguy: I was going to submit a new ticket for this effort, but thankfully @kencu already took care of that:

[60644 - gobject_introspection portgroup: support meson](https://trac.macports.org/ticket/60644)

evanmiller:
Still seeing my `ERROR: Dependency "appleframeworks" not found, tried framework` using the shipping `pango-devel`. The message itself is entirely unhelpful – however, after doing some research, it looks like Meson can't load frameworks using GCC:

https://github.com/mesonbuild/meson/blob/1dbb6d6b8c83cdad0f478b15395103589cbfb9b3/mesonbuild/dependencies/platform.py#L38-L46

Can you see if you can reproduce the issue by compiling (configuring) with GCC? Might need to patch the build file, or require Clang for this port...

mascguy:
Update: Downgraded `pango` to 1.48.9. Ditto for `pango-devel`; the latter committed to master.

evanmiller:
One more issue with pango-devel and meson: the GObject Introspection dylib location is incorrect:

```
$ g-ir-inspect --print-shlibs Pango
shlib: ./pango/libpango-1.0.0.dylib
```

It should look like GLib's, with the full installed path instead of the relative build path:

```
$ g-ir-inspect --print-shlibs GLib
shlib: /opt/local/lib/libglib-2.0.0.dylib
shlib: /opt/local/lib/libgobject-2.0.0.dylib
```

I am working on a patch for Meson to address this.

x11/pango/Portfile
i0ntempest: That cflag was indeed needed to build on my system when I tried, did another version came out? If yes it might not be needed anymore.

Gcenx: I have glib-2.68.4 installed on my end and haven't needed this cflag on Mojave

evanmiller:
The weeds get pretty thick with the GObject Introspection bug – which affects many, many ports – but #12209 should work around it for pango-devel and other Meson-based projects.

evanmiller:
All of my PPC-related concerns have now been addressed 👍

mascguy:
Observers: Since I'm now taking on co-maintainership of these ports - and in the interest of moving forward - I'm merging this. And I'll certainly take ownership of any fallout.

PR #11951, opened 2021-08-21
dovecot: update to 2.3.16

mojca:
Closes: https://trac.macports.org/ticket/63241

#### Description

These changes were mostly written by @gctwnl, but there were still some issues left with the patches which I later cleaned up a bit.

I didn't actually test the functionality, I leave that up to @essandess and @gctwnl.
Please don't merge the PR until either of them or the maintainer have confirmed that this actually works.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [x] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/mail/dovecot/files/dovecot-core-pushnotify.patch
+++ b/mail/dovecot/files/dovecot-core-pushnotify.patch
@@ -1,6 +1,5 @@
-diff -Naur ./README.md ./README.md
---- ./README.md	2020-09-18 20:09:11.000000000 -0400
-+++ ./README.md	2020-09-18 19:57:02.000000000 -0400
+--- ./README.md
++++ ./README.md
 @@ -1,3 +1,53 @@
 +Apple Push Notification Service (APNS) Plugin
 +=============================================
@@ -55,10 +54,9 @@ diff -Naur ./README.md ./README.md
  Installation
  ============
  
-diff -Naur ./configure.ac ./configure.ac
---- ./configure.ac	2020-09-18 20:09:11.000000000 -0400
-+++ ./configure.ac	2020-09-18 19:57:02.000000000 -0400
-@@ -930,6 +930,7 @@
+--- ./configure.ac
++++ ./configure.ac
+@@ -906,6 +906,7 @@
  src/plugins/notify/Makefile
  src/plugins/notify-status/Makefile
  src/plugins/push-notification/Makefile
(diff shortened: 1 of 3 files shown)


mail/dovecot/files/dovecot-core-pushnotify.patch
```diff
@@ -55,619 +54,56 @@ diff -Naur ./README.md ./README.md
...
-++	unsigned long msg;
-++	unsigned long pid;
-++
-++	char d1[128];
-++	char d2[512];
-++	char d3[512];
-++	char d4[512];
-++} msg_data_t;
-++
-++#endif
```
mojca, line 623: @essandess: it looks like this was added by accident?

essandess: This code is necessary for the `apns` variant:

https://github.com/macports/macports-ports/blob/26f5adbdc746e77befbb60579b5bc94ade1c660c/mail/dovecot/Portfile#L240-L339

It's only useful for people who go through the trouble outlines in the notes of keeping their APNS mail push notification certificates up-to-date, so should not be added unconditionally, only as part of the `apns` variant. 

It looks like the thing to do is to update this patch by updating dovecot core and computing a new diff that adds this code. See: https://github.com/essandess/core

I haven't had to update these patches at all because they've all worked with the latest dovecot updates.

Is there a reason the `apns` patches are being updated in this PR?

mojca: I understand that there are patches needed for the `apns` variant, but your patch file is recursive. It first adds a patch for `configure.ac`, but then also creates a **_new_** file `dovecot-core-pushnotify.patch` to patch `configure.ac` in the same way. I'm referring to the super long block that I deleted from `dovecot-core-pushnotify.patch`. Everything from the deleted block was still kept as a regular patch elsewhere in that file. This block can go, right?

> It's only useful for people who go through the trouble outlined in the notes of keeping their APNS mail push notification certificates up-to-date, so should not be added unconditionally, only as part of the apns variant.

While they might only be useful for these people: do the patches hurt the others who don't want to go through the troubles outlined in the notes (and don't activate the `apns` variant)? I'm asking because it's all too easy to update the port without testing all the variant combinations, and miss the fact that the patch no longer applies.

> It looks like the thing to do is to update this patch by updating dovecot core and computing a new diff that adds this code. See: https://github.com/essandess/core

I already did that, and I would be grateful if you were willing to test this PR. (Well, I manually changed the diff rather than patching dovecot and creating a new diff, but that doesn't make any difference.)

> Is there a reason the `apns` patches are being updated in this PR?

* The patch no longer applies to the latest version (for this reason it is really necessary).
* There is a long recursive section (which is likely harmless, but super super confusing) that called for deletion.
* I changed updated line number to make the patch apply cleanly (which is again not strictly necessary, but it's nice to do).
* I removed some bogus timestamps like `1969-12-31 19:00:00.000000000 -0500` (which is not needed to be done at all, but these timestamps are not useful for anything either, and not accurate either).

gctwnl: Yes, it is needed that the `apns` patch is updated. Without Mojca's fixes, dovecot 2.3.16 will not compile because the existing `apns` patch fails.

gctwnl: @mojca @essandess I am now running dovecot 2.3.16 with APNS based on the way Mojca created the patch. So, I can report it works. 

I did encounter the problem again where pushnotify.pl of APNS freezes the file system of my macOS Mojave + Server. When this happens, everything that requires file system access locks up. E.g. a simple 'ls' command will just hang. In the past, I had a period I could not use push notifications because this happened. When I was slowly bringing the mail server live, it happened when I loaded the dovecot-apns launch daemon. My suspicion then was that it had to do with the kernel filesystem subsystem locking up in some way, the culprit being the socket used for APNS in combination with how my setup is configured. For instance, I tend to run things in chroot jails when possible for security reasons. My postfix, for instance, runs chrooted. pushnotify.pl also drops privileges. When I ran pushnotify.pl as root, I did not have this problem. It also might have to do with the fact that pushnotify.pl runs from launchd and starts waiting for access.

So, I needed to hard reset my server 10 minutes ago again, and after the reboot, the system is running, and APNS works, but I am somewhat nervous as in the past I had this 'file system locking up' problem too and it was triggered not immediately. The last year I have been running without trouble and my guess was that for this not to happen, I needed to make sure that the socket for APNS had the right permissions.

My working notes say:

```
BUGHUNTING push notfications

I've seen different permissions on this socket, but this works

bash-3.2# ls -l /opt/local/var/run/dovecot/push_notify
srw-rw-rw-  1 root  wheel  0 Nov 28 15:21 /opt/local/var/run/dovecot/push_notify

This works with pushnotify dropping privileges:

bash-3.2# ls -al /opt/local/etc/dovecot-apns/
total 16
drwxr-x---   5 _dovecot  _dovenull   160 Nov 29 13:48 .
drwxr-xr-x  44 root      wheel      1408 Nov 23 17:20 ..
-rw-r--r--   1 root      _dovenull     0 Nov 23 16:36 .turd_dovecot
-rw-r--r--   1 _dovecot  _dovenull  1952 Sep 13 15:23 com.apple.mail.cert.pem
-rw-r-----   1 _dovecot  _dovenull  1704 Sep 13 15:23 com.apple.mail.key.pem

```

essandess: Okay got it, thanks.

This will require more work on my end because the commits that modify dovecot to use APNS conflict with the latest version. I'll do my best to fix these, generate a new patch file (also without the recursion issue), make sure it works, and post back.

Also, there's an alternate APNS solution that could be used if these Apple-based dovecot patches ever become to difficult to merge with the latest dovecot version: [dovecot-xaps-plugin](https://github.com/st3fan/dovecot-xaps-daemon/issues/46).

> ```
> git pull --rebase --autostash origin master
> From github.com:essandess/core
 > * branch                  master     -> FETCH_HEAD
> Auto-merging src/plugins/Makefile.am
> CONFLICT (content): Merge conflict in src/plugins/Makefile.am
> Auto-merging src/lib-lda/mail-deliver.h
> Auto-merging src/lib-lda/mail-deliver.c
> Auto-merging src/imap/imap-settings.c
> CONFLICT (content): Merge conflict in src/imap/imap-settings.c
> Auto-merging src/imap/imap-client.c
> Auto-merging src/imap/Makefile.am
> Auto-merging src/imap-login/imap-login-settings.c
> CONFLICT (content): Merge conflict in src/imap-login/imap-login-settings.c
> Auto-merging src/imap-login/imap-login-client.c
> Auto-merging src/auth/auth-settings.c
> CONFLICT (content): Merge conflict in src/auth/auth-settings.c
> Auto-merging configure.ac
> Auto-merging README.md
> error: could not apply edacd60ff7... Dovecot core Apple Push Notification Service (APNS) Plugin
> Resolve all conflicts manually, mark them as resolved with
> "git add/rm <conflicted_files>", then run "git rebase --continue".
> You can instead skip this commit: run "git rebase --skip".
> To abort and get back to the state before "git rebase", run "git rebase --abort".
> Could not apply edacd60ff7... Dovecot core Apple Push Notification Service (APNS) Plugin
> ```

essandess: Just to make sure that we’re all on the same page with the APNS patch, please note:
* I have been maintaining this patch with a branch `apns` off a clone of dovecot core that has the working commits. See https://github.com/essandess/core/tree/apns. The MacPorts patch file for the `apns` variant is made from a diff between master and this branch.
* Up until the last version, these `apns` commits did not conflict with dovecot core updates.
* There are now several conflicts in the actual code base, possibly suggesting that the patch file mods in the current PR are inadequate. I haven’t dug into the code yet to determine of these are trivial or substantive conflicts.

mojca: > Okay got it, thanks.
> 
> This will require more work on my end because the commits that modify dovecot to use APNS conflict with the latest version. I'll do my best to fix these, generate a new patch file (also without the recursion issue), make sure it works, and post back.

OK, but can you please at least go carefully through my notes?

The patch above **ALREADY** works. It has already been adapted to the latest release. You are just hitting the exact same conflicts that I already resolved. I can "port" them to your fork of dovecot if you want (I guess you want/need to do that for the future anyway), but there's nothing else that needs to be done about this particular PR other than testing and potentially fixing further issues that might have popped up with the latest version.

(The fastest way for you to "rebase" the changes might be to apply the patch from this PR.)

There are two types of conflicts, both trivial. One file was removed from `Makefile.am` and there was a minor change in the syntax that occurs at two locations.

> Also, there's an alternate APNS solution that could be used if these Apple-based dovecot patches ever become to difficult to merge with the latest dovecot version: [dovecot-xaps-plugin](https://github.com/st3fan/dovecot-xaps-daemon/issues/46).

I leave that up to you to decide.

One further question: is this port a good candidate for port abandonement procedure?

> I have been maintaining this patch with a branch apns off a clone of dovecot core that has the working commits. See https://github.com/essandess/core/tree/apns. The MacPorts patch file for the apns variant is made from a diff between master and this branch.

I didn't know that earlier, but I can give you a commit that fixes that.

> Up until the last version, these apns commits did not conflict with dovecot core updates.
>
> There are now several conflicts in the actual code base

Yes, precisely, this is what me and Gerben have been trying to explain to you since the beginning of the thread.

> possibly suggesting that the patch file mods in the current PR are inadequate.

The patch in this PR should work. What I cannot say is whether there were any further incompatibilities introduced in the latest version beyond what has already been patched be you before.

Btw: did you try to submit the patch upstream?

> I haven’t dug into the code yet to determine of these are trivial or substantive conflicts.

They are all straightforward. See above.

mojca: > @mojca @essandess I am now running dovecot 2.3.16 with APNS based on the way Mojca created the patch. So, I can report it works.

That's great news, thank you.

> I did encounter the problem again where pushnotify.pl of APNS freezes the file system of my macOS Mojave + Server.

From what I understand that's unrelated to the update though, right?

essandess:
I confirm that I can build and run dovecot with working APNS using this patch file:

https://github.com/essandess/core/blob/apns/dovecot-core-pushnotify.patch

This patch file  is from a branch [apns](https://github.com/essandess/core/tree/apns) off a clone of dovecot core, generated with the diff command:
```
diff -Naur --exclude=dovecot-core-pushnotify.patch --exclude=.git core-2.3.16 core-2.3.16-pushnotify | sed -E -e 's/core-([[:digit:]]+\.?)+(-pushnotify)?\//.\//g' > dovecot-core-pushnotify.patch
```

I expect that this is the same functionality as the edited patch file in this PR. I believe we want to generate the patch from git branches, not edit `.diff` files by hand, so I suggest copying the file https://github.com/essandess/core/blob/apns/dovecot-core-pushnotify.patch into `./files`.

Also, would you please change the comment on this line to the diff command above to avoid the patch file issue you mentioned>

https://github.com/macports/macports-ports/blob/061217e00e1be65ba96aeca7d7eb855991e410bc/mail/dovecot/Portfile#L242

> Btw: did you try to submit the patch upstream?

I have not. I expect that upstream will have the same question that I do: is Apple's support of APNS stable enough to warrant adding and maintaining this capability as a part of dovecot? I don't know, and am uncertain enough not to want expend effort to elevate this patch upstream.

Apple currently supports APNS and it works great, so thank you Apple, but the best thing looking forward from an inside perspective is that we "[are not yet doomed :)](https://github.com/st3fan/dovecot-xaps-daemon/issues/46#issuecomment-659190675)."

mojca:
> This patch file is from a branch [apns](https://github.com/essandess/core/tree/apns) off a clone of dovecot core, generated with the diff command:
> 
> ```
> diff -Naur --exclude=dovecot-core-pushnotify.patch --exclude=.git core-2.3.16 core-2.3.16-pushnotify | sed -E -e 's/core-([[:digit:]]+\.?)+(-pushnotify)?\//.\//g' > dovecot-core-pushnotify.patch
> ```

Why would you want to use a highly non-trivial diff command with tons of stuff that you need to document in order to be able to remember it (and a special directory structure that needs to be set up correctly), and which requires special handling, `sed`-ing etc. to avoid recursive inclusion of the file, when you could simply use
```
git diff --no-prefix
```
instead and get everything handled automatically, nicely, out-of-the-box?

mail/dovecot/files/dovecot-core-pushnotify.patch
```diff
@@ -876,28 +309,26 @@ diff -Naur ./src/imap/imap-commands.h ./src/imap/imap-commands.h
...
- 	DEF(SET_BOOL, imap_metadata),
- 	DEF(SET_BOOL, imap_literal_minus),
-@@ -101,6 +102,7 @@
+--- ./src/imap/imap-settings.c
++++ ./src/imap/imap-settings.c
+@@ -75,6 +75,7 @@
+ 	DEF(STR, imap_logout_format),
+ 	DEF(STR, imap_id_send),
+ 	DEF(STR, imap_id_log),
++	DEF(STR, aps_topic),
```
mojca, line 318: @essandess: everything else in your own patch is the same as in my file, except here.

Is adding `DEF(STR, aps_topic)` no longer needed then?

essandess:
Okay your git fu is great than mine. Please use this instead, which excludes the patch file in the branch:
```
git diff --no-prefix master -- ':(exclude)dovecot-core-pushnotify.patch'
```

> Why would you want to use a highly non-trivial diff command with tons of stuff that you need to document in order to be able to remember it (and a special directory structure that needs to be set up correctly), and which requires special handling, `sed`-ing etc. to avoid recursive inclusion of the file, when you could simply use
> 
> ```
> git diff --no-prefix
> ```
> 
> instead and get everything handled automatically, nicely, out-of-the-box?

mail/dovecot/files/dovecot-core-pushnotify.patch
essandess: Oops, thanks for catching that. It's fixed now. Not sure why it builds and works without that definition—I'll go back and sanity check compilation.

essandess:
Whoops, this patch approach throws this error during the patch phase:
> ```
> missing header for unified diff at line 5 of patch
> can't find file to patch at input line 5
> Perhaps you used the wrong -p or --strip option?
> The text leading up to this was:
> --------------------------
> |diff --git README.md README.md
> |index 842f723475..0157bcc16e 100644
> |--- README.md
> |+++ README.md
> ```

> Okay your git fu is greater than mine. Please use this instead, which excludes the patch file in the branch:
> 
> ```
> git diff --no-prefix master -- ':(exclude)dovecot-core-pushnotify.patch'
> ```
> 
>  > Why would you want to use a highly non-trivial diff command with tons of stuff that you need to document in order to be able to remember it (and a special directory structure that needs to be set up correctly), and which requires special handling, `sed`-ing etc. to avoid recursive inclusion of the file, when you could simply use
> > ```
> > git diff --no-prefix
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead and get everything handled automatically, nicely, out-of-the-box?

mojca:
> Whoops, this patch approach throws this error during the patch phase:

I'm sorry, I'm usually editing that part in my text editor, so I wasn't paying a close enough attention. This one should work:
```
git diff master..apns --src-prefix=./ --dst-prefix=./
```
I made some minor changes compared to yours in [my fork](https://github.com/mojca/dovecot-core/tree/apns).

mail/dovecot/files/dovecot-core-pushnotify.patch
gctwnl: No, I am not certain it is unrelated to the update. The update may have reverted something to a situation which triggers the 'filesystem deadlock' issue on my Mojave system.

I recall that there were two approaches to APNS. I also recall that Apple was planning to change the API and retire something, which required a different approach. Getting that different solution to work was something @essandess did. So, I have been wondering if my current situation has reverted to something that can trigger the problem again. So it might be related. It is good that @essandess has a look at it to check if my current situation (with your cleaned patch) is APNS-wise like it was supposed to be with 2.3.13.

My guess is that the problem is triggered by the socket in combination with the fact that privileges are dropped and that that leads to some sort of deadlock, but I have never been able to create the exact situation that will reliably trigger it and I am loath to do so as this is a main production system which is also used for backups for instance.

gctwnl: But it _is_ important for @essandess to know that my current situation with 2.3.16 and APNS does work and does provide push notifications.

gctwnl:
So, is there something I should test now? Stable enough to try? And I assume this _should_ result vin exactly the same situation as I am running now?

mojca:
> So, is there something I should test now?

No, there were absolutely no changes in the patches as far as I have seen.

mail/dovecot/files/dovecot-core-pushnotify.patch
essandess: > that Apple was planning to change the API and retire something, which required a different approach.

No, Apple simply dropped support of the legacy binary protocol to its APNS API. This is independent of the two approaches, and the `apns` variant of this port supports the new API.

https://developer.apple.com/news/?id=c88acm2b

essandess: > But it _is_ important for @essandess to know that my current situation with 2.3.16 and APNS does work and does provide push notifications.

I confirm that APNS works with this update on my mail server. The issues you refer to look like they may be on your end. I recommend comparing your configuration with what the Portfile specifies and/or chasing down logs.

gctwnl:
This morning it turned out I had made a mistake yesterday and my DNS resolver was not running (forgot to turn it on) as well as redis. So a lot of things went wrong (e.g. grey listing by rspamd). So, I did another stop and start of the mail server set (after having made sure `unbound` was running.

Again, I ran into the file system freeze issue and I had to reboot the machine. Starting it up when the system is just after a restart seems to work OK.

What I was wondering, `port load dovecot` does not start the pushnotify service. which listens on that socket. So, I tend to start it 'by hand'. Here is my startup sequence:

```
launchctl load -w /Library/LaunchDaemons/org.macports.dovecot-apns.plist
for i in 	redis \
	dcc \
	rspamd \
	clamav-server \
	apache-solr8 \
	dovecot \
	postfix
do
	port load $i
done

```

Stopping does the reverse.

Now, what I was wondering is this. There is no `port load dovecot-apns` while there is a launch daemon. Am I missing something?

essandess:
> There is no `port load dovecot-apns` while there is a launch daemon. Am I missing something?

Correct. The Portfile currently injects its own plists. Perhaps this should be changed to create them and load both. I honestly forget the design rational for the current setup, but there was some good reason …

https://github.com/macports/macports-ports/blob/26f5adbdc746e77befbb60579b5bc94ade1c660c/mail/dovecot/Portfile#L413-L458

essandess:
@mojca I believe that this PR is ready to merge.

reneeotten:
@mojca @gctwnl @essandess everyone happy with this to be merged or is there anything left to do?

gctwnl:
I have been running this for a few weeks now. It has worked without glitches.

I did encounter the (serious) issue with the APNS part (starting the APNS daemon) but I had the same issue with earlier releases and it is something that hasn’t been triggered on any other system (almost nobody uses this I think), so no reason not to release. It might be specific to my setup or my actions. 

G

essandess:
Yes, I have been running the upgrade with apns without issue for a while. lgtm, please merge.

PR #11955, opened 2021-08-22
tortoisehg: update to 5.8.1

mojca:
#### Description

@danchr: please try to remember to update tortoisehg whenever you update mercurial.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G9016 x86_64
Xcode 10.1 10B61

###### Verification 
Have you

### DIFF
--- a/devel/tortoisehg/Portfile
+++ b/devel/tortoisehg/Portfile
@@ -6,7 +6,7 @@ PortGroup           app 1.0
 PortGroup           gitlab 1.0
 
 gitlab.instance     https://foss.heptapod.net
-gitlab.setup        mercurial/tortoisehg thg 5.8rc0
+gitlab.setup        mercurial/tortoisehg thg 5.8.1
 
 name                tortoisehg
 
@@ -21,9 +21,9 @@ description         A set of graphical tools for Mercurial
 long_description    A set of graphical tools for the Mercurial distributed \
                     source control management system.
 
-checksums           rmd160  d060b630a813c2ea98fdf845e8e74a551711e00a \
-                    sha256  a5ffe4bd6f43bee72199db9664650b4bc4c2afe431d51005a2ba3e1ce130979e \
-                    size    7512399
+checksums           rmd160  f4c9ad52fda164e01f4b60aa6356ec93273c8bbe \
+                    sha256  ccd69bd224c9b1af85e1d8e7c71476a8b2196b11ae18c0cef7a859388e47e638 \
+                    size    7523863
 
 python.default_version 39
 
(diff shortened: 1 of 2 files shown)


devel/tortoisehg/Portfile
```diff
@@ -6,7 +6,7 @@ PortGroup           app 1.0
 PortGroup           gitlab 1.0
 
 gitlab.instance     https://foss.heptapod.net
-gitlab.setup        mercurial/tortoisehg thg 5.8rc0
+gitlab.setup        mercurial/tortoisehg thg 5.8.1
```
mojca, line 9: Just to make sure: does this count as version increase?

danchr: Not sure what you mean? IIRC thg 5.8 wasn't out yet when I bumped Mercurial, so I had to use the rc, although that's generally a bad practice. Any 5.8.x _should_ be compatible, but I assume you tested that?

danchr:
Thanks for doing this! I'm not on my computer right now, but could you check if at least Mercurial and hg-evolve are also up-to-date?

danchr (approved):

mojca:
> Thanks for doing this! I'm not on my computer right now, but could you check if at least Mercurial and hg-evolve are also up-to-date?

You updated mercurial in f96c475273fd499365668027baa0c156fdcd1934.
```
Date:   Fri Jul 9 10:04:46 2021 +0200

    mercurial: update to 5.8.1
```

I have no clue how to test `hg-evolve` or what the version dependencies are. You updated it in 7b308c9fb64f8809c0014eeb0ffcf6ffc4a19b9f to 10.3.2 and now `livecheck` says
```
hg-evolve seems to have been updated (port version: 10.3.2, new version: 10.3.3)
```
but if it worked fine back then when you updated it, I guess it must still be working fine? I would prefer to leave the updating of that port to you.

A newer version of Mercurial exists, but it's still in RC, so I wouldn't want to update it just yet.

devel/tortoisehg/Portfile
mojca: (That question wasn't about why you changed anything. Yes, ideally we would want to avoid switching to RC, but here my question is different.)

Assuming someone already had 5.8rc0 installed: does MacPorts see 5.8.1 as an update compared to 5.8rc0 or not?

I didn't test whether tortoisehg 5.8rc0 was compatible with mercurial 5.8.1, I just remember that thg was very often broken when mercurial was updated without updating thg as well. Maybe that was true for major version changes only, but since the update is there, there's no reason not to update it to a totally compatible version.

danchr: I just ran a similar check; Mercurial 5.9 is actually released, it appears — at least to PyPI. I had an earlier, uncommited version of Mercurial 5.9rc0 installed, and it didn't upgrade it. In this case, we should probably either upgrade Mercurial to 5.9 now, or wait until TortoiseHg 5.9 is released, and upgrade _then_.

danchr:
Could you perhaps test #11962 for me? Pango is currently broken with the `+quartz` variant, so I haven't been able to test TortoiseHg 5.9 myself.

danchr:
I've merged #11962 instead, updating everything to 5.9.

PR #11980, opened 2021-08-25
kustomize: update to 4.3.0

herbygillot:
###### Tested on

macOS 10.15.7 19H1323 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/devel/kustomize/Portfile
+++ b/devel/kustomize/Portfile
@@ -3,13 +3,13 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/kubernetes-sigs/kustomize 4.2.0 kustomize/v
+go.setup            github.com/kubernetes-sigs/kustomize 4.3.0 kustomize/v
 revision            0
 
 categories          devel
+installs_libs       no
 maintainers         {breun.nl:nils @breun} openmaintainer
 platforms           darwin
-supported_archs     x86_64
 license             Apache-2
 
 description         Customize raw, template-free YAML files for multiple purposes, like Kubernetes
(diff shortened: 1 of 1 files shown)


breun:
Looks good to me, but I'm not familiar with `installs_libs` and I don't see that keyword documented in the MacPorts guide.

Also, what is the reason for setting `github.tarball_from archive`? According to https://guide.macports.org/chunked/reference.portgroup.html#reference.portgroup.github maintainers are encouraged to use `github.tarball_from releases`.

herbygillot:
> Looks good to me, but I'm not familiar with `installs_libs` and I don't see that keyword documented in the MacPorts guide.

Basically `installs_libs` says that this package does not export any shared libraries, which may matter when linking between software with potentially incompatible licenses.  This probably should be set to `no` for all Go ports.  It is documented in the `portfile` man page:
```
     installs_libs
         By default, it is assumed that ports may install libraries or headers that can be incorporated into their dependents.  If this is not the case, set
         installs_libs to no.  This means that this port's dependents need not check that it is installed for the same architectures as them; that it is permis-
         sible to distribute binaries of the dependents even if their licenses conflict with the license of this port; and that updates to this port can never
         result in broken dynamic linking in its dependents.
```

> Also, what is the reason for setting `github.tarball_from archive`? According to https://guide.macports.org/chunked/reference.portgroup.html#reference.portgroup.github maintainers are encouraged to use `github.tarball_from releases`.

The standard download links used by the `github` portgroup are "tarball"-style URLs, which Github wants to move away from.  The current default links generated to download assets from repositories in Github are `archive` style URLs.  There are also releases, but releases are manually triggered by repository owners, where archive URLs will automatically exist for every branch by default.

If you take a look at the source tgz download link for the [current `kustomize` release](https://github.com/kubernetes-sigs/kustomize/releases/tag/kustomize%2Fv4.3.0), you'll see that it is an archive-style URL: `https://github.com/kubernetes-sigs/kustomize/archive/refs/tags/kustomize/v4.3.0.tar.gz`, and that's what

breun:
Ok, thanks for the explanation.

PR #12005, opened 2021-08-28
python3*: add dtrace variant

jrabinow:
#### Description

Adds a variant to use dtrace as described in
https://docs.python.org/3/howto/instrumentation.html

Also see https://github.com/macports/macports-ports/pull/6308

Fixes https://trac.macports.org/ticket/54443

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 11.5.2 20G95 x86_64
Xcode 12.5.1 12E507

###### Verification 
Have you

### DIFF
--- a/lang/python310-devel/Portfile
+++ b/lang/python310-devel/Portfile
@@ -232,6 +232,10 @@ variant optimizations description {Compile with LTO and PGO. Build time greatly 
     configure.args-append   --enable-optimizations
 }
 
+variant dtrace description {enable DTrace support} {
+    configure.args-append   --with-dtrace
+}
+
 livecheck.type      regex
 livecheck.url       ${homepage}downloads/source/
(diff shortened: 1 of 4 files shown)


jmroot:
Did you test all of the versions with the variant enabled? The configure option doesn't seem to exist in < 3.6.

jrabinow:
It looks like you're right, https://docs.python.org/3/howto/instrumentation.html says that this option was added in 3.6. I had built 3.2 and 3.9 with the flag, assuming that if it worked for both it would work for everything in between. The builds went through just fine but I must have checked improperly for 3.2.
I've updated the commit to only include the flag on 3.6+, and built and tested each version separately. Apologies for the mix-up.

jmroot:
Thanks. I added a Fixes: line to the commit message so the ticket will be closed automatically.

PR #12013, opened 2021-08-29
libuv: update pre-10.7 to 1.42.0

evanmiller:
#### Description

Update the pre-Lion felines to libuv 1.42.0 for version parity across OSes. Passes a great majority of the tests on a Tiger box, but needs testing on the Leopards so leaving WIP for the moment. I've consolidated existing patches and updated to use `MAC_OS_X_VERSION_MIN_REQUIRED` instead of `MAC_OS_X_VERSION_MAX_ALLOWED` to permit back-compiling to older OSes. Upstream needs a couple of changes in this respect.

(However, I did not modify patch-libuv-unix-core-close-nocancel.diff with `MIN_REQUIRED` because I didn't understand the implications.)

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you


kencu:
I'll give you the rundown on MAX ALLOWED in a bit. It was correct as I wrote it, I believe.

Thanks!

kencu:
hmm. no real changes at all from what I had? It just updated with a version bump?

That would be a surprise! libuv has not been so friendly in the past!

I'll likely want to hang on to the 8, 9-10, and 11+ setup because of frequent breakage in the past (and likely future) btw.

kencu:
Here's a teaser though - busy day so I can get this later -- but why do you think ```MAC_OS_X_VERSION_MAX_ALLOWED``` exists, if in every case you want to use ```MAC_OS_X_VERSION_MIN_REQUIRED``` instead?

When would it be correct to use ```MAC_OS_X_VERSION_MAX_ALLOWED```  ever? 

The best explanation for this somewhat confusing set of macros that I have ever read comes from Chris Espinosa, one of Apple's first 10 employees, here:

https://lists.apple.com/archives/xcode-users/2005/Aug/msg00399.html

evanmiller:
> hmm. no real changes at all from what I had? It just updated with a version bump?

`patch-libuv-legacy.diff` consolidates your previous work and adds changes to darwin-proctitle.c, udp.c, and test-fs.c

evanmiller:
> When would it be correct to use `MAC_OS_X_VERSION_MAX_ALLOWED` ever?

Sure, MAX_ALLOWED is for the SDK and MIN_REQUIRED is for the run-time environment. For system calls, MIN_REQUIRED is necessary or you'll get run-time linkage problems. Eg. we definitely want MIN_REQUIRED for `pthread_setname_np`. Happy to switch the others back since it looks like they're just structs and #defines.

kencu:
So I'm going to keep three separate version setups, for darwin 8, 9-10, and 11+. Every time Michael updated libuv, everything broke. So that will stay.


The MAC_OS_X_VERSION_MAX_ALLOWED macro refers to what is in the SDK you are using. The MAC_OS_X_VERSION_MN_REQUIRED refers to the system you are deploying on.

see for example this patch I made for python for how it is supposed to be done to properly support current and deploying on older systems:

https://github.com/macports/macports-ports/blob/master/lang/python39/files/patch-threadid-older-systems.diff

I want to leave all the other patchfiles just as they are, in general, for now at least. We can talk about  your thoughts on tweaking min required and consolidating patches, etc, at another time.

BUT - if you have an update for Tiger, and a patchfile change if needed, I'd be happy to look at that! Much appreciate the help!

kencu:
Evan, thanks. I've downloaded your patch, and I'll see if I can use it to integrate the changes needed.

I'll give you full credit for the patches of course, and thanks a lot!

Will close this PR now while I work on seeing what was needed to make the newer version go on Tiger, and push that, rather than go back-and-forth for days about it.

evanmiller:
@kencu Works for me, happy to help

evanmiller:
@michaelld Would you mind taking another look at this? It appears that Ken has stepped away from this package, and the changes we discussed were never merged.

michaelld:
> I'm sure this version will work (although I didn't actually test it) on Tiger.
> 
> I was just asking Evan to keep the sequestered version for Tiger to make life simpler for Michael, as it breaks on every libuv update and I knew Michael would not want to have to deal with that every time he updates libuv.

Ah ... OK that makes a LOT of sense to me. I agree with Ken here!

@evanmiller Let's keep the older-specific version code around and update it as testing shows that it works. Let's _not_ merge it into the general Portfile.

kencu:
Michael, I'll be happy to PR something for you once I test it on a few systems.

kencu:
Michael -- as it goes, most of my older systems have been converted to various flavours of linux over this past week. I had hoped to set this up for you to be less trouble on updates, but I don't seem to be likely to get to it.

You might then just merge this as Evan wants, and if it breaks on the next libuv update (as was the case for so long when it broke on every update), Evan will probably be available to fix it. Maybe you'll get  lucky, and it will be less fragile for a while!

evanmiller:
> Sure -- I upstreamed the last 5 updates or so to libuv, and they were receptive in general.

I'll start with the 10.5/10.6 bits. So far most upstream projects have been sufficiently astonished that their code runs on Tiger that they've been amenable.

I'm sorry we're losing you to the light side of Linux – for my last trick I was hoping to use your GCC8 work to get an ancient Rust / LLVM combo rolling on PPC. Right after `webkit2-gtk` compiles...

PR #12016, opened 2021-08-30
openssl: Fix build on 10.4, RNG on 10.10-10.11

fhgwright:
The patch for pre-Yosemite OS versions relied on Availability.h, which
didn't exist in 10.4.  This updates the patch by substituting the
__ENVIRONMENT... versions of the macros to avoid the need for the
header, as well as eliminating the superfluous defined() checks.

In addition, the RNG from Apple crypto apparently doesn't work
correctly prior to 10.12, so this updates the version threshold for
its use.  Because this changes code on 10.10 and 10.11, it includes a
revbump.  Dependents should only require revbumps to pick up the fix
if they link statically.

It also renames and adjusts the patch to conform to MacPorts style.

Closes: https://trac.macports.org/ticket/63421

TESTED:
Built on 10.4-10.5 ppc, 10.4-10.6 i386, and 10.6-10.15 x86_64.
Verified that libcrypto references _CCRandomGenerateBytes in OS
versions >=10.12 and not otherwise.
Tests pass on 10.4-10.5 and 10.9-10.15.
Test framework has preexisting issues on 10.6-10.8.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you


fhgwright:
I just pushed a rebased version on general principles, though the content was unchanged.

kencu:
looks good. As per the ticket where this was discussed, we were just going to add legacysupport and be done with it, but this fix is also neat and tidy. There are always three ways to do anything.
 
I like it.

evanmiller:
Perhaps a separate PR but 10.11 might need this flag too:

https://trac.macports.org/ticket/63421

kencu:
for that ticket, it looks like the patch might best be tweaked by one system and that fix upstreamed, if it's reproducible.

fhgwright:
@kencu 
> for that ticket, it looks like the patch might best be tweaked by one system and that fix upstreamed, if it's reproducible.

My original fix here was to make the `Availability.h` include conditional on `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` (with its iOS equivalent blindly thrown in as well), but that didn't work for some nonobvious reason.  Rather than debug it, I decided to switch to the approach here, which has the advantage of being highly focused and low-risk.  But the other approach would be better as an upstream fix, to minimize the need for "MacPorts magic" when building.

I probably would have discovered this if I'd run `port test openssl` on 10.11, but I didn't see the need under the circumstances.

It would also make sense for upstream to have just one place (perhaps as a configure test) where it decides whether Apple crypto is available (or perhaps what version of Apple crypto), and then uses that result everywhere.  Meanwhile, the existing patch could be extended to cover the test as well, and the tweak here would make it work for 10.4.

In any case, _this_ PR fixes the more serious (in some sense) problem of not building at all on 10.4.

fhgwright:
I realized today why my earlier patch didn't work, so I'm trying that now, since it's more appropriate as an upstream fix.  It's a larger change to the port, but a smaller end result since it simplifies the patch.

I also tried "port test" across a bunch of OS versions, with some rather weird results:  10.4-10.5 pass, 10.6-10.8 fail, 10.9-10.11 pass, 10.12 fails, 10.13 passes.  10.14-10.15 not yet tested since I have to reboot for those, and the PowerMac tests aren't finished.  Note that 10.11 passed here.

A bit of perusing of the log shows that it uses perl for the tests, but doesn't declare any perl dependency, so it probably just uses whatever perl happens to be the default.  That may matter more than the OS version.

fhgwright:
Well, the perl version theory doesn't hold up, since all my VMs are on v5.28.3.

fhgwright:
I replaced this change with the other one mentioned above, which is a larger change but a smaller end result, and more suitable for upstreaming.  A better view of the change is here:
[https://github.com/fhgwright/openssl/commit/0c83990acaf20e0f69d25b8c61ae3c0f1d47cd9b](url)

This doesn't address the test failures mentioned in [https://trac.macports.org/ticket/63421](url),
but I think those are a separate issue.

kencu:
@fhgwright -- I think your ```ENVIRONMENT`` defines are not spelled right.

You have:
```
__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED
```
but clang and gcc look like this:
```
 % clang -dM -E - < /dev/null | grep ENVIRONMENT
#define __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ 110000
#define __ENVIRONMENT_OS_VERSION_MIN_REQUIRED__ 110000
```

fhgwright:
@kencu 
> @fhgwright -- I think your `ENVIRONMENT` defines are not spelled right.
> 
> You have:
> 
> ```
> __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED
> ```
> 
> but clang and gcc look like this:
> 
> ```
>  % clang -dM -E - < /dev/null | grep ENVIRONMENT
> #define __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ 110000
> #define __ENVIRONMENT_OS_VERSION_MIN_REQUIRED__ 110000
> ```

You're right, even though my own test program for definitions of that form has the correct spelling.  I verified that the change here unbroke the 10.4 build, but didn't verify that it didn't disable the use of Apple Crypto in >=10.10.

Based on what I found online, the iPhone versions of those macros do *not* have the trailing underscores, though I don't have any iOS build setups for testing.

I'll have to fix that and retest.

kencu:
Thanks. I note that ooks like upstream also found out that their patch needed tweaking, as we did in our ticket:

https://trac.macports.org/ticket/63421

Please see:

https://github.com/openssl/openssl/commit/2f3b120401533db82e99ed28de5fc8aab1b76b33

and 

https://github.com/openssl/openssl/issues/16517

basically, they upgraded their cutoff to ```101200``` . 

I have no interest in making you do more work here, but what do you think?

fhgwright:
@kencu 
> Thanks. I note that looks like upstream also found out that their patch needed tweaking, as we did in our ticket:
> 
> https://trac.macports.org/ticket/63421
> 
> Please see:
> 
> [openssl/openssl@2f3b120](https://github.com/openssl/openssl/commit/2f3b120401533db82e99ed28de5fc8aab1b76b33)
> 
> and
> 
> [openssl/openssl#16517](https://github.com/openssl/openssl/issues/16517)
> 
> basically, they upgraded their cutoff to `101200` .
> 
> I have no interest in making you do more work here, but what do you think?

I think it's worth looking into, though I don't have time tonight.  I don't see a need to mark this PR as draft, since merging it as is wouldn't make anything worse, but an updated version may be better.  If the other adjustment is needed, then there should be a revbump to fix the broken 10.10 and 10.11 cases.

Oddly enough, the previous "broken" version of my fix would have fixed this problem by avoiding Apple crypto across the board (as previous OpenSSL versions did, anyway).

I see the OpenSSL folks still haven't figured out that the defined() tests are redundant. :-)

fhgwright:
Indeed there was an RNG test failure on 10.10 and 10.11, which is fixed by moving the version threshold.

All tests are broken on 10.6-10.8, including in the previous version, but this doesn't address that (old) problem.

kencu:
@neverpanic -- any objections to committing @fhgwright 's fixes here?

devel/openssl/Portfile
```diff
@@ -52,8 +53,8 @@ if {${os.platform} eq "darwin" && ${os.major} < 11} {
 
 # https://github.com/macports/macports-ports/commit/9ce0f93609adf300e84bdd599366a3d4ab05215d
 # https://github.com/devnexen/openssl/commit/cef404f1e7a598166cbc2fd2e0048f7e2d752ad5
```
neverpanic, line 55: These two lines can probably be removed now that you're removing the patch referenced here?

neverpanic (changes requested):
Small nitpick on the comment, lg2m otherwise. Thanks for the extensive testing!

devel/openssl/Portfile
fhgwright: The first one seems to simply reference the previous update to the port, which seems like a strange thing to have included, anyway.  But the second one references the quasi-upstream patch, which is modified but not removed here, hence the "patch updated" comment.

fhgwright: I updated it to reference the two relevant upstream commits from the actual upstream repo, plus the comment about fixing the upstream version for 10.4.

I only retested the comment-only change on one system. :-)

kencu:
OK, let's go.

PR #12018, opened 2021-08-30
ruby26: Fix build issues on 10.4, 10.5.

fhgwright:
This includes a patch for some OS version issues common to earlier
Ruby versions:

The OS version conditional in dln.c is comparing an undefined macro to
a sometimes undefined (and essentially useless) macro.  However, the
relevant code doesn't seem to be built for any OS version, making the
bug irrelevant.  Nevertheless, it's now fixed, since another, similar
fix was needed, anyway, and it might become relevant if something else
changes.

The OS version conditionals in error.c were using another useless
macro for the comparison, this time with a real effect, though only in
a couple of messages related to crash reports on < 10.6.  This is now
fixed.

It also includes an updated patch for a 10.4 issue:

The context register naming issue on 10.4 was sort of fixed upstream
but tied to an inappropriate config flag.  The patched version uses an
SDK version macro to make the decision.

A newly discovered issue, common to some older Ruby versions, is:

The 10.5 x86_64 build fails if libunwind-headers is active.  Since
this is an unusual configuration, and the need for libunwind-headers
seems to be waning, it's not worth fixing "right".  This just adds the
usual failure for such cases.  It's unknown why this only affects 10.5
x86_64 (and perhaps 10.4 x86_64, though that's untested).

The remaining issues were new in ruby26:

The syntax-checking compiler invocations in transform_mjit_header.rb
require a C99 or later compiler, but don't include compiler flags from
the build.  There's code to detect the non-C99 case and add -std=c99
to the real uses, but it relies on matching the error message from the
test case.  Recent versions of gcc use locale-dependent quotes which
may break such comparisons, depending on the current locale settings.
The fix recommended by gcc documentation is to run it with LC_ALL=C to
get traditional ASCII quotes.  This adds code to apply that
environment to the compiler test invocation.

For some strange reason, using gcc-4.2 for Ruby's coroutine-related
assembler code on x86_64 passes "-arch i386" to the assembler, thereby
causing failures on x86_64 register references.  To fix that, cases
with gcc-4.2 and x86_64 targeted now pass the correct architecture in
ASFLAGS.

Also corrects a small typo in the description.

Also fixes some lint errors in the Portfile.

Since this affects the installed code for < 10.6, it includes a
revbump.

Universal builds still fail on 10.4 i386 due to a weird configure bug,
and 10.4 ppc has broken universal dependencies.

TESTED:
Built -universal on 10.4-10.5 ppc, 10.4-10.6 i386, and 10.6-10.15
x86_64.  Built +universal on 10.5-10.13.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

- [N/A] tried existing tests with `sudo port test`?

### DIFF
--- a/lang/ruby26/Portfile
+++ b/lang/ruby26/Portfile
@@ -5,7 +5,7 @@ PortGroup           select 1.0
 
 name                ruby26
 version             2.6.8
-revision            1
+revision            2
 
 categories          lang ruby
 maintainers         {kimuraw @kimuraw} openmaintainer
@@ -16,7 +16,7 @@ long_description    Ruby is the interpreted scripting language for quick \
                     and easy object-oriented programming. It has many \
                     features to process text files and to do system \
                     management tasks (as in Perl). It is simple, \
-                    straight-forward, extensible, and portable.
+                    straightforward, extensible, and portable.
 
 homepage            https://www.ruby-lang.org/
 license             {Ruby BSD}
@@ -29,7 +29,8 @@ dist_subdir         ruby26
 checksums           md5 c53761123d17e929cfe248f50429bcab \
                     rmd160 734337856050cb238e0b81280364b614ee02b56d \
                     sha1 7d38cacb6a0779f04b9f19f94406da97e95bbec4 \
-                    sha256 dac96ca6df8bab5a6fc7778907f42498037f8ce05b63d20779dce3163e9fafe6
+                    sha256 dac96ca6df8bab5a6fc7778907f42498037f8ce05b63d20779dce3163e9fafe6 \
+                    size 14131671
 
 use_parallel_build  no
 
@@ -59,12 +60,20 @@ configure.args      --enable-shared \
 # patch-configure_cxx11.diff: fix "invalid suffix on literal" with C++11
 #                             from RUBY_ARCH and RUBY_PLATFORM in config.h
 #                             https://trac.macports.org/ticket/58255
+# Note that this patches the output of autoconf, rather than the input, but
+# at present, rerunning autonconf produces a broken configure.
 patchfiles          patch-configure_cxx11.diff
 # Fix test failure at ext/fiddle
 patchfiles-append   patch-test-fiddle-helper.rb.diff
 # avoid build error with apple clang 12 (implicit declaration of function)
 # https://bugs.ruby-lang.org/issues/17777
 patchfiles-append   patch-ruby_issue_17777.diff
+# Fix build on 10.4
+patchfiles-append   patch-tiger.diff
+# Fix build on < 10.6
+patchfiles-append   patch-osversions.diff
+# Fix build with pre-C99 compilers and some locale settings
+patchfiles-append   patch-c99check.diff
 
 # [NOTE] workaround for mismatch of sdk versions on macOS 11.x,
 # such as MacOSX11.0.sdk (buildbot) <=> MacOSX11.1.sdk (user's Mac).
@@ -77,11 +86,11 @@ post-build {
         # rewrite MacOSX11.1.sdk -> MacOSX.sdk in rbconfig.rb
         file copy ${worksrcpath}/rbconfig.rb ${worksrcpath}/rbconfig.rb.orig
         reinplace -E -q {s|(/MacOSX)[0-9\.]+(\.sdk[\'\"]?[[:blank:]])|\1\2|g} \
-		    ${worksrcpath}/rbconfig.rb \
-		    ${worksrcpath}/ruby-2.6.pc
+            ${worksrcpath}/rbconfig.rb \
+            ${worksrcpath}/ruby-2.6.pc
         reinplace -E -q {s|(/MacOSX)[0-9\.]+(\.sdk[\'\"]?)$|\1\2|g} \
-		    ${worksrcpath}/rbconfig.rb \
-		    ${worksrcpath}/ruby-2.6.pc
+            ${worksrcpath}/rbconfig.rb \
+            ${worksrcpath}/ruby-2.6.pc
     }
 }
 
@@ -161,6 +170,21 @@ platform darwin {
         build.cmd               ${prefix}/bin/gmake
         configure.args-append   --disable-dtrace
         configure.cflags-append -std=c99
+    }
+
+    # Xcode gcc 4.2 building x86_64 assembler code passes "-arch i386" to the
+    # assembler, breaking x86_64 builds, unless we override it.
+    if {[string match *gcc-4.2 ${configure.compiler}]
+        && "x86_64" in [get_canonical_archs]} {
+        configure.env-append    ASFLAGS=[get_canonical_archflags]
+    }
+
+    # Building for 10.5 x86_64 fails if libunwind-headers is active.
+    # It's unknown why only this one case fails.
+    if {${os.major} == 9 && "x86_64" in [get_canonical_archs]} {
+        PortGroup    conflicts_build 1.0
+
+        conflicts_build-append libunwind-headers
     }
 
     if {${os.major} == 8} {
--- a/lang/ruby26/files/patch-c99check.diff
+++ b/lang/ruby26/files/patch-c99check.diff
@@ -1,0 +1,13 @@
+--- tool/transform_mjit_header.rb.orig	2021-07-07 03:38:58.000000000 -0700
++++ tool/transform_mjit_header.rb	2021-08-17 16:03:15.000000000 -0700
+@@ -176,7 +176,9 @@ module MJITHeader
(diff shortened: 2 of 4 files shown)


kencu:
I use something like this:

https://github.com/kencu/TigerPorts/blob/master/lang/ruby25/files/patch-ruby25-tigeri386.diff

This PR looks very complicated.

kencu:
BTW, I locally overrode the changes made last year in the rubys for Tiger, as I think the patch I referenced is the more correct fix for those Tiger hiccups.

fhgwright:
@kencu:
> I use something like this:
> 
> https://github.com/kencu/TigerPorts/blob/master/lang/ruby25/files/patch-ruby25-tigeri386.diff
> 
> This PR looks very complicated.

Why am I not surprised that you chimed in on this ? :-)

As far as this particular fix goes, mine is simpler than yours.  The overall PR has a lot because there were multiple issues to fix, all of which are explained in the commit message.  I originally had multiple commits (starting with one inheriting the patches from `ruby22-25`), but squashed them for the PR in keeping with MacPorts policy.

My Tiger patch simply extends the scope of the existing conditional for an optional feature, to cover a non-working case that it had missed.  Because of the nature of the change, a successful build is a sufficient test.  Your fix involves actually making the context stuff work on 10.4 i386, which then requires actually testing that feature.  I don't know how to test that, and of course there's no "port test ruby26" at all.  Do you have a test for it?  _I_ don't want to push a change that _I_ don't know how to test properly, and it doesn't make sense to make the perfect the enemy of the good by leaving `ruby26` in its current state where it (and consequently asciidoctor) is broken in all 10.4/10.5 cases.  You can always follow up with a better fix, if you have a way to properly test it, though if it's really that important, it ought to be fixed for PPC as well, particularly given that 10.4 ppc is probably far more common than 10.4 i386.

About a year ago, I attacked `ruby24` because it was needed for `asciidoctor`.  I figured that as long as I was working on `ruby24`, I might as well fix as many other ruby versions as possible.  I wound up fixing `ruby22-25`, but `ruby26-27` presented additional problems that looked hard to fix, so I left it at that.  But then someone moved `asciidoctor` to `ruby26` for arm64 compatibility, reintroducing the old 10.4/10.5 breakage.  I was right about "hard to fix", since the `transform_mjit_header.rb` bug was a royal PITA to track down, including a time when the bug was literally staring me in the face and I didn't see it.  After the dust has settled, I may revisit `ruby27` in light of what I've learned.


BTW, it occurs to me that the issue with `libunwind-headers` on 10.5 x86_64 may also apply to 10.4 x86_64, but I don't have a VM set up to test that case.  I believe that 10.4 x86_64 is theoretically possible, but too weird to be worth the hassle.  As it is, both 10.5 x86_64 and 10.6 i386 have proven to be "interesting".

fhgwright:
I rebased this with a new revbump, due to the collision with the libffi-induced revbump.

kencu:
I've been running ruby26 on Leopard and Tiger for several years, using 2 simple fixes in this repo:

https://github.com/kencu/TigerPorts/tree/master/lang/ruby26

please take a look. It's much much much much simpler than this, and I think a better fix.

kencu:
converted to draft because I don't think this should be accidentally merged until other options are looked at that are much simpler to maintain going forward

kencu:
OK, so according my last test, ruby26 is already working on 10.4 PPC, 10.5 PPC, and (I think) 10.5 Intel (have to triple check that) and to fix the build on 10.4 Intel I did only this:

```
    if {${os.major} < 10} {
        depends_build-append    port:gmake
        build.cmd               ${prefix}/bin/gmake
        configure.args-append   --disable-dtrace
        configure.cflags-append -std=c99
    }

    if {${os.major} == 8} {
        configure.cppflags-append -DCPU_SUBTYPE_MASK=0xff000000

+        # force old register names #11509 
+       configure.cppflags-append -D__DARWIN_UNIX03=0
    }
```
I didn't push that last tiny bit to MacPorts as yet.

None of the rest of the PR was needed, and the test suite passes nicely for me

So I will have to see what is up here with this PR and if something has changed somehow to require all this.

kencu:
> @kencu:
> 
> > I use something like this:
> > https://github.com/kencu/TigerPorts/blob/master/lang/ruby25/files/patch-ruby25-tigeri386.diff
> > This PR looks very complicated.
> 
> Why am I not surprised that you chimed in on this ? :-)

Nice.

I simply don't think you have so far understood the different register names in Tiger i386 vs newer systems (a very simple change) that is much much simpler than your ruby fixes to date.

kencu:
actually I change my mind. I don't use any of these ruby ports in MacPorts on older systems anyway, I use my own which work perfectly and I can understand them, so it matters not at all to me if this gets merged or not as I won't be using it anyway.

Fire away.

kencu:
https://github.com/macports/macports-ports/pull/12181

kencu:
Fred, I was using a slightly older ruby26. I'll try my fix on the current version and see if it still works now.

kencu:
everywhere else in MacPorts, we use the proper regnames fix for the Tiger SDK.

eg.
https://github.com/macports/macports-ports/commit/f9641f5f5cb7287e6787b566945fff460a80bc39#diff-c5cf17e504a1536e91626cc4225a8486d18080f47a3e3f6f1653cb4fddadea37

can you please take a look at that, and other similar fixes throughout MacPorts, eg cctools, and possibly chang yours to match that, rather than incorrectly blicking out the entire section with a system version block?

I would be most appreciative if you did thst here, and whenever else you come across this issue, and it will prevent me from having to come back later and fix it, or worse judt ignore it and leave it wrong.

kencu:
typos due to ipad fingers :) I'm splurging for a new ipad this Xmas, and then I hopefully will be able to edit in github, which I cannot presently do.

fhgwright:
I'll take a look, but probably not today.

Overall, it isn't just about the register names - every change here represents something I found, fixed, and explained.

Calling the `libunwind-headers` issue new in `ruby26` is actually inaccurate - it's just that I hadn't encountered it before (it also affects `ruby[2-5]` at least).  Whether I currently have `libunwind-headers` active has tended to depend on whether I last built `cctools` or some version of `gcc`, though now I'm starting to make a point of having it active when testing ports in order to catch such things.  For `ruby2[2-6]` (and perhaps others), I have no idea why it only matters for 10.5 x86_64, but that's what I see.

fhgwright:
I've updated this.

The 10.4 context register bug had an upstream fix (that happened later than when I was working on this last year), but it was tied to `__DARWIN_UNIX03`, which is normally set up by `cdefs.h`, but not in an appropriate way for this use.  Defining it directly has potentially far-reaching side effects, so I replaced that check with an SDK version check.

I removed the `configure.ac` consistency patch, since rerunning `autoconf` results in a completely broken `configure` anyway.  I added a comment to that effect in the Portfile.  Note that the current `configure` patch isn't upstreamable, since patching the output of `autoconf` isn't acceptable as an upstream fix.

It turns out that `configure` isn't the only generated file that doesn't rebuild correctly on the Mac, and make rebuilds them (often incorrectly) any time they look outdated based on mtime comparisons.  Fortunately, the tarball has consistent mtimes that don't suffer from this, but it causes intermittent failures when working directly in a git repo, due to git's propensity for clobbering mtimes (which will never be fixed since Linus considers it to be a feature).

There are at least two bugs in the tests that make it premature to enable the test phase:

1. Something which isn't even strictly a test issue is that using `Dir.mktmpdir` to create Unix domain socket paths easily exceeds the hard-coded OS socket-name length limit when the current directory has a verbose pathname (as is usually true of MacPorts builds).

2. I've never seen the tests run to completion without the test runner hanging.

@kencu 

> everywhere else in MacPorts, we use the proper regnames fix for the Tiger SDK.
> 
> eg.
> [f9641f5#diff-c5cf17e504a1536e91626cc4225a8486d18080f47a3e3f6f1653cb4fddadea37](https://github.com/macports/macports-ports/commit/f9641f5f5cb7287e6787b566945fff460a80bc39#diff-c5cf17e504a1536e91626cc4225a8486d18080f47a3e3f6f1653cb4fddadea37)

Note that that patch is not in an upstreamable form, since it relies on a conditionally applied patch.

kencu:
> Note that that patch is not in an upstreamable form, since it relies on a conditionally applied patch.

Tiger is unusable without MacPorts' newer compilers, cctools, ld64, meson, and gcc7/libgcc7 integration anyway, so I am not going to hold people's feet to the fire about Tiger patches being upstreamable -- I'd be too embarrassed to open the PR :>

kencu:
this looks good now, thanks for redoing it.

@kimuraw -- LGTM.

kencu:
Thanks, @fhgwright !

PR #12044, opened 2021-09-01
brlaser: new port

evanmiller:
#### Description

[`brlaser`](https://github.com/pdewacht/brlaser) is an open-source CUPS driver for Brother monochrome laser printers, useful on systems where the proprietary driver is not available. The driver is packaged on FreeBSD as [print/brlaser](https://www.freshports.org/print/brlaser) and on Debian as [printer-driver-brlaser](https://packages.debian.org/sid/text/printer-driver-brlaser).

This port differs from the above packages in that it compiles and installs the PPD files themselves (with help from [`cups-ppdc`](https://ports.macports.org/port/cups-ppdc/details/)) instead of leaving it as a task to the user. The PPD files are installed to the same location that `cups-pdf` uses.

The notes field instructs end users on setting up a printer for the first time. Patches are included to assist with compilation and to enhance duplex printing – submitted upstream but the project is not especially active.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/print/brlaser/Portfile
+++ b/print/brlaser/Portfile
@@ -1,0 +1,81 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem      1.0
+PortGroup       github 1.0
+PortGroup       cmake 1.1
+
+github.setup    pdewacht brlaser 6 v
+revision        0
+categories      print
+license         GPL-2+
+maintainers     nomaintainer
+description     Driver for Brother laser printers
+
+long_description \
+    brlaser is an open-source driver for Brother-branded monochrome laser \
+    printers, useful on systems where the proprietary driver is not \
+    available.
+
+notes "
+    **************************************************************************
+    brlaser has installed a collection of PostScript Printer Description (PPD)
+    files to the following directory:
+
+      ${prefix}/share/cups/model
+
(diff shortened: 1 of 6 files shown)


kencu:
test.run yes
test.target check

kencu:
because this is nomantainer, and therefore will likely come up with tickets etc that we have to deal with somehow, could you take a few minutes and put a note of some kind into the patches you added (other than the path one, which is obvious) about why the patch is needed, and what it is doing exactly?

That will help someone like me to come along later and possibly fix whatever issues arises in the tickets later.

I have a monochrome brlaser, and it is working nicely with the Brother drivers now. Has been since 2005 in fact, on Tiger. But I guess some newer models might not be supported, etc, so this seems useful.

Is the path that the model ppds are being installed to something that other cups ports might want to also install stuff do (some HP cups port, Canon cups port, etc)? I guess there is no likelihood of trampling?

evanmiller:
@kencu Updated the patches with some additional context.

I suppose a PPD name conflict could occur. We could provide a `brother` subdirectory, or prefix the file names with `brother_` if there is concern.

Debian's approach (or their 2006 approach) is described here: https://wiki.debian.org/PpdFileStructureSpecification

evanmiller:
One other question: Would gzipped PPD files be preferred? Mac's print system can handle them directly back to 10.4 at least. It's as simple as adding `-z` to the `ppdc` command. The only drawback is that ".ppd.gz" files might cause some user confusion.

kencu:
LGTM. Let's see how it flies in the wild.

PR #12081, opened 2021-09-04
cmake: update to 3.21.2

FnControlOption:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you


FnControlOption:
`patch-restore-no-default-rpath-added-on-Tiger.diff` is not needed anymore because of [cmake commit 1af23c4d](https://gitlab.kitware.com/cmake/cmake/-/commit/1af23c4de2c7d58c06171d70b37794b5c860d5f4)

devel/cmake/Portfile
```diff
@@ -36,18 +36,18 @@ subport cmake-devel {
     PortGroup       obsolete 1.0
     replaced_by     cmake
     version         20200323-8638c49f
-    revision        2
+    revision        0
```
michaelld, line 39: undo this ... this section is not to be touched except for removal entirely (which is coming)

michaelld (changes requested):
I've been using the 3.21 series since it was released with no issues, so I have no issues with this PR except the tweak to the `devel` subport. Also: please clean up the remaining patches so that they apply cleanly.

michaelld:
Tagging @kencu ... maybe you have thoughts on how/if this new CMake will work better or worse on older macOS? I know from discussion upstream that they accepted the Tiger fix, and in that discussion was noted that there's no reason CMake won't work on MacOS X back to 10.4. But, I don't think they do CI / QA testing beyond 10.12... hence the question here before this gets merged.

kencu:
Thanks Michael. I'll build this today and report back when done.

kencu (approved):

kencu:
builds without issue on 10.6. 

If something is broken before that, we can straighten it out later on.

Thanks for asking.

For the future, if cmake has tests available (must have ... ), perhaps we might figure out how to enable and run the test suite.

michaelld (approved):
LGTM; thx!

michaelld:
> builds without issue on 10.6.
> 
> If something is broken before that, we can straighten it out later on.
> 
> Thanks for asking.
> 
> For the future, if cmake has tests available (must have ... ), perhaps we might figure out how to enable and run the test suite.

Excellent all around! Yes cmake does have built-in tests. In my handling of the cmake port, I have never seen this pass at 100% ; there are almost always 10-15 tests that fail for various reasons -- out of something like 350, so that's pretty decent. Some I could fix easily, some would take serious work. I have, obviously, never taken time for either & right now I really don't have time. That said, it should be straight forward to add a `test` phase to this port ... hmmm ... have to look into that.

Anyway: Merging!

PR #12104, opened 2021-09-06
hamlib: add patch to use correct python*-config

michaelld:
#### Description

I don't `select` a Python3, so my setup does not provide `python3-config`, which is used by the `configure` script in HamLib ... and thus configuration fails. We should not rely on any `select` files inside MP -- those are for user's convenience; instead, we should use the port's files. That's what this patch does.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.5 20G71
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/science/hamlib/Portfile
+++ b/science/hamlib/Portfile
@@ -110,6 +110,12 @@ foreach py_ver ${python_versions} {
         configure.args-append \
             --with-python-binding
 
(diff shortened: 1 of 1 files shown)


ra1nb0w:
LGTM. thank you very much!

PR #12186, opened 2021-09-10
mozjs60: Support for 10.4, 10.5, and PowerPC

evanmiller:
#### Description

Add PPC and <10.6 support to mozjs60, which is required by policykit and thus a large fraction of the Gnome universe. Performed basic testing + JIT testing of the JavaScript interpreter with one failure out of the first ~1,000 tests run.

To get it to compile and run on Tiger, I disabled the International API (`--without-intl-api`) on that platform. The ICU that ships with mozjs60 contains some assembly code that Tiger's assembler doesn't like, and even after fixing that there was a run-time failure. Disabling ICU may also be necessary on Leopard, but I don't have a Leopard machine to test with.

Also removed some obsolete compiler blacklist logic (this is taken care of by `compiler.cxx_standard`).

Leaving as draft until someone can test basic functionality on Leopard/PPC.

See: https://trac.macports.org/ticket/57032
Closes: https://trac.macports.org/ticket/62350

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you


evanmiller:
Additional notes: In order to get mozjs60 to compile on GCC/PowerPC, I added the `-mone-byte-bool` flag. By default, this platform uses four-byte booleans, which have long gone out of style. The addition of this flag fixed various struct layout issues and the incorrect assumption that "a bool is a char wherever you are".

One side effect however is that there are many, many bools in the exposed JS API (`<jsapi.h>` and friends), so anything compiling against mozjs60 will now need to also specify one byte booleans. Of course, right now on PowerPC, nothing is compiling against mozjs60, but I anticipate that updates will be necessary to the `policykit` and `gjs` ports, the only two direct consumers of mozjs in the MacPorts universe right now.

Based on my cursory research, any one-byte booleans exposed by policykit and gjs should binary-compatible with four-byte booleans, so I do not anticipate a cascading effect – but it's something to be aware of for future projects that want to compile against `jsapi.h` on PowerPC. I will plan to add something to the "notes" field about all this.

evanmiller:
Following up some more: I've added logic to slip `-mone-byte-bool` into the `pkg-config` and `js-config` output on PowerPC, which should obviate the need for further Portfile changes to `gjs` or `policykit`. However, as the lengthy comment explains, it's possible there will be some side effects for client projects that use the pkg-config output in a subset of compilation units and then expose `bool` pointers or packed `bool` struct elements in their private or public APIs.

I don't think there's a perfect solution, but the `notes` field is intended to make software developers aware of the issue.

kencu:
I'm testing this out today, but I think before we could consider committing it, we'll need to clean up the older system support so it is quite out of the way of the mainstream port logic.

For an idea what I mean, take a look at what I did with the webkit2-gtk Portfile to try to make it ... maintainable for others. The older systems support is all tucked away and does not affect (or interfere with the sightlines) of the main portfile for the 99% of users who are using a reasonably current system.

evanmiller:
@kencu OK - I have reorganized my changes to move the 10.4/10.5 stuff to the end.

kencu:
Yeah, that is getting there. 

So because people don't have hours and hours to spend on reviewing PRs, the way this goes smoothly is for any given step to be one, bite-sized chunk. If too much is changed that is not actually relevant to what you are actually trying to do, people get spooked and will not commit it. Or if someone like me does commit it, it can get heavily reviewed afterwards.

The way to do this is to leave all of Dave's original portfile, and all his patches, completely alone. No cleaning up his Portfile, no renaming patchfiles, etc, in your one bit PR.

Someone looking at your PR and changes should more or less look and see "This changes nothing I care about".

If all your stuff is added in a block that cannot run on anything but your PPC systems, then nobody has anything to worry about. The ports are d**mned fragile, and they break so easily in unexpected ways that you would be very surprised.

In another step, probably a different PR completely, you can suggest general cleanups to Dave's existing port. One commit per bite-sized cleanup. Then those can be reviewed separately, each one individually, and people can see more clearly what is happening. Then your PR gets committed quite quickly.

Also, as a general rule, "style" fixes are generally left to the maintainer. In this case, Dave has been out of circulation for a while, so it's a bit different, but in general that is what most people expect.

kencu:
One more thing -- your additions are very well done and high-level, and we're lucky to have someone with your obvious skill set interested in helping out with MacPorts in general and with these older systems in particular.

Thanks for all you are doing! It is very much appreciated!

(BTW I added a buildable xorg-server to TigerPorts while I sort out what to do about that issue in MacPorts, so if you need one,  there is one there now to install).

kencu:
build went through without a hitch. Looks really really good. Running the test suite now, will take a bit but so far all in the green except the same timezone error I saw on SnowLeopard Intel.

Wowza. Great.

evanmiller:
Great to hear. Peeling off the Draft label, since all I wanted to hear was that it compiles and can compute 1+1 on Leopard. PPC support was hiding in there, just needed a little bit of coaxing :-). I'll keep my fingers crossed for your test suite.

I appreciate the advice about submitting successful PRs. Since you're pretty deep in the weeds with me on this one, I won't plan to split off the ancillary changes into separate PRs, but I will take your perspective into account in the future.

Happy to help make some of these old Macs a little less obsolete. I am enjoying the community here.

kencu:
actually leopard is building now. I put the Tiger test results in the ticket.

I tweaked your PR a bit locally to do what I described above as the minimal change, so no eyebrows would raise. I know you're tired of changing things, so it's here:

https://github.com/kencu/TigerPorts/commit/21e5f379034e27eee8cf4f4ca7c5e1da4f30c1ab

kencu:
Leopard also barfs on the ICU assembly, so I moved the disabler out of the Tiger block, and we'll try Leopard again.

kencu:
disabling ICU for Leopard PPC fixes it as well, build completed overnight. running the test suite now.

evanmiller:
@kencu Great to hear. Since you have access to Tiger and Leopard and specific ideas for how the final PR should look, can I hand you the `mozjs60` ball to get over the line here? Trying to wrap up my MacPorts work so I can move on to some other things for the time being.

By the way, once #12192 is merged, no ports will rely on `mozjs52` – perhaps a good candidate for deprecation. Then on to `mozjs78`...

kencu:
pushed and committed.

there is a slight inconsistency we might run up against. The ```-mone-byte-bool``` flag will be in the pkg-config files, but there is no way to test if the compiler that is using that flag is gcc or not. So if someone builds with clang on 10.4 or 10.5 Intel, they will get that flag added.

We will just have to see what happens with that, I guess, for now.

evanmiller:
@kencu The pkg-config patch is only applied on PPC, so the only danger is the PPC + clang combo. They need a C++11 (14?) compiler to consume `jsapi.h`, so maybe that's an impossible combination? Anyway, Intel shouldn't catch any cross-fire here.

Looking forward to reports from the field – thank you for shepherding this one through.

kencu:
> @kencu The pkg-config patch is only applied on PPC, so the only danger is the PPC + clang combo.

No, the patch is applied on all systems < 10.6. 

We could change it to be applied only for PowerPC systems though. That might make sense.

kencu:
Yeah, I think that does make sense. I'll do that.

evanmiller:
See line 146 of the Portfile, the pkg-config patch is protected by a PPC check

kencu:
oh yeah, you're right. There are two similarly-named patches and I momentarily confused the two. 

We're good then!

evanmiller:
> (BTW I added a buildable xorg-server to TigerPorts while I sort out what to do about that issue in MacPorts, so if you need one, there is one there now to install).

Thanks re: `xorg-server`, I'll check it out. I installed 1.18 here, but the performance is significantly degraded compared to factory X11.

PR #12192, opened 2021-09-10
gjs: update to 1.54.3

evanmiller:
#### Description

This update:

* Upgrades gjs version from 1.52.5 to 1.54.3
* Bumps the mozjs dependency to mozjs60
* Removes the obsolete compiler blacklist

Originally this PR included PowerPC support requiring #12186, but that change is still WIP and may not require alterations to this Portfile so I will keep it separate.

Debian's 1.52 -> 1.54 migration notes can be found here:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=906016

The conclusion of that discussion was that the included mozjs52 -> mozjs60 transition does not require a soname bump, so the reverse dependents should not require a rebuild.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11 8S165 Power Macintosh
Component versions: DevToolsCore-798.0; DevToolsSupport-794.0

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [ ] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 
- [x] checked your Portfile with `port lint`?
- [ ] tried existing tests with `sudo port test`?
- [ ] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?

PR #12194, opened 2021-09-10
policykit: PowerPC support

evanmiller:
#### Description

Along with #12186 and #12192, this is the third musketeer of mozjs on PowerPC. The only functional change is to compile policykit with the same `bool` size that the updated `mozjs60` port will be expecting. Policykit doesn't expose any `bool`s to the outside world (good boy, *pat pat*), so this should be the end of the necessary changes.

To avoid a rev-bump, this PR triplet should be merged on the same day. Leaving as draft until the primary PR #12186 is tested and approved.

Also remove outdated blacklist logic since it's made redundant by `compiler.cxx_standard`.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/security/policykit/Portfile
+++ b/security/policykit/Portfile
@@ -2,7 +2,6 @@
 
 PortSystem          1.0
 PortGroup           gobject_introspection 1.0
-PortGroup           compiler_blacklist_versions 1.0
 PortGroup           legacysupport 1.0
 
 # pin version at 0.116 for now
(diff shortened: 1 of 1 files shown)


evanmiller:
With some further experimenting, it looks like the magic `-mone-byte-bool` flag can be provided via pkg-config, rather than explicitly in every Portfile, and so this pull request may be unnecessary. Leaving as draft until I can confirm everything works with the new approach.

PR #12209, opened 2021-09-12
meson: Fix dylib install paths for typelib files

evanmiller:
#### Description

This is designed to work around the "zillion tickets" issue of typelib files having the wrong paths to dylibs, and GObject Introspection being broken for a large number of ports. A small sampling:

* https://trac.macports.org/ticket/62391
* https://trac.macports.org/ticket/61792
* https://trac.macports.org/ticket/61934

This patch ignores the presumed bug in `g-ir-scanner` and supplies the install path explicitly to `g-ir-compiler` in Meson projects. It ensures that for any Meson project that uses `gnome.generate_gir`, the Meson-supplied install location will be used in the final typelib file.

Note that this is intended as a workaround – the installed plain-text `gir` files will still have the incorrect location, as the underlying bug in `g-ir-scanner` will still need to be addressed. Still, using the explicit install location is likely a reliable approach, and this patch won't necessarily need to be reverted in the future.

A few ports that this fixes locally in my testing:

* libgweather
* geocode-glib
* appstream-glib
* pango-devel (👋 @mascguy)

If merged, these and any port that invokes Meson's [`gnome.generate_gir`](https://mesonbuild.com/Gnome-module.html#gnomegenerate_gir) function will need a revbump.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/devel/meson/Portfile
+++ b/devel/meson/Portfile
@@ -57,6 +57,9 @@ patchfiles-append   patch-meson-32bit-apple.diff
 # https://github.com/mesonbuild/meson/pull/9211
 patchfiles-append   patch-meson-gcc-appleframeworks.diff
 
+# Compiled typelib files need their dylibs' full install path
+patchfiles-append   patch-meson-gnome.diff
+
 # disable warning not accepted by older clang versions
 # this manifests currently on systems up to 10.9
 # https://github.com/mesonbuild/meson/issues/8307
@@ -74,16 +77,18 @@ platform darwin 8 {
 
     github.setup        mesonbuild meson 0.57.2
     github.tarball_from releases
-    revision            0
+    revision            1
     checksums           rmd160  8cf66e2b6cd4214cf6166d67fd4820de1f7776fb \
                         sha256  3a83e7b1c5de94fa991ec34d9b198d94f38ed699d3524cb0fdf3b99fd23d4cc5 \
                         size    1853721
 
     patchfiles-delete   patch-meson-clang-unknown-optimization-error.diff \
                         patch-meson-objc-accept-gnu89.diff \
-                        patch-meson-32bit-apple.diff
+                        patch-meson-32bit-apple.diff \
+                        patch-meson-gnome.diff
     patchfiles-append   patch-meson57-tiger-no-rpath-fix.diff \
-                        patch-meson-32bit-apple-tiger.diff
+                        patch-meson-32bit-apple-tiger.diff \
+                        patch-meson57-gnome.diff
 }
 
 # add a search path for crossfiles in our prefix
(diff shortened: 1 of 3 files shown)


kencu:
this issue certainly has been an intermittent PITA!

A question. It appears your fix assumes that every shared library will always go into

/opt/local/lib

is that correct? If so, are we worried that a shared lib might ever go somewhere else?

evanmiller:
It assumes the shared library will go wherever the Meson build file is told that shared libraries will go. I.e. if an individual project is configured with `--prefix=/opt/local --libdir=lib` then /opt/local/lib, if `--prefix=/foo --libdir=bar` then /foo/bar. So this should work fine with non-MacPorts builds.

I suppose unusual builds may not respect this, but based on my understanding of the problem, Meson is going to have more information about the final install location than `g-ir-scanner` will on its own.

kencu:
@michaelld what do you think about Evan's approach here, vs your mentioned fix in https://trac.macports.org/ticket/62391 that I think you haven't pushed yet?

michaelld (approved):
LGTM in a general sense. I have not tested specifically.

michaelld:
@kencu I like this approach as part of an overall fix. "we" (I) still need to work on GO-I itself to fix `gi-scanner` yet again. PITA, it is! I think the issue is multi-port.

michaelld:
It is not clear to me whether this PR's tweaks will work for all ports that use `gi-scanner` ... and i doubt it does ... and if not then GO-I need to be (re)fixed as well.

evanmiller:
@michaelld There are several tickets about autotools and CMake based projects with the same problem (e.g. [`webkit2-gtk`](https://trac.macports.org/ticket/62508)) – this PR does not address those.

evanmiller:
This PR will need to be tested on a project (maybe GLib?) that houses multiple shared libraries in a single typelib file. This will ensure that multiple `--shared-library` arguments are handled correctly. (All of my test projects used 1 dylib per typelib.)

evanmiller:
> This PR will need to be tested on a project (maybe GLib?) that houses multiple shared libraries in a single typelib file. This will ensure that multiple `--shared-library` arguments are handled correctly. (All of my test projects used 1 dylib per typelib.)

Tested with `harfbuzz-devel`, which includes 2 dylibs in a single typelib file, and it builds correctly. The `gir` files in this case do not manifest the original problem, but the important thing is the final result is correct since we are overriding the `gir` values.

kencu:
I'm planning on leaving this for Michael to decide. Also, there is some kind of a conflict now.

michaelld:
Needs a rebase, but after that I'm fine with merging. This will take care of some major issues; not all, but it's a good start!

evanmiller:
Rebased, let me know if you need anything else from me here

kencu:
OK, Michael says go.

PR #12234, opened 2021-09-14
exa: Fix build for arm64, add doc for macOS 10.12

judaew:
Changes:
* Replace `pandoc` by `go-md2man`. `pandoc` cannot be build because it dependency stack only support the arch x86_64. It's maybe fix build for arm64.
* Man-pages now be available on macOS 10.12 and below.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.5.2 20G95 x86_64
Xcode 12.5.1 12E507

###### Verification 
Have you

### DIFF
--- a/sysutils/exa/Portfile
+++ b/sysutils/exa/Portfile
@@ -5,7 +5,7 @@ PortGroup           github 1.0
 PortGroup           cargo 1.0
 
 github.setup        ogham exa 0.10.1 v
-revision            0
+revision            1
 categories          sysutils
 platforms           darwin
 maintainers         {catlett.info:chad @chadcatlett} \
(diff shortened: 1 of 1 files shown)


harens (commented):
Thanks @judaew for not only adding the amazing  go-md2man port, but for implementing this bugfix! 🎉 My main comment is the following warning during the build:

```
WARNING: go-md2man does not handle node type HTMLBlock
```

This results in things like the following:

image

...which isn't ideal, since this isn't an issue with pandoc. At the same time, it's not a dealbreaker either compared with not providing the man pages in the first place for certain systems.

I think the best solution is that go-md2man should be used where pandoc can't, and we accept that there are some minor formatting errors on those systems. Alternatively, we could look for a different program that does support code blocks. Thoughts?

Either way, thanks @judaew for sending this PR. Greatly appreciated :) 👍

sysutils/exa/Portfile
```diff
@@ -40,20 +40,15 @@ build.pre_args-replace   --frozen --offline
 variant git description {Build with git functionality} {}
 
 variant doc description {Build man pages} {
-    depends_build-append port:pandoc
+    depends_build-append port:go-md2man
     post-build {
-        system -W ${worksrcpath} "${prefix}/bin/pandoc --standalone -f markdown -t man man/exa.1.md > man/exa.1"
-        system -W ${worksrcpath} "${prefix}/bin/pandoc --standalone -f markdown -t man man/exa_colors.5.md > man/exa_colors.5"
+        system -W ${worksrcpath}/man "go-md2man -in=${name}.1.md -out=${name}.1"
+        system -W ${worksrcpath}/man "go-md2man -in=${name}_colors.5.md \
```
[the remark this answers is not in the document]
judaew: Thanks so much, I remember it!

judaew:
Usually, man pages written in markdown format don't use HTML tags. So go-md2man, ronn, or ronn-ng will do fine, but not in this case.

@harens I agree with you and also think the best solution is to use go-md2man for macOS 10.12 and below or Apple Silicon. I made the change as you requested. Can you check it?

sysutils/exa/Portfile
```diff
@@ -40,20 +40,32 @@ build.pre_args-replace   --frozen --offline
...
+        post-build {
+            system -W ${worksrcpath}/man "${prefix}/bin/go-md2man \
+                -in=${name}.1.md -out=${name}.1"
+            system -W ${worksrcpath}/man "${prefix}/bin/go-md2man \
+                -in=${name}_colors.5.md -out=${name}_colors.5"
+        }
+        notes "
+            The man pages for macOS 10.12 and below or Apple Silicon on may \
+            have minor formatting errors as they were built with go-md2man \
+            instead of pandoc.
```
harens, line 55: This is good, I like these notes. Just a minor nitpicky typo with _"Apple Silicon on may"_ and then it LGTM.

judaew: Thanks, that's better? I am not a native English speaker and therefore have some difficulties.

harens (approved):
Thanks @judaew for this PR!

I hope you don't mind, but I just realised that the notes will have double spacing (e.g. https://github.com/macports/macports-ports/pull/11604#discussion_r671569886), so to speed things up I thought I'd correct it quickly.

Thank you so much for your work on this. 👍

sysutils/exa/Portfile
harens: That's perfect. Thanks.

judaew:
Thanks @harens 🌟

PR #12269, opened 2021-09-17
[WIP] godot: patch macOS version detection

jasonliu--:
#### Description

I just noticed that the Godot source code is using `MAC_OS_X_VERSION_MAX_ALLOWED` in a few places. MacPorts uses `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` to tell source codes what version of macOS is being used, and `MAC_OS_X_VERSION_MAX_ALLOWED` may not get reliably set by MacPorts, and should not be used.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/games/godot/Portfile
+++ b/games/godot/Portfile
@@ -24,6 +24,7 @@ long_description    Godot Engine is a cross-platform game engine for \
 
 if {$subport eq ${name}} {
     github.setup    godotengine ${name} 3.3.1 "" -stable
+    revision        1
 
     checksums       rmd160  7b64622072972ff7d5e2fda3bd55980d7add2049 \
                     sha256  ee3bdd6355854957041cdc293d1d58431edfb9b80d6a96c5008b937cd86104e3 \
@@ -67,6 +68,11 @@ post-patch {
 \\      \"${prefix}/bin/yasm\",\\
 " \
         ${worksrcpath}/modules/webm/libvpx/SCsub
+
+    if {${os.platform} eq "darwin" && ${os.major} <= 14} {
+        reinplace {s/MAC_OS_X_VERSION_MAX_ALLOWED/__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__/} \
+            ${worksrcpath}/platform/osx/os_osx.mm
+    }
 }
 
 use_configure       no


kencu:
That is not quite true.

```
MAC_OS_X_VERSION_MAX_ALLOWED
```
is often the correct macro to test for, and it is not correct to substitute 

```
__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__
```
and expect the same behaviour.

MAX_ALLOWED looks at the SDK you are building against, and what SDK-related names and defines are in there. 

MIN_REQUIRED on the other hand looks at the deployment target you are building to, and indicates what symbols you might expect to find in the system libraries.

If the SDK is a match for the DEPLOYMENT_TARGET then the two macros are identical, and there is no difference between them in use. To avoid confusion, that is what MacPorts tries to do (build against the SDK that matches the system). 

But well-written Apple-supporting code will use MAX_ALLOWED to branch on the SDK, and MIN_REQUIRED to branch on the symbols expected, like this patch I wrote that follows one from an Apple engineer:

```
-- Python/thread_pthread.h.orig	2020-03-07 12:42:06.000000000 -0800
+++ Python/thread_pthread.h	2020-03-07 12:48:22.000000000 -0800
@@ -325,9 +325,19 @@
 {
     if (!initialized)
         PyThread_init_thread();
 #ifdef __APPLE__
     uint64_t native_id;
-    (void) pthread_threadid_np(NULL, &native_id);
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 1060
+    native_id = pthread_mach_thread_np(pthread_self());
+#elif MAC_OS_X_VERSION_MIN_REQUIRED < 1060
+    if (&pthread_threadid_np) {
+        (void) pthread_threadid_np(NULL, &native_id);
+    } else {
+        native_id = pthread_mach_thread_np(pthread_self());
+    }
+#else
+    (void) pthread_threadid_np(NULL, &native_id);
+#endif
 #elif defined(__linux__)
     pid_t native_id;
     native_id = syscall(SYS_gettid);
```

You see how it uses MAX_ALLOWED and MIN_REQUIRED correctly?

I have not read in detail the source file that you are doing the block reinplace on, but in general, if upstream has written it properly, we should not change it.

kencu:
I think we should close this unless we can show that the current upstream code is broken somehow. And if it is broken, which it might be, we should probably upstream that.

kencu:
One final comment: our legacysupport library and headers change some of these old suppositions, and what was once true may no longer be true if legacysupport is used.

That's the tradeoff we make for using legacysupport.

jasonliu--:
Yes, I see what you are talking about. However, what I am trying to fix is the following:

```
#if MAC_OS_X_VERSION_MAX_ALLOWED >= 101200
#include <os/log.h>
#endif
```

which is resulting in the following build error on the [10.10 through 10.8 builds](https://ports.macports.org/port/godot/details/):

```
platform/osx/os_osx.mm:52:10: fatal error: 'os/log.h' file not found
#include <os/log.h>
         ^~~~~~~~~~
1 error generated.
scons: *** [platform/osx/os_osx.osx.opt.tools.x86_64.o] Error 1
scons: building terminated because of errors.
```

I couldn't think of how any way to fix it other than replacing `MAC_OS_X_VERSION_MAX_ALLOWED` with `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__`, since I know that `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` is getting properly set and used in my other ports, like Blender.

Any suggestions for how I can resolve this build error?

kencu:
First, let's see in which SDK <os/log.h> actually exists, so we know where it should be.

If MAC_OS_X_VERSON_MAX_ALLOWED is not being defined, it would be 0, and so <os/log.h> would never be included. So that is not exactly the explanation for what is happening here.

kencu:
looks like it is indeed defined in 101200 and up:

https://github.com/alexey-lysiuk/macos-sdk/search?q=__os_log_h&type=Code

kencu:
It is hard to imagine that they would not be including the proper:
```
#include <Availability.h> 
```
somewhere in their headers, as that is an extremely basic header to include for any Apple-related software build...

jasonliu--:
> It is hard to imagine that they would not be including the proper:
> 
> ```
> #include <Availability.h> 
> ```
> 
> somewhere in their headers, as that is an extremely basic header to include for any Apple-related software build...

Grepping through the Godot source code, it indeed appears that they don't ever include `<Availability.h>`... although as you also pointed out, if MAC_OS_X_VERSON_MAX_ALLOWED is not being defined, then theoretically <os/log.h> would never be included. Maybe SCons is doing something strange? I'm not really familiar with SCons or anything, it's just a wild guess.

jasonliu--:
As I am looking over their macOS-related source code, and based on our discussions thus far, I'm starting to wonder whether I should file an issue upstream with them not using `<Availability.h>` anywhere....

kencu:
A couple of ```#error wheres my header??``` in the source would tell you if it is working at all.

But once again, 0 < 101200 so ...

By the way, they have removed all that fallback code from master. No more accomodations for older systems are in the file any longer, I see.

So you will not likely get a positive response to older system support now....

jasonliu--:
> A couple of #error wheres my header?? in the source would tell you if it is working at all.

I'll run a couple tests on my machine to see what the build believes is the value of `MAC_OS_X_VERSION_MAX_ALLOWED`, and report back. My machine is a macOS 10.11, and not 10.9 or 10.8, but it might at least give us some clues.

> By the way, they have removed all that fallback code from master. No more accomodations for older systems are in the file any longer, I see.

The master branch is the current development branch for Godot 4.0 (where they are rewriting a large chunk of Godot to replace OpenGL with Vulkan). You should be looking at either the `3.3` or `3.x` branch (they're currently working on Godot 3.4 beta in the `3.x` branch) to see code history that is more relevant to what we're discussing here. :smiley:

> So you will not likely get a positive response to older system support now....

Very true, but if I'm not mistaken, including `<Availability.h>` is still the canonical way of adding the OS/platform detection macros, is it not?

jasonliu--:
Okay, if I replace my code from this PR with the following:

```
    reinplace "/^static void get_key_modifier_state/i\\
#define a <.MAC_OS_X_VERSION_MAX_ALLOWED>\\
#include a\\
" \
        ${worksrcpath}/platform/osx/os_osx.mm
```

then this is the output from the (MacPorts) build:

```
:info:build Compiling ==> platform/osx/os_osx.mm
:info:build platform/osx/os_osx.mm:90:10: fatal error: '.101202' file not found
:info:build #include a
:info:build          ^
:info:build platform/osx/os_osx.mm:89:11: note: expanded from macro 'a'
:info:build #define a <.MAC_OS_X_VERSION_MAX_ALLOWED>
:info:build           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
:info:build <scratch space>:16:1: note: expanded from here
:info:build <.101202>
:info:build ^~~~~~~~~
:info:build scons: *** [platform/osx/os_osx.osx.opt.tools.x86_64.o] Error 1 
:info:build scons: building terminated because of errors.
```

So it looks like the build _is_ seeing a value for `MAC_OS_X_VERSION_MAX_ALLOWED`, at least on my local machine running macOS 10.11 and an Xcode 10.12 SDK.

(By the way, for reference and posterity, the debugging code I added was based on the idea here: https://stackoverflow.com/a/48489697)

This makes me even more curious to know what the value would be if the Portfile, using my debugging code, were to be run on a 10.10-10.8 machine. What value could be returned for `MAC_OS_X_VERSION_MAX_ALLOWED` that would make it think that it's > 101200?

kencu:
> You should be looking at either the `3.3` or `3.x` branch (they're currently working on Godot 3.4 beta in the `3.x` branch) to see code history that is more relevant to what we're discussing here. 😃

I did check those -- if they are still taking fixes for the 3.x branches they may be receptive.

> Very true, but if I'm not mistaken, including `<Availability.h>` is still the canonical way of adding the OS/platform detection macros, is it not?

Absolutely. It should be included, esp in a *.m file.

kencu:
> This makes me even more curious to know what the value would be if the Portfile, using my debugging code, were to be run on a 10.10-10.8 machine. What value could be returned for `MAC_OS_X_VERSION_MAX_ALLOWED` that would make it think that it's > 101200?

10.7 returns 1070, 10.8 returns 1080, 10.9 returns 1090, and then 10.10 returns 101000, etc.

It is not making much sense so far, fer sure.

jasonliu--:
So the mystery deepens. Looking at http://packages.macports.org/godot/, it seems that binaries for `godot-3.2.3_1` were successfully built around 2021-06-13 by the 10.8 and 10.9 builders. That would seem to correspond to the PR committed here: 52bf3cc221a51b8612c58e4edceab7332a085158. Originally, I thought the problem may have been because I rewrote the Portfile in order to add the `godot-3.2` subport, but looking at http://packages.macports.org/godot-3.2/, it seems that binaries for `godot-3.2-3.2.3_0` were successfully built by the 10.8 and 10.9 builders on 2021-09-06, which corresponds to my PR that added the subport: 31d180609ad129a25b6f07cdfb4a7e08996a3d76. This would seem to imply that the problem was not because I rewrote the Portfile to accommodate the new subport.

When I look into Godot's source code, `MAC_OS_X_VERSION_MAX_ALLOWED` was already being used in the Godot 3.2.3 source... so you are very likely correct, and this PR should be closed in favor of a proper solution. I will close this PR once we have sufficiently documented the problem, since this PR is now essentially acting as a record for posterity.

There's also a remote possibility that a regression bug may have been introduced by my PR that was committed here: 550675ed20316b9b450151359db42b77b38e56fa, but I'm not really sure how modifying the clang version checking would result in `MAC_OS_X_VERSION_MAX_ALLOWED` suddenly not working. If I look at the build log on the builder for the most recent failed builds for 10.8 and 10.9, it seems like they're trying to build using MacPorts Clang, as I would expect:

https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/159687/steps/install-port/logs/stdio
```
DEBUG: compiler clang 600.0.57 blacklisted because it matches {clang < 802}
DEBUG: universal_variant is false, so not adding the default universal variant
DEBUG: Executing variant clang9 provides clang9
DEBUG: Running callback portconfigure::add_automatic_compiler_dependencies
DEBUG: Chosen compiler macports-clang-9.0 is provided by a port, adding dependency
DEBUG: Adding depends_build port:clang-9.0
DEBUG: Adding depends_skip_archcheck clang-9.0
<snip>
DEBUG: configure phase started at Thu Sep 16 20:14:24 UTC 2021
--->  Configuring godot
DEBUG: Preferred compilers: clang macports-clang-9.0 macports-clang-8.0 macports-clang-7.0
macports-clang-6.0 macports-clang-5.0 macports-clang-12 macports-clang-11 macports-clang-10
macports-clang-3.7 macports-clang-3.4
DEBUG: Using compiler 'MacPorts Clang 9.0'
<snip>
MACOSX_DEPLOYMENT_TARGET='10.9'
Executing:  cd "/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_games_godot/godot/work/godot-3.3.1"
&& /opt/local/bin/scons -j8 platform=osx arch=x86_64 target=release_debug macports_clang=9.0 
DEBUG: system:  cd "/opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_games_godot/godot/work/godot-3.3.1"
&& /opt/local/bin/scons -j8 platform=osx arch=x86_64 target=release_debug macports_clang=9.0 
scons: Reading SConscript files ...
Detected iOS SDK version older than 13
Building for macOS 10.9+, platform x86-64.
```

jasonliu--:
A Trac ticket has been submitted which may (or may not) be related to the problems discussed in this PR:

https://trac.macports.org/ticket/63543

jasonliu--:
It doesn't look like [Trac ticket #63543](https://trac.macports.org/ticket/63543) is directly related to this PR. I am closing this PR in order to deal with the issue reported in the Trac ticket.

PR #12295, opened 2021-09-19
gjs: Fix build on 32-bit systems

evanmiller:
#### Description

Originally part of #12192. For a thorough upstream discussion of the underlying issue, see [here](https://gitlab.gnome.org/GNOME/gjs/-/merge_requests/680). The gjs project has sent the issue further upstream to GLib; a fix there will not be available until 2.71. In the meantime, this patch allows for 32-bit builds of gjs 1.54.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/gnome/gjs/Portfile
+++ b/gnome/gjs/Portfile
@@ -38,6 +38,9 @@ autoreconf.args     -fvi
 
 compiler.cxx_standard 2014
 
+# Fix gsize/size_t mismatch on 32-bit systems
+patchfiles          patch-gjs-gsize.diff
+
 # profiler currently only supported on Linux
 configure.args      --disable-profiler \
                     --disable-silent-rules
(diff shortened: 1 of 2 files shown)


kencu:
@jmroot -  would you take a look here? Based in this and the upstream discussions I'm 99.99% sure @evanmiller has nailed it here.

evanmiller:
FYI, up-up-stream has indicated that a future GLib will just do `typedef size_t gsize` and be done with it. (The same issue affect WebAssembly, so there is renewed interest in 32-bit.) I guess we could backport that change into MP GLib, but these individual patches seem safer for now.

kencu:
The trick with managing 25,000 interconnected ports is remembering exactly what you did and need to undo, and doing the exact small tweak (if possible) to fix the thing you are trying to fix without accidentally breaking 500 other things you didn't think would break, but will.

kencu:
And it would be perhaps simpler if every port was stock as upstream delivered it, but as you have seen, almost every port has some kind of workaround, tweak, or "improvement" somebody added -- so it's a helluva crapshoot sometimes.

PR #12322, opened 2021-09-21
Update to omniORB and py-omniORBpy 4.2.4

lockhart:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H1323 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

Current tickets #63523 and #63524 cover these changes.
Oops. Opened a ticket for each port in the pull request. Can cancel if needed.


lockhart:
OK awesome. So some of it fails in the build test because the omniORB prerequisite needs to be built with the same pythonxx variant as py-omniORBpy. The test tries all of the variants for python version but does not go back and install the matching omniORB variant.

When I’m installing locally I usually just install omniORB and then py-omniORBpy and specify the variants myself.

How do I specify a variant for a prerequisite package within the port file?

Thanks for any info…

- Tom

lockhart:
And without omniORB already installed it only does the right thing for py39-omniORBpy; any other variant builds omniORB +python39 rather than the version actually needed.

The portfile is included below since someone with more experience might be able to spot the solution with a quick glance.

TIA!

- Tom

reneeotten:
> OK awesome. So some of it fails in the build test because the omniORB prerequisite needs to be built with the same pythonxx variant as py-omniORBpy. The test tries all of the variants for python version but does not go back and install the matching omniORB variant. When I’m installing locally I usually just install omniORB and then py-omniORBpy and specify the variants myself. How do I specify a variant for a prerequisite package within the port file? Thanks for any info… - Tom

You cannot. You specify correctly what it needs using the `require_active_variants omniORB python${python.version}` line, but you cannot explicitly depend on a variant. So I think you've done this all correctly! Please do amend the commit messages as requested above and then it should be probably be good to go.

python/py-omniORBpy/Portfile
```diff
@@ -1,14 +1,13 @@
...
+PortGroup           active_variants 1.1
 
 name                py-omniORBpy
-version             4.2.2
-revision            1
-python.versions     27 36
-python.default_version 36
+version             4.2.4
+python.versions     27 36 37 38 39
+# python.default_version 39
```
reneeotten, line 10: can you please just remove this line instead of commenting it out

devel/omniORB/Portfile
```diff
@@ -64,8 +64,35 @@ variant python36 conflicts python27 description {Build Python 3.6 IDL support} {
     notes "Install py36-omniORBpy for omniORBpy."
 }
 
-if {![variant_isset python27]} {
-    default_variants +python36
+variant python37 conflicts python27 python36 python38 python39 description {Build Python 3.6 IDL support} {
```
reneeotten, line 67: the description should say "Python 3.7"

devel/omniORB/Portfile
```diff
@@ -64,8 +64,35 @@ variant python36 conflicts python27 description {Build Python 3.6 IDL support} {
...
+    configure.python            ${prefix}/bin/python3.7
+    set python_framework        ${frameworks_dir}/Python.framework/Versions/3.7/lib
+    depends_lib-append          port:python37
+    configure.pkg_config_path   ${python_framework}/lib/pkgconfig
+
+    notes "Install py37-omniORBpy for omniORBpy."
+}
+
+variant python38 conflicts python27 python36 python37 python39 description {Build Python 3.6 IDL support} {
```
reneeotten, line 76: the description should say "Python 3.8"

devel/omniORB/Portfile
```diff
@@ -64,8 +64,35 @@ variant python36 conflicts python27 description {Build Python 3.6 IDL support} {
...
+    configure.python            ${prefix}/bin/python3.8
+    set python_framework        ${frameworks_dir}/Python.framework/Versions/3.8/lib
+    depends_lib-append          port:python38
+    configure.pkg_config_path   ${python_framework}/lib/pkgconfig
+
+    notes "Install py38-omniORBpy for omniORBpy."
+}
+
+variant python39 conflicts python27 python36 python37 python38 description {Build Python 3.6 IDL support} {
```
reneeotten, line 85: the description should say "Python 3.9"

reneeotten:
@lockhart are you able to take care of the changes to the commit messages and the few minor changes requested above? If you prefer I am happy to do this for you as well and get the PR merged - just let us know!

PR #12339, opened 2021-09-23
libgweather: update to 40.0

evanmiller:
#### Description

With the new `libgweather` library version, some programs (e.g. `gnome-weather`) may emit an error like

```
GWeather-CRITICAL **: 09:39:32.216: gweather_info_update: assertion 'info->contact_info != NULL' failed.
```

This does not affect the library functionality, but the client programs will need to be updated at some point.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

###### Verification 
Have you

### DIFF
--- a/net/libgweather/Portfile
+++ b/net/libgweather/Portfile
@@ -4,8 +4,8 @@ PortSystem          1.0
 PortGroup           meson 1.0
 
 name                libgweather
-version             3.36.1
-set branch          [join [lrange [split ${version} .] 0 1] .]
+version             40.0
+set branch          [lrange [split ${version} .] 0 0]
 maintainers         {devans @dbevans} openmaintainer
 categories          net gnome
 license             GPL-2+ LGPL-2.1+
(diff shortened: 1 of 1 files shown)


evanmiller:
This may be a more complicated update than I realized... From the NEWS file:

```
================
Version 40.alpha
================

This version contains a number of API and ABI changes that applications
will need to be modified for, including:
- An application-id as well as contact information must be provided for
  every application to use online weather services. For GApplication-based
  applications, the application ID is prefilled.
- All online weather providers are disabled by default, applications
  should enable those that it has verified it can follow the terms of
  service for. Note, for example, the requirement to subscribe to the
  met.no API users mailing-list.
- A lot of the getter functions for the location database now return
  references to objects which the application will need to handle. Applications
  that use JavaScript or Python bindings should not require special handling
  for this change.
```

So this update may need to be coordinated with updates to all of the client applications.

PR #12377, opened 2021-09-25
py-gobject3: add patch to address closure / tramp on ARM64

michaelld:
#### Description

py-gobject3: add patch to address closure / tramp on ARM64

NOTE: This patch is not official! It is a "best effort" workaround for the issue, and for the Python interface to GObject-Introspection only. See also [the actual commit](https://github.com/Damenly/pygobject/commit/01013b1003bac67061e595cd857be9cf7b69d7fd) for its message.

Closes: https://trac.macports.org/ticket/62180

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.5 20G71 [Intel]
Xcode 12.3 12C33

macOS 11.6 20G165 [ARM64]
Xcode 13.0 13A233

###### Verification 
Have you

### DIFF
--- a/python/py-gobject3/Portfile
+++ b/python/py-gobject3/Portfile
@@ -7,7 +7,7 @@ PortGroup           meson 1.0
 name                py-gobject3
 set my_name         pygobject
 version             3.38.0
-revision            1
+revision            2
 set branch          [join [lrange [split ${version} .] 0 1] .]
 categories-append   gnome
 license             LGPL-2.1+
@@ -38,6 +38,11 @@ if {${name} ne ${subport}} {
                             port:gobject-introspection
 
     compiler.c_standard     2011
+
+    # https://github.com/Damenly/pygobject/commit/01013b1003bac67061e595cd857be9cf7b69d7fd
+    # this patch is designed to work around PyGObject's use of GOI's use of LIBFFI
+    patchfiles              patch-pygobject_fix_arm64_closure_exec.diff
+    patch.pre_args -p1
 
     use_configure           yes
 
--- a/python/py-gobject3/files/patch-pygobject_fix_arm64_closure_exec.diff
+++ b/python/py-gobject3/files/patch-pygobject_fix_arm64_closure_exec.diff
@@ -1,0 +1,620 @@
+From 01013b1003bac67061e595cd857be9cf7b69d7fd Mon Sep 17 00:00:00 2001
+From: Su Yue <l@damenly.su>
+Date: Fri, 3 Sep 2021 20:38:20 +0800
+Subject: [PATCH] Hack for segmentation faults on Apple Silicon(Macos arm64)
+
+There are bugs in pyobject and gobject-introspection.
+The API g_callable_info_prepare_closure() only returns the address
+of exec not closure.
+On linux arm64 and x86_64 platforms, exec address is same as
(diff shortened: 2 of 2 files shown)


kencu:
Well. Not an easy one to sort out or test. I read through https://gitlab.gnome.org/GNOME/pygobject/-/issues/455 and I can see the troubles people are having.

It's times like this that we will feel the pain from losing Jeremy's attentions...

michaelld:
> Well. Not an easy one to sort out or test. I read through https://gitlab.gnome.org/GNOME/pygobject/-/issues/455 and I can see the troubles people are having.
> 
> It's times like this that we will feel the pain from losing Jeremy's attentions...

Indeed! (all around ...) The gist of how I read the patch is that they are replicating the underlying behavior but fixing it to actually do the right thing. It's a bit of a rabbit hole to go down (the comparison), but in my read their fix is legit -- even if just for this very specific case of GO-I and LIBFFI. I'm not sophisticated enough to work out how to patch GO-I to generically do the right thing. Hence, trying this route to see if there's any traction.

kencu (approved):
so this is one of those situations where we leap and then fix whatever comes next,  I believe.

michaelld:
> so this is one of those situations where we leap and then fix whatever comes next, I believe.

Yup that's my take too. It's like my M1 patches for NumPy and SciPy: make it work and hope upstream fixes things soonish!

michaelld:
Checks passes, which is always a good sign. Let's see what @dbevans says ...

michaelld:
This port is openmaintainer, and this hack/fix is critical for GNURadio's GRC to execute on M1. Per @kencu's note: this is one of those situations where we leap and then fix whatever comes next. Thus I'm merging & we'll see what happens. 🤞

gergomez:
This is still happening. I've found an easy way to reproduce it without gnuradio:

```
import gi
gi.require_version("Gtk", "3.0")
from gi.repository import Gtk

window = Gtk.Window(title="Hello World")
window.show()
window.connect("destroy", Gtk.main_quit)
Gtk.main()
```

using 
py39-gobject3@3.38.0
py39-cairo@1.20.1
libffi@3.4.2_2
glib2@@2.64.6

Running this on macOS 12.2 in a M1 Max,  I get the same behavior as reported. Backtrace attached.

If I do:

`gi.require_version("Gtk", "2.0")`

everything works as expected.

[bt.txt](https://github.com/macports/macports-ports/files/7990079/bt.txt)

gergomez:
In fact the bug I encountered is not because of this but because of glib2 (https://trac.macports.org/ticket/64377#ticket). Solved by removing the universal.patch and config.ed and recompiling everything from source.

michaelld:
@gergomez thanks for that info! I'll give that a try & see if I can reduce the issue to some specific part of those files ... they are usually there for a reason, but sometimes need updating for newer releases than they were originally created

michaelld:
Excellent! For my install I have to disable that patch, then reinstall glib2, then reinstall gtk3. Then GRC works. Nice!

wifwucite:
Cool, sounds like you guys have found a feasible work-around. Unfortunately I am not able to follow along. Could someone post some instructions on how to get GRC up and running on an ARM64 Mac with MacPorts? I think quite a few people would appreciate this. :-) Thanks in a advance.

PR #12382, opened 2021-09-26
s6: submission

saj:
#### Description

[s6](https://skarnet.org/software/s6/) is an actively maintained spiritual descendant of djb's [daemontools](https://cr.yp.to/daemontools.html).

This pull includes s6, [execline](https://skarnet.org/software/execline/), and [skalibs](https://skarnet.org/software/skalibs/).  The former depend on the latter.  All are from the same upstream author.  For the sake of simplicity, this s6 port is delivered with all features enabled (namely execline support) without variants.  execline is technically optional, though working to exclude it would probably only benefit people on space constrained embedded devices.

`--disable-shared` is an [upstream directive](https://git.skarnet.org/cgi-bin/cgit.cgi/s6/tree/README.macos) for the Mac platform.

I have been testing this port, albeit with an older upstream release, since mid August.  No problems so far.

Closes: https://trac.macports.org/ticket/58116

Thank you for MacPorts!

###### Type(s)

- [x] enhancement

###### Tested on

macOS 11.6 20G165 x86_64
Xcode CLT 12.5.1.0.1.1623191612

###### Verification
Have you

s6 and execline include dozens of small programs (by design).  I have not had a reason to use all of them.  I have tested the most significant programs used for service supervision, including `s6-svscan`, `s6-supervise`, `s6-svscanctl`, `s6-svc`, `s6-log`, `s6-tai64nlocal`, `s6-envdir`, and possibly a few others.

### DIFF
--- a/devel/skalibs/Portfile
+++ b/devel/skalibs/Portfile
@@ -1,0 +1,62 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                skalibs
+version             2.11.0.0
+revision            0
+categories          devel
+platforms           darwin
+license             ISC
+maintainers         nomaintainer
+description         general-purpose libraries for C system programming
+
+long_description \
+    skalibs is a package centralizing the free software / open source \
+    C development files used for building all software at skarnet.org: \
+    it contains essentially general-purpose libraries. skalibs can \
+    also be used as a sound basic start for C development. There are a \
+    lot of general-purpose libraries out there\; but if your main goal \
+    is to produce small and secure C code with a focus on system \
+    programming, skalibs might be for you.
+
+homepage            https://skarnet.org/software/skalibs/
+master_sites        ${homepage}
+
+checksums           rmd160  f033bb3054865561f86edc93c0d2f265e6a72829 \
+                    sha256  98dfc8a02a333f5b12d069d84471c0d51ab5a421c4292963048b3652563d34d9 \
+                    size    216201
+
+configure.args      --disable-shared
+
+post-destroot {
+    set docdir ${prefix}/share/doc/${name}
+    xinstall -d ${destroot}${docdir}
+    xinstall -m 0644 -W ${worksrcpath} \
+        AUTHORS \
+        COPYING \
+        NEWS \
+        README \
+        README.macos \
+        ${destroot}${docdir}
+    xinstall -m 0644 {*}[glob ${worksrcpath}/doc/*.html] \
+        ${destroot}${docdir}
+    xinstall -d ${destroot}${docdir}/libdatastruct
+    xinstall -m 0644 {*}[glob ${worksrcpath}/doc/libdatastruct/*.html] \
+        ${destroot}${docdir}/libdatastruct
+    xinstall -d ${destroot}${docdir}/libposixplz
+    xinstall -m 0644 {*}[glob ${worksrcpath}/doc/libposixplz/*.html] \
+        ${destroot}${docdir}/libposixplz
+    xinstall -d ${destroot}${docdir}/librandom
+    xinstall -m 0644 {*}[glob ${worksrcpath}/doc/librandom/*.html] \
+        ${destroot}${docdir}/librandom
+    xinstall -d ${destroot}${docdir}/libstdcrypto
+    xinstall -m 0644 {*}[glob ${worksrcpath}/doc/libstdcrypto/*.html] \
+        ${destroot}${docdir}/libstdcrypto
+    xinstall -d ${destroot}${docdir}/libstddjb
+    xinstall -m 0644 {*}[glob ${worksrcpath}/doc/libstddjb/*.html] \
+        ${destroot}${docdir}/libstddjb
+    xinstall -d ${destroot}${docdir}/libunixonacid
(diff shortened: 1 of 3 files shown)


devel/skalibs/Portfile
```diff
@@ -0,0 +1,62 @@
...
+    skalibs is a package centralizing the free software / open source \
+    C development files used for building all software at skarnet.org: \
+    it contains essentially general-purpose libraries. skalibs can \
+    also be used as a sound basic start for C development. There are a \
+    lot of general-purpose libraries out there\; but if your main goal \
+    is to produce small and secure C code with a focus on system \
+    programming, skalibs might be for you.
+
+homepage            https://skarnet.org/software/skalibs/
+master_sites        https://skarnet.org/software/skalibs/
```
herbygillot, line 24: You can avoid repeating yourself here by using a variable:
```
homepage            https://skarnet.org/software/skalibs/
master_sites        ${homepage}
```

sysutils/s6/Portfile
```diff
@@ -0,0 +1,53 @@
...
+    and daemons. It is meant to be a toolbox for low-level process and \
+    service administration, providing different sets of independent \
+    tools that can be used within or without the framework, and that \
+    can be assembled together to achieve powerful functionality with \
+    a very small amount of code.
+
+homepage            https://skarnet.org/software/s6/
+master_sites        https://skarnet.org/software/s6/
+
+depends_build       port:execline port:skalibs
```
herbygillot, line 27: Preferred to have one dependency per line, using line continuations:
```
depends_build       port:execline \
                    port:skalibs
```

herbygillot:
Hello @saj, thank you for this quite thorough and well put-together submission.

I'm curious, does `s6` have any facility to allow it to start its supervisor daemon at startup in the same way that `launchd` would?

saj:
> does `s6` have any facility to allow it to start its supervisor daemon at startup in the same way that `launchd` would?

You are probably wondering why I did not include a `startupitem`.  I promise to circle around to something resembling an explanation. :-)  It might help if I explain how I use this software.

My current approach, on macOS, is to automatically start [`s6-svscan`](https://skarnet.org/software/s6/s6-svscan.html) from `launchd` on login.  As nothing here need execute as the superuser -- or as any other particular system user -- and because all of my jobs only perform I/O on my own personal data, I developed a preference for the following arrangement:

- execute the entire s6 supervision tree as my own UID
- store all s6 service directories beneath my own home directory

I manually dropped a `.plist` into `~/Library/LaunchAgents` with the following argv:

```
<key>ProgramArguments</key>
<array>
  <string>/opt/local/bin/s6-svscan</string>
  <string>/Users/saj/svc</string>
</array>
```

Most other server ports I have installed from MacPorts, like nginx, are arranged as s6 services in my supervision tree.  (Which is to say that I do not use MacPorts StartupItems.)

```
saj@rowlf % s6-procs | grep nginx
 |-+- 00470 saj s6-supervise nginx/log
 |-+- 00472 saj s6-supervise nginx
 | \-+= 00484 saj nginx: master process nginx -c /Users/saj/svcs/nginx/d/nginx.conf -e stderr
 |   \--- 00490 saj nginx: worker process
```

My goal is to touch `launchd` only once -- then leave everything else to s6.

I figure my arrangement is at least a little idiosyncratic.  Perhaps that is fitting.  After all, s6 was designed as a collection of  goodies that can be used in many different ways.

That got me wondering whether it would make any sense to provide a canned `startupitem` for this port.  Unlike, say, a popular database management system, I think would be fair to assume that anyone installing this port is going to know their way around unix, and will already have a specific goal (and idiosyncratic servicedir and supervision tree layout) in mind.  I was unsure whether a dummy `startupitem` would provide any value to users.  In my case, it would have only posed a nuisance.

On the other hand, the existing daemontools port does ship with a dummy servicedir and `startupitem`, so there is a valid counter-argument to be made on the basis of consistency.

I can add a `startupitem` if you feel it is necessary.

saj:
@herbygillot, latest push includes a `startupitem`.  I used the existing `daemontools` port as inspiration.  Paths are similar -- but we use `s6-svscan` here instead of `svscan` so that the two ports will peacefully coexist if installed on the same machine.  (Users may want to do this temporarily when migrating from daemontools to s6.)

I loaded the `LaunchDaemons` link as a quick test.  The s6 supervisor started without a problem.

If we are going to provide users with a global scandir (beneath the MacPorts `${prefix}`), I think that supervisor tree should execute as root -- as is the case with the existing `daemontools` port.  Users can easily drop privileges with `s6-setuidgid` from root.  That would not be possible if we forced the global supervisor tree to execute as some other UID, which would make the MacPorts `startupitem` all but useless.

Thanks for the review!

herbygillot:
Just tried this out, and it looks good, @saj.  Very good point about the supervisor needing to run as root.

Just a quick note you might already be aware of: for ports that use `startupitem` directives to automatically create startupitems, you can load the port into `launchd` using `port load` (`sudo port load s6` - I think the notes for your port will automatically mention this whenever you specify a `startupitem` in your Portfile).

This looks good and thank you not only for this submission, but also for your patience in making all the requested changes.

saj:
> I think the notes for your port will automatically mention this whenever you specify a `startupitem` in your Portfile

Your recollection is correct.  During testing, I noted the `startupitem` triggered an automated message guiding the user towards `port load`; the `ui_msg` (with port-specific guidance) was printed separately.

MacPorts defaults are quite nice.  I see now that it is typical to provide a `startupitem` for convenience, but leave it to the user to explicitly enable the service if that is what they want.  This is the right balance that some other distributions fail to meet. 👍  Cheers.

PR #12462, opened 2021-10-03
OpenBLAS: update to 0.3.18, update minimal architectures

NicosPavlov:
#### Description

Update of OpenBLAS to the latest version 0.3.18, and update of *-devel port to latest commit.
The minimal architectures to build upon have been increased based on the machines supported by the different OS versions, and dependencies have been cleaned up.
The different steps are in different commits to ease comparison, they should be squashed upon merging.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6 20G165 x86_64
Xcode 13.0 13A233

###### Verification 
Have you

### DIFF
--- a/math/OpenBLAS/Portfile
+++ b/math/OpenBLAS/Portfile
@@ -27,22 +27,12 @@ compiler.thread_local_storage yes
 if {${os.major} >= 11 && ${os.major} <= 17} { 
     set blas_arch "CORE2"
 } 
+#OS 10.14 supports down to Mac Pro 5,1 that has Nehalem architecture
 if {${os.major} >= 18} {
     if {${os.arch} eq "arm"} {
         set blas_arch "ARMV8"
     } else {
-        #OS 10.14 supports down to Mac Pro 5,1 that has Nehalem architecture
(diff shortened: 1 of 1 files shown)


michaelld (approved):
LGTM

michaelld:
Trying to rerun the CI / QA ... just for the sake of it. I don't think this PR is the issue with those failing, but it's worth trying them again just in case.

NicosPavlov:
Thanks for running them again. I noticed there is actually an error in there with "Real BLAS Test", SROT instruction. Considering that OpenBLAS-devel (from Sep 17th) builds fine on all bots, this could be related to a recent change in the code, but I wonder. I'll probably try to revert the new architectures in the PR to see if it changes, that might be a weird case with virtual environments.

NicosPavlov:
So, it appears it was due to the change in CPU architecture, as the checks now go through. I may have been wrong in my checks, but I suspect that the virtual environments may cause issues. 
In any case, it is not worth it if bots are prevented to build, so I'll probably push that PR without these changes.

PR #12488, opened 2021-10-06
osm2pgsql: Specify lua as the default variant

frankdean:
#### Description

@Veence I've create a Portfile for `mod_tile` (PR #12391), a map tile server for OpenStreetMaps.  It requires `osm2pgsql` with the `lua` variant.  Without it as a default, the `mod_tile` port does not pass checks as it builds `osm2pgsql` without the `lua` variant.  It's also a poor user experience in the same scenario.

Ryan Schmidt has suggested some possible solutions [1]; either the variant being a default or alternatively being broken out into another port or sub-port.

Is it OK to have it as a default variant?  If not, what solution would you prefer?

[1]: https://lists.macports.org/pipermail/macports-dev/2021-October/043788.html

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6 20G165 x86_64
Xcode 12.5 12E262

###### Verification 
Have you

### DIFF
--- a/gis/osm2pgsql/Portfile
+++ b/gis/osm2pgsql/Portfile
@@ -7,12 +7,11 @@ PortGroup           active_variants             1.1
 PortGroup           boost                       1.0
 
 github.setup        openstreetmap osm2pgsql 1.4.1
-revision            2
 
 categories          gis
 maintainers         {vince @Veence} openmaintainer
-description         osm2pgsql is an OSM data injector for PostGIS
-long_description    osm2pgsql is a command line tool to import\
+description         ${subport} is an OSM data injector for PostGIS
+long_description    ${subport} is a command line tool to import\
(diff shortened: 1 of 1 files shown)


Veence:
I’m not sure specifying Lua as a default variant would be the right path to take, because Lua is not needed for the tool to work fine in 99% of its application cases. Creating a subport would definitely be the way I'd choose. What do you think?

frankdean:
@Veence – Is this subport approach Ok?

Veence:
I've committed it. Unfortunately, I’m currently without a Mac, so it’s some sort of blind commit. I hope everything goes well – until I can put my hands on a new computer.

frankdean:
Thank you 🤞

PR #12514, opened 2021-10-07
Add more generic support for OpenSSL versions

cjones051073:
#### Description

This MR does a few things

 - It updates the openssl10 and openss11 ports to install *directly* into their isolated libexec areas, and then install sym links back into the primary prefix as required. This is better than the other way around, as it creates a more 'regular' install area which some builds prefer, is a bit simpler (does not require as much moving of files around post-destroot) and allows things the old builds did not, like have the docs available for all builds.

- It makes the openssl port a basic shim port, that just installs a few sym-links in the primary prefix to the libexec areas for the 'default' openssl version. The end result is everything is still there, where is was before, so all ports using openssl are blind to the change.

- There is a new 'openssl' PG, largely evolved from the old_openssl PG, that handles defining the default openssl branch (currently still 1.1) and configuring builds to use the libexec isolated builds as required.

- As a proof of principle, all ports previously using old_openssl PG are migrate to this new one targeting the 1.0 branch.

- Provides a new openssl3 port for the 3.0.0 release (note upstream have changed their versioning scheme, see

https://www.openssl.org/policies/releasestrat.html

  hence it is called `openssl3` and not `openssl30`

What this MR *does not* do is make any statement of when openssl 3 should be made the default openssl version (which can be done by just changing the default in the openssl PG). What it does is make it available in the stack, so we can see how well openssl3 builds on all our systems (I cannot test much in the way off older systems than macOS11), and also makes it possible for any port maintainer that wishes to to switch to openssl3 now, using the new PG, before it is made the default.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/openssl11/Portfile
+++ b/devel/openssl11/Portfile
@@ -133,14 +133,6 @@ post-destroot {
     }
     # shared/man dir seems to still end up in ${prefix} and not libexec...
     move ${destroot}${prefix}/share/man ${destroot}/${my_prefix}/share/
-    # base only compresses man file automatically if under ${prefix}/share/man
-    # so do by hand here...
-    set gzip [findBinary gzip ${portutil::autoconf::gzip_path}]
-    fs-traverse manpage ${destroot}${my_prefix}/share/man {
-        if { [file isfile ${manpage}] } {
-            system "${gzip} -9vf ${manpage}"
-        }
-    }
 }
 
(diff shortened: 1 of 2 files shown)


cjones051073:
Any comments, in particular from the openssl maintainers ?

PR #12580, opened 2021-10-14
graphviz-devel: add build dep 'autoconf-archive'

eborisch:
Building graphviz 2.49.1 failed for me [10.15.7] without these changes.

Wondering if anyone else is running into this or just me; or if others can confirm that this at least doesn't break it for them.

### DIFF
--- a/graphics/graphviz-devel/Portfile
+++ b/graphics/graphviz-devel/Portfile
@@ -58,7 +58,8 @@ revision                        0
 
 conflicts                       graphviz${otherbranch}
 
-depends_build                   port:pkgconfig
+depends_build                   port:pkgconfig \
+                                port:autoconf-archive
 
 depends_lib                     path:include/turbojpeg.h:libjpeg-turbo \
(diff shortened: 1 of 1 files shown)


mascguy:
Eric/Anyone, is the use of `autoconf-archive` prevalent enough, that we might want to install it along with `autoconf`? (Either by folding it into the latter, or some other means?)

mascguy:
Nit-picky suggestion: Maybe include "add build dep for autoconf-archive" in the commit message.

mascguy:
Or as an alternative: I'm wondering if we should update `autoconf` with a variant (enabled by default), which adds a lib dependency on `autoconf-archive`: If the macros in the latter are widely-used, why not just ensure it's always installed by default? Particularly since `autoconf-archive` doesn't have any dependencies, there aren't any binaries (and is architecture-independent), and the build/install is lightning-quick.

@Schamschula Marius, what do you think?

PR #12734, opened 2021-10-30
ffmpeg-devel: some tweaks

Gcenx:
#### Description

- `ld64` was only required for Xcode11 bug (should be handled by `xcode_workaround`)

Removed;
```
if {[lsearch [get_canonical_archs] i386] != -1} {
    compiler.blacklist-append {clang < 422.1.7}
}
```
That would already be covered by;
```
compiler.blacklist-append {clang < 800}
```

###### Type(s)
- [x] enhancement

###### Tested on
macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505

###### Verification 

### DIFF
--- a/multimedia/ffmpeg-devel/Portfile
+++ b/multimedia/ffmpeg-devel/Portfile
@@ -19,7 +19,7 @@ conflicts           ffmpeg
 # Please increase the revision of mpv whenever ffmpeg's version is updated.
 epoch               0
 version             4.4.1
-revision            2
+revision            3
 license             LGPL-2.1+
 categories          multimedia
 maintainers         {devans @dbevans} {jeremyhu @jeremyhu} {mascguy @mascguy} openmaintainer
@@ -68,7 +68,6 @@ checksums           rmd160  ebb1f042b2ba4f13be86339d30522cd73eb6da3e \
 
 depends_build       port:pkgconfig \
                     port:gmake \
-                    port:ld64 \
                     port:cctools \
                     port:texinfo
 
@@ -111,12 +110,8 @@ minimum_xcodeversions {9 3.1}
 # requires a C11 compiler
 compiler.c_standard 2011
 
-if {[lsearch [get_canonical_archs] i386] != -1} {
-    # clang-3.1 hits https://trac.macports.org/ticket/30137 (<rdar://problem/11542429>)
-    # clang-139 hits https://trac.macports.org/ticket/38141
-    compiler.blacklist-append {clang < 422.1.7}
-}
-
+# clang-3.1 hits https://trac.macports.org/ticket/30137 (<rdar://problem/11542429>)
+# clang-139 hits https://trac.macports.org/ticket/38141
 # warning: unknown warning option '-Werror=partial-availability'; did you mean '-Werror=availability'? [-Wunknown-warning-option]
 # warning: unknown warning option '-Wno-bool-operation'; did you mean '-Wno-bool-conversion'? [-Wunknown-warning-option]
 compiler.blacklist-append {clang < 800}


multimedia/ffmpeg-devel/Portfile
```diff
@@ -67,9 +67,6 @@ checksums           rmd160  ebb1f042b2ba4f13be86339d30522cd73eb6da3e \
                     size    9557516
 
 depends_build       port:pkgconfig \
-                    port:gmake \
-                    port:ld64 \
-                    port:cctools \
```
mascguy, line 72: Hey Dean, `cctools` was a recent necessary addition, as the port opportunistically uses those components. And without this, the port will be broken on the next `libffi` update.

Why would we want to remove this...?

Gcenx: I should have added some more note for this one, it would be better to only add `cctools` when a macports compiler is required to avoid installing when not needed just unsure how best to handle that.

mascguy: Hmmm. Without diving way into the weeds on this, my gut feel is that this would be risky.

Perhaps if there's a few specific platforms where the removal would be beneficial - say, on 10.4 and 10.5 - then perhaps that would be fine. But even then, I'd like guidance from a few others.

@evanmiller Any thoughts on this, from a 10.4/10.5 perspective?

@kencu @jmroot Your folks' thoughts on this in-general, particularly in the realm of a future `libffi` update, etc?

Gcenx: If the problem is ffmpeg opportunistically links to libffi wouldn’t it make more sense to then add libffi as an ffmpeg dep?

Then on the next libffi bump ffmpeg would also get a revision bump ensuring it doesn’t break.

mascguy: No, the problem is that the components installed by `cctools` are broken by a `libffi` update. And without a dependency on the former, the `ffmpeg` build will attempt to opportunistically use those broken components, before they're updated/fixed.

Does that make sense?

mascguy: As for the removal of dep `ld64`, if we're absolutely sure that `xcode_workaround` eliminates that requirement (?), I'm fine with that.

However, I'm not sure I agree that it makes sense to remove the dependency on `gmake`. Personally I prefer to use that, vs. the outdated macOS version, across-the-board. And I don't see the benefit, given that it's still necessary for 10.4 anyway.

multimedia/ffmpeg-devel/Portfile
```diff
@@ -111,12 +108,6 @@ minimum_xcodeversions {9 3.1}
 # requires a C11 compiler
 compiler.c_standard 2011
 
-if {[lsearch [get_canonical_archs] i386] != -1} {
-    # clang-3.1 hits https://trac.macports.org/ticket/30137 (<rdar://problem/11542429>)
-    # clang-139 hits https://trac.macports.org/ticket/38141
-    compiler.blacklist-append {clang < 422.1.7}
-}
-
```
mascguy, line 119: I'm okay with this removal, given the later blacklisting of `clang < 800`. Seems like this is outdated cruft that can be removed, unless I'm missing something.

multimedia/ffmpeg-devel/Portfile
mascguy: Overall, though, don't let any of my push-back dissuade you: Your contributions and insight are very much appreciated, and please don't be discouraged by any of this!

Gcenx: > Does that make sense?

That makes more sense now.

> As for the removal of dep ld64, if we're absolutely sure that xcode_workaround eliminates that requirement (?), I'm fine with that.

The `xcode_workaround` should handle the Xcode11 weirdness that's what that group was created for I remember the fallout from that.

> However, I'm not sure I agree that it makes sense to remove the dependency on gmake. Personally I prefer to use that, vs. the outdated macOS version, across-the-board. And I don't see the benefit, given that it's still necessary for 10.4 anyway.

More trying to keep inline with other Ports that mostly only add `gmake` if required as most upstream bugreports endup coming from brew these days.

mascguy: Your rationale for removing `gmake` is certainly reasonable. However, I'm reluctant to remove it, given that we've been utilizing it for so long. And the idea of encountering some unusual bug in macOS' older `make`, somewhere down the line, is enough to cause my PTSD to flare up! Lol

mascguy: So if you want, keep `cctools` and `gmake`, and only remove `ld64`. Once that's done, I'll quickly run some local tests, and merge this.

Gcenx: Keeping `cctools` & `gmake` around does seem reasonable just need to later confirm if Macports-port `cctools` is now behaving correctly (ld64 likes to force llvm even on Mojave)

I’ll modify the PR accordingly when I’m in front of a computer.

mascguy: I'll go ahead and push the changes, to save time.

And we can certainly revisit the other topics later. Every bit of cleanup helps!

mascguy (approved):

PR #12736, opened 2021-10-30
jumpcut: Update to 0.75

judaew:
#### Description

Closed: [trac#60092](https://trac.macports.org/ticket/60092)

Update to 0.75 and switch to use releases. The latest pre-build package supports arm64 and x86_64 architectures. For macOS below 10.10 (Yosemite), the author recommends using version 0.63. Pre-build package for 0.63 supports i386 and ppc architectures.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.6 20G165 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/aqua/jumpcut/Portfile
+++ b/aqua/jumpcut/Portfile
@@ -1,33 +1,65 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           xcode 1.0
 
 name                jumpcut
-version             0.63
 categories          aqua
-maintainers         nomaintainer
+maintainers         {judaew @judaew} openmaintainer
 license             MIT
-supported_archs     i386 ppc
 
 description         Jumpcut: Minimalist Clipboard Buffering for OS X
-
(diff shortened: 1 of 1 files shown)


aqua/jumpcut/Portfile
```diff
@@ -1,33 +1,53 @@
...
-                    rmd160  ddf4f7fee02d84de62cec037a286d5f9c8a704f1 \
-                    sha256  48a0d44673138a97a2e734cd991bb065437ec66e1653a44d0e2679aa1087c9d9 \
-                    size    486246
-
-patchfiles          patch-AppController.m.diff
+homepage            https://snark.github.io/jumpcut/
+
+if {${os.major} < 14 } {
+    version             0.63
+    revision            0
```
ryandesign, line 21: It looks like you're changing this port from build-from-source to install-binary. Since the pre-built binary is probably not identical to the one we would have built from source, probably a good idea to increase the revision here so that everyone has the same version.
```suggestion
    revision            1
```

I assume the installed binary is universal. If so, ensure that the universal variant is enabled, or better yet, strip out the architectures that were not requested if the user installs without the universal variant.

judaew:
Thanks for the help. I suppose `ui_error` is a good option to indicate that the port does not need to be installed without the `+universal` variant.

PR #12757, opened 2021-11-02
apple-pki-bundle: Add macOS System Roots, utility scripts, and notes

essandess:
* See: https://trac.macports.org/ticket/63740

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.1 20G224 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you


RJVB:
One question: why do you go through all the trouble to put all certificates in a single .pem file (or maybe two, now) if afterwards you need to extract them again in order to add them automatically?

I've run into this same question on Linux when I tried to add the bundle; it seems that multi-certificate .pem files aren't supported by most utilities for manipulating certificate stores (even if Firefox will download certificate chains that way).

So, the exported `System Roots` store aside, why not simply install individual files with explicative names, like the ca-certificate package does on Linux (Debian/Ubuntu)? Your scripts could then (mostly) iterate over files rather than over the output from a pipe command.

essandess:
> why do you go through all the trouble to put all certificates in a single .pem file

That’s the basic format required by numerous applications for which a trust store must be provided: `curl`, `privoxy`, and so forth.

RJVB:
My impression is that all these can also work with a collection of .pem files in the same location. Is there any reason why they'd work differently on Mac than they do under Linux? Because there I simply couldn't get the .pem bundle to be accepted, not even by installing it manually into /etc/ssl/certs and then running update-ca-certificates (`gcr` even balked on each individual certificate but that's probably because of operator unawareness).

essandess:
> My impression is that all these can also work with a collection of .pem files in the same location.

Not for most applications, whether on BSD or Linux. Even the webpage you're reading now starts with sending a PEM chain.

This is the standard format. Anyone who wants to split the PEM files can simply use `gcsplit`.

catap:
@essandess at least you need one more hack of old macOS: https://catap.medium.com/how-to-fix-internet-on-old-macos-c7061401ab35

essandess:
> @essandess at least you need one more hack of old macOS: https://catap.medium.com/how-to-fix-internet-on-old-macos-c7061401ab35

Please see the linked trac discussion.

catap:
@essandess an issue is much-much deeper to be honest, for example azure pipeline is broken is also because it can't download anything ;)

RJVB:
>> My impression is that all these can also work with a collection of .pem files in the same location.
>Not for most applications, whether on BSD or Linux. Even the webpage you're reading now starts with sending a PEM chain.

Again, not my experience. I kind of expected it'd just mention/reference the signing certificate, leaving it up to the client to have a copy in store, and of the certificates required down the chain. But aybe the webpage does indeed send a single stream (chain) of certificates. Even so we wouldn't be having this discussion if that were all that's required. The items of that chain still have to be in some local store, with a trust setting. I strongly doubt that most applications will do the corresponding checking themselves. But whatever :)

essandess:
@RJVB May we please limit discussion on this PR to specific issue pertaining to the PR.

mascguy:
Let's merge this, as it'll be great to release this into the wild. And we can certainly refine later, to the Nth degree, as folks wish.

Thanks so much Steve!

RJVB:
On Wednesday November 03 2021 07:32:26 Christopher Nielsen wrote:
>Assigned #12757 to @RJVB.

Why??

PR #12760, opened 2021-11-02
py-wheel: enable pep517 build

jmroot:
Also remove setuptools_scm dependency. The change that required it was
reverted upstream in version 0.34.1.


Schamschula:
It looks like there is an error for several python versions:
`ModuleNotFoundError: No module named 'importlib_metadata'`

jmroot:
> It looks like there is an error for several python versions:
> `ModuleNotFoundError: No module named 'importlib_metadata'`

Fixed now.

PR #12773, opened 2021-11-02
py-importlib-resources: enable pep517 build

jmroot:
Also add missing setuptools_scm dep.

This is a dependency of py-python-install with some python versions, so it needs to use py-bootstrap-modules to avoid circular deps.

### DIFF
--- a/python/py-importlib-resources/Portfile
+++ b/python/py-importlib-resources/Portfile
@@ -74,6 +74,19 @@ if {${name} ne ${subport}} {
                     port:py${python.version}-zipp
     }
 
+    if {${python.version} >= 36} {
+        depends_build-append \
+                    port:py${python.version}-setuptools_scm
+        python.pep517   yes
+        # break circular dependency with python-install
+        python.add_dependencies no
+        depends_build-append   port:py-bootstrap-modules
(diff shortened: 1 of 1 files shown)


rubendibattista:
>importlib_resources supplies a backport of importlib.resources, enabling early access to features of future Python versions and making functionality available for older Python versions. Users are encouraged to use the Python standard library where suitable and fall back to this library for future compatibility. Developers looking for detailed API descriptions should refer to the standard library documentation.

Theoretically this should be needed only for Python < 3.1. Why do we still ship it for modern Python? Wouldn't be better to fix dependents to use the standard lib?

jmroot:
`importlib` has been around for a while as you say, but `importlib.resources` was added in 3.7. https://docs.python.org/3/library/importlib.html#module-importlib.resources

rubendibattista:
Ok, then I think we should probably ship ip until that version and try to fix the dependents for more recent versions of Python?

jmroot:
It's possible that there are fixes or new features in the newer stdlibs that you might want when using an older python. I don't know.

rubendibattista:
Ok then!

PR #12791, opened 2021-11-03
beets-devel: update to 20211103

catap:
-------

#### Description

@mascguy I guess that this port will build for all supported macOS ;)

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6 20G165 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/audio/beets/Portfile
+++ b/audio/beets/Portfile
@@ -39,13 +39,13 @@ if {$subport eq $name} {
 subport ${name}-devel {
     conflicts       $name
 
-    github.setup    beetbox beets 8fb1c03ca5a10abbf4eccd88b0499b533023c120
-    version         20211102
+    github.setup    beetbox beets 3b531811b982119c6e598433e413e8ca9f2ad068
+    version         20211103
     revision        0
 
-    checksums       rmd160  cc913ce6fb592d374f5c196c36a0db32d15c7bc1 \
-                    sha256  7bbe4a770e96015b92f5a5a0b97fad5bab31b2bd7d2a418b4c314cc7aa42d429 \
-                    size    1663855
+    checksums       rmd160  be782a980a9f3a53054cd22aeec05e929b2bf69f \
+                    sha256  1b4289eed1b63bb05450540dc3f8d31de981497196b49965e53623646e947abf \
+                    size    1665854
 
     depends_build-append \
                     port:py${python.version}-sphinx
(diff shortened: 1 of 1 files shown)


audio/beets/Portfile
```diff
@@ -84,6 +84,7 @@ if {${name} eq ${subport} || "${name}-devel" eq ${subport}} {
 
     if {${name} eq ${subport}} {
         depends_lib-append \
+                    port:py${python.version}-gmusicapi \
```
mascguy, line 87: It looks like this is a duplicate, as you're also adding `py-gmusicapi` as a dependency on line 67.

catap: @mascguy indeed. Because I mean `depends_lib-delete`

mascguy:
I'll validate this locally, later today, when I can find some time.

audio/beets/Portfile
```diff
@@ -83,7 +83,8 @@ if {${name} eq ${subport} || "${name}-devel" eq ${subport}} {
                     port:py${python.version}-yaml
 
     if {${name} eq ${subport}} {
-        depends_lib-append \
+        depends_lib-delete \
+                    port:py${python.version}-gmusicapi \
                     port:py${python.version}-six
```
mascguy, line 88: I don't see `py-six` in the earlier dependency list. Is this supposed to be an add, or...?

mascguy: As for the dependency logic in-general, it would be preferable to invert it: Only add deps that are shared between both, and then selectively add others for the Devel port:

```
if {${name} ne ${subport}} {
    depends_lib-append \
        port:py${python.version}-gmusicapi
}
```

Does that make sense?

catap: @mascguy it was removed from the last update. I've introduced a typo with `-append` instead of `-delete` before.

catap: Hm, it should be here. Let me double check it.

catap: @mascguy you are defently right. Thanks to catch this issue!

catap: @mascguy to make things clear:
 - the last release required both `py-gmusicapi` and `py-six`;
 - the current master is removed this dependency.
 
 I've removed `py-six` at b6fb1fd0e12d3bfb27dd4ccd501233d5ec1a4b42 but here introduced a very bad typo.

mascguy:
Kirill, I tried testing locally. But when attempting to install `beets-full`, I encountered this issue:

[63803 - py-bottlenose/py-anytree: both ports are installing ../Python.framework/Versions/<version>/LICENSE](https://trac.macports.org/ticket/63803)

catap:
@mascguy ah, I'll take a look for this right away.

Meanwhile I'd love to share with you another example of similar issue: https://trac.macports.org/ticket/63481

catap:
@mascguy can you try again?

mascguy (approved):

mascguy:
Just finished retesting, and looks good. Nice work!

PR #12802, opened 2021-11-04
ld64: avoid +llvm90 for ld64_xcode

Gcenx:
#### Description

Avoid forcing `llvm-9.0` when `ld64` defaults to `+ld64_xcode`

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505

###### Verification 

### DIFF
--- a/devel/ld64/Portfile
+++ b/devel/ld64/Portfile
@@ -6,7 +6,7 @@ PortGroup               compiler_blacklist_versions 1.0
 name                    ld64
 epoch                   2
 version                 3
-revision                3
+revision                4
 set dyld_version        655.1.1
 categories              devel
 platforms               darwin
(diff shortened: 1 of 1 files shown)


kencu:
This shouldn't be needed. The ld64-xcode variant has a different code path that doesn't involve any distfiles, llvms, etc.

https://github.com/macports/macports-ports/blob/bdd500177762a9bfb1e676b97bf6fea867418fee/devel/ld64/Portfile#L304

kencu:
what issue are you having that you feel this should fix? Not sure how you would get to that point...

Gcenx:
I've experienced some weirdness where a dependency of a dependency requires `ld64` ended up adding `+llvm90`, but when I directly install `ld64` there no issue.

kencu:
Oh, I never thought of that one. Yes, if some port was called in with what happened to be an +llvm90 variant, it would indeed pass that variant up the line to all the ports it called in.

What a wacky situation! There is just no anticipating everything that might happen, it seems.

mascguy:
Dean, the lint errors - at least those that are easily fixable, such as missing checksum sizes - were bugging me. So I fixed those in a subsequent commit.

Since we're making changes anyway, let's fix these old pesky issues! Lol

Gcenx:
Christopher thanks I wasn’t sure if I should touch anything other than my intended changes since the lint issues have been around for a long time.

It’s unfortunate `ld64-latest` is rather outdated now.\
Haven’t checked cctools etc for a while now since [cctools-port](https://github.com/tpoechtrager/cctools-port) is so easy to build.

mascguy:
@kencu Any concerns? And should we wait for Jeremy's blessing?

kencu:
No opinion. 

My own needs for cctools and ld64-latest are satisfied by our current versions, on my remaining SnowLeopard machine that runs MacPorts. All my other systems have moved to Ubuntu or Debian now, or have used OpenCore to get to BigSur or Monterey.

I don't think we're going to see Jeremy any longer on MacPorts, which is very unfortunate. That was the end of an era. If he does come along, I suspect it will be very spotty, and he won't take ownership of these issues any longer.

PR #12807, opened 2021-11-04
Update to OpenSSL 3 by default

cjones051073:
- update openssl shim port and PG to use openssl 3 by default.
- rev bump all ports depending on openssl (using attached script).

Closes: [63801 - python ports: update to use openssl 3, allowing dependent binaries to be redistributed](https://trac.macports.org/ticket/63801)

[macports-bump-rev.sh.txt](https://github.com/macports/macports-ports/files/7478378/macports-bump-rev.sh.txt)

### DIFF
--- a/net/openssh/Portfile
+++ b/net/openssh/Portfile
@@ -6,7 +6,7 @@ PortGroup           compiler_blacklist_versions 1.0
 
 name                openssh
 version             8.8p1
-revision            0
+revision            1
 categories          net
 platforms           darwin
 maintainers         nomaintainer
--- a/security/cyrus-sasl2/Portfile
+++ b/security/cyrus-sasl2/Portfile
@@ -3,22 +3,10 @@
 PortSystem              1.0
 PortGroup               github 1.0
 PortGroup               muniversal 1.0
-PortGroup               legacysupport 1.0
+PortGroup               legacysupport 1.1
 
 github.setup            cyrusimap cyrus-sasl 2.1.27 cyrus-sasl-
-revision                3
-
-platform darwin 20 {
-    if {${version} eq "2.1.27"} {
-        incr revision
-    }
-    pre-configure {
-        if {[vercmp [macports_version] 2.7.0] < 0} {
-            ui_error "${subport} @${version} requires MacPorts 2.7.0 or greater"
-            return -code error "incompatible MacPorts version"
-        }
-    }
-}
+revision                5
 
 name                    cyrus-sasl2
 categories              security net
--- a/security/yara/Portfile
+++ b/security/yara/Portfile
@@ -8,7 +8,7 @@ PortGroup               legacysupport 1.0
 legacysupport.newest_darwin_requires_legacy 10
 
 github.setup            VirusTotal yara 4.1.3 v
-revision                0
+revision                1
 
 categories              security
 license                 GPL-2+
--- a/sysutils/john/Portfile
+++ b/sysutils/john/Portfile
@@ -76,7 +76,7 @@ destroot {
 subport john-jumbo {
     conflicts           john
     version             1.9.0
-    revision            1
+    revision            2
     set jumbo_patchlvl  1
     license             GPL-2 GPL-3+ Apache-2 Restrictive
     master_sites        https://www.openwall.com/john/k/
--- a/www/nghttp2/Portfile
+++ b/www/nghttp2/Portfile
@@ -47,6 +47,7 @@ if {${subport} eq ${name}} {
 }
 
 subport nghttp2-tools {
+    incr revision
     description         Tools for nghttp2, an implementation of HTTP/2 in C.
     long_description    HTTP/2 client, server and proxy tools, as well as a \
                         load test and benchmarking tool for HTTP/2.
--- a/_resources/port1.0/group/openssl-1.0.tcl
+++ b/_resources/port1.0/group/openssl-1.0.tcl
@@ -27,7 +27,7 @@ default openssl_cache_configure    ""
 proc openssl::default_branch {} {
     # NOTE - Whenever the default branch is bumped, the revision
     # in the openssl shim port should be reset back to 0.
-    return 1.1
+    return 3
 }
 
 proc openssl::branch {} {
--- a/aqua/Chmox/Portfile
+++ b/aqua/Chmox/Portfile
@@ -3,7 +3,7 @@ PortGroup xcode 1.0
 
 name		Chmox
 version		0.4
-revision	4
+revision	5
 categories	aqua textproc
 maintainers	nomaintainer
 description	Read CHM documents on your Mac
--- a/aqua/DateLine/Portfile
+++ b/aqua/DateLine/Portfile
@@ -3,7 +3,7 @@ PortGroup               xcode 1.0
 
 name                    DateLine
 version                 0.61
-revision                2
+revision                3
 categories-append       amusements
 maintainers             nomaintainer
 description             displays linear calendar on desktop
--- a/aqua/MacBiff/Portfile
+++ b/aqua/MacBiff/Portfile
@@ -3,7 +3,7 @@ PortGroup               xcode 1.0
 
 name                    MacBiff
 version                 1.1.16
-revision                1
+revision                2
 categories              aqua net
 supported_archs         i386 x86_64
 maintainers             nomaintainer
--- a/aqua/barrier/Portfile
+++ b/aqua/barrier/Portfile
@@ -7,6 +7,7 @@ PortGroup           qt5 1.0
 PortGroup           app 1.0
 
 github.setup        debauchee barrier 2.3.3 v
+revision            1
 categories          aqua net sysutils
 platforms           darwin
 license             GPL-2
--- a/aqua/osx2x/Portfile
+++ b/aqua/osx2x/Portfile
@@ -6,7 +6,7 @@ PortGroup       xcode 1.0
 name            osx2x
 version         2.4.0
 set my_version  2.4
-revision        8
+revision        9
 set git_hash    3cc708236898ab789bb99a5fba7420ff76ede9f7
 license         BSD
 platforms       darwin
--- a/aqua/pgAdmin3/Portfile
+++ b/aqua/pgAdmin3/Portfile
@@ -5,7 +5,7 @@ PortGroup           wxWidgets 1.0
 
 name                pgAdmin3
 version             1.22.2
-revision            1
+revision            2
 
 categories          aqua
 maintainers         nomaintainer
--- a/aqua/qt4-mac/Portfile
+++ b/aqua/qt4-mac/Portfile
@@ -11,7 +11,7 @@ PortGroup           compiler_blacklist_versions 1.0
 
 name                qt4-mac
 version             4.8.7
-revision            12
+revision            13
 set branch          [join [lrange [split ${version} .] 0 1] .]
 
 categories          aqua
--- a/aqua/qt5/Portfile
+++ b/aqua/qt5/Portfile
@@ -169,7 +169,7 @@ array set modules {
         {"Qt Core" "Qt GUI" "Qt Network" "Qt SQL" "Qt Test" "Qt Widgets" "Qt Concurrent" "Qt D-Bus" "Qt OpenGL" "Qt Platform Headers" "Qt Print Support" "Qt XML"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtcharts {
@@ -919,7 +919,7 @@ foreach {module module_info} [array get modules] {
 
     subport ${name}-${module} {
 
-        distname        ${module}-${middle_name}-src-${version}
+        distname ${module}-${middle_name}-src-${version}
 
         revision ${revision_string}
 
@@ -1836,7 +1836,7 @@ foreach {sql_names sql_info} [array get sql_plugins] {
 
         PortGroup qmake5 1.0
 
-        distname        qtbase-${middle_name}-src-${version}
+        distname qtbase-${middle_name}-src-${version}
 
         revision ${revision_string}
 
@@ -1916,10 +1916,10 @@ foreach {sql_names sql_info} [array get sql_plugins] {
 subport ${name}-docs {
     # meta-port to install documentation for various modules
 
-    revision            0
+    revision          0
 
-    description         Documentation for Qt Tool Kit ${qt_major}
-    long_description    Documentation for Qt Tool Kit ${qt_major}
+    description       Documentation for Qt Tool Kit ${qt_major}
+    long_description  {*}${description}
 
     master_sites
     distfiles
--- a/aqua/qt511/Portfile
+++ b/aqua/qt511/Portfile
@@ -164,7 +164,7 @@ array set modules {
         {"Qt Core" "Qt GUI" "Qt Network" "Qt SQL" "Qt Test" "Qt Widgets" "Qt Concurrent" "Qt D-Bus" "Qt OpenGL" "Qt Platform Headers" "Qt Print Support" "Qt XML"}
         ""
         "variant overrides: "
-        "revision 6"
+        "revision 7"
         "License: "
     }
     qtcanvas3d {
@@ -1715,10 +1715,10 @@ foreach {sql_names sql_info} [array get sql_plugins] {
 subport ${name}-docs {
     # meta-port to install documentation for various modules
 
-    revision            0
+    revision          0
 
-    description         Documentation for Qt Tool Kit ${qt_major}
-    long_description    Documentation for Qt Tool Kit ${qt_major}
+    description       Documentation for Qt Tool Kit ${qt_major}
+    long_description  {*}${description}
 
     master_sites
     distfiles
--- a/aqua/qt513/Portfile
+++ b/aqua/qt513/Portfile
@@ -164,7 +164,7 @@ array set modules {
         {"Qt Core" "Qt GUI" "Qt Network" "Qt SQL" "Qt Test" "Qt Widgets" "Qt Concurrent" "Qt D-Bus" "Qt OpenGL" "Qt Platform Headers" "Qt Print Support" "Qt XML"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtcharts {
@@ -1734,10 +1734,10 @@ foreach {sql_names sql_info} [array get sql_plugins] {
 subport ${name}-docs {
     # meta-port to install documentation for various modules
 
-    revision            0
+    revision          0
 
-    description         Documentation for Qt Tool Kit ${qt_major}
-    long_description    Documentation for Qt Tool Kit ${qt_major}
+    description       Documentation for Qt Tool Kit ${qt_major}
+    long_description  {*}${description}
 
     master_sites
     distfiles
--- a/aqua/x2goclient/Portfile
+++ b/aqua/x2goclient/Portfile
@@ -7,7 +7,7 @@ qt4.debug_variant       no
 
 name                    x2goclient
 version                 4.1.2.2
-revision                0
+revision                1
 checksums               sha256  c9953267c40fa67119ad96a73bacb1f266196da2059f0cdcd1b8d5199421d12a \
                         rmd160  bc8d71587e5c278632150db0a84836963b791b3e \
                         size    2576404
--- a/archivers/dar/Portfile
+++ b/archivers/dar/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                dar
 version             2.7.2
-revision            0
+revision            1
 checksums           rmd160  0a39355abc53699f54c6d49ea81d24302b4364f7 \
                     sha256  973fa977c19b32b1f9ecb62153c810ba8696f644eca048f214c77ad0e8eca255 \
                     size    2294187
--- a/archivers/minizip-ng/Portfile
+++ b/archivers/minizip-ng/Portfile
@@ -5,7 +5,7 @@ PortGroup           cmake 1.1
 PortGroup           github 1.0
 
 github.setup        zlib-ng minizip-ng 3.0.3
-revision            0
+revision            1
 checksums           rmd160  66ffb92ec13c31d9b1870d557ff20ab00ca0fc07 \
                     sha256  5f1dd0d38adbe9785cb9c4e6e47738c109d73a0afa86e58c4025ce3e2cc504ed \
                     size    638966
--- a/archivers/wimlib/Portfile
+++ b/archivers/wimlib/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                wimlib
 version             1.13.4
-revision            0
+revision            1
 checksums           rmd160  3309c5b9a3107c448ea1cb8759829282f563d2ab \
                     sha256  4b87dd0ad9cc1a58cee5721afebb98011dab549e72f2b55533f315f08b2ede12 \
                     size    1040411
--- a/archivers/xar/Portfile
+++ b/archivers/xar/Portfile
@@ -6,7 +6,7 @@ PortGroup           clang_dependency 1.0
 name                xar
 set apple_version   452
 version             1.8.0.${apple_version}
-revision            0
+revision            1
 
 categories          archivers sysutils
 platforms           darwin freebsd linux
--- a/audio/faust-devel/Portfile
+++ b/audio/faust-devel/Portfile
@@ -7,7 +7,7 @@ github.setup            grame-cncm faust 96f22a55e86dac30cbf2e38ba115f5723ed6b09
 # When updating faust-devel to a new version, please rebuild faustlive-devel
 # simultaneously by increasing its revision or updating it to a new version.
 version                 2.11-20181109
-revision                1
+revision                2
 
 name                    faust-devel
 conflicts               faust
--- a/audio/faust/Portfile
+++ b/audio/faust/Portfile
@@ -4,7 +4,7 @@ PortSystem              1.0
 PortGroup               github 1.0
 
 github.setup            grame-cncm faust 2.5.23
-revision                1
+revision                2
 github.tarball_from     releases
 
 conflicts               faust-devel
--- a/audio/faustlive-devel/Portfile
+++ b/audio/faustlive-devel/Portfile
@@ -5,7 +5,7 @@ PortGroup               github 1.0
 
 github.setup            grame-cncm faustlive f7edb2ff41d4987fb302fbce51723616eb728f19
 version                 2.46-20181019
-revision                2
+revision                3
 
 name                    faustlive-devel
 categories              audio
--- a/audio/gmpc/Portfile
+++ b/audio/gmpc/Portfile
@@ -4,7 +4,7 @@ PortSystem 1.0
 
 name            gmpc
 version         11.8.16
-revision        4
+revision        5
 categories      audio
 platforms       darwin
 license         GPL-2+
--- a/audio/icecast2/Portfile
+++ b/audio/icecast2/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                icecast2
 version             2.4.4
-revision            1
+revision            2
 categories          audio net
 license             GPL-2
 maintainers         nomaintainer
--- a/audio/libshout2/Portfile
+++ b/audio/libshout2/Portfile
@@ -5,7 +5,7 @@ PortSystem      1.0
 name            libshout2
 set my_name     [strsed ${name} {g/[0-9]//}]
 version         2.4.5
-revision        0
+revision        1
 checksums       rmd160  4605469f44989a65b163a1df8559ba3414fd7632 \
                 sha256  d9e568668a673994ebe3f1eb5f2bee06e3236a5db92b8d0c487e1c0f886a6890 \
                 size    543991
--- a/audio/opusfile/Portfile
+++ b/audio/opusfile/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                opusfile
 version             0.12
-revision            0
+revision            1
 categories          audio
 license             BSD
 platforms           darwin
--- a/audio/xmms2/Portfile
+++ b/audio/xmms2/Portfile
@@ -5,7 +5,7 @@ PortGroup               waf 1.0
 
 name                    xmms2
 version                 0.8DrO_o
-revision                22
+revision                23
 categories              audio
 # Mostly LGPL, some plugins and clients are GPL
 license                 LGPL-2.1+ GPL-2+ GPL-2
--- a/benchmarks/polygraph/Portfile
+++ b/benchmarks/polygraph/Portfile
@@ -4,7 +4,7 @@ PortSystem              1.0
 
 name                    polygraph
 version                 4.3.2
-revision                3
+revision                4
 categories              benchmarks www
 platforms               darwin
 maintainers             nomaintainer
--- a/comms/c3270/Portfile
+++ b/comms/c3270/Portfile
@@ -2,7 +2,7 @@ PortSystem          1.0
 
 name                c3270
 version             3.3.15ga9
-revision            2
+revision            3
 set branch          [join [lrange [split ${version} .] 0 1] .]
 categories          comms
 platforms           darwin
--- a/comms/sofia-sip/Portfile
+++ b/comms/sofia-sip/Portfile
@@ -4,7 +4,7 @@ PortSystem      1.0
 
 name            sofia-sip
 version         1.12.11
-revision        2
+revision        3
 license         LGPL-2.1
 categories      comms
 maintainers     nomaintainer
--- a/comms/telepathy-idle/Portfile
+++ b/comms/telepathy-idle/Portfile
@@ -4,7 +4,7 @@ PortSystem 1.0
 
 name            telepathy-idle
 version         0.2.0
-revision        3
+revision        4
 license         LGPL-2.1
 description     The ${name} IRC connection manager component of Telepathy - a Flexible Communications Framework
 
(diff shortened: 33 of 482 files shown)


mascguy (approved):

jmroot:
Shouldn't only the ports using the default openssl version be rev bumped?

cjones051073:
> Shouldn't only the ports using the default openssl version be rev bumped?

my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...

cjones051073:
> > Shouldn't only the ports using the default openssl version be rev bumped?
> 
> my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...

to be clear, I have no planes to do any more or less than my script does. if that rev bumps ports that do not need it (unlikely) or misses ports, so be it..

mascguy:
In terms of the logistics of ultimately merging this, do we expect the buildbots to handle this without dying?

cjones051073:
> In terms of the logistics of ultimately merging this, do we expect the buildbots to handle this without dying?

no. they will always certainly timeout...

> At the very least, the watchers are going to be tied up for a LONG time, processing the full dependency list...

ultimately I believe it comes down to a judgement call. do we 'merge and deal with the aftermath as is' or .... personally I am not sure there is much of an or...

cjones051073:
> > In terms of the logistics of ultimately merging this, do we expect the buildbots to handle this without dying?
> 
> no. they will always certainly timeout...

Just be 100%B clear I assume you are referring to the CI test builds here. these will probably timeout. The macports buildbots will handle then update though just fine...

mascguy:
> > > In terms of the logistics of ultimately merging this, do we expect the buildbots to handle this without dying?
> > 
> > 
> > no. they will always certainly timeout...
> 
> Just be 100% clear, I assume you are referring to the CI test builds here ? these will probably timeout. The macports buildbots will handle then update though just fine...

Nope, I'm talking about the buildbots. If it's one giant commit, the watchers will be tied up for an eternity, whilst leaving the builders idle.

mascguy:
My recommendation - barring more definitive guidance from Josh and Ryan - would be to limit the number of rev-bumps to no more than 50 per commit.

cjones051073:
> My recommendation - barring more definitive guidance from Josh and Ryan - would be to limit the number of rev-bumps to no more than 50 per commit.
> 
> And while we're not dealing with thousands of rev-bumps, I'm concerned we might have an issue like this from a few months ago:
> 
> https://lists.macports.org/pipermail/macports-dev/2021-June/043605.html

I am not going to do that. I don’t thing this PR is that large that merging it as is is that much of an issue.

jmroot:
> my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...

That doesn't appear to be the case. For example python26 is bumped and it only depends on openssl10.

cjones051073:
> > my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...
> 
> That doesn't appear to be the case. For example python26 is bumped and it only depends on openssl10.

My script is probably not perfect. Bar doing it by hand, which I am not going to do, its better than nothing.

cjones051073:
> > my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...
> 
> That doesn't appear to be the case. For example python26 is bumped and it only depends on openssl10.



> > > my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...
> > 
> > 
> > That doesn't appear to be the case. For example python26 is bumped and it only depends on openssl10.
> 
> My script is probably not perfect. Bar doing it by hand, which I am not going to do, its better than nothing.



> > > my script rev-bumps ports that depend on the openssl shim port. nothing more nothing less...
> > 
> > 
> > That doesn't appear to be the case. For example python26 is bumped and it only depends on openssl10.
> 
> My script is probably not perfect. Bar doing it by hand, which I am not going to do, its better than nothing.

```
Oberon ~/Downloads > port echo depends:openssl | grep python
python26                        
python27                        
python27-bootstrap              
python32                        
python33                        
python34                        
python35                        
python36                        
python37                        
python38                        
python39                        
python310                       
python311-devel 
```

if python26 should not appear in the list above, thats a bug in base.

jmroot:
It's a bug in your regex. Try `depends:':openssl(\s|$)'`.

mascguy:
Hey Chris, bear in mind that we're not trying to be critical of what you've done here. And we appreciate you taking the initiative, which is often the most important step!

Others are happy to assist, to help refine all of this. Whether that means breaking up the commits into smaller chunks, trimming the list of rev-bumps, or whatever. As long as you don't block commits to this PR by your fellow members, we'll help too!

cjones051073:
> It's a bug in your regex. Try `depends:':openssl(\s|$)'`.

thanks. TBH I still view this as a conceptual bug in base that I have to do this to avoid unintentional matches...

cjones051073:
> Hey Chris, bear in mind that we're not trying to be critical of what you've done here. And we appreciate you taking the initiative, which is often the most important step!
> 
> Others are happy to assist, to help refine all of this. Whether that means breaking up the commits into smaller chunks, trimming the list of rev-bumps, or whatever. As long as you don't block commits to this PR by your fellow members, we'll help too!
> 
> p.s. We love ya Man!

thats fine, but just to be 100% clear I am not going to waste time myself breaking this up into smaller MRs. I don't see this as being necessary. I generated the changes here by a) updating the openssl Pam to default to 3 and b) running my attached script to auto-rev bump dependent ports. thats all I am prepared to do, but you want more fine grained than that please take the script and knock yourself out ;)...

kencu:
Well well well. Deja vu.

Tatsh:
I am approving of this OpenSSL update for OpenRCT2 😄

RJVB:
On Thursday November 04 2021 15:36:15 Chris Jones wrote:

>thats fine, but just to be 100% clear I am not going to waste time myself breaking this up into smaller MRs.

I suppose one could get the rev-bump approval here, but drop the actual change like it's already been done elsewhere AFAIK. If shouldn't take a lot of time to adapt your script so it applies only to approved ports, and applies the changes to your working copy of the actual ports tree. It could even do the pushing at some kind of interval (but I doubt the build bots will be less stressed by a series of commits than a single big commit).

What I'm missing is a little bit of background. I wasn't even aware of OpenSSL3 (what ever happened to OpenSSL2?), only vaguely about the OpenSSL PG. I think the least would be to touch a few words on why the rev-bump is necessary. Is it because of an ABI incompatibility or some potentially less stringent reason?

lhaeger:
https://wiki.openssl.org/index.php/OpenSSL_3.0#Main_Changes_in_OpenSSL_3.0_from_OpenSSL_1.1.1

> Am 05.11.2021 um 09:55 schrieb René Bertin ***@***.***>:
> 
> What I'm missing is a little bit of background.

ryandesign:
> > In terms of the logistics of ultimately merging this, do we expect the buildbots to handle this without dying?
> 
> no. they will almost certainly timeout...

Buildbot will not time out. The only situation in which our buildbot will time out a portbuilder build is if it goes for 60 minutes without producing any output, and I'm not aware of any ports in our collection that do that. There are no timeouts associated with portwatcher builds; they are allowed to run indefinitely. Barring a power outage, it should work.

> > At the very least, the watchers are going to be tied up for a LONG time, processing the full dependency list...

Yes, it will take a long time, hours most likely. But I don't think it can be avoided, so if we're doing this change, go ahead and do it.

> So before merging, we might want to break this up into at a few separate commits. (Preferably no more than 100 rev-bumps each.)

That probably won't be very effective, unless you commit a batch of revbumps, wait for all builders to begin working on those, then commit another batch, wait for buildbot to get to that batch, etc. It would also be tedious to try to construct small batches of ports in which there are no dependencies on other ports that use openssl.

Buildbot always updates to the latest code in the macports-ports repository before any build. So if you push 5 commits each at least 5 seconds apart, and assuming buildbot was idle when you did so, then buildbot will get to work on the first batch immediately, but by the time it gets to the second batch, the third through fifth batches are already waiting so the ports tree that the second batch uses is current up to the fifth batch, or whatever subsequent commits came in. If any of the ports in commit 2 depend on any of the ports modified in commits 3-5, those ports will get scheduled in response to commit 2.

In addition, all commits made within 5 seconds of one another are merged into one portwatcher build. This applies for example if you have multiple commits in a single pull request or if you push multiple commits to master manually at once.

I would probably not bother with separate small batches and just do it all at once and accept that subsequent builds will have to wait.

RJVB:
To save others a bit of clicking:

> OpenSSL 3.0 is a major release and consequently any application that currently uses an older version of OpenSSL will at the very least need to be recompiled in order to work with the new version. It is the intention that the large majority of applications will work unchanged with OpenSSL 3.0 if those applications previously worked with OpenSSL 1.1.1. However this is not guaranteed and some changes may be required in some cases. Changes may also be required if applications need to take advantage of some of the new features available in OpenSSL 3.0 such as the availability of the FIPS module. 


> https://wiki.openssl.org/index.php/OpenSSL_3.0#Main_Changes_in_OpenSSL_3.0_from_OpenSSL_1.1.1
> […](#)
> 05.11.2021 09:55 René Bertin ***@***.***>: What I'm missing is a little bit of background.

cjones051073:
> I suppose one could get the rev-bump approval here, but drop the actual change like it's already been done elsewhere AFAIK. If shouldn't take a lot of time to adapt your script so it applies only to approved ports, and applies the changes to your working copy of the actual ports tree. It could even do the pushing at some kind of interval (but I doubt the build bots will be less stressed by a series of commits than a single big commit). What I'm missing is a little bit of background. I wasn't even aware of OpenSSL3 (what ever happened to OpenSSL2?), only vaguely about the OpenSSL PG. I think the least would be to touch a few words on why the rev-bump is necessary. Is it because of an ABI incompatibility or some potentially less stringent reason?

OpenSSL changed their version strategy with version 3. It always kinda followed a MAJOR.MINOR.PATCH scheme, but they only guaranteed ABI compatibility on MAJOR.MINOR. This is why the previous two 'major' versions were 1.0.x and 1.1.x. The other side affect of this was having only the PATCH level left to play with was not enough granularity, which is why we got the weird alpha-numeric patch versions, like 1.1.1k etc.

Now with version 3, ABI compatibility is only on the MAJOR version. So the MINOR can (and will) change whilst remaining ABI compatible. We also (hopefully) will not need the alpha-numeric patch levels any more.

So, to answer your question, there was no version 2 as effectively 1.1.x was the second 'major' version.

kimuraw:
@cjones051073 , all of versions of ruby (ruby/openssl) do not support openssl-3.
please remove lang/ruby* from this PR.

I'll set openssl.branch to 1.1 for ruby24 .. ruby30.

- ruby/openssl do not support version 2.2.x or earlier
  - https://github.com/ruby/openssl/issues/369
- latest version of ruby/openssl is 2.2.1
  - ruby-3.0.2 includes 2.2.0, build fails with port:openssl3

cjones051073:
> @cjones051073 , all of versions of ruby (ruby/openssl) do not support openssl-3. please remove lang/ruby* from this PR.
> 
> I'll set openssl.branch to 1.1 for ruby24 .. ruby30.
> 
> * ruby/openssl do not support version 2.2.x or earlier
>   
>   * [Support OpenSSL 3.0 ruby/openssl#369](https://github.com/ruby/openssl/issues/369)
> * latest version of ruby/openssl is 2.2.1
>   
>   * ruby-3.0.2 includes 2.2.0, build fails with port:openssl3

I cannot 'remove' them from this PR in the sense that this PR updates the default openssl version to 3.

If/when we do make the switch to use openssl3 by default, the ruby ports which do not  support this version will need to be adapted to directly use the openssl PG to peg themselves back to openssl 1.1.

cjones051073:
@kimuraw I have removed the rev-bump of all ruby ports here, so this PR no long touches them. you can update them to directly use the openssl PG to peg themselves to 1.1 as and when you want, it doesn't have to wait for this change.

RJVB:
port:VLC2 should be removed too. It doesn't build currently so there's no point in revbumping.

multimedia/mythtv.28/Portfile
```diff
@@ -639,7 +639,7 @@ if {$subport eq "mythtv-plugins${majorversion}"} {
 
 if {$subport eq "mythtv${majorversion}"} {
     version             0${majorversion}${minorversion}-Fixes-${lastcommit}
-    revision            1
+    revision            2
```
jmroot, line 642: The mythtv.28 port doesn't need to be bumped but the mythtv-core.28 and mythtv-plugins.28 subports do.

net/openssh/Portfile
```diff
@@ -254,7 +254,7 @@ if {${name} eq ${subport}} {
 }
 
 subport ssh-copy-id {
-    revision            0
+    revision            1
```
jmroot, line 257: The ssh-copy-id subport doesn't need to be bumped.

security/cyrus-sasl2/Portfile
```diff
@@ -6,7 +6,7 @@ PortGroup               muniversal 1.0
 PortGroup               legacysupport 1.0
 
 github.setup            cyrusimap cyrus-sasl 2.1.27 cyrus-sasl-
-revision                3
+revision                4
```
jmroot, line 9: It would be nice to bump the revision to 5 and remove the platform darwin 20 block below that increments the revision.

security/yara/Portfile
```diff
@@ -74,7 +74,7 @@ foreach b ${python.branches} {
 
     subport py${python.version}-${name} {
         github.setup            VirusTotal yara-python 4.1.3 v
-        revision                0
+        revision                1
```
jmroot, line 77: The py*-yara subports don't need to be bumped.

sysutils/john/Portfile
```diff
@@ -5,7 +5,7 @@ PortSystem          1.0
 name                john
 conflicts           john-jumbo
 version             1.9.0
-revision            1
+revision            2
```
jmroot, line 8: The john port does not need to be bumped (just the john-jumbo subport).

www/nghttp2/Portfile
```diff
@@ -5,7 +5,7 @@ PortGroup           github 1.0
 PortGroup           legacysupport 1.0
 
 github.setup        nghttp2 nghttp2 1.46.0 v
-revision            0
+revision            1
```
jmroot, line 8: The nghttp2 port doesn't need to be bumped, but the nghttp2-tools subport does.

jmroot (changes requested):
These Portfiles appear to need a rev bump but are not currently included:
```
databases/mariadb-10.2/Portfile
databases/mariadb-10.3/Portfile
databases/mariadb-10.4/Portfile
databases/mariadb-10.5/Portfile
databases/mariadb-10.6/Portfile
databases/mysql5/Portfile
databases/mysql56/Portfile
databases/percona/Portfile
perl/p5-crypt-openssl-ec/Portfile
perl/p5-crypt-openssl-x509/Portfile
perl/p5-git-raw/Portfile
perl/p5-net-dns-sec/Portfile
perl/p5-net-ssh2/Portfile
```
@dbevans @michaelld @ryandesign

kimuraw:
done at 36878b9a51aa0e184f9c1282c631fbedeac3b05b

ports ruby24 .. ruby30 uses openssl11.

> @cjones051073 , all of versions of ruby (ruby/openssl) do not support openssl-3. please remove lang/ruby* from this PR.
> 
> I'll set openssl.branch to 1.1 for ruby24 .. ruby30.
> 
> * ruby/openssl do not support version 2.2.x or earlier
>   
>   * [Support OpenSSL 3.0 ruby/openssl#369](https://github.com/ruby/openssl/issues/369)
> * latest version of ruby/openssl is 2.2.1
>   
>   * ruby-3.0.2 includes 2.2.0, build fails with port:openssl3

gctwnl:
I do maintain the nsd port (updates etc.) but this is the first time I have seen such a request as this.

I have a local clone of my own fork, so my guess s something like

- create a git branch for this in my own clone of my own fork
- git pull upstream name-of-branch-that-has-this-PR
- compile nsd (I don’t know of nsd has a testing suite that can run, nor do I use SSL-like stuff in my own nsd setup)

Is that it?

Gerben Wierda (LinkedIn <https://www.linkedin.com/in/gerbenwierda>)
R&A Enterprise Architecture <https://ea.rna.nl/> (main site)
Book: Chess and the Art of Enterprise Architecture <https://ea.rna.nl/the-book/>
Book: Mastering ArchiMate <https://ea.rna.nl/the-book-edition-iii/>

cjones051073:
> I do maintain the nsd port (updates etc.) but this is the first time I have seen such a request as this. I have a local clone of my own fork, so my guess s something like - create a git branch for this in my own clone of my own fork - git pull upstream name-of-branch-that-has-this-PR - compile nsd (I don’t know of nsd has a testing suite that can run, nor do I use SSL-like stuff in my own nsd setup) Is that it? Probably not.

Depends what you aim is ? What do you have in min d doing ?

cjones051073:
> These Portfiles appear to need a rev bump but are not currently included:
> 
> ```
> databases/mariadb-10.2/Portfile
> databases/mariadb-10.3/Portfile
> databases/mariadb-10.4/Portfile
> databases/mariadb-10.5/Portfile
> databases/mariadb-10.6/Portfile
> databases/mysql5/Portfile
> databases/mysql56/Portfile
> databases/percona/Portfile
> perl/p5-crypt-openssl-ec/Portfile
> perl/p5-crypt-openssl-x509/Portfile
> perl/p5-git-raw/Portfile
> perl/p5-net-dns-sec/Portfile
> perl/p5-net-ssh2/Portfile
> ```
> 
> @dbevans @michaelld @ryandesign

Thanks! These look like portfiles that set the revision in 'non standard' ways and the script I used was unable to handle them.

cjones051073:
> done at [36878b9](https://github.com/macports/macports-ports/commit/36878b9a51aa0e184f9c1282c631fbedeac3b05b)
> 
> ports ruby24 .. ruby30 uses openssl11.
> 
> > @cjones051073 , all of versions of ruby (ruby/openssl) do not support openssl-3. please remove lang/ruby* from this PR.
> > I'll set openssl.branch to 1.1 for ruby24 .. ruby30.
> > 
> > * ruby/openssl do not support version 2.2.x or earlier
> >   
> >   * [Support OpenSSL 3.0 ruby/openssl#369](https://github.com/ruby/openssl/issues/369)
> > * latest version of ruby/openssl is 2.2.1
> >   
> >   * ruby-3.0.2 includes 2.2.0, build fails with port:openssl3

Apologies but I did not notice you had pushed commits here, and just force pushed some changes that wiped them out.

The ruby changes do not need to be tied to this PR, so can I suggest you make a separate PR for those, or just push them directly to master.

cjones051073:
> > done at [36878b9](https://github.com/macports/macports-ports/commit/36878b9a51aa0e184f9c1282c631fbedeac3b05b)
> > ports ruby24 .. ruby30 uses openssl11.
> > > @cjones051073 , all of versions of ruby (ruby/openssl) do not support openssl-3. please remove lang/ruby* from this PR.
> > > I'll set openssl.branch to 1.1 for ruby24 .. ruby30.
> > > 
> > > * ruby/openssl do not support version 2.2.x or earlier
> > >   
> > >   * [Support OpenSSL 3.0 ruby/openssl#369](https://github.com/ruby/openssl/issues/369)
> > > * latest version of ruby/openssl is 2.2.1
> > >   
> > >   * ruby-3.0.2 includes 2.2.0, build fails with port:openssl3
> 
> @kimuraw Apologies but I did not notice you had pushed commits here, and just force pushed some changes that wiped them out.
> 
> The ruby changes do not need to be tied to this PR, so can I suggest you make a separate PR for those, or just push them directly to master.

Ah, my mistake, the changes are in master already....

cjones051073:
@jmroot Thanks for your comments, I hope I have now addressed them all.

gctwnl:
> > I do maintain the nsd port (updates etc.) but this is the first time I have seen such a request as this. I have a local clone of my own fork, so my guess s something like - create a git branch for this in my own clone of my own fork - git pull upstream name-of-branch-that-has-this-PR - compile nsd (I don’t know of nsd has a testing suite that can run, nor do I use SSL-like stuff in my own nsd setup) Is that it? Probably not.
> 
> Depends what your aim is ? What do you have in mind doing ?

Well, I got these messages (in fact I got all these mentions in my email box) so I assumed I had to do something. I'm trying to find out what. If nothing, that's OK too.

(I noticed the nds version could use a bump too, so it was a good reason to look into that and now I'm trying to fix the expired root certificate problem on my macOS Mojave main workhorse, which Apple just dropped from patching and which I cannot migrate until I have time for it late December)

RJVB:
On Saturday November 06 2021 03:43:26 gctwnl wrote:
> I'm trying to fix the expired root certificate problem on my macOS Mojave main workhorse, which Apple just dropped from patching and which I cannot migrate until I have time for it late December)

Check out the updated port:apple-pki-bundle port for that!

cjones051073:
> > > I do maintain the nsd port (updates etc.) but this is the first time I have seen such a request as this. I have a local clone of my own fork, so my guess s something like - create a git branch for this in my own clone of my own fork - git pull upstream name-of-branch-that-has-this-PR - compile nsd (I don’t know of nsd has a testing suite that can run, nor do I use SSL-like stuff in my own nsd setup) Is that it? Probably not.
> > 
> > 
> > Depends what your aim is ? What do you have in mind doing ?
> 
> Well, I got these messages (in fact I got all these mentions in my email box) so I assumed I had to do something. I'm trying to find out what. If nothing, that's OK too.

You get automatic messages when you are mentioned in a PR, e.g. if a port you maintain is touched. It does not necessarily imply you have to do anything specific.

mascguy:
Hey Chris, given that these changes will tie up the CI build jobs until timeout, go ahead and add `[skip ci]` to the commit comment. Documented via the following blog post:

https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

In the interim, I've been manually cancelling them. But that's a pain. LOL

gctwnl:
> On Saturday November 06 2021 03:43:26 gctwnl wrote: I'm trying to fix the expired root certificate problem on my macOS Mojave main workhorse, which Apple just dropped from patching and which I cannot migrate until I have time for it late December)
> Check out the updated port:apple-pki-bundle port for that!

I wonder how that could help. If the problem is that www.nlnetlabs.nl (the source for nsd) uses a letsencrypt certificate that has a complete chain including the cross-siding of the correct root certificate by the expired root certificate, then changing anything at my end doesn't work.

Instead, I've asked NLnet Labs to update their certificate according the the MacPorts instructions at https://trac.macports.org/wiki/ProblemHotlist#letsencrypt and they will do so.

chrstphrchvz:
Pinging those who might not have been earlier in case of [limit of 50 mentions per comment](https://trac.macports.org/ticket/63821) on behalf of @macportsbot:

> @alfredh for port baresip, libre, librem, libsrtp.
> @michaelld for port qt4-mac, dvisvgm.
> @ryandesign for port dar, minizip-ng, wimlib, faust-devel, faust, faustlive-devel, eet, libhsplasma, libmongo-client, PlasmaClient, bahamut, php, curl, docsis, kerberos5, net-snmp, rabbitmq-c, rsync, shellinabox, php-event, php-mongo, php-stomp, php-swoole, aircrack-ng, openvas-client, rsyncrypto, NetSurf, QupZilla, apache2, lighttpd, slowhttptest, winetricks.
> @jolan78 for port cryfs.
> @ryanakca for port barrier, opensmtpd.
> @markemer for port Io.
> @dgilman for port redis.
> @g5pw for port pev, radare2, rust, esniper, lnav, socat.
> @arjanvandervelde for port poco.
> @davidnich for port qore-asn1-module, qore-json-module, qore-xml-module, qore.
> @danielluke for port bind9, nmap, sysmon, clamav, ntp.
> @lpsinger for port py-twilio, ds9, htcondor, lscsoft-deps.
> @agraef for port faust-devel, faust, faustlive-devel.
> @roederja for port msodbcsql, msodbcsql17, mico, sequoia-pgp.
> @gpanders for port netdata.
> @nareshov for port git-crypt.
> @schiebel for port mico.
> @tobypeterson for port ldmud, epic4, epic5.
> @jandemter for port py-trio, py-trustme.
> @fhgwright for port libaes_siv, ntpsec.
> @jvirkki for port dupd.
> @Schamschula for port xar, xeus-sqlite, xeus, lftp, tcpdump, zeek, py-acme, py-ndg_httpsclient, certbot, stunnel, apache2, goaccess, nghttp2, nginx.
> @mascguy for port openldap-devel, openldap.
> @sikmir for port lagrange.
> @jrjsmrtn for port wpa_passphrase.
> @tenomoto for port nco.
> @Raimondi for port lastpass-cli.
> @toy for port stressdrive.
> @mklein-de for port cadaver.
> @raimue for port gmpc, hexchat, sslscan.
> @dbevans for port opusfile, telepathy-idle, libgda5, libgit2, podofo, deluge, libtorrent-rasterbar, p5-crypt-smime.
> @mohd-akram for port crystal, websocketpp, cpprestsdk.
> @jcvernaleo for port alpine.
> @nano103 for port libp11.
> @amake for port yubikey-manager.
> @catap for port retdec.
> @cardi for port weechat.
> @alexeyt820 for port librdkafka.
> @millerdev for port openfortivpn.
> @bwalle for port crackpkcs12.
> @Gminfly for port libssh2, medusa.
> @NicosPavlov for port kdelibs4, smokekde.
> @danchr for port pypy, p5-net-eboks.
> @essandess for port mail-server, apple-pki-bundle, calendar-contacts-server, privoxy.
> @akkornel for port munge.
> @Jakker for port softhsm.
> @jerryyhom for port libevent, alpine.
> @Veence for port gdal, qgis, qgis3.

danchr (approved):
I do not expect any particular breakage for PyPy, but I will check the port once it's built. Also: Yes, please! This will matter greatly for the Mercurial ports I also maintain.

jmroot (approved):

judaew (approved):
LGTM for my ports

lbschenkel:
I'm happy with whatever is the consensus and if the ports build.

jmroot:
> You get automatic messages when you are mentioned in a PR, e.g. if a port you maintain is touched. It does not necessarily imply you have to do anything specific.

However, you can test if your port builds against openssl 3 by adding `PortGroup openssl 1.0` and in most cases simply setting `openssl.branch 3` (some extra portgroup options may be needed to help some ports find the right openssl version). If it fails to build, you can set `openssl.branch 1.1` instead for the time being, and ask for the port to be excluded from this PR.

toy:
I've managed to check stressdrive against openssl-3, but for that I needed to change include and lib paths in `Portfile` (`${prefix}/include` to `${prefix}/include/openssl-3` and `${prefix}/lib` to `${prefix}/lib/openssl-3`). Do I understand right that with this PR the version of openssl that will be symlinked directly to include and lib will change to version 3?

cjones051073:
> I've managed to check stressdrive against openssl-3, but for that I needed to change include and lib paths in `Portfile` (`${prefix}/include` to `${prefix}/include/openssl-3` and `${prefix}/lib` to `${prefix}/lib/openssl-3`). Do I understand right that with this PR the version of openssl that will be symlinked directly to include and lib will change to version 3?

Please do not hardcode the paths above, but use the proc methods provided in the PG to access the correct include and lib paths. i.e. `[openssl::include_dir]` and `[openssl::lib_dir]`.

But yes, there are two main ways you can use openssl. 

 1. The most common is to use the `openssl` shim port, that installs sym-links to the default openssl version (currently 1.1, to become 3 with this PR)  into the primary macports prefix. 
 2. Use the openssl PG to configure your port to use the isolated build areas, under `libexec`. If you do this, but do not give a version you will use the default, which is the same as in 1. If you do give a version that will be used regardless of whatever the default is. 

Right now without this PR to test against openss3 you have to follow 2., as the default is 1.1. Once this PR is committed though you are correct the default will become 3, so if your port is OK with this you can if you prefer carry on with approach 1. and just use the shim port, with a rev-bump to trigger a rebuild.

cjones051073:
> I'm happy with whatever is the consensus and if the ports build.

I do not guarantee all the ports build as a) many don't build anyway, regardless of the openssl version and b) If a port needs to be pinned back to openssl 1.1 this will need to be done separately to this PR, I am not going to do that here. This can be tested and do before, or after, this PR is merged.

cjones051073:
> Hey Chris, given that these changes will tie up the CI build jobs until timeout, go ahead and add `[skip ci]` to the commit comment. Documented via the following blog post:
> 
> https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/
> 
> In the interim, I've been manually cancelling them. But that's a pain. LOL

I was not aware of that - Useful to know thanks...

Schamschula:
I got clean rebuilds on all my packages.

However, I found two issues unrelated to this PR:

1. zeus-sqlite fails to build under Mojave (and likely any older macOS version) due to an issue with Apple clang and c++-17 (missing the `filesystem` include file)
2. py-ndg_httpsclient is missing the py-setuptools build dependency. I'm not sure how this could have been missed. This has to be a separate commit.

cjones051073:
I think I am going to go ahead and merge this, as given the number of ports it touches I am constantly having to rebase it to fix conflicts as those ports get independent updates. We seem to have approval here and I have not heard anything on the mailing lists objecting. I know it hasn't been a very long period, but we are going to have to hit go on this sometime to switch to the latest openssl so it might as well be now...

catap:
I'm testing both `retdec` ports, but it requires couple of hours to build each. It should work, but let me be sure.

RJVB:
On Sunday November 07 2021 04:38:34 Chris Jones wrote:
>Merged #12807 into master.

No offense, but you *could* have left a little bit more reacting time for those of us who've been too busy with more (sic) urgent matters, esp. on a sunday...

I was under the impression that I was co-maintainer on port:qca (and then only port:qca-qt5) so didn't say anything about it. I'll have to take a look how this is going to work out for the qca*-ossl and -pkcs11 subports.

mascguy:
Geez Chris, thanks for tying up the Buildbot Watchers for hours...  ;-)

Just kidding...

ryandesign:
> @ryandesign for port ... php ...

Looks like the relevant php subports did not end up getting revbumped in this PR after all; I'll do them. afflib was missed too; see f08053a254f3222198b15bf473dbb4b066f3fcd4. I wonder if any others were missed.

mascguy:
Thus far, the failures have been few and far between, so that's good!

There is one notable failure though, for `cargo-c`. It appears that it's being built prior to `openssl` perhaps...?

```
--->  Building cargo-c
DEBUG: system -W /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_cargo-c/cargo-c/work/cargo-c-0.7.3: /opt/local/bin/cargo update
dyld: Library not loaded: /opt/local/libexec/openssl11/lib/libssl.1.1.dylib
  Referenced from: /opt/local/bin/cargo
  Reason: image not found
Command failed: /opt/local/bin/cargo update
```

https://ports.macports.org/port/cargo-c/builds/

The port has a dependency on `openssl`, so I'd expect that to be built first... preventing this issue. @cjones051073 any thoughts/ideas?

CC: @MarcusCalhoun-Lopez

mascguy:
Two more Cargo-dependent ports, exhibiting the same issue (due to a broken openssl dylib on the buildbots):

* amp
* angle-grinder

CC: @herbygillot

pmetzger:
As I mentioned on the bug tracker, ruby30 seems to be broken. It's mysterious because it's part of portgroup openssl 1.0

https://trac.macports.org/ticket/63845

onodera-punpun:
Cargo-c indeed seems to be failing, coincidentally the port is very outdated.

cjones051073:
> Thus far, the failures have been few and far between, so that's good!
> 
> There is one notable failure though, for `cargo-c`. It appears that it's being built prior to `openssl` perhaps...?
> 
> ```
> --->  Building cargo-c
> DEBUG: system -W /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_devel_cargo-c/cargo-c/work/cargo-c-0.7.3: /opt/local/bin/cargo update
> dyld: Library not loaded: /opt/local/libexec/openssl11/lib/libssl.1.1.dylib
>   Referenced from: /opt/local/bin/cargo
>   Reason: image not found
> Command failed: /opt/local/bin/cargo update
> ```
> 
> https://ports.macports.org/port/cargo-c/builds/
> 
> The port has a dependency on `openssl`, so I'd expect that to be built first... preventing this issue. @cjones051073 any thoughts/ideas?
> 
> CC: @MarcusCalhoun-Lopez
> 
> p.s. I'm also seeing a build failure locally, as is at least one of our users. Note that this is definitely not specific to Monterey, as my failure is occuring on 10.13:
> 
> [63848 - cargo-c @0.7.3_1 error: pasting formed 'RUST_VERSION_OPENSSL_](https://trac.macports.org/ticket/63848)

the cargo port needed a rev-bump, it was missed here. I have already submitted that.

essandess:
OpenSSL3 breaks the port `tor` build: https://trac.macports.org/ticket/63870.

RJVB:
So, mostly out of curiosity, now that I'm merging these changes into my own ports tree:

Am I right that there was no immediate need to rev-bump the dependents because they would have continued to find the OpenSSL 1.1.1 libraries in the same locations?

I also have my questions as to the use of an all-in subprefix in libexec with a bunch of symlinks, instead of using `configure` options  to specify the same versioned directories (like lib/openssl-XY) to install just the categories that contain files which have the same name for all opensslXY ports. I've been doing that in my own implementation and it works just fine, but well, whatever :)

gctwnl:
Now that NLNetLabs has fixed their cert I wanted to go on with an NSD update and that also means OpenSSL3 (as that has been released in MacPorts). I started a compile of the existing NSD version.

OpenSSL3 breaks the _existing_ NSD port (v4.2.1) as out has removed ERR_GET_FUNC:

NSD 4.3.8 builds fine. I am testing (lightly) and will prepare a pull request for NSD asap.

PR #12837, opened 2021-11-06
qt5*-qtbase: explicitly disable Vulkan support

chrstphrchvz:
Do not opportunistically enable Vulkan support (maybe it can be explicitly enabled later if desired)
Fixes: https://trac.macports.org/ticket/62104

###### Type(s)

- [x] bugfix

###### Tested on

Only configure phase is tested.
macOS 10.15.7
Xcode 12.4 command line tools (I disabled `use_xcode yes`)

###### Verification 
Have you

### DIFF
--- a/aqua/qt5/Portfile
+++ b/aqua/qt5/Portfile
@@ -169,7 +169,7 @@ array set modules {
         {"Qt Core" "Qt GUI" "Qt Network" "Qt SQL" "Qt Test" "Qt Widgets" "Qt Concurrent" "Qt D-Bus" "Qt OpenGL" "Qt Platform Headers" "Qt Print Support" "Qt XML"}
         ""
         "variant overrides: "
-        "revision 3"
+        "revision 4"
         "License: "
     }
     qtcharts {
@@ -1292,6 +1292,12 @@ foreach {module module_info} [array get modules] {
                 -no-kms                 \
                 -no-libinput            \
                 -no-system-proxies
+
+            # do not opportunistically enable Vulkan support
+            # (TODO: is Vulkan support desirable?)
+            # see https://trac.macports.org/ticket/62104
+            configure.args-append \
+                -no-feature-vulkan
 
             # MacOS/iOS options:
             configure.args-append    \
--- a/aqua/qt511/Portfile
+++ b/aqua/qt511/Portfile
@@ -1149,6 +1149,12 @@ foreach {module module_info} [array get modules] {
                 -no-libinput            \
                 -no-system-proxies
 
+            # do not opportunistically enable Vulkan support
+            # (MoltenVK support unavailable in Qt < 5.12)
+            # see https://trac.macports.org/ticket/62104
+            configure.args-append \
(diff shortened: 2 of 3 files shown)


chrstphrchvz:
@MaddTheSane would you be able to test this change?

I would like to at least confirm that after configure for `qt5-qtbase` finishes, main.log contains:

```
:info:configure   Vulkan ................................. no
```

MaddTheSane:
Yes, it does change the Vulkan configure to no.

mascguy (approved):

chrstphrchvz:
Applied and tested same change to qt513-qtbase.

chrstphrchvz:
Applied and tested same change to qt511-qtbase, even though it does not support MoltenVK (Vulkan on macOS).

chrstphrchvz:
I am under the impression that revision bumps are not necessary in this case. I think they may only be needed if someone managed to configure and build qtbase successfully using a compatible Vulkan/MoltenVK installation, which MacPorts does not seem to have offered—I don‘t know whether the scope of revision bumping includes avoiding stuff in locations which may/are known to interfere with MacPorts (e.g. /usr/local).

mascguy:
> I am under the impression that revision bumps are not necessary in this case. I think they may only be needed if someone managed to configure and build qtbase successfully using a compatible Vulkan/MoltenVK installation, which MacPorts does not seem to have offered—I don‘t know whether the scope of revision bumping includes avoiding stuff in locations which may/are known to interfere with MacPorts (e.g. /usr/local).

Given that the end result will be different - at least for users who have a Vulkan/MoltenVK installation exists - it probably makes sense to rev-bump.

Anyone else?

Gcenx:
> I am under the impression that revision bumps are not necessary in this case. I think they may only be needed if someone managed to configure and build qtbase successfully using a compatible Vulkan/MoltenVK installation, which MacPorts does not seem to have offered—I don‘t know whether the scope of revision bumping includes avoiding stuff in locations which may/are known to interfere with MacPorts (e.g. /usr/local).

It’s lightly safe to assume nobody has a recent version of MoltenVK installed unless they themselves installed it.

My own updated MoltenVK Port doesn’t install any headers and due to an upstream bug MoltenVK.xcframework is also missing the headers.

By default Macports ignore /usr/local so even having a working version of MoltenVK installed using the VulkanSDK installer it won’t be found by Macports.

Gcenx:
@mascguy better to do the rev-bump just to be safe.

SDL2 also does this as well

chrstphrchvz:
Revisions incremented for qt5-qtbase and qt513-qtbase, since I haven’t proven they will always fail to build Vulkan support.

chrstphrchvz:
Can whoever merges this please also close https://trac.macports.org/ticket/63822 as duplicate?

mascguy:
It looks like the CI jobs are failing. Is that expected?

chrstphrchvz:
> It looks like the CI jobs are failing. Is that expected?

Yes, because Qt takes too long to build.

PR #12848, opened 2021-11-07
openssl3: Fix i386 compile

Gcenx:
#### Description

Currently unable to compile for i386 arch, using a draft upstream patch resolves the issue.

```
:info:build ${LDCMD:-/usr/bin/clang} -arch i386 -pipe -Os -arch i386 -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX10.13.sdk -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.13.sdk -Wl,-search_paths_first -L/opt/local/lib -Wl,-headerpad_max_install_names -arch i386 -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX10.13.sdk -L/opt/local/lib -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX10.13.sdk \
:info:build 		-o test/algorithmid_test \
:info:build 		test/algorithmid_test-bin-algorithmid_test.o \
:info:build 		test/libtestutil.a libcrypto.a -lz 
:info:build ld: warning: The i386 architecture is deprecated for macOS (remove from the Xcode build setting: ARCHS)
:info:build ld: warning: The i386 architecture is deprecated for macOS (remove from the Xcode build setting: ARCHS)
:info:build Undefined symbols for architecture i386:
:info:build   "___atomic_is_lock_free", referenced from:
:info:build       _CRYPTO_atomic_or in libcrypto.a(libcrypto-lib-threads_pthread.o)
:info:build       _CRYPTO_atomic_load in libcrypto.a(libcrypto-lib-threads_pthread.o)
:info:build Undefined symbols for architecture i386:
:info:build   "___atomic_is_lock_free", referenced from:
:info:build       _CRYPTO_atomic_or in libcrypto.a(libcrypto-lib-threads_pthread.o)
:info:build       _CRYPTO_atomic_load in libcrypto.a(libcrypto-lib-threads_pthread.o)
:info:build ld: symbol(s) not found for architecture i386
:info:build ld: symbol(s) not found for architecture i386
:info:build clang: error: linker command failed with exit code 1 (use -v to see invocation)
:info:build clang: error: linker command failed with exit code 1 (use -v to see invocation)
:info:build make[1]: *** [test/algorithmid_test] Error 1
:info:build make[1]: *** Waiting for unfinished jobs....
:info:build make[1]: *** [fuzz/cmp-test] Error 1
```

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505

###### Verification 
Have you

### DIFF
--- a/devel/openssl3/Portfile
+++ b/devel/openssl3/Portfile
@@ -58,6 +58,13 @@ if {${os.platform} eq "darwin" && ${os.major} < 11} {
 #   To be seen if needed with 3.0.0, and if dodes needs updating.
 #patchfiles-append   patch-pre-Sierra.diff
 
+# https://github.com/openssl/openssl/pull/16584
+# https://github.com/openssl/openssl/issues/16551
+# Fixes "Undefined symbols for architecture i386: ___atomic_is_lock_free"
+if { ${configure.build_arch} eq "i386" } || "i386" in ${configure.universal_archs} {
+    patchfiles-append   patch-fix-i386-atomics.diff
+}
+
 set my_name         openssl-${major_v}
 set my_prefix       ${prefix}/libexec/${name}
 
--- a/devel/openssl3/files/patch-fix-i386-atomics.diff
+++ b/devel/openssl3/files/patch-fix-i386-atomics.diff
@@ -1,0 +1,276 @@
+From 47574c32d31fc3c57baebf06a47b096191f9f2cd Mon Sep 17 00:00:00 2001
+From: xtkoba <69125751+xtkoba@users.noreply.github.com>
+Date: Sun, 12 Sep 2021 00:23:16 +0900
+Subject: [PATCH 1/8] Update crypto.h.in
+
+---
+ include/openssl/crypto.h.in | 11 +++++++++--
+ 1 file changed, 9 insertions(+), 2 deletions(-)
+
+diff --git include/openssl/crypto.h.in include/openssl/crypto.h.in
+index 724e2ca5da79..c5e668ef1e3a 100644
+--- include/openssl/crypto.h.in
++++ include/openssl/crypto.h.in
+@@ -85,10 +85,17 @@ __owur int CRYPTO_THREAD_write_lock(CRYPTO_RWLOCK *lock);
+ int CRYPTO_THREAD_unlock(CRYPTO_RWLOCK *lock);
+ void CRYPTO_THREAD_lock_free(CRYPTO_RWLOCK *lock);
+ 
++# ifdef __GNUC__
++typedef __attribute__((__aligned__(8))) uint64_t CRYPTO_ATOMIC_U64;
++# else
++typedef uint64_t CRYPTO_ATOMIC_U64;
++# endif
++
+ int CRYPTO_atomic_add(int *val, int amount, int *ret, CRYPTO_RWLOCK *lock);
+-int CRYPTO_atomic_or(uint64_t *val, uint64_t op, uint64_t *ret,
++int CRYPTO_atomic_or(CRYPTO_ATOMIC_U64 *val, uint64_t op, CRYPTO_ATOMIC_U64 *ret,
+                      CRYPTO_RWLOCK *lock);
+-int CRYPTO_atomic_load(uint64_t *val, uint64_t *ret, CRYPTO_RWLOCK *lock);
++int CRYPTO_atomic_load(CRYPTO_ATOMIC_U64 *val, CRYPTO_ATOMIC_U64 *ret,
++                       CRYPTO_RWLOCK *lock);
+ 
+ /* No longer needed, so this is a no-op */
+ #define OPENSSL_malloc_init() while(0) continue
+
+From e4ad3833d14a02aeed44aed5bc14a7a9fec1f241 Mon Sep 17 00:00:00 2001
+From: xtkoba <69125751+xtkoba@users.noreply.github.com>
+Date: Sun, 12 Sep 2021 00:24:17 +0900
+Subject: [PATCH 2/8] Update init.c
+
+---
+ crypto/init.c | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git crypto/init.c crypto/init.c
+index 6a27d1a8e440..e254251b4c09 100644
+--- crypto/init.c
++++ crypto/init.c
+@@ -34,7 +34,7 @@
+ #include <openssl/trace.h>
+ 
+ static int stopped = 0;
+-static uint64_t optsdone = 0;
++static CRYPTO_ATOMIC_U64 optsdone = 0;
+ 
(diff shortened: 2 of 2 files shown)


mascguy (approved):

pmetzger:
It would be more conservative to apply the patch only if the architecture is i386.

Gcenx:
My compiles are always +universal (i386 x86_64) when done on my Intel system and no issues, as this was taken directly from upstream and won't be applied conditionally is that a good idea?

pmetzger:
The patch isn't in an official release and changes the exposed API a bit so I'm just suggesting maximal conservatism. I'm not sure what the right thing is though; someone else may want to chime in.

mascguy:
Given that this is a Devel port anyway, I'm not too overly worried about breaking anything. So we can start with the less conservative approach, and see how it looks.

And we'll certainly vet this out thoroughly for at least a few days - or perhaps a week-ish - before applying the change to the non-Devel version.

pmetzger:
openssl3 is not a devel port. It's the default openssl in MacPorts now.

mascguy:
My apologies, I was just looking at the open PR for `librsvg-devel`. Too many tabs open at the moment! LOL

Never Mind...  😉

Gcenx:
Let me run another compile of openssl3 with the patch applies conditionally to i386, if the +universal destroot finishes without issue I'll rebase this PR

Gcenx:
```
if { ${configure.build_arch} eq "i386" } || "i386" in ${configure.universal_archs} {
    patchfiles-append   16584.patch
}
```
The above will only apply the patch when build for i386 target, however that means it's applied for both slices when compiling for i386 x86_64.

The patch is to fix a bug with how clang handles atomics so I'd think it's better to just apply it regardless until upstream handles this properly.

pmetzger:
Does the patch leave the ABI intact? I couldn't figure that out.

Gcenx:
> Does the patch leave the ABI intact? I couldn't figure that out.

See https://github.com/openssl/openssl/issues/16551#issuecomment-917427776
The patch works around the suspected clang bug.

pmetzger:
Reading https://github.com/openssl/openssl/pull/16584 does not make me feel comfortable about how baked this is.

Gcenx:
The only real take away from that (to me) is slower atomics operations, the issue with the `make update` perl script shouldn't be an issue for macports.

_I'd rather not need to add this into macports-wine as it might end up living there long term_

Gcenx:
@cjones051073 the patch you recently added should need to be applied on Mojave as openssl3 builds using XCode10 and later.

I didn’t touch the change in this rebased PR however.

cjones051073:
> @cjones051073 the patch you recently added should need to be applied on Mojave as openssl3 builds using XCode10 and later.
> 
> I didn’t touch the change in this rebased PR however.

I don’t know what exactly you are referring to ?

cjones051073:
> My compiles are always +universal (i386 x86_64) when done on my Intel system and no issues, as this was taken directly from upstream and won't be applied conditionally is that a good idea?

As this is a draft patch from upstream it would be better to apply it only when needed, for now, abd then when in a public release can be removed.

Gcenx:
> I don’t know what exactly you are referring to ?

`patch-disable-avx512.diff` doesn't need to be applied on macOS Mojave only below as Xcode10 compiles this without issue (Xcode10 Command Line Tools)

> As this is a draft patch from upstream it would be better to apply it only when needed, for now, and then when in a public release can be removed.

While I don't exactly agree as from reading more into this it restores how atomics worked in prior versions of openssl I'm waiting on a compile to complete before pushing the change.

Gcenx:
Compile with `patch-disable-avx512.diff` not applied using Xcode10 Command Line Tools [main.log](https://github.com/macports/macports-ports/files/7501586/main.log)

FMudanyali (approved):

pmetzger:
@cjones051073 What do you suggest here?

cjones051073:
Seems fine I guess. I'm not really in a position to tst anything w.r.t. i386 (really not keen to start building things universal if I don't have to).

pmetzger:
My inclination is to commit this now that it is applied only for the i386. I'm going to do so as it fixes a broken piece of key software for a large number of users.

devel/openssl3/Portfile
```diff
@@ -58,6 +58,13 @@ if {${os.platform} eq "darwin" && ${os.major} < 11} {
 #   To be seen if needed with 3.0.0, and if dodes needs updating.
 #patchfiles-append   patch-pre-Sierra.diff
 
+# https://github.com/openssl/openssl/pull/16584
+# https://github.com/openssl/openssl/issues/16551
+# Fixes "Undefined symbols for architecture i386: ___atomic_is_lock_free"
+if { ${configure.build_arch} eq "i386" } || "i386" in ${configure.universal_archs} {
```
ryandesign, line 64: Note that this patch will apply on x86_64 systems even when the universal variant is not requested, so long as i386 is in universal_archs. That would warrant a revision increase since it changes things for people who already had openssl3 x86_64 installed.

Gcenx: I’d recent done a similar change for librsvg and the fallback wasn’t triggered when not compiling as +universal (cargo/rust got added to deps)

When compiling as +universal the fallback was forced.

ryandesign: I can't speak to that, but what I stated remains true.

ryandesign: It's easy enough to verify by putting a `ui_msg "patch applied"` inside that conditional:
```
$ port info --name openssl3 build_arch=x86_64 -universal
patch applied
name: openssl3
$ port info --name openssl3 build_arch=x86_64 +universal
patch applied
name: openssl3
$ port info --name openssl3 build_arch=i386 -universal
patch applied
name: openssl3
$ port info --name openssl3 build_arch=i386 +universal
patch applied
name: openssl3
```

Gcenx: Not near my system but what happens if you change the first part of the if statement to `${build_arch} eq "i386"`

Gcenx: Or more what would be the correct way to handle this so it’s only getting applied for i386 or a universal build that contains i386?

ryandesign: I would not expect that to change anything.

If you want to restrict something to when the universal variant is set, you have to add the condition `${universal_possible} && [variant_isset universal]`.

```
if {(${configure.build_arch} eq "i386") || (${universal_possible} && [variant_isset universal] && "i386" in ${configure.universal_archs})} {
```

Since some users may already have built the previous version, the revision should be increased when this change is made.

Gcenx: So ether this will need to wait on me getting home and opening a PR or a Macports maintainer to push the the change and bump the revisit.

PR #12851, opened 2021-11-08
libreoffice: update to 7.2.3.1

Tatsh:
#### Description

Bump LibreOffice to 7.2.3.1

Also attempting to use system libcmis in the second commit. I will remove it if it does not work in CI.

I cannot test this locally as glm does not build on Monterey.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/glm/Portfile
+++ b/devel/glm/Portfile
@@ -25,6 +25,8 @@ github.tarball_from releases
 use_7z              yes
 worksrcdir          ${name}
 
+patchfiles          patch-remove-werror.diff
+
 post-extract {
     # The directories in the archive are restricted so that only the extracting
     # user can look inside them.
--- a/devel/glm/files/patch-remove-werror.diff
+++ b/devel/glm/files/patch-remove-werror.diff
@@ -1,0 +1,11 @@
+--- test/CMakeLists.txt.old	2021-11-10 03:46:06.000000000 -0500
++++ test/CMakeLists.txt	2021-11-10 03:46:29.000000000 -0500
+@@ -197,7 +197,7 @@
+ 		message("GLM: Clang - ${CMAKE_CXX_COMPILER_ID} compiler")
+ 	endif()
+ 
+-	add_compile_options(-Werror -Weverything)
++	add_compile_options(-Weverything)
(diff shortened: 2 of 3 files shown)


pmetzger:
CI failed. @Tatsh, what would you like to do?

Tatsh:
Issue is related to a dep:

>Installing dependency (243 of 293) 'cargo-c' with variants '' (requesting '') ... [FAIL]

Tatsh:
Which is related to OpenSSL 3. I guess it needs to rerun?

pmetzger:
I'll try the CI again. I assume you've checked the build yourself?

Tatsh:
> I'll try the CI again. I assume you've checked the build yourself?

I can't because the dependency `glm` is broken on Monterey.

pmetzger:
Should we fix the dependency? This would seem like an important port. We're also not getting good feedback from the CI so being able to test locally would be good.

Tatsh:
I'm willing to wait till glm is fixed. There has been no activity on that port and the issue has a ticket.

pmetzger:
@Tatsh Sometimes the best way to get something to happen is to do it yourself...

Tatsh:
Fixed glm with this PR, and not going to use system CMIS as it won't get past configure phase.

Tatsh:
@pmetzger if you don't want to wait for CI, this built perfectly on my Monterey machine.

pmetzger:
Merged. That said, I think the glm fix is a bit brute force. Which is to say, if there are warnings, they probably should get fixed.

Tatsh:
> Merged. That said, I think the glm fix is a bit brute force. Which is to say, if there are warnings, they probably should get fixed.

If you want to do it differently let me know. The fix in code would be to explicitly cast, and the other possible fix is to add another exception to the flags: `-Wno-implicit-int-conversion`. You can see on lines 9 and 10 of the file upstream already sets a lot of exceptions after `-Werror`.

pmetzger:
Is upstream maintained? Perhaps you should report the issue to them?

Tatsh:
Not really sure what happened but they disabled issues on GitHub. I can make a PR to get their attention. https://github.com/g-truc/glm

PR #12873, opened 2021-11-09
Revert "py-parsing: upgrade to 3.0.5"

catap:
#### Description

3.0.5 has a regression https://github.com/pyparsing/pyparsing/issues/329

Also fixes: https://trac.macports.org/ticket/63872

This reverts commit a83905f74d75383236de152c43264983fce58e02

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/python/py-parsing/Portfile
+++ b/python/py-parsing/Portfile
@@ -5,9 +5,11 @@ PortGroup           python 1.0
 
 name                py-parsing
 python.rootname     pyparsing
-version             3.0.5
+# 3.0.5 is broken
+# See: https://github.com/pyparsing/pyparsing/issues/329
+version             3.0.4
 revision            0
-epoch               1
+epoch               2
 
 categories-append   lang
 platforms           darwin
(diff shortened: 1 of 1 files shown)


python/py-parsing/Portfile
```diff
@@ -5,7 +5,7 @@ PortGroup           python 1.0
 
 name                py-parsing
 python.rootname     pyparsing
-version             3.0.5
+version             3.0.4
 revision            0
 epoch               1
```
chrstphrchvz, line 10: Decreasing the version requires incrementing the epoch.

I suggest adding a comment near here to inform others not to attempt updating the port to 3.0.5.

chrstphrchvz (changes requested):

python/py-parsing/Portfile
catap: @chrstphrchvz good point, thanks!

chrstphrchvz: > Decreasing the version requires incrementing the epoch.

The only possible exception I can think of for this is when the port always fails to install, which would not apply here.

nerdling: 3.0.6 will likely be out before anyone re-attempts 3.0.5 in macports.

PR #12886, opened 2021-11-10
mysql8: remove patch for 10.10 and older macOS

laggardkernel:
#### Description

mysql8 only support macOS 10.13 and newer now.
`patch-mysql8-workaround-no-SC_PHYS_PAGES.diff` is redundant
after the commit 1cdb3d25.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix
- [x] cleanup

###### Tested on

Sorry, I didn't test it. Current mysql8 port can't be built on macOS 10.14, which is the OS I'm running.

###### Verification 
Have you


chrstphrchvz (commented):
The legacysupport portgroup (only effective up to 10.12 Sierra) can also be dropped.

laggardkernel:
> The legacysupport portgroup (only effective up to 10.12 Sierra) can also be dropped.

Thanks. I deleted the legacysupport and force pushed the commit.

chrstphrchvz:
> ###### Tested on
> Sorry, I didn't test it. Current mysql8 port can't be built on macOS 10.14, which is the OS I'm running.

Since this is for dropping a patch for older OSes, I don’t think there is much to test, other than at a minimum making sure the port parses and that the patch phase still works (e.g. in case of dependencies between patches), which it does. The CI failures were due to `openssl.branch 1.1` not yet being specified.

I think this is ready to merge.

PR #12918, opened 2021-11-12
mysql56, mysql57, mysql8: use MacPorts’ libevent

chrstphrchvz:
…instead of building bundled libevent

#### Description
See https://github.com/macports/macports-ports/pull/12915#issuecomment-966753694: the bundled libevent may be disabling a feature during configure due to Xcode ≥ 12 implicit function declaration errors, which MacPorts’ libevent seems unaffected by.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

Untested.

###### Verification 
Have you

### DIFF
--- a/databases/mysql56/Portfile
+++ b/databases/mysql56/Portfile
@@ -12,7 +12,7 @@ set name_mysql      ${name}
 version             5.6.51
 # Set revision_client and revision_server to 0 on
 # version bump.
-set revision_client 2
+set revision_client 3
 set revision_server 1
 set version_branch  [join [lrange [split ${version} .] 0 1] .]
 categories          databases
@@ -65,6 +65,7 @@ if {$subport eq $name} {
                         size    32411131
 
     depends_lib-append  port:ncurses \
+                        port:libevent \
                         port:tcp_wrappers \
                         port:zlib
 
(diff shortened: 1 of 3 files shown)


pmetzger:
Not sure how to review this, but it looks okay to me? @chrstphrchvz is there anyone in particular you feel should look at this before we merge it?

herbygillot:
As mentioned in #12915, can we verify that the bundled `libevent` isn't including custom changes that wouldn't be found in the standard `libevent` installed by MacPorts?

chrstphrchvz:
Looking in https://github.com/mysql/mysql-server/commits/mysql-8.0.27/extra/libevent/libevent-2.1.11-stable, I haven’t noticed any substantial changes made to the bundled libevent: they mostly seem to be for disabling certain features or compiler options that are a problem on other platforms or with the rest of the MySQL compilation process, and for removing things not needed in the MySQL source tree (example code, documentation, autotools build); MySQL hasn’t added new features to libevent, for example—so as long as MacPorts’ libevent builds fine, I would think it is usable for MySQL.

I primarily think of this as another case where MacPorts would probably prefer maintaining its own libevent rather than potentially also maintaining the bundled libevent in MySQL.

herbygillot:
I think it's important to exercise caution before replacing a project's bundled version of a particular dependency with the standard version provided by MacPorts.

A good example is Rust and LLVM - the LLVM bundled in Rust has a number of tweaks made by the project that are specific to Rust.

Of course it's good in general to use the MacPorts version where we can, but I think we should always ensure we have the full context behind why something might be bundled before doing so.

Thanks for looking into this.

PR #12929, opened 2021-11-12
libreoffice: build on arm64

catap:
#### Description

Here a fix of two ports which is required to build an Libreoffice on M1.

~Anyway, libreoffice supports macOS 12 only in 7.3.0.0 https://github.com/LibreOffice/core/commit/71b227dfa46fbc9fbf9b9f2046705939ab0a02bf which hasn't been releaed yet.~

~This is a sort fixing in advance I guess.~

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/clucene/Portfile
+++ b/devel/clucene/Portfile
@@ -1,13 +1,13 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem 1.0
-PortGroup               cmake 1.0
+PortGroup               cmake 1.1
 PortGroup               conflicts_build 1.0
 PortGroup               muniversal 1.0
 
 name                    clucene
 version                 2.3.3.4
-revision                2
+revision                3
 set branch              [join [lrange [split ${version} .] 0 1] .]
 categories              devel
 platforms               darwin
@@ -22,7 +22,8 @@ homepage                http://clucene.sourceforge.net/
 master_sites            sourceforge:project/clucene/clucene-core-unstable/${branch}
 
 checksums               rmd160  5acfc9c8acd167b3684cfc731a60fd9c5465cc9b \
-                        sha256  ddfdc433dd8ad31b5c5819cc4404a8d2127472a3b720d3e744e8c51d79732eab
+                        sha256  ddfdc433dd8ad31b5c5819cc4404a8d2127472a3b720d3e744e8c51d79732eab \
+                        size 2241498
 
 depends_lib             port:zlib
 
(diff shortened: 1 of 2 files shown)


herbygillot:
> This is a sort fixing in advance I guess.

Does that mean these changes will need to be removed when that version comes out?

catap:
@herbygillot nope it fixes dependency of libreoffice. But it is still impossible to build libreoffice because it haven't been released yet.

catap:
@herbygillot anyway, it should allows to build libreoffice for BigSur.

catap:
@herbygillot I was wrong. I can build libreoffice on M1. The root cause of issue was that I've used old ports tree — before https://github.com/macports/macports-ports/pull/12851 was merged.

Shall I bump a revision to trigger rebuild?

herbygillot:
> @herbygillot I was wrong. I can build libreoffice on M1. The root cause of issue was that I've used old ports tree — before #12851 was merged.
> 
> Shall I bump a revision to trigger rebuild?

Yes.  If `libreoffice` depends on both of these ports and you are updating them, then `libreoffice` will have to be revision-bumped in order to rebuild against these updated ports. 

I also notice that `box2d` is missing a revision bump as well.  Since its contents are changing, we will want to do that.

catap:
@herbygillot here it is.

herbygillot:
> @herbygillot here it is.

Thanks @catap. Let's leave it open for the maintainers of these ports to review.

devel/clucene/Portfile
```diff
@@ -1,13 +1,13 @@
...
 PortSystem 1.0
-PortGroup               cmake 1.0
+PortGroup               cmake 1.1
 PortGroup               conflicts_build 1.0
 PortGroup               muniversal 1.0
 
 name                    clucene
 version                 2.3.3.4
-revision                2
+revision                0
```
harens, line 10: Since you're not updating the version number, I would've thought that the revision number should be incremented to 3 rather than set to 0.

office/libreoffice/Portfile
```diff
@@ -5,7 +5,7 @@ PortGroup           boost 1.0
 
 name                libreoffice
 version             7.2.3.1
-revision            0
+revision            1
```
harens, line 8: I don't think this qualifies as a revision bump, since we're not changing the contents of the port itself. The [docs](https://guide.macports.org/#reference.keywords/) also suggest not to rev bump to fix build failures.

The only reason I know this is since I've been corrected on this myself :) https://github.com/macports/macports-ports/commit/65512ee30539ec36aaae353035929eda248f67d1

harens (commented):
Thanks for your work on this @catap! ⭐ Minor comments from me below.

Also CC @Tatsh as the libreoffice port maintainer.

office/libreoffice/Portfile
Tatsh: Yes let's undo this revbump. No need.

devel/clucene/Portfile
catap: @harens thanks, this was a typo based on a lot of rebase.

office/libreoffice/Portfile
catap: @harens I wasn't sure should I update libreoffice revision or not. Without it, it won't trigger rebuild of libreoffice => user still needs to rebuild it by themself (and it requires huge time).

if it is ok for @Tatsh I'd love to keep revision bump here, to force buildbot's to rebuild it for users.

harens: From https://github.com/macports/macports-ports/commit/65512ee30539ec36aaae353035929eda248f67d1:

> exactly, for fixing build failure one shouldn't increase the revision. Ideally, the build should just be started on the buildbot, but for that you need an account. I think that eventually it will retry failed builds automatically, so it will get done at some point.

A revision bump would force everyone to rebuild this, including machines where it was working fine beforehand. My understanding is that some authorised users i.e. not me ;) can get the arm buildbots to rebuild the binary without a revision bump.

Is that ok @catap? Thanks again for your work on this, greatly appreciated.

catap: @harens let make things faster and simpler when

catap:
@herbygillot I guess everything is clear now ;)

harens (approved):
Thanks @catap for the changes. LGTM.

catap:
@herbygillot can I ask you to merge it?

PR #12940, opened 2021-11-13
msieve: add devel subport

catap:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6 20G165 x86_64
Xcode 13.1 13A1030d

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/math/msieve/Portfile
+++ b/math/msieve/Portfile
@@ -12,14 +12,16 @@ long_description    ${description}
 
 platforms           darwin
 
+conflicts           ${name}-devel
+
 homepage            https://sourceforge.net/projects/msieve/
 master_sites        sourceforge:project/msieve/msieve/Msieve%20v${version}
 distname            ${name}[strsed ${version} {g/\.//}]
-worksrcdir          ${name}-${version}
 
 distfiles           ${distname}_src${extract.suffix}
 checksums           rmd160  81170f6ee363140c44bd07b9937cab0ab8c75f1a \
-                    sha256  c5fcbaaff266a43aa8bca55239d5b087d3e3f138d1a95d75b776c04ce4d93bb4
+                    sha256  c5fcbaaff266a43aa8bca55239d5b087d3e3f138d1a95d75b776c04ce4d93bb4 \
+                    size 976135
 
 depends_lib         port:gmp
 
@@ -27,9 +29,25 @@ patchfiles          patch-Makefile.diff \
                     patch-time.diff
 
 post-patch {
+    reinplace "s|@VER@|${version}|g" ${worksrcpath}/Makefile
     reinplace "s|@CC@|${configure.cc}|g" ${worksrcpath}/Makefile
     reinplace "s|@PREFIX@|${prefix}|g" ${worksrcpath}/Makefile
     reinplace "s|-march=native||g" ${worksrcpath}/Makefile
+}
+
+subport ${name}-devel {
+    conflicts       ${name}
+
+    fetch.type      svn
+    svn.url         https://svn.code.sf.net/p/msieve/code/trunk/
+    version         1043
+    revision        0
+
+    patchfiles-delete \
+                    patch-Makefile.diff
+
+    patchfiles-append \
+                    patch-Makefile-trunk.diff
 }
 
 use_configure       no
--- a/math/msieve/files/patch-Makefile-trunk.diff
+++ b/math/msieve/files/patch-Makefile-trunk.diff
@@ -1,7 +1,7 @@
---- Makefile.orig	2016-11-12 00:40:27.000000000 +0900
-+++ Makefile	2016-11-14 22:04:27.000000000 +0900
-@@ -12,13 +12,15 @@
- #  $Id: Makefile 989 2016-01-04 15:03:27Z jasonp_sf $
+--- Makefile.orig	2021-11-13 13:56:43.000000000 +0100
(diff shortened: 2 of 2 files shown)


catap:
@ryandesign if you have any comments about this port, I'd like to fix it.

pmetzger:
LGTM but presumably @NicosPavlov should weigh in.

NicosPavlov:
As far as I can see, it is not possible to fetch the code with the current changes (fails with 404 not found). Is there a reason to rely on a snapshot generated by the host instead of directly fetching the code from the repository with `fetch.type svn` ?

I also take this opportunity to ask about the previous PR: what was the reason to remove the `-march=native` flag? In general, it seems to me that this flag is preferred to allow for finer optimization.

catap:
@NicosPavlov let me start from the second point. I've removed `-march=native` because it prevents to build on macOS 12 on M1. It fails on error:
```
clang: error: the clang compiler does not support '-march=native'
```

I don't have an idea why, probably M1 toolchain is more like iOS toolchain which doesn't support `-march=native` because it is impossible to build anything on iPhone :)

Regarding `fetch.type svn`: the root cause of issue that I can't find the right URL to use SourceForge's SVN :)

NicosPavlov:
> @NicosPavlov let me start from the second point. I've removed `-march=native` because it prevents to build on macOS 12 on M1. It fails on error:
> 
> ```
> clang: error: the clang compiler does not support '-march=native'
> ```
> 
> I don't have an idea why, probably M1 toolchain is more like iOS toolchain which doesn't support `-march=native` because it is impossible to build anything on iPhone :)

Ok. I am not sure what the issue is and I cannot test on macOS 12, but clang has supported this flag for quite some time to my knowledge. I am just concerned that this may affect performance, especially for a port for which optimization can be important. This is not for this PR anyway, but it could be preferable to offer compiler variants as other ports do.

> Regarding `fetch.type svn`: the root cause of issue that I can't find the right URL to use SourceForge's SVN :) Anyway, snapshot works well for me and for github/azure CI which sucessfuly build new port.

I get that, but I could not get it to fetch, and it is still an approach that relies on the sourceforge server to have the snapshot. 
The svn URL is https://svn.code.sf.net/p/msieve/code/trunk/, this URL format is used reliably by other ports to my knowledge.

catap:
@NicosPavlov regarding `clang`. I do have access to M1 => I can easy share something and test it. For example list of supported architectures by clang:
```
catap@Kirills-mini-m1 ~ % clang --print-supported-cpus
Apple clang version 13.0.0 (clang-1300.0.29.3)
Target: arm64-apple-darwin21.1.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
Available CPUs for this target:

	a64fx
	apple-a10
	apple-a11
	apple-a12
	apple-a13
	apple-a14
	apple-a7
	apple-a8
	apple-a9
	apple-latest
	apple-m1
	apple-s4
	apple-s5
	carmel
	cortex-a34
	cortex-a35
	cortex-a53
	cortex-a55
	cortex-a57
	cortex-a65
	cortex-a65ae
	cortex-a72
	cortex-a73
	cortex-a75
	cortex-a76
	cortex-a76ae
	cortex-a77
	cortex-a78
	cortex-a78c
	cortex-r82
	cortex-x1
	cyclone
	exynos-m3
	exynos-m4
	exynos-m5
	falkor
	generic
	kryo
	lightning
	neoverse-e1
	neoverse-n1
	neoverse-n2
	neoverse-v1
	saphira
	thunderx
	thunderx2t99
	thunderx3t110
	thunderxt81
	thunderxt83
	thunderxt88
	tsv110
	vortex

Use -mcpu or -mtune to specify the target's processor.
For example, clang --target=aarch64-unknown-linux-gui -mcpu=cortex-a35
catap@Kirills-mini-m1 ~ % 
```

Maybe put `apple-m1` is good idea, But it is defently separated PR and I personally have no idea regarding performance impact for this optimization.

Thus, thanks for link, I'll fix it right away.

NicosPavlov:
Thanks, I'll have a look at the arch flag thing and try to make a PR, it would surely be helpful if you have the ability to test. 

This PR looks good to me. One eventual last detail just to make port lint happy would be to set version to 1043, and let svn.revision be set automatically.

catap:
@NicosPavlov here it is.

And feel free to contact me directly or via PR when you need help to test.

ryandesign:
> Is there a reason to rely on a snapshot generated by the host instead of directly fetching the code from the repository with `fetch.type svn` ?

Fetching from a VCS should be a last resort. If one needs to fetch the latest code and snapshot tarballs are available and reliable (i.e. do not disappear or change over time) they should be used. I do not recall whether SourceForge's snapshot tarballs are reliable.

> what was the reason to remove the `-march=native` flag? In general, it seems to me that this flag is preferred to allow for finer optimization.

`-march=native` is not available on Apple Silicon.

No port should use `-march=native` by default because it would mean that a port built on one Intel Mac may not work on another Intel Mac if it uses a different processor generation. Any use of `-march=native` should be confined to a non-default "perf" variant to ensure that it builds on the user's processor (rather than on our build farm's processors which may not be the same model as the user's).

PR #12975, opened 2021-11-15
py-tomli: fix file timestamp issue on Tiger

evanmiller:
#### Description

Closes: https://trac.macports.org/ticket/63926

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/python/py-tomli/Portfile
+++ b/python/py-tomli/Portfile
@@ -21,6 +21,12 @@ checksums           rmd160 e37cff10eb7e7f050049585b3d7224dd907b463c \
 
 python.versions     36 37 38 39 310
 
+# https://trac.macports.org/ticket/63926
+platform darwin 8 {
+    extract.post_args "| ${prefix}/bin/gnutar --no-same-owner -xf -"
+    depends_extract port:gnutar
+}
+
 if {${name} ne ${subport}} {
     python.pep517 yes
     python.add_dependencies no


mascguy (approved):

mascguy:
LGTM, but we should let Toby review and merge.

tobypeterson:
I thought the error had to do with timestamp, not ownership...

evanmiller:
The `--no-same-owner` stuff just mimics the current behavior. The important bit is just using an updated `gnutar` compared to what ships with 10.4.

tobypeterson:
I see. Is it possible to fix by just doing `touch ${worksrcpath}/PKG-INFO` in post-extract? That seems like it'd be a nice solution that doesn't require any logic specific to an OS version.

evanmiller:
I will try it out... Right now the timestamps on most of the extracted files is coming out as 1/1/1970.

tobypeterson:
Huh, I see. For me, it's only PKG-INFO with a timestamp of 0:
`-rw-r--r--  0 0      0        1072 Oct 25 02:50 tomli-1.2.2/LICENSE
-rw-r--r--  0 0      0        7994 Oct 25 02:50 tomli-1.2.2/README.md
-rw-r--r--  0 0      0        5106 Oct 25 02:50 tomli-1.2.2/pyproject.toml
-rw-r--r--  0 0      0         299 Oct 25 02:50 tomli-1.2.2/tomli/__init__.py
-rw-r--r--  0 0      0       21659 Oct 25 02:50 tomli-1.2.2/tomli/_parser.py
-rw-r--r--  0 0      0        2817 Oct 25 02:50 tomli-1.2.2/tomli/_re.py
-rw-r--r--  0 0      0         126 Oct 25 02:50 tomli-1.2.2/tomli/_types.py
-rw-r--r--  0 0      0          26 Oct 25 02:50 tomli-1.2.2/tomli/py.typed
-rw-r--r--  0 0      0        9089 Dec 31  1969 tomli-1.2.2/PKG-INFO
`

mascguy:
Toby/Evan, where do we go from here?

evanmiller:
I think the PR as it stands is the simplest and most sustainable solution.

tobypeterson:
Did you confirm that doing `touch` in post-extract _doesn't_ work?

evanmiller:
@tobypeterson Touching PKG-INFO post-extract does not solve the problem. Every file in the archive has a timestamp of zero. I suppose I could do a recursive touch but this seems convoluted compared to requiring a newer gnutar on affected platforms.

jmroot:
Rather than adding a dependency to every port that cares about source file timestamps, it seems like it would be a better use of your time to fix this once in base.

evanmiller:
Looks like the build error was fixed in https://github.com/macports/macports-ports/commit/11b2758fc9b36a88b0ad3177cf6c78ecb56604e2, closing

tobypeterson:
Huh, that's one way to do it. I suppose if this issue is widespread, it's better than patching many ports like this.

jmroot:
"It's never just the one thing." Tar files generated by python's `tarfile` module set the mtime correctly in the pax header but zero out the mtime field in the ustar header. So unarchivers that don't support the pax header, like the old gnutar versions, extract everything with a 0 mtime. And that only caused an error because the files are being put into a wheel, which uses zip, which uses a different epoch.

PR #12976, opened 2021-11-15
godot: Add old AppKit compatibility wrapper

jasonliu--:
#### Description

This PR adds the compatibility wrapper for old versions of AppKit to the Godot port. It is identical to the file that I have created for my Blender and MaterialX ports. Adding the file is in preparation for the fixes needed in order to resolve https://trac.macports.org/ticket/63033. Since I've also modified a comment in the equivalent file in the Blender and MaterialX ports, I've decided to make this a separate PR from the other Godot fixes.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you


evanmiller:
I'm a little bit confused by this PR – it looks like you are defining the old name to the new name, but only for the old deployment target? Generally if it's a newer SDK deploying to an older target, nothing needs to support compile-time integer constants. However, if the purpose is to support an older SDK, then I would expect the defines to happen in the opposite direction (define the new name, used in the Godot source code, to be the old value understood by the old SDK).

Compile-time SDK targeting by the way will want to use `MAC_OS_X_VERSION_MAX`.

jasonliu--:
> I'm a little bit confused by this PR – it looks like you are defining the old name to the new name, but only for the old deployment target? Generally if it's a newer SDK deploying to an older target, nothing needs to support compile-time integer constants. However, if the purpose is to support an older SDK, then I would expect the defines to happen in the opposite direction (define the new name, used in the Godot source code, to be the old value understood by the old SDK).
> 
> Compile-time SDK targeting by the way will want to use `MAC_OS_X_VERSION_MAX`.

@evanmiller , please read the comments in the file. There's a very specific reason for defining the old name to the new name, and why it's needed.

mascguy:
> @evanmiller , please read the comments in the file. There's a very specific reason for defining the old name to the new name, and why it's needed.

Evan, haven't taken too close a look at this yet, but do the comments address your concern?

evanmiller:
I understand the logic better – sorry for not reading through the comment. However, I believe the condition should be testing `MAC_OS_X_VERSION_MAX` (the SDK version) rather than `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` (the deployment target).

jasonliu--:
> I understand the logic better – sorry for not reading through the comment. However, I believe the condition should be testing `MAC_OS_X_VERSION_MAX` (the SDK version) rather than `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` (the deployment target).

First off, allow me to be pedantic and say that it should be `MAC_OS_X_VERSION_MAX_ALLOWED`, and not `MAC_OS_X_VERSION_MAX`.

When I first started working on the AppKit wrapper header file over a year ago, I too believed that something like `MAC_OS_X_VERSION_MAX_ALLOWED` should be used. However, if you look at the [MacPorts legacy support](https://github.com/macports/macports-legacy-support) package, more specifically, [MacPortsLegacySupport.h](https://github.com/macports/macports-legacy-support/blob/master/include/MacportsLegacySupport.h), you will see that all of the macros are `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__`. After fairly lengthy discussions with @cjones051073, @kencu, and other long-time MacPorts veterans, I learned that they use `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` because it is the only availability macro that has proven, over time, to behave consistently across multiple versions of macOS, even from the early versions of Mac OS X.

Until I experienced strange behavior myself, I had a hard time believing that something which seems like it's supposed to be an internally-used macro (due to the double-underscores), should be used instead of something like `MAC_OS_X_VERSION_MAX_ALLOWED`. However, as discussed in PR #12269, and subsequently patched in PR #12893, there have been documented cases where `MAC_OS_X_VERSION_MAX_ALLOWED` is actually being incorrectly interpreted by the preprocessor. After having seen first-hand how `MAC_OS_X_VERSION_MAX_ALLOWED` can behave in unexpected and incorrect ways, I now firmly trust the others' judgment in deciding to use `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` instead. And remember that this wrapper file is destined to eventually get added to the MacPorts legacy support package, when I get around to it. When that time comes, it will likely also get an entry in `MacPortsLegacySupport.h`, and I will also use `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` as well.

Thank you suggesting using `MAC_OS_X_VERSION_MAX_ALLOWED`, but unless @cjones051073, @kencu, or @michaelld speak up and tell me otherwise, I have no plans to change my usage of `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` in the file.

kencu:
I use ```MAC_OS_X_VERSION_MAX_ALLOWED``` if I know that ```Availability.h``` or ```AvailabilityMacros.h``` is for sure included and if I want to test things that are related to the SDK I am building against (certain defines being defined, etc).

I use ```MAC_OS_X_VERSION_MIN_REQUIRED``` if I know that ```Availability.h``` or ```AvailabilityMacros.h``` is for sure included and if I want to test things that are related to the target I am deploying against (certain functions being available, etc).

I use ```__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__``` if I am not sure if  ```Availability.h``` or ```AvailabilityMacros.h``` are for sure included, or if I want to test against the deployment target and don't want to bother  adding in the header (eg one or two of the tests in libuv, or the ```LegacySupport.h``` header where we didn't want to risk adding in  ```Availability.h``` or ```AvailabilityMacros.h``` and changing something unexpected.

The authority on this would appear to be Chris Espinosa, Apple employee #8,  certainly a multimillionaire by now therefore I hope, who outlined it about 20 years ago here:

https://lists.apple.com/archives/xcode-users/2005/Aug/msg00399.html

"
Where am I supposed to define MAC_OS_X_VERSION_MIN_REQUIRED and
MAC_OS_X_VERSION_MAX_ALLOWED? In 'other C flags' and 'other C++ flags'
or in my .pch? Or either? I thought there was an actual setting in the
Xcode UI, but I can't find it.

Both are set implicitly by other settings; you probably never need to set them directly. Both are set by AvailabilityMacros.h according to certain criteria.

MAC_OS_X_VERSION_MAX_ALLOWED is set according to the version of SDK or system headers you are using. (Note that the name is somewhat of a misnomer: your object code will run on Mac OS X versions later than the version specified, of course.) It controls what APIs you can compile against: if it's 1030, for example, APIs that are only available in Tiger are not present in the headers you're building against. So you can use this in your code to #if out a passage of code using newer APIs.

MAC_OS_X_VERSION_MIN_REQUIRED is set according to you Mac OS X Deployment Target build setting. It sets the earliest version you expect to compile against. Symbols that are declared in this version and earlier will link normally; symbols defined in the SDK or system headers (up to MAC_OS_X_VERSION_MAX_ALLOWED) but later than this version will be imported weak, and you must check at runtime.

So in truth these are read-only macros. The actual settings are the SDK choice (SDKROOT) and the Mac OS X Deployment Target (MACOSX_DEPLOYMENT_TARGET).
"

jasonliu--:
Ken's comment reminds me of another good point: `Availability.h` and `AvailabilityMacros.h` are for sure _NOT_ included anywhere in Godot's source code.

michaelld:
If whatever the issue here being solved is generic enough across different ports and relates to changes in macOS / OSX API/ABI, then is this not better put into LegacySupport & then used in that form? That way we have 1 version of it to maintain & any fixes to it will be universally applied to ports that require using it.

jasonliu--:
Paraphrasing @kencu's comment in the [Track ticket](https://trac.macports.org/ticket/63033#comment:14), when I first came up with the mapping file, the concern with immediately adding it to the MacPorts legacy support package was that every piece of software that used LegacySupport was at risk of being very affected by my redefines that, at the time, was very hard to predict. So instead, I used my Blender and MaterialX ports as the testbeds, since the mapping file was originally meant to resolve problems specifically in those two ports.

Assuming I can successfully incorporate the mapping file into the Godot port, that would make three relatively complex software projects that are using the file, and having used it for a year across multiple versions of the ports, it may be ready to be declared "ready for prime time".

michaelld:
Fair enough. That said, since the project must `#include <AppKit/AppKit.h>` while using LegacySupport, that seems to me to be a pretty small number that will have conflicts -- if any! I actually know of no ports that use that header, though I'm sure some do & I just don't know it.

jasonliu--:
> Fair enough. That said, since the project must `#include <AppKit/AppKit.h>` while using LegacySupport, that seems to me to be a pretty small number that will have conflicts -- if any! I actually know of no ports that use that header, though I'm sure some do & I just don't know it.

It's exceedingly rare for a project to `#include <AppKit/AppKit.h>` directly. Instead, it's much more common to `#include <Cocoa/Cocoa.h>`, which itself is a very short file that simply contains the following three lines:

```
#import <Foundation/Foundation.h>
#import <AppKit/AppKit.h>
#import <CoreData/CoreData.h>
```

ra1nb0w:
Can we merge as is and re-valuate LegacySupport in the future?

mascguy (approved):

mascguy:
LGTM. Anyone else?

ra1nb0w:
ok. merge.

PR #12977, opened 2021-11-15
python3{9,10}: Use openssl PG instead of shim port

cjones051073:
Avoids dependency on openssl shim port by using openssl PG to directly configure builds.

Avoiding having openssl include/libs in primary prefix is useful in preventing accidental usage, e.g. when a port should be using openssl 1.1 from the isolated libexec area, but build is poorly configured and finds the shim port in the primary prefix instead.

### DIFF
--- a/lang/python310/Portfile
+++ b/lang/python310/Portfile
@@ -2,12 +2,13 @@
 
 PortSystem 1.0
 PortGroup select 1.0
+PortGroup openssl 1.0
 
 name                python310
 
 # Remember to keep py310-tkinter and py310-gdbm's versions sync'd with this
 version             3.10.0
-revision            3
+revision            4
 
 set branch          [join [lrange [split ${version} .] 0 1] .]
 categories          lang
@@ -48,7 +49,6 @@ depends_lib         port:bzip2 \
                     port:libedit \
                     port:libffi \
                     port:ncurses \
-                    path:lib/libssl.dylib:openssl \
                     port:sqlite3 \
                     port:xz \
                     port:zlib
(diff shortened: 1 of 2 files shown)


jmroot:
I don't see any reason to do this. Having to use this portgroup should be the exception, not the rule.

cjones051073:
Hi jmroot,

I understand this view point, and I am not suggesting we make it a rule. Its just in some scenarios it will help avoid issues.

the problem is more ports than I  thought (was lead to believe) seem to need to be pinned back to openssl11, and for a fraction of these ports they fail to install if the shim port is active as they opportunistically find and use the shim port links, rather than the specific openssl11 install area they have been configured to use. Yes, by the book this maybe should probably be regarded as a bug in their build configurations that should be resolved, but this is fair from trivial in a lot of cases. For now the only option is to add a builds conflict statement to these ports that at least gives the user some instructions how to deal with it, but is not perfect as eg prevents binary tarballs being produced in the buildbots, so prevents these ports, and any port depending on them, from being distributed in binary tarball form.

so yes, switching these python builds, to use the PG is not needed to build the ports themselves, but will really help with a handful of ports that currently fail to build in the buildbots because they have e.g. python39 in the dependency tree, which currently drags in openssl, and have zero impact on the functionally of the ports themselves.

just as an example

https://build.macports.org/builders/ports-10.14_x86_64-builder/builds/105909

jmroot:
That's unfortunate, but this change would just hide the problem in a few specific cases. I don't think it's a good idea.

PR #12981, opened 2021-11-15
Remove cxx11 portgroup

jmroot:
This has been deprecated for a long time.

- [x] enhancement


RJVB:
On Monday November 15 2021 08:27:46 MacPorts Bot wrote:
>Notifying maintainers:

***@***.*** for port audacity, dbusmenu-qt, Vc.

You seem to have added the required C++ standard specifications so these ports should be fine.

HOWEVER: please don't change anything else to port:audacity! I'm pretty certain I've had to unset `sdk_version` to avoid build failures.

jmroot:
> HOWEVER: please don't change anything else to port:audacity! I'm pretty certain I've had to unset `sdk_version` to avoid build failures.

OK, but configure.sdk_version isn't currently unset in audacity, and clearing configure.sdkroot prevents the configure.sdk_version value from doing anything except print a warning because the sdkroot doesn't match the requested SDK version.

peternewman (approved):
This looks like it should be fine for OLA assuming it builds fine as part of the CI.

mojca:
As far as my ports are concerned the changes seem fine (I didn't test anything yet), just some additional comments:
* `gate` is probably broken and may need some love & care to bring it up to date. Just to let you know that there's no need to worry if it doesn't currently build.
* I would also remove commented-out references from `_resources/port1.0/group/wxWidgets-1.0.tcl`. (I should probably remove the extra `-cxx11` subport anyway, it hardly makes any sense after we switched to `libc++` on older systems.)

RJVB:
> (I should probably remove the extra `-cxx11` subport anyway, it hardly makes any sense after we switched to `libc++` on older systems.)

On that subject: there's an option to build wxWidgets with c++14 support or just using that syntax. I have no idea about the pros and cons, and can't say I've noticed issues using a non-C++14 build in software that does require that standard. Food for thought that's off-topic here anyway...

PR #12985, opened 2021-11-15
haproxy: Currently requires OpenSSL 1.1

judaew:
#### Description

`haproxy` currently does not support OpenSSL 3. Therefore, now this port is using OpenSSL 1.1.

See https://trac.macports.org/ticket/6395

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

@aphor

### DIFF
--- a/net/haproxy/Portfile
+++ b/net/haproxy/Portfile
@@ -1,12 +1,17 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
+# `haproxy` currently does not support OpenSSL 3. Therefore, now this port is
+# using OpenSSL 1.1.
+#
+# See https://trac.macports.org/ticket/6395
 PortSystem          1.0
+PortGroup           openssl 1.0
 PortGroup           legacysupport 1.1
 PortGroup           makefile 1.0
 
 name                haproxy
 version             2.4.8
-revision            1
+revision            2
 
 set branch          [join [lrange [split ${version} .] 0 1] .]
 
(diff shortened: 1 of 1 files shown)


net/haproxy/Portfile
```diff
@@ -52,6 +56,8 @@ build.target        TARGET=osx
 pre-build {
     build.args      LD="${configure.cc}"\
                     LDFLAGS="${configure.ldflags}" \
+                    SSL_INC=${prefix}/include/openssl-1.1 \
+                    SSL_LIB=${prefix}/lib/openssl-1.1 \
```
herbygillot, line 60: The `openssl` portgroup provides some convenience variables for use here.

First off is `[openssl::install_area]`.  This is the base directory for whichever OpenSSL is currently being provided by the portgroup.

Then you have `[openssl::include_dir]` and `[openssl::lib_dir]`, which both build off of `[openssl::install_area]`. You would use both of these for your `SSL_INC` and `SSL_LIB` args, respectively:

```
    build.args      LD="${configure.cc}"\
                    LDFLAGS="${configure.ldflags}" \
                    SSL_INC=[openssl::include_dir] \
                    SSL_LIB=[openssl::lib_dir] \
                    ...
```

net/haproxy/Portfile
```diff
@@ -52,6 +56,8 @@ build.target        TARGET=osx
 pre-build {
     build.args      LD="${configure.cc}"\
                     LDFLAGS="${configure.ldflags}" \
+                    SSL_INC=[openssl::include_dir] \
+                    SSL_LIB=[openssl::install_area] \
```
herbygillot, line 60: Per documentation, `SSL_LIB` should be pointing at the `.../openssl-.../lib` directory, so this should be set to `[openssl::lib_dir]`.

judaew: Fix

PR #12996, opened 2021-11-16
nodejs16/17: link against system zlib

Tatsh:
#### Description

Fixes https://trac.macports.org/ticket/63774

My guess is that some code from Node's variant of zlib is getting optimised incorrectly with the new Apple Clang version, but I did not look into it much. Normally we link against system libs anyway, and this fixes the issue we are having.

Also there are more opportunties for system lib linking:

* brotli
* c-ares
* http-parser
* libuv
* nghttp2
* nghttp3 (no port yet)
* ngtcp2 (no port yet)

Only bumping these two versions as these are the versions supported around the time of Monterey's release and only Monterey is affected by the zlib static code issue.

- nodejs16: link against system zlib
- nodejs17: link against system zlib

Revbump is to force a rebuild for those affected (and unfortunately those who aren't, but I have never seen `revision` used conditionally).

I do not know how far back the patch should go if we decide to patch nodejs15 and before that. However, linking against system libs is still seen as an improvement.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/nodejs16/Portfile
+++ b/devel/nodejs16/Portfile
@@ -12,7 +12,7 @@ compiler.cxx_standard   2014
 
 name                    nodejs16
 version                 16.13.0
-revision                1
+revision                2
 
 categories              devel net
 platforms               darwin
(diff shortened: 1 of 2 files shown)


mascguy (approved):

mascguy:
Folks have had eight days to provide feedback, so merging.

PR #13010, opened 2021-11-17
qca-qt5: fix build on macOS 10.7-10.11

chrstphrchvz:
#### Description

Fix builds for 10.7 through 10.11.

Fixes: https://trac.macports.org/ticket/63980
Fixes: https://trac.macports.org/ticket/63983

###### Tested on
Testing on older macOS releases - at a minimum, 10.7 and 10.9 - currently in-progress. The PR description will be updated, once everything has been validated.

###### Verification 
Have you

### DIFF
--- a/devel/qca/Portfile
+++ b/devel/qca/Portfile
@@ -61,7 +61,7 @@ switch ${Qt_Major} {
         }
         patchfiles-append \
                         patch-qca-ossl.diff \
-                        patch-unique_ptr.diff \
+                        patch-missing-headers.diff \
                         qt5/patch-qca210-qt550.diff \
                         qt5/patch-support-older-qt5.diff
         configure.args-append \
--- a/devel/qca/files/patch-missing-headers.diff
+++ b/devel/qca/files/patch-missing-headers.diff
@@ -50,3 +50,16 @@ index 9c81746844f9d8543f7e9eff8603662f495d7a1e..aa8c1dd78eac220278834f0ff340f26e
  
  #include <openssl/err.h>
  #include <openssl/opensslv.h>
+diff --git a/src/qca_default.cpp b/src/qca_default.cpp
+index 294b90565e50c493da36eb6a5ba998fe2cb48645..7c017d7d0a3d49ee85dd58bd2eaefc387ba5d784 100644
+--- a/src/qca_default.cpp
++++ b/src/qca_default.cpp
+@@ -30,6 +30,8 @@
+ #include "qca_systemstore.h"
+ #endif
+ 
++#include <cstdlib>
++
+ #define FRIENDLY_NAMES
+ 
+ namespace QCA {
--- a/devel/qca/files/qt5/patch-support-older-qt5.diff
+++ b/devel/qca/files/qt5/patch-support-older-qt5.diff
@@ -13,10 +13,24 @@ index 6e290245cb9d4c7b6a208dbe245aab3105f4fd0a..ffaeaf43d6d8da08084dabe9a6498274
  
  set(CMAKE_AUTOMOC ON)
 diff --git a/plugins/qca-gnupg/gpgaction.cpp b/plugins/qca-gnupg/gpgaction.cpp
-index c739dd1a7f63c18fe68f256ac91622299460215e..22efab0d2a0171da55173269d4a785d36cc2d501 100644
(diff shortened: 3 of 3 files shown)


RJVB:
LGTM. We might consider rolling this new patchfile and the one for `std::unique_ptr` into a single file, maybe (`patch-missing-headers.diff`)? My guess would be that the number of missing headers will only increase because upstream clearly develops on much more recent systems than the ones we support. And possibly only tests with GCC which might be doing all kinds of includes as a courtesy (and despite what the C++ standard seems to suggest).

chrstphrchvz:
Combining the patches sounds fine, though I do think they should be submitted upstream, as it is improper to not be using features from `<cstdlib>`, `<memory>`, etc. without include the headers.

mascguy (approved):

mascguy:
Rene, are you interested in merging this as-is, or would you prefer to have additional changes made?

RJVB:
You could extend the fix down to OS X 10.7 (right?) and include the updated patchfile I just attached to ticket 63983. Other than that I see nothing else to change.

mascguy:
Folks, I added a 2nd commit, with the latest patch from [63983 - qca-qt5: error: no member named 'toSecsSinceEpoch' in 'QDateTime'](https://trac.macports.org/ticket/63983). The author is set to Rene for the subsequent change, and the commit message will close that ticket.

mascguy:
p.s. I'm also planning to test this on a few older macOS releases - 10.7 and 10.9, at a minimum - before we merge.

mascguy:
Folks, I can confirm that `qca-qt5` builds successfully on both 10.7 and 10.9, with all of these changes. (I also installed `qca` as well, just for completeness.)

While it's not practical to try and test this across every macOS release, are there any other scenarios we should test against?

RJVB:
>While it's not practical to try and test this across every macOS release, are there any other scenarios we should test against?

Is there any interest in getting 10.6 or even lower supported as well? In other words, which OS versions where port:qca-qt5* were in use are now left out? The easiest way to re-support them would be to bump them back down to v2.2.1

mascguy:
> Is there any interest in getting 10.6 or even lower supported as well? In other words, which OS versions where port:qca-qt5* were in use are now left out? The easiest way to re-support them would be to bump them back down to v2.2.1

If you folks are willing to support 10.6 as well, I'll happily help validate it!

I'm wondering if we should do that as a separate PR though, so we can release these fixes? Thoughts?

RJVB:
>I'm wondering if we should do that as a separate PR though, so we can release these fixes? Thoughts?

Yes that would be better because doing this would take time I can't spend on it right now (whether I manage to add the support to v2.3.4 or downgrade the port for the earlier OS versions)

mascguy:
> Yes that would be better because doing this would take time I can't spend on it right now (whether I manage to add the support to v2.3.4 or downgrade the port for the earlier OS versions)

Sounds great, I'll merge this PR then. And let me know when we're ready to add support for 10.6, and I'll help test.

And thanks again for all the work on this folks, this is great for our users!

chrstphrchvz:
Upstream pull request for patch-missing-headers.diff opened: https://invent.kde.org/libraries/qca/-/merge_requests/65

RJVB:
>Upstream pull request for patch-missing-headers.diff opened: https://invent.kde.org/libraries/qca/-/merge_requests/65

Well, you beat me to it! And 't was merged already, without as much as a wisecrack that maybe Xcode 8.2.1 couldn't be used anyway because from a system not supported by the minimum required Qt version. I expected more vigilance from the KDE guys ;)

RJVB:
ALERT ALERT ...

A syntax error slipped into the Qt5 support patch: the first hunk for qca-ossl.cpp introduces a `##else` which somehow got past all of us.

It causes a build failure port:qca-qt5-ossl so it can be fixed by a simple commit, without any revbumping. Can you take care of that please, Christopher?

chrstphrchvz:
> Can you take care of that please, Christopher?

Which Christopher—me or @mascguy (the one with commit rights)? :^)

See #13072

chrstphrchvz:
`qca-qt5-ossl` **did** fail on the CI builds, so maybe we should have caught it earlier.

RJVB:
>Which Christopher—me or @mascguy (the one with commit rights)? :^)

I'll admit I was a bit confused myself but I meant the latter ;)

Little point in asking another "mere mortal" maintainer, right...

PR #13027, opened 2021-11-18
py-pyside2: remove qtwebengine on arm architecture

abey79:
#### Description

The port py-qtwebengine is currently broken on arm architecture (see [ticket 63725](https://trac.macports.org/ticket/63725)). This is being addressed by #12595 but the fix is still plagued with at least one arm-specific issue.

Would it make sense to remove qtwebengine from py-pyside2 _on arm targets_ while py-qtwebengine is being fixed? If so, this is what this PR attempts to do. If this is not appropriate for some reason, I'll happily close this PR.

Note: this is my first contribution attempt and it is likely that I oversaw something.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/python/py-pyside2/Portfile
+++ b/python/py-pyside2/Portfile
@@ -90,7 +90,6 @@ if {${name} ne ${subport}} {
         qtsvg \
         qttools \
         qtwebchannel \
-        qtwebengine \
         qtwebsockets
 
     if {[vercmp ${qt5.version} 5.6] >=0 } {
@@ -101,6 +100,13 @@ if {${name} ne ${subport}} {
     if {[vercmp ${qt5.version} 5.9] >=0 } {
         qt5.depends_component \
             qtremoteobjects
+    }
+
+    # qt5-qtwebengine is currently broken on arm64 and Monterey,
+    # see https://trac.macports.org/ticket/62059
+    if { ${os.arch} ne "arm" && ${os.arch} ne "arm64" && ${os.major} < 21 } {
+        qt5.depends_component \
+            qtwebengine
     }
 
     livecheck.type      none


pmetzger:
I'd expect some downstream users of the python library probably expect to be able to use py-qtwebengine, and this would break them.

abey79:
Well the rationale is that they are broken regardless on ARM architecture for the time being. This is about having at least something for those who don't need qtwebengine. But I will agree that this is rather inelegant.

pmetzger:
https://github.com/macports/macports-ports/pull/12595 proposes to fix the entire problem.

python/py-pyside2/Portfile
```diff
@@ -103,6 +102,12 @@ if {${name} ne ${subport}} {
             qtremoteobjects
     }
 
+    # qt5-qtwebengine is currently broken on arm64, see https://trac.macports.org/ticket/62059
+    if { ${os.arch} ne "arm" && ${os.arch} ne "arm64" } {
+        qt5.depends_component \
+            qtwebengine
+    }
+
```
chrstphrchvz, line 110: `qt5-qtwebengine` is also broken on Monterey as well (x86-64 and arm64: https://trac.macports.org/ticket/63725), so that would seem like another case to disable here (if disabling at all is what should be done). With #12595 there could actually be a greater chance of `qt5-qtwebengine` working soon on Big Sur arm64 than at all on Monterey.

abey79: I will happily add this check if the disabling is deemed desirable.

mascguy:
Antoine, just started the CI jobs for you. And since this is your first contribution, we wish you a warm welcome from the entire MacPorts family!

python/py-pyside2/Portfile
abey79: Check added.

abey79:
I just added the Monterey check and rebased the commit.

mascguy:
Just re-approved the CI jobs again. My apologies for the delay: Apparently this is necessary after every push, since this is your first PR for MacPorts.

Thankfully this won't be necessary for any of your subsequent PRs, though.

abey79:
With #12595 getting very close to completion, this PR becomes irrelevant.

PR #13043, opened 2021-11-19
gcc11: patch for output failure in gfortran (darwin21)

remkos:
#### Description

The issue that triggered this patch is that an output of a simple program like
write ( * , * ) "hello, world!"
end
comes out on the terminal, but when redirecting to a file (./a.out > file.txt) results in a zero-size file.
For the original issue:
https://fortran-lang.discourse.group/t/gfortran-code-is-not-proving-data-in-output-file-macos-monterey/2297/6
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102992
The patch as accepted for gcc12 works fine with gcc11 as well: https://gcc.gnu.org/g:fabe8cc41e9b01913e2016861237d1d99d7567bf

The patch is to circumvent an apparent bug/limitation in the C library for MacOS 12.0 (still there in 12.0.1). Until fixed in either MacOS 12 or gcc11, we can apply the new patch.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.0.1 21A559 x86_64
Xcode Command Line Tools 13.1.0.0.1.1633545042

###### Verification 
Have you

- x ] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 

### DIFF
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -13,7 +13,7 @@ name                gcc11
 
 # Note, ARM builds have their own version below...
 version             11.2.0
-revision            1
+revision            2
 
 platforms           darwin
 categories          lang
@@ -69,6 +69,9 @@ if { ${os.platform} eq "darwin" } {
     # respective version is used by macports.
     # See also: https://github.com/iains/gcc-darwin-arm64/commit/20f61faaed3b335d792e38892d826054d2ac9f15
     patchfiles-append patch-darwin21-support.diff
+    # This following patch works around a problem on MacOS 12.0 with fortran
+    # See also: https://gcc.gnu.org/git/?p=gcc.git;a=patch;h=fabe8cc41e9b01913e2016861237d1d99d7567bf
+    patchfiles-append patch-darwin21-fortran.diff
 }
 
 if { ${os.platform} eq "darwin" && ${os.major} >= 17 } {
--- a/lang/gcc11/files/patch-darwin21-fortran.diff
+++ b/lang/gcc11/files/patch-darwin21-fortran.diff
@@ -1,0 +1,388 @@
+From fabe8cc41e9b01913e2016861237d1d99d7567bf Mon Sep 17 00:00:00 2001
+From: Iain Sandoe <iain@sandoe.co.uk>
+Date: Thu, 4 Nov 2021 09:37:14 +0000
+Subject: [PATCH] IPA: Provide a mechanism to register static DTORs via
+ cxa_atexit.
+
+For at least one target (Darwin) the platform convention is to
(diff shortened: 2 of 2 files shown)


mascguy:
LGTM, at least in principle. But we need the blessing of at least one of our toolchain gurus.

@cjones051073 and @kencu, any thoughts/comments?

mascguy (approved):

kencu:
I'm not really in a position to vet this at present. I think Chris will be up for it.

remkos:
Thanks, in advance, for any reviewers to check this out. Two additional comments:
- The fix is in the runtime library, so any code already compiled with gcc11 will get the fix in runtime
- This might also be backportable to gcc10 or earlier.

pmetzger:
@remkos does the gfortran test suite still pass after this patch is in place?

remkos:
> @remkos does the gfortran test suite still pass after this patch is in place?

Yes, it does!

pmetzger:
Then to the extent that I can review this (the contents of the patch are beyond my knowledge) it seems okay. I agree that we should have a couple of toolchain experts weigh in if possible.

remkos:
I would like to point out again that upstream the same patch has already been approved for GNU CC 12, so I think it's pretty safe to apply it to our gcc11 port, as long as this issue is not fixed either upstream or in MacOS 12.

pmetzger:
That sounds quite reasonable, and I have the expectation that this will go in very soon. I'm not myself an expert in the area, and when it comes to a port that is a commonly used prerequisite for loads of other ports, it's always nice to get another opinion. (If this were an obscure leaf of the dependency tree and not part of the trunk, I'd have merged it already.)

kencu:
> @kencu, any thoughts/comments?

Only tht Iain is the lead gcc developer for Darwin, so as the patch comes directly from Iain, it is highly likely to be reasonable.

remkos:
@pmetzger: yes, of course. I just wanted to add this as information to whomever would additionally review. It works nicely for me, otherwise nearly none of my Fortran programs on Monterey would provide proper output when standard output was redirected. I am just trying to help others before they start (like me) trying to revert to Big Sur because of it. 

The links provided near the top give some good analyses by the GCC experts on this issue and also provided the patch to upstream GCC 12, which is not out yet.

remkos:
> > @kencu, any thoughts/comments?
> 
> Only that Iain is the lead gcc developer for Darwin, so as the patch comes directly from Iain, it is highly likely to be reasonable.
> 
Indeed, it comes from the highest authority 😀

kencu (commented):
well done!

PR #13054, opened 2021-11-20
privoxy: Add variant ecc to use Elliptic Curve Keys

essandess:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.1 20G224 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/www/privoxy/files/patch-ssl_common.h.diff
+++ b/www/privoxy/files/patch-ssl_common.h.diff
@@ -1,0 +1,27 @@
+--- ./ssl_common.h	2021-11-20 08:43:20.000000000 -0500
++++ ./ssl_common.h	2021-11-20 08:48:53.000000000 -0500
+@@ -32,8 +32,24 @@
+ 
+ #include "project.h"
+ 
++/*
++ * Use ECC, i.e. OpenSSL EVP_PKEY_EC; otherwise use EVP_PKEY_RSA
++ */
++#define USE_EVP_PKEY_EC
++
++/*
++ * Use EVP_PKEY_RSA by default
++ */
++#ifndef USE_EVP_PKEY_EC
+ #define RSA_KEY_PUBLIC_EXPONENT          65537             /* Public exponent for RSA private key generating */
+ #define RSA_KEYSIZE                      2048              /* Size of generated RSA keys */
++#else
++/*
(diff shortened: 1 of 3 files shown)


www/privoxy/files/patch-ssl_common.h.diff
```diff
@@ -0,0 +1,27 @@
+--- ./ssl_common.h	2021-11-20 08:43:20.000000000 -0500
++++ ./ssl_common.h	2021-11-20 08:48:53.000000000 -0500
+@@ -32,8 +32,24 @@
+ 
+ #include "project.h"
+ 
++/*
++ * Use ECC, i.e. OpenSSL EVP_PKEY_EC; otherwise use EVP_PKEY_RSA
++ */
++#define USE_EVP_PKEY_EC
```
Tim-Magee, line 10: Suggestion: #include "config.h" up top, then guard this with #ifdef FEATURE_EC and add to configure.in a block like:
`AC_ARG_ENABLE(ec,
[  --enable-ec                Use 384-bit ECC instead of 2048-bit RSA for HTTPS inspection.],
[if test $enableval = yes; then
  AC_DEFINE(FEATURE_EC)
fi])
`
Which makes the feature selectable at config time, but not default. To make it the default you'd add a fourth argument to AC_ARG_ENABLE so the last line reads: `fi], AC_DEFINE(FEATURE_EC))`

As I say, suggestion only. Also I just typed that stuff so although it looks right to me it's worth checking https://autotools.io/autoconf/arguments.html

mascguy (approved):

www/privoxy/files/patch-ssl_common.h.diff
essandess: Thank you for the suggestion.

If upstream adopts this feature, I will suggest that it become a configurable run-time option, set in the `config` file.

Therefore, whether we add this now using the C preprocessor or `autoconf` is a detail, and I prefer the former at this point because it’s simpler and I know the existing code works.

My preference is to keep the patch as it stands and defer the final design to upstream, if adopted.

Also, please see https://sourceforge.net/p/ijbswa/bugs/933/.

PR #13057, opened 2021-11-20
mysql56, mysql8: use MacPorts’ libedit

chrstphrchvz:
…instead of bundled libedit (as already done for `mysql57` in 38aa1a0ec8)
Fixes: https://trac.macports.org/ticket/64006

###### Type(s)

- [x] bugfix

###### ~~Tested on~~

Only configure phase is tested. I would like @laggardkernel or other knowledgable user to confirm the editline issue is resolved.

###### Verification 
Have you

### DIFF
--- a/databases/mysql56/Portfile
+++ b/databases/mysql56/Portfile
@@ -12,7 +12,7 @@ set name_mysql      ${name}
 version             5.6.51
 # Set revision_client and revision_server to 0 on
 # version bump.
-set revision_client 3
+set revision_client 4
 set revision_server 1
 set version_branch  [join [lrange [split ${version} .] 0 1] .]
 categories          databases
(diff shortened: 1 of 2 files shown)


chrstphrchvz:
MySQL does not allow overriding the CMake variables `EDITLINE_INCLUDE_DIR` and `EDITLINE_LIBRARY` (at least without patching); but if I understand correctly, `EDITLINE_INCLUDE_DIR` will be set by `FIND_PATH()` based on `CMAKE_SYSTEM_PREFIX_PATH="${prefix};/usr"` passed to configure, and `EDITLINE_LIBRARY` set by `FIND_LIBRARY()` based on the `LIBRARY_PATH='${prefix}/lib'` environment variable. So it should only ever use libedit headers and libraries from MacPorts, and not from system/Homebrew/etc.

PR #13069, opened 2021-11-21
gocryptfs: Fix build for arm64

MarcelBochtler:
#### Description

* Introduce doc variant, which is enabled by default.
* Use go-md2man instead of pandoc on arm64 as this arch is not yet
supported by upstream yet [1].

[1]: https://github.com/jgm/pandoc/issues/6960

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.1 20G224 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/fuse/gocryptfs/Portfile
+++ b/fuse/gocryptfs/Portfile
@@ -123,9 +123,6 @@ go.vendors          gopkg.in/yaml.v3 \
 set builddate       "build_date_not_set"
 set ldflags         "-X \"main.GitVersion=${version}\" -X \"main.GitVersionFuse=${gitversionfuse}\" -X \"main.BuildDate=${builddate}\""
 
-depends_build-append \
-                    port:pandoc
-
 build.args          -ldflags="${ldflags}"
 
 # According to [1] there is no benefit to build gocryptfs with openssl on M1 Macs.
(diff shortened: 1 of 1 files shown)


pmetzger:
Would md2man work better than pandoc on all platforms?

MarcelBochtler:
> Would md2man work better than pandoc on all platforms?

No, because the manpage generated by md2man has some formatting issues, which is why I'd like to use it only when pandoc is not available.

pmetzger:
Makes sense.

kencu:
We previously had arm64 copy over and install the x86_64 binary of pandoc, which runs just fine on arm64 and solves this issue until ghc is updated to support arm macs. It was packaged like this:

https://packages.macports.org/pandoc/pandoc-2.10.1_0.darwin_20.arm64.tbz2


but somewhere along the line that was scrubbed out, and I guess we're just broken again.

kencu:
Actually, Josh said that MacPorts would automatically see the x86_64 version as runnable on arm64 Macs, and install that x86_64 version without any intervention by the user if it existed on the packages server.

Or that was supposed to be how it worked, if I understood the logic correctly.

AlexBasile123:
Thanks @MarcelBochtler ! The change is working on M1 now

MarcelBochtler:
@kencu I investigated this and it seems to be that direct installations of x86_64 ports works on ARM64 devices.
However, when trying to install ports that has a x86_64 dependency it doesn't work. I tried with [scrcpy](https://ports.macports.org/port/scrcpy/):
```
❯ sudo port install scrcpy
--->  Computing dependencies for scrcpy
Error: Cannot install scrcpy for the arch 'arm64' because
Error: its dependency android-platform-tools only supports the arch 'x86_64'.
Error: Processing of port scrcpy failed

~
❯ sudo port install android-platform-tools
--->  Computing dependencies for android-platform-tools
--->  Fetching archive for android-platform-tools
--->  Attempting to fetch android-platform-tools-31.0.3_0.darwin_21.x86_64.tbz2 from https://nue.de.packages.macports.org/android-platform-tools
--->  Attempting to fetch android-platform-tools-31.0.3_0.darwin_21.x86_64.tbz2 from https://fra.de.packages.macports.org/android-platform-tools
--->  Attempting to fetch android-platform-tools-31.0.3_0.darwin_21.x86_64.tbz2 from https://packages.macports.org/android-platform-tools
--->  Fetching distfiles for android-platform-tools
--->  Verifying checksums for android-platform-tools
--->  Extracting android-platform-tools
--->  Configuring android-platform-tools
--->  Building android-platform-tools
--->  Staging android-platform-tools into destroot
--->  Installing android-platform-tools @31.0.3_0
--->  Activating android-platform-tools @31.0.3_0
--->  Cleaning android-platform-tools
--->  Updating database of binaries
--->  No broken files found.
--->  No broken ports found.
```

This issue is also outlined in this ticket: https://trac.macports.org/ticket/63092

kencu:
Hmm. I wonder if we might "depends_skip_archcheck pandoc" then, as it is just a tool and we don't care if it has a matching architecture to the port being built.

that may not be the exact line to use in the Portfile, but it is something similar to that..

PR #13099, opened 2021-11-23
blender: Fix `OSL_SHADER_DIR-NOTFOUND` compile error

jasonliu--:
#### Description

This PR fixes the case where the compile fails with a `OSL_SHADER_DIR-NOTFOUND` error. The error occurs because Blender's CMake scripts are unable to find the directory that contains the OSL shaders, due to the search paths being hard-coded to search in `/usr/local`.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/graphics/blender/Portfile
+++ b/graphics/blender/Portfile
@@ -8,7 +8,7 @@ PortGroup           boost 1.0
 
 name                blender
 version             2.91.2
-revision            4
+revision            5
 categories          graphics multimedia
 platforms           darwin
 license             GPL-2+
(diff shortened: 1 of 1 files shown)


graphics/blender/Portfile
```diff
@@ -248,6 +248,27 @@ list(APPEND USD_LIBRARIES\\
...
+            "s|/opt/share|${prefix}/share|" \
+        ] {
+            reinplace -q -E $re $f
+        }
+    }
+    reinplace -E "s|(${prefix}/lib)|\\1/opencollada|" \
+        ${worksrcpath}/build_files/cmake/Modules/FindOpenCOLLADA.cmake
+    reinplace -E "s|(${prefix}/lib)|\\1/usd|" \
+        ${worksrcpath}/build_files/cmake/Modules/FindUSD.cmake
+
```
pmetzger, line 271: It's a bit of a hammer, but I imagine it works. I assume really is no way to explicitly set these locations during configuration?

jasonliu--: > It's a bit of a hammer, but I imagine it works. I assume really is no way to explicitly set these locations during configuration?

It's possible, but about as equal of a hack as what I'm doing. The upstream devs make it easy to turn the feature completely off by using `-DWITH_USD=OFF`, but the source code generally relies on the Find*.cmake scripts to figure out the paths. MacPorts' cmake PortGroup does a pretty good job of providing hints to CMake in terms of trying to tell it to look in `/opt/local`, but it does occasionally miss certain locations.

PR #13117, opened 2021-11-25
scrcpy: Set supported_archs to noarch

MarcelBochtler:
#### Description

This enables scrcpy to be installed on arm64 devices, even if its
dependency android-platform-tools is only available for x86_64 devices.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d
and 
macOS 11.6.1 20G224 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/multimedia/scrcpy/Portfile
+++ b/multimedia/scrcpy/Portfile
@@ -38,7 +38,22 @@ depends_build-append \
 depends_lib-append  path:lib/libavcodec.dylib:ffmpeg \
                     port:libsdl2
 
-depends_run-append  port:android-platform-tools
+variant adb description {Install android-platform-tools for adb} {
+    depends_run-append  port:android-platform-tools
+}
+
+if {${configure.build_arch} eq "x86_64"} {
+    default_variants +adb
+}
+
+if {![variant_isset adb]} {
+    notes "
+        adb has to be available in your PATH at runtime. You can either:
+        * Install it via MacPorts: `sudo port install android-platform-tools`
+        * Install it via Android Studio and add it to your PATH
+        * Install scrcpy using the +adb variant: `sudo port install scrcpy +adb`
+    "
+}
 
 configure.args-append \
                     --buildtype release \


MarcelBochtler:
Hi @herbygillot,

thanks for your review.

> If `scrcpy` is building for `x86_64`, but **not** for `arm64`, what you may be able to do is specify `installs_libs no` if it is not installing any libraries.

`scrcpy` builds on `arm64`. The dependency `android-platform-tools` does not (and it also specifies the `supported_arch` `x86_64` explicitly.

This causes this error when trying to install `scrcpy` on `arm64`:
```
❯ sudo port install
--->  Computing dependencies for scrcpy
Error: Cannot install scrcpy for the arch 'arm64' because
Error: its dependency android-platform-tools only supports the arch 'x86_64'.
Error: Processing of port scrcpy failed
```

`installs_libs no` does not solve this.

What about a workaround:
1. Add a variant `+adb` which adds the runtime dependency `port:android-platform-tools`.
2. Default to `-adb` on `arm64` and print a hint that says that the user has to make `adb` available from `$PATH` that `scrcpy` works, either via macports or from Android Stuido. [This is how homebrew](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/scrcpy.rb) is installing `scrcpy`:
> At runtime, adb must be accessible from your PATH.

3. Default to `+adb` on other platforms

What do you think?

herbygillot:
> Hi @herbygillot,
> 
> thanks for your review.
> 
> > If `scrcpy` is building for `x86_64`, but **not** for `arm64`, what you may be able to do is specify `installs_libs no` if it is not installing any libraries.
> 
> `scrcpy` builds on `arm64`. The dependency `android-platform-tools` does not (and it also specifies the `supported_arch` `x86_64` explicitly.
> 
> This causes this error when trying to install `scrcpy` on `arm64`:
> 
> ```
> ❯ sudo port install
> --->  Computing dependencies for scrcpy
> Error: Cannot install scrcpy for the arch 'arm64' because
> Error: its dependency android-platform-tools only supports the arch 'x86_64'.
> Error: Follow https://guide.macports.org/#project.tickets if you believe there is a bug.
> Error: Processing of port scrcpy failed
> ```
> 
> `installs_libs no` does not solve this.
> 
> What about a workaround:
> 
> 1. Add a variant `+adb` which adds the runtime dependency `port:android-platform-tools`.
> 2. Default to `-adb` on `arm64` and print a hint that says that the user has to make `adb` available from `$PATH` that `scrcpy` works, either via macports or from Android Stuido. [This is how homebrew](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/scrcpy.rb) is installing `scrcpy`:
> 
> > At runtime, adb must be accessible from your PATH.
> 
> 3. Default to `+adb` on other platforms
> 
> What do you think?

I think you can try adding `installs_libs no` to `android-platform-tools`, as long as we know that `android-platform-tools` does not provide any shared libraries.

What I am trying to say is the MacPorts knows that it can run `x86_64` binaries on an Apple `arm64` platform.

It will choose **not** to do so if it thinks the `x86_64` port includes shared libraries, because then things become messy: you cannot compile an `arm64` binary against `x86_64` shared libs.

So the point of the workaround I mentioned asserts to MacPorts that this `x86_64` port has no shared libraries, it should be safe to use on an `arm64` platform.  So MacPorts will download the already built `x86_64` package and use it as the dep for the `arm64` port.

MarcelBochtler:
Thanks for the explanation. Closing this in favor of https://github.com/macports/macports-ports/pull/13186

PR #13156, opened 2021-11-28
kcov: add rpath for LLDB

bcully:
#### Description

On at least OS X 11 and 12 (all I have available for testing now),
kcov currently fails to start:

```
dyld[3855]: Library not loaded: @rpath/LLDB.framework/Versions/A/LLDB
  Referenced from: /opt/local/bin/kcov
  Reason: tried: '/opt/local/lib/LLDB.framework/Versions/A/LLDB' (no such file), '/opt/local/lib/LLDB.framework/Versions/A/LLDB' (no such file), '/Library/Frameworks/LLDB.framework/Versions/A/LLDB' (no such file), '/System/Library/Frameworks/LLDB.framework/Versions/A/LLDB' (no such file)
```

LLDB.framework isn't in the system path on these versions of OS X. The
only copy on my systems is in
`/Library/Developer/CommandLineTools/Library/PrivateFrameworks`

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/kcov/Portfile
+++ b/devel/kcov/Portfile
@@ -9,7 +9,7 @@ PortGroup           legacysupport 1.0
 legacysupport.newest_darwin_requires_legacy 14
 
 github.setup        SimonKagstrom kcov 38 v
-revision            0
+revision            1
 checksums           rmd160  7bc2dfb463ec46c5ae541894fc522c18a615e9f2 \
                     sha256  b37af60d81a9b1e3b140f9473bdcb7975af12040feb24cc666f9bb2bb0be68b4 \
                     size    306098
@@ -36,7 +36,8 @@ depends_build-append \
 depends_lib-append  port:curl \
                     port:zlib
 
-patchfiles          bin-to-c-source.py.use-python38.patch
+patchfiles          bin-to-c-source.py.use-python38.patch \
+                    kcov-lldb-rpath.patch
 
 post-patch {
     reinplace "s|@PREFIX@|${prefix}|g" ${worksrcpath}/src/bin-to-c-source.py
--- a/devel/kcov/files/kcov-lldb-rpath.patch
+++ b/devel/kcov/files/kcov-lldb-rpath.patch
@@ -1,0 +1,11 @@
+--- src/CMakeLists.txt.orig	2021-12-07 21:15:06.000000000 -0800
++++ src/CMakeLists.txt	2021-12-07 21:32:38.000000000 -0800
+@@ -443,7 +443,7 @@
+       set(LLDB_LIBRARY_INSTALL_NAME "${LLDB_LIBRARY}/Versions/Current/LLDB")
+ 	  add_custom_command(TARGET ${KCOV}
+         POST_BUILD COMMAND
+-        "${CMAKE_INSTALL_NAME_TOOL}" -change @rpath/LLDB.framework/LLDB "${LLDB_LIBRARY_INSTALL_NAME}" $<TARGET_FILE:${KCOV}>
++        "${CMAKE_INSTALL_NAME_TOOL}" -change @rpath/LLDB.framework/LLDB "${LLDB_LIBRARY_INSTALL_NAME}" -change @rpath/LLDB.framework/Versions/A/LLDB "${LLDB_LIBRARY_INSTALL_NAME}" $<TARGET_FILE:${KCOV}>
+         COMMENT "Changing LLDB install name from @rpath/LLDB.framework/LLDB to ${LLDB_LIBRARY_INSTALL_NAME}")
+ 	endif()
+ endif()


devel/kcov/Portfile
```diff
@@ -41,3 +41,9 @@ patchfiles          bin-to-c-source.py.use-python38.patch
 post-patch {
     reinplace "s|@PREFIX@|${prefix}|g" ${worksrcpath}/src/bin-to-c-source.py
 }
+
+post-destroot {
+    # kcov links to LLDB as @rpath/LLDB.framework/Versions/A/LLDB but doesn't add an rpath to find it
+    # On at least OS X 11 and 12 this causes kcov to fail to start with a dynamic linker error
+    system "install_name_tool -add_rpath /Library/Developer/CommandLineTools/Library/PrivateFrameworks ${destroot}${prefix}/bin/kcov"
```
mascguy, line 48: Seems reasonable, at least on the surface. But unfortunately I don't have a Big Sur or Monterey install to test against.

Also, is there a preferred/better way to handle this type of situation? @kencu and others?

kencu:
The fix in this PR looks like one way to make it work. The software looks like it wants to do the right thing:

https://github.com/SimonKagstrom/kcov/blob/271ebb77b9101e7f7290d23735b8a1188c203b9a/src/CMakeLists.txt#L473

but I guess something is preventing it from doing so (possible the cmake PortGroup turning off @rpath support on MacPorts installs?). 

I have never tried to build this, myself.

mascguy:
> The fix in this PR looks like one way to make it work. The software looks like it wants to do the right thing:
> 
> https://github.com/SimonKagstrom/kcov/blob/271ebb77b9101e7f7290d23735b8a1188c203b9a/src/CMakeLists.txt#L473
> 
> but I guess something is preventing it from doing so (possible the cmake PortGroup turning off @rpath support on MacPorts installs?).

Thanks Ken, glad to hear that it's a reasonable approach.

Do you think we should merge this as-is, or perhaps investigate the issue you mentioned?

kencu:
the fix in this PR is not MacPorts usual way of fixing this problem.

If it is committed like this, there will probably be complaints. I could not approve it or do it, but perhaps you might get away with it.

The usual Macports fix is to get rid of the rpath use, like the build tries to do itself. So the Macports fix is to fix that, or do that.

mascguy:
> the fix in this PR is not MacPorts usual way of fixing this problem.
> 
> If it is committed like this, there will probably be complaints. I could not approve it or do it, but perhaps you might get away with it.
> 
> The usual Macports fix is to get rid of the rpath use, like the build tries to do itself. So the Macports fix is to fix that, or do that.

Good feedback. And I'd certainly prefer to avoid merging a temporary fix, if there's a better way to resolve it.

@ryandesign and @jmroot, thoughts regarding how we proceed here?

kencu:
You didn't like the suggestion to just fix it to not use rpaths? That is the only suggestion you will get from Ryan or Josh or anyone else, really.

How to fix it to not use rpaths is described above -- make it use the build's internal fix mechanism, or do that manually if you can't figure out how to make the build do it, as a last option.

mascguy:
> You didn't like the suggestion to just fix it to not use rpaths? That is the only suggestion you will get from Ryan or Josh or anyone else, really.
> 
> How to fix it to not use rpaths is described above -- make it use the build's internal fix mechanism, or do that manually if you can't figure out how to make the build do it, as a last option.

I'm definitely fine with the rpath suggestion!

mascguy:
> Yeah, that comment was pre-coffee, sorry :>

No worries, mine were as well! LOL

bcully:
I don't know the downside to adding an rpath (does it make the linkage less deterministic?) but based on your feedback I've patched the cmake script instead. The problem on my system was just that the source path given to install_name_tool was incorrect.

It is perhaps possible that the name was once correct, or is correct on older versions of macOS. Given that I suppose an alternative might be to try to patch both source paths?

mascguy:
> I don't know the downside to adding an rpath (does it make the linkage less deterministic?) but based on your feedback I've patched the cmake script instead. The problem on my system was just that the source path given to install_name_tool was incorrect.
> 
> It is perhaps possible that the name was once correct, or is correct on older versions of macOS. Given that, I've modified the build script to try both locations.

Sounds good.

@kencu , I'm not yet caffeinated enough to assess the latest changes, but they sound good in principle.

What do you think?

kencu:
When I started contributing to MacPorts in 2016 I learned that there was a policy / strong suggestion to not use rpaths for library linking. Nothing has changed since then.

I believe the basic issue with rpaths is mostly that it is at times more difficult to debug issues related to which library is actually being used on a remote users system, when such users are often unskilled. To really know which library is in use by the dynamic linker requires setting some environment variables to turn on dyld link tracing. Very few people know how to do this. 

Otherwise, you have to know in what order dyld will search various paths for rpath-specified libraries (almost nobody knows this) and then how to properly set the rpath search order in the executable (nobody knows this). So for example we like to (sometimes) use a newer libc++.dylib on some systems, and install one in ```/opt/local/lib/libcxx/libc++.dylib```. If this is fully hard-coded, then we know which one is being used. If this is referenced by an rpath, then it will be used if it is there, otherwise the executable will fall back to the one in ```/usr/lib/libc++.dylib```. But there is no ideal what is going on in a submitted ticket, and debugging it is a nightmare.

At present, rpath use is tolerated in MacPorts only in very very few scenarios, where there is a relocatable executable perhaps, and only one library it can use, or some such. In fact, the llvm/clang ports are one of the few places where it has to be used to allow things to work properly.

Debian has had similar issues with rpaths, and rpaths are not allowed on debian:

https://wiki.debian.org/RpathIssue

It might, however, be coming that we have no choice but to accept rpath use; the darwin gcc maintainer has been quite strong in his support for rpath use by gcc, based in large part on the difficulty in running the test suite on a not-yet-installed gcc build when there is already an installed gcc build -- the test suite picks up the installed libgcc instead of the one you want to test, and this is difficult to work around with tooling other than rpath.

cmake and meson use the same mechanism to allow in-tree testing (with rpaths) but they rewrite the paths to full paths on installation in most cases. Perhaps we can sort out how to make gcc do that too -- in the meantime, the rpath use is gcc-devel is causing plenty of heartburn.

For this port, if you would please just show us the otool output of the final product has no rpaths in there, this can be committed.

Thanks for reading!

bcully:
Thanks very much for all the context, I really appreciate it. I wonder if it would be helpful to include the rpath policy in the ports guide, perhaps in https://guide.macports.org/chunked/development.practices.html ?

The otool output:
```
❯ otool -L /opt/local/bin/kcov
/opt/local/bin/kcov:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 905.6.0)
	/opt/local/lib/libcurl.4.dylib (compatibility version 12.0.0, current version 12.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
	/Library/Developer/CommandLineTools/Library/PrivateFrameworks/LLDB.framework/Versions/Current/LLDB (compatibility version 1.0.0, current version 1205.0.27)
```

PR #13164, opened 2021-11-28
qemu: build with MacPorts meson

evanmiller:
#### Description

qemu's internal meson contains a number of bugs that have been fixed either upstream or by MacPorts patches. One bug prevents building with GCC (#12069); others prevent building with PowerPC and on Tiger. To work around these bugs, supply the MacPorts meson at to qemu's configure, and also skip over an error that disallows both `--python` and `--meson` to be specified.

(For the full context of the commented-out error, see:

https://github.com/qemu/qemu/blob/dd4b0de45965538f19bb40c7ddaaba384a8c613a/configure#L1519-L1530

Both qemu and meson use MacPorts-supplied Python 3.9, so the error is not applicable.)

This is not a complete fix for any particular OS/compiler combination; it is rather a prerequisite to other PRs which will follow.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -31,20 +31,22 @@ checksums               rmd160  00f21e102c9ca70c154b770d0094fd48568bbdd5 \
 
 depends_build           port:texinfo \
                         port:libtool \
+                        port:meson \
                         port:ninja \
                         port:pkgconfig \
                         port:py39-sphinx
 
-# python/perl5 is only used for build scripts, no linking
-depends_build-append    port:python39 \
-                        port:perl5
+# perl5 is only used for build scripts, no linking
+depends_build-append    port:perl5
 
-license_noconflict      python39 \
-                        perl5
+license_noconflict      perl5
 
 depends_lib             path:lib/pkgconfig/glib-2.0.pc:glib2 \
                         port:zlib \
                         path:lib/pkgconfig/pixman-1.pc:libpixman
+
+# This patch sets the python interpreter from meson's shebang line
+patchfiles              patch-qemu-configure.diff
 
 # "ERROR: You need at least GCC v7.5 or Clang v6.0 (or XCode Clang v10.0)"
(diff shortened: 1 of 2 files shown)


mascguy (approved):

raimue:
The problem is that we have to ensure qemu and meson will always use the same python version. This is the case at the moment, but as soon as the meson port is updated to the next python release, we also need to update the qemu port at the same time. So... how are we going to do maintain this relationship?

evanmiller:
@raimue I have modified the patch to extract the python path from meson's shebang line. Does that address your concern?

raimue:
This is a nice hack! :+1: 

By deleting the `license_noconflict` I was a bit afraid the port can no longer be distributed as binary archive. However, I think the problem was that Python linked with OpenSSL 1.x with its license being incompatible with the GPL. But with OpenSSL >= 3.0 I expect this to be a thing of the past now anyway. However, I cannot confirm this right now with `port_binary_distributable.tcl` as I am not on my Mac.

mascguy:
Looks good now!

```
$ ../macports-infra/jobs/port_binary_distributable.tcl -v qemu
"qemu" is distributable
```

evanmiller:
One final note: I'm not sure how exactly `python.versions` works, but all of the versions listed in the `meson` Portfile should be compatible with qemu (3.6+, defaulting to 3.9).

mascguy:
Sounds good Evan.

Rainer, are we ready to merge?

raimue (approved):
Thanks for working on this! It makes a lot of sense to use the patched meson provided by MacPorts. 👍

evanmiller:
Thanks for the quick attention. I have a few more PRs planned for qemu as I work my way backward through the OS versions. (I've sent a couple patches upstream, but it's a pretty busy project and I haven't heard anything from them yet...)

PR #13168, opened 2021-11-28
mint: new port

klanchman:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/mint/Portfile
+++ b/devel/mint/Portfile
@@ -1,0 +1,40 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+
+github.setup            yonaskolb Mint 0.17.1
+revision                0
+github.tarball_from     archive
+
+name                    mint
+categories              devel
+license                 MIT
+platforms               darwin
+maintainers             {kylelanchman.com:macports @klanchman} openmaintainer
+
+description             A package manager that installs and runs executable Swift packages
+long_description        {*}${description}
+
+checksums               rmd160  9e5d01ecb931cb17d88e29c146ebf2a4ba77b463 \
+                        sha256  0e3ab23e548a752f6eee3a7b98d1c137a30371e4a0ec9212840baaa56741d2e4 \
+                        size    23308
+
+if {${os.platform} eq "darwin" && ${os.major} < 19} {
+    known_fail yes
+    pre-fetch {
+        ui_error "${name} requires macOS 10.15 or later."
+        return -code error "incompatible macOS version"
+    }
+}
+
+# Universal support enabled by default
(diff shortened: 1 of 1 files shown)


mascguy:
Kyle, looks great overall. And welcome to MacPorts!

One thought though: Would it be better to run the build during the `build` phase, rather than `destroot`?

mascguy:
> Kyle, looks great overall. And welcome to MacPorts!
> 
> One thought though: Would it be better to run the build during the `build` phase, rather than `destroot`?

To be fair though, I haven't looked closely at the numerous ports that build via Xcode. So if this is the established convention, then I'm fine with it.

Anyone else?

mascguy:
> I initially tried splitting it up, but I found that running `make install` caused it to build again during `destroot`. Would it be more typical to `xinstall` during `destroot` instead of using the Makefile in this case? (I was borrowing from the [Carthage Portfile](https://github.com/macports/macports-ports/blob/master/devel/carthage/Portfile) in this case since it was similar and simple 😄)

Yes indeed, so no need to run multiple targets.

> Aside, is the build failure on Xcode 11 something I should deal with? I glanced at it, and the Makefile is using a newer Swift compiler option (`--arch`, Xcode ~12.5?), but I think the project is supposed to build with Xcode 10.2+. If so, should I actually make it work, or just add a minimum Xcode version that works with the Makefile?

For now, we can require macOS 10.15 (or 10.14, if that's the minimum version). But if you can update the port to support older macOS/Xcode releases, that would definitely be preferable!

Also, just pushed some minor changes. Now we're building during that phase, with nothing to be done during phase `destroot`.

Take a look, and let me know if it all makes sense!

mascguy:
Let's merge as-is, as a starting point. And then if you want to refine via a subsequent PR - such as expanding macOS/Xcode support, if possible - we can certainly do that.

And thanks so much for your contribution Kyle!

klanchman:
Sounds good! Thanks for the cleanup, I think it all makes sense 👍

What would be a good way to test if I don’t have easy access to older versions of macOS? Make a draft PR and fiddle with it so the checks here run? Is there some way to try it out against all the configurations shown in the “Port Health” section of the port detail page?

mascguy:
> What would be a good way to test if I don’t have easy access to older versions of macOS? Make a draft PR and fiddle with it so the checks here run? Is there some way to try it out against all the configurations shown in the “Port Health” section of the port detail page?

One of the GitHub CI jobs cover macOS 10 14, so you'll get that validation automatically when creating a PR.

For macOS releases earlier than that, I'm happy to help: I have a suite of macOS VMs covering 10.6 through 10.15, making it easy to validate most anything.

To get started, create a draft PR, and get it working for 10.14. Once that's done, I can assist with validation on earlier releases.

How does that sound?

klanchman:
Excellent! I’ll try to start playing around with this in the next few days.
For some reason it hadn’t really struck me how far back MacPorts tries to support! 🤯

PR #13214, opened 2021-12-01
rust: fix build on macOS 10.7 and 10.8

catap:
#### Description

Closes: https://trac.macports.org/ticket/62639
Closes: https://trac.macports.org/ticket/58848

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.8.5 12F45 x86_64
Xcode 5.1.1 5B1008

macOS 10.7.5 11G63 x86_64
Xcode 4.6.3 4H1503

###### Verification 
Have you

### DIFF
--- a/lang/rust/Portfile
+++ b/lang/rust/Portfile
@@ -153,7 +153,7 @@ checksums-append    rust-std-${rustc_version}-x86_64-apple-${os.platform}${extra
                     sha256  6ed30275214e956ee10b03db87b0b4297948fd102d39896cece01669555047ef \
                     size    5918639
 
-set min_darwin 13
+set min_darwin 12
 if {${os.platform} eq "darwin" && ${os.major} < ${min_darwin}} {
     known_fail yes
     pre-fetch {
@@ -166,7 +166,21 @@ set rust_platform       ${arch}-apple-${os.platform}
 set rust_root           ${worksrcpath}/build/stage0-${arch}
 
 if { ${subport} ne ${ccwrap} } {
+    if {${os.platform} eq "darwin" && ${os.major} < 13} {
+        depends_build-append \
+                        port:curl
+    }
+
     post-extract {
+        # Relink agains MacPort's libcurl allows to fix `error: Unknown option`
+        # which happened because cargo tries to use HTTP/2 and it fails
+        # See: https://github.com/rust-lang/cargo/blob/2e2a16e983f597da62bc132eb191bc3276d4b1bb/src/cargo/core/package.rs#L423
+        if {${os.platform} eq "darwin" && ${os.major} < 13} {
+            set sLib /usr/lib/libcurl.4.dylib
+            set lLib ${prefix}/lib/libcurl.dylib
+            system "install_name_tool -change ${sLib} ${lLib} ${workpath}/cargo-${rustc_version}-${rust_platform}/cargo/bin/cargo > /dev/null 2>&1 ; true"
+        }
+
         file mkdir ${rust_root}
         system "cp -r ${workpath}/rust-std-${rustc_version}-${rust_platform}/rust-std-${rust_platform}/* ${rust_root}"
         system "cp -r ${workpath}/rustc-${rustc_version}-${rust_platform}/rustc/* ${rust_root}"
--- a/lang/rust/Portfile
+++ b/lang/rust/Portfile
@@ -154,6 +154,8 @@ checksums-append    rust-std-${rustc_version}-x86_64-apple-${os.platform}${extra
                     sha256  6ed30275214e956ee10b03db87b0b4297948fd102d39896cece01669555047ef \
                     size    5918639
 
+# The only way to introduce support of macOS 10.6 is reimplement __thread
+# See: https://github.com/rust-lang/rust/issues/10842
 set min_darwin 11
 if {${os.platform} eq "darwin" && ${os.major} < ${min_darwin}} {
     known_fail yes
@@ -191,6 +193,8 @@ if { ${subport} ne ${ccwrap} } {
 
 # Rust claims to only need c++14 but uses some c++17 features relating to std::unique_ptr
 compiler.cxx_standard  2017
+# Rust never claims to need it, but it uses `__thread`
+compiler.thread_local_storage yes
 # Use recent macports clang
 compiler.blacklist-append {macports-clang-[4-9].0}
 # Want to force a specific version, as it is baked into wrapper scripts, so do
--- a/lang/rust/Portfile
+++ b/lang/rust/Portfile
@@ -8,6 +8,7 @@ PortGroup           legacysupport               1.1
 legacysupport.use_static                        yes
 legacysupport.newest_darwin_requires_legacy     15
 
+# keep in mind that you also need to update cargo.crates at the end of this file
 name                rust
 version             1.57.0
 revision            0
@@ -153,7 +154,7 @@ checksums-append    rust-std-${rustc_version}-x86_64-apple-${os.platform}${extra
                     sha256  6ed30275214e956ee10b03db87b0b4297948fd102d39896cece01669555047ef \
                     size    5918639
 
-set min_darwin 12
+set min_darwin 11
 if {${os.platform} eq "darwin" && ${os.major} < ${min_darwin}} {
     known_fail yes
     pre-fetch {
@@ -309,8 +310,7 @@ if { ${subport} ne ${ccwrap} } {
 }
 envs_append       LD=${ld}
 
-configure.args    --enable-vendor \
-                  --default-linker=${ld} \
+configure.args    --default-linker=${ld} \
                   --disable-codegen-tests \
                   --disable-docs \
                   --release-channel=stable \
@@ -339,7 +339,7 @@ post-configure {
     # place config file in the HOME directory
     # see https://trac.macports.org/wiki/UsingTheRightCompiler
     xinstall -d -m 0755 ${workpath}/.home/.cargo
-    set config [open ${workpath}/.home/.cargo/config w]
+    set config [open ${workpath}/.home/.cargo/config a]
 
     puts ${config} "\[target.${rust_platform}\]"
     puts ${config} "linker = \"${ld}\""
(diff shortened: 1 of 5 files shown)


catap:
Probably `__thread` can be fixed for existed build by introducing a wrapper which calls the right function from macports-libcxx.

mascguy:
This is an exciting proposition. Great stuff Kirill!

catap:
I'd love to ask @kencu, maybe he has some ideas how to emulate `__tlv_bootstrap`. It is needed only on bootstrap stage and as soon as it done, I thinkg, macports has rust for 10.6 (and maybe before).

kencu:
If you have rust building on 10.7+, that is quite a feat! Well done indeed.

tlv_bootstrap is kinda tricky to emulate on < 10.7, I found. There are two methods of doing thread_local_storage on Darwin. Apple's version is done by dyld, and involves tlv_* functions.

The other method is "emulated_tls" which is done by gcc on all versions of MacOS in fact, and is done by llvm on systems that don't have intrinsic thread_local_storage. 

I modified llvm (and clang and libc++abi) to provide that emulated_tls on < 10.7 to give us what we have now. It works surprisingly well. But I have not, as yet, mapped out exact function-by-function correlates between the tlv_* functions and the emulated_tls* functions. I don't think they exactly map over 1::1 , but I did not dig deeply into trying to make it work, to be honest. 

perhaps it could be done, if someone focused just on that for a bit.

great work with 10.7+!

catap:
@kencu, the tricky part that it should be emulated somehow in the way which can be used to relink pre built version of rust :)

Anyway, it leads to the next issue: rust uses its version of llvm and your patches, probably, should be backported.

All of this seems durable but I feel that it should be the next step.

Gcenx:
Unrelated but since it seems your gotten rust working on older versions of mac do you happen to be versed enough to deal with cross-building rust for i386?

catap:
@Gcenx I haven’t tried it.

kencu:
rust is supposed to be able to cross-compile (via llvm) pretty effortlessly. 

Our problem has been trying to build an i386 rust, rather than just use an x86_64 rust to compile i386 code.

It's a MacPorts limitation in +universal logic, really -- overcomeable though.

Gcenx:
> rust is supposed to be able to cross-compile (via llvm) pretty effortlessly.
> 
> Our problem has been trying to build an i386 rust, rather than just use an x86_64 rust to compile i386 code.
> 
> It's a MacPorts limitation in +universal logic, really -- overcomeable though.

That could be the case but rust documentation is rather annoying, guess that something to look into later.

cjones051073:
@catap Its not so easy to work out the reasons behind what you have done here, just looking at the changes. Can you please write a short description of the changes and put that in the PR description above ?

catap:
@cjones051073 before it was inside commit, but I’ve tried explained everything that I’ve done and why via comments.

Anyway, I’ll summarize changes.

Or I can split into two commits: for 10.8 and 10.7

cjones051073:
> @cjones051073 before it was inside commit, but I’ve tried explained everything that I’ve done and why via comments.
> 
> Anyway, I’ll summarize changes.
> 
> Or I can split into two commits: for 10.8 and 10.7

you can certainly use more than one commit, if you think it would help documenting the changes in each those commits.

cjones051073:
For instance one thing no clear to me is why we now have the `cargo.crates` section ? Is this to just avoid rust triggering the downloads itself during build (and thus to do the fetch correctly in the MacPorts fetch stage) ?

catap:
@cjones051073 here it is.

mascguy:
> Unrelated but since it seems your gotten rust working on older versions of mac do you happen to be versed enough to deal with cross-building rust for i386?

@evanmiller, would you be interested in collaborating as well? Between you, Dean, and Kirill, I bet you folks could get it working!

cjones051073:
Perfect thanks, those comments really help.

catap:
@cjones051073 I'm testing a naive approach to introduce macOS 10.6, and I need a few more hours to test it. Anyway, the first two commits is ready for review :)

lang/rust/Portfile
```diff
@@ -166,7 +169,21 @@ set rust_platform       ${arch}-apple-${os.platform}
...
+                        port:curl
+    }
+
     post-extract {
+        # Relink agains MacPort's libcurl allows to fix `error: Unknown option`
+        # which happened because cargo tries to use HTTP/2 and it fails
+        # See: https://github.com/rust-lang/cargo/blob/2e2a16e983f597da62bc132eb191bc3276d4b1bb/src/cargo/core/package.rs#L423
+        if {${os.platform} eq "darwin" && ${os.major} < 13} {
+            set sLib /usr/lib/libcurl.4.dylib
+            set lLib ${prefix}/lib/libcurl.4.dylib
```
cjones051073, line 183: Minor point, but what happens here if curl is update in the future to no longer provide v4 dylib here. maybe make this just `${prefix}/lib/libcurl.dylib` ?

lang/rust/Portfile
```diff
@@ -384,3 +403,445 @@ subport rust-src {
...
+
+            # Since https://github.com/rust-lang/rust/commit/8a6ff90a3a41e6ace18aeb089ea0a0eb3726dd08
+            puts ${export_list} "___muloti4"
+
+            close ${export_list}
+
+            system "${configure.cc} -dynamiclib -exported_symbols_list ${export_list_file} -o ${scalpel_dylib} ${prefix}/lib/libcxx/libc++.a"
+
+            build.env-append DYLD_FORCE_FLAT_NAMESPACE=1 \
+                        DYLD_INSERT_LIBRARIES=${scalpel_dylib}
```
cjones051073, line 440: I think it would be nice if maybe in the future this `___muloti4` could be factored into say legacysupport in some way, just to avoid the need to build a small helper lib here...

cjones051073 (approved):
Looks good. Just a few minor points...

I don't myself have my old VMs running at the moment so cannot test build this myself. I am sure @catap has well tested it himself, but it would be nice if someone else with access to the old OSes could also give it a try. As the CI tests aren't much help here ;)

mascguy (approved):

lang/rust/Portfile
catap: @cjones051073 well, it has a few mul options which can be extracted as separated library. Honestly, I feel that the right things to do is keep llvm's compiler-rt as dedicated port and use it when someone needed.

catap:
@cjones051073 regarding tests: I can build a rust and cargo for example by 10.7 and 10.8

evanmiller:
> > Unrelated but since it seems your gotten rust working on older versions of mac do you happen to be versed enough to deal with cross-building rust for i386?
> 
> @evanmiller, would you be interested in collaborating as well? Between you, Dean, and Kirill, I bet you folks could get it working!

I briefly looked into porting Rust to PPC - because Rust has been self-hosting for a long time, you need an ancient version of Rust in order to start bootstrapping. These are distributed in binary form for a limited number of platforms; I couldn't find instructions for supporting a new platform. (You'd want either a cross-compile story or an old Rust compiler written in C/C++.)

lang/rust/Portfile
catap: fixed

lang/rust/Portfile
cjones051073: what I mean is if what is done here is something other ports could benefit from, then we should avoid each port having to build the library themselves.

kencu:
rust is currently working great on my powerpc 32 and 64 bit systems, with llvm-13 and clang-13, running debian11.  There is tooling available to cross-compile to darwin using linux. That is probably someone's only hope for getting rust on powerpc darwin. But as rust needs a functioning recent llvm and needs to be able to output into the darwin powerpc ABI, that project is likely doomed.

catap:
@kencu this can be solved by shiping rust boostrap as huge `.ll` (LLVM-IR) file which just need to be compiled on target platform.

Anyway, as soon as it need to be linked agains something, the hell begin.

kencu:
right now there is no tool that can compile llvm ir into coherent darwin powerpc abi code. The ELF lowering works properly and is fully supported by upstream llvm, which is why debian is sailing along. The darwin powerpc parts never worked right, and what was there has been gradually stripped out staring with llvm/clang 8.0.

catap:
@kencu defently doomed

lang/rust/Portfile
```diff
@@ -295,8 +314,7 @@ if { ${subport} ne ${ccwrap} } {
 }
 envs_append       LD=${ld}
 
-configure.args    --enable-vendor \
```
herbygillot, line 298: Why the removal of `--enable-vendor`?

catap: @herbygillot because it creates the second `source.crates-io`

lang/rust/Portfile
catap: Seems I figured out how to build this library after all. I'll include it inside this PR.

catap: @cjones051073 well, I have no idea how to make such port.

A naive approach:
```
clang -dynamiclib -Wl,-all_load -o /tmp/test.dylib /opt/local/libexec/llvm-11/lib/clang/11.1.0/lib/darwin/libclang_rt.osx.a
```
fails with duplicated symbols
```
duplicate symbol '___bswapdi2' in:
    /opt/local/libexec/llvm-11/lib/clang/11.1.0/lib/darwin/libclang_rt.osx.a(bswapdi2.c.o)
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/13.0.0/lib/darwin/libclang_rt.osx.a(bswapdi2.c.o)
duplicate symbol '___bswapsi2' in:
    /opt/local/libexec/llvm-11/lib/clang/11.1.0/lib/darwin/libclang_rt.osx.a(bswapsi2.c.o)
    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/13.0.0/lib/darwin/libclang_rt.osx.a(bswapsi2.c.o)
```

Can I use `${prefix}/libexec/llvm-11/lib/clang/11.1.0/lib/darwin/libclang_rt.osx.a` as donor instead of `${prefix}/lib/libcxx/libc++.a`? I don't like because I need to hardcore llvm version also here and current approach seems better for me.

Maybe it is a good idea to move this hack with creating a dynamic library and load it to legacysupport but I have no idea how to do it on the right way.

kencu: The usual way to make symbols from compiler_rt available to a build, when needed, is to directly link in the static library, as you mentioned. It would be probably much better to do this, if you can add static libraries to a cargo/rust build (I'm sure you must be able to) rather than to make a dylib and insert it. It would also not be a very good idea (IMHO) to add random comiler_rt symbols to the legacysupport.dylib, as -- they shouldn't need to be there.

kencu:
I used the Portfile version as of about 11:00 PM Pacific DST last night, and let it run all night on a 10.7 hardware system I have. It did this, perhaps not surprisingly, so have to still sort that (or maybe it has been):
```
   Compiling either v1.6.0
   Compiling remove_dir_all v0.5.3
   Compiling crossbeam-utils v0.7.2
   Compiling memoffset v0.5.5
   Compiling crossbeam-epoch v0.8.2
   Compiling rayon v1.3.1
   Compiling lzma-sys v0.1.16
   Compiling textwrap v0.11.0
   Compiling miniz_oxide v0.4.0
dyld: lazy symbol binding failed: Symbol not found: ___muloti4
  Referenced from: /opt/local/var/macports/build/_opt_macportsofficial_macports-ports_lang_rust/rust/work/rustc-1.56.1-src/build/stage0-x86_64/lib/librustc_driver-98178664a8620a8c.dylib
  Expected in: /opt/local/var/macports/build/_opt_macportsofficial_macports-ports_lang_rust/rust/work/rustc-1.56.1-src/build/stage0-x86_64/lib/libstd-35b14ca5e62c7d66.dylib

dyld: Symbol not found: ___muloti4
  Referenced from: /opt/local/var/macports/build/_opt_macportsofficial_macports-ports_lang_rust/rust/work/rustc-1.56.1-src/build/stage0-x86_64/lib/librustc_driver-98178664a8620a8c.dylib
  Expected in: /opt/local/var/macports/build/_opt_macportsofficial_macports-ports_lang_rust/rust/work/rustc-1.56.1-src/build/stage0-x86_64/lib/libstd-35b14ca5e62c7d66.dylib

   Compiling walkdir v2.3.1

Did not run successfully: signal: 5
rustc exited with signal: 5
error: could not compile `miniz_oxide`

```

lang/rust/Portfile
catap: @kencu the issue that I need it to run `rustc` which was prebuilt. `rustc` tries to looks for `___muloti4` inside its std library which is also ships as prebuilt binary.

The only way is create a small dynamic library which I've forced to load and switched everything to flat namespace.

Honest, I doubt that such hack is needed for any another port.

catap:
@kencu this is an issue which I've solved by my small dynamic library.

kencu:
I used a patch with the dylib fix, but perhaps it didn't take right. 

 32593ecbbe15cb35639b6f6e8638bd303a85f6f5

I think the commits have been updated since this

lang/rust/Portfile
kencu: Yes, you're right. I forgot for a moment that this was needed to get the downloaded prebuilt binary to work. Right.

catap:
Seems, that better to disable `CARGO_LOG=debug` because right now it produces huge amount of logs:
```
macOS-Lion:~ catap$ du -sh /opt/local/var/macports/logs/_Volumes_SharedFolders_Home_src_macports-ports_lang_rust/rust/main.log 
 54M	/opt/local/var/macports/logs/_Volumes_SharedFolders_Home_src_macports-ports_lang_rust/rust/main.log
macOS-Lion:~ catap$ 
```

catap:
@kencu thanks, I've reproduced the issue that you pointed on 10.7

evanmiller:
> rust is currently working great on my powerpc 32 and 64 bit systems, with llvm-13 and clang-13, running debian11. There is tooling available to cross-compile to darwin using linux. That is probably someone's only hope for getting rust on powerpc darwin. But as rust needs a functioning recent llvm and needs to be able to output into the darwin powerpc ABI, that project is likely doomed.

On the PPC front I identified some early post-1.0 Rust releases that are advertised to work with the LLVMs we have working on Darwin/PPC. So my hope was just to package up one of those releases as `rust-legacy` so we could have a hope of compiling small Rust projects on 10.4/10.5 (e.g. the Rust component of mozjs78). But I hit a wall with the bootstrap issue.

kencu:
Although there are llvms that build on PowerPC (I had up to llvm-5.0 building there) none of them work properly to lower to binary code that adheres to the Darwin PowerPC ABI.

So before you spend too much time with rust, you need to get something llvm-wise actually working right (not just building) on PowerPC.

Your best bet as a starting point (still broken, but the least broken) is Iain's llvm-7.0 fork for PowerPC. But even he fully acknowledges it is broken, and has not worked on it in more that two years now.

Buyer beware!

catap:
@kencu I've deep a bit more with thread local inside rust... Just a link: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/thread_local_dtor.rs#L57

I doubt that your way with emulation make any good here.

:(

kencu:
in clang I replaced tlv_atexit functions with cxa_atexit for darwin < 10.7. See line 38 of this patch:

https://github.com/macports/macports-ports/blob/master/lang/llvm-13/files/0019-10.6-and-less-use-emulated-TLS-before-10.7.patch

Such a hack could possibly be made to work in rust, maybe...if someone has the time and dedication to do that project.

catap:
@kencu this hack can be applied for sources, yes. Anyway, how to apply such hack for binary I haven't got any clue.

kencu:
you would need to write shim functions for the tlv_* functions, that use the emulated_tls functions to do the implementation, and put those into a dylib.

but as I mentioned above, I don't believe they map cleanly 1::1 and I never tried hard to do it...

catap:
I just rebased this PR to the last master and triple checked that I've included everything. I've tried to rebuild from scratch on clean system for macOS 10.8 and macOS 10.7 with no changes in macports tree. As prove that it works I runt `port install cargo` on both system. It required huge time.

As prove I'd love to attached two screenshots:
**macOS 10.7.5**
Screen Shot 2021-12-02 at 7 50 14 PM

**macOS 10.8.5**
Screen Shot 2021-12-02 at 7 50 21 PM

mascguy:
Awesome work Kirill!

catap:
@mascguy I feel that better to move away my hacks from PG. Do you agree?

mascguy:
> @mascguy I feel that better to move away my hacks from PG. Do you agree?

Do you mean for the changes to the `cargo` and `cargo-fetch` portgroups?

catap:
@mascguy yep, let me do it.

mascguy:
> @mascguy yep, let me do it.

I'm fine with whatever you think makes the most sense.

Does anyone else have a strong opinion either way?

catap:
@mascguy here it is. Let me keep it inside ports. Maybe I'll figure out how to build it without this hacks.

catap:
Lool. 1.57.0 was released. Let me take care of it :)

cjones051073:
Many thanks for the great  work here @catap , ita really nice to see macports getting the attention from someone likw you dedicated to keeping it going on older OSes.

i cannot reqlly test this myself, so if things like good once you rebase aginst the 1.57.0 update lets go with merging..

catap:
> Many thanks for the great work here @catap , ita really nice to see macports getting the attention from someone likw you dedicated to keeping it going on older OSes.

thanks! This was quite enjoyable plus I haven't done something like this for a while and digging inside old system and track down almost unique bugs inside well designed system was a good way to switch from crazy life :)

> i cannot reqlly test this myself, so if things like good once you rebase aginst the 1.57.0 update lets go with merging..

I'm rebuilding it on 10.7. It requires a few hours, as soon as it done I'll mark PR as ready for review.

kencu (commented):
no issues to mention.

catap:
Ok, it is built and seems to work on both systems => seems nice to have it as merged.

If someone is interesting I've started to bring it to 10.6: https://github.com/catap/macports-ports/commit/28415a90e15d7c6a3cf44cc71036830659d42afc

kencu:
So, just a couple of thing to mention.

Just changing the dylib from the system libcurl.4.dylib to whatever is currently  /opt/local/lib/libcurl.dylib  seems to work... but it is just incredibly fragile. It could easily break on any libcurl update at all -- the fact that it does work is pretty lucky. 

So if you have identified a libcurl that is newer but can be plugged in without breaking, you might do well to remember what it is, because you can expect it to stop working very shortly. Maybe. or maybe it will work for years, by the Zen of the Zen Gods -- you really can't say. But make a notation of the libcurl version that worked for this hack, for later use.

The missing function from compiler_rt is in every compiler_rt from clang-3.4 onwards, so just to be aware that there is nothing terribly special about the one from the macports-libcxx port. That is an easy one to locate -- it is a real PITA to find the clang function that prints out the full path to the compiler_rt that is in use by that clang (I found this, for another use). But it would be perhaps simpler to use some already-installed clang, maybe. OTOH, the macport-libcxx is a small, reliable build dep, so easy enough, and probably should never change it's version to any newer clang, by the way, at least not without a lot of checking things -- probably never changing it is safer -- so it's I guess as good a donor for compiler_rt as any... actually, are we stealing from compiler_rt? 

It might be simpler to just steal the "c" source file for the one missing function from compiler_rt (any version) and compile that one "c" file into a little dylib, because at least then you own it forever, but that could be pondered if this breaks someday.

Looking forward to rust working again on 10.7+. Upstream has currently committed to supporting 10.7+ in what I have read, so they really should finish getting their act together for this, so these kind of fragile hacks won't be needed.

kencu:
> If anyone is interesting I've started to bring it to 10.6: [catap@3933af0](https://github.com/catap/macports-ports/commit/3933af039150ba6c49e9a75224baeeb3e2191f00)

Great start!

catap:
> Just changing the dylib from the system libcurl.4.dylib to whatever is currently /opt/local/lib/libcurl.dylib seems to work... but it is just incredibly fragile. It could easily break on any libcurl update at all -- the fact that it does work is pretty lucky.
> 
> So if you have identified a libcurl that is newer but can be plugged in without breaking, you might do well to remember what it is, because you can expect it to stop working very shortly. Maybe. or maybe it will work for years, by the Zen of the Zen Gods -- you really can't say. But make a notation of the libcurl version that worked for this hack, for later use.

Yes, I agree with you and this is the second reason why I've switched build to offline mode. I've used libcurl only to start it. It haven't use any network to build rust right now (if offline works as it should).

=> seems safe enough.
 
> The missing function from compiler_rt is in every compiler_rt from clang-3.4 onwards, so just to be aware that there is nothing terribly special about the one from the macports-libcxx port. That is an easy one to locate -- it is a real PITA to find the clang function that prints out the full path to the compiler_rt that is in use by that clang (I found this, for another use). But it would be perhaps simpler to use some already-installed clang, maybe. OTOH, the macport-libcxx is a small, reliable build dep, so easy enough, and probably should never change it's version to any newer clang, by the way, at least not without a lot of checking things -- probably never changing it is safer -- so it's I guess as good a donor for compiler_rt as any... actually, are we stealing from compiler_rt?
> 
> It might be simpler to just steal the "c" source file for the one missing function from compiler_rt (any version) and compile that one "c" file into a little dylib, because at least then you own it forever, but that could be pondered if this breaks someday.

this function is for 128 bit numbers, and somewhere here the hell is started :)

> Looking forward to rust working again on 10.7+. Upstream has currently committed to supporting 10.7+ in what I have read, so they really should finish getting their act together for this, so these kind of fragile hacks won't be needed.

I guess they don't care about old platforms.

To be honest 10.8 isn't enough to run it. You need something new.

From another hand we may introduce (relatively easy) support of 10.6 by introducing a few rust-legacy ports which is used to bootstrap it.

As far as I recall rust 0.5 can be build by clang/gcc :)

The 1.5x port may use 1.4x (or that ever it really need to be build) as stage0, and 1.4x is uses (for example) 1.2, and so on until it can't be build from scratch by clang.

It allows to rebuild everything on the target platform with macport's libraries.

catap:
> > If anyone is interesting I've started to bring it to 10.6: [catap@3933af0](https://github.com/catap/macports-ports/commit/3933af039150ba6c49e9a75224baeeb3e2191f00)
> 
> Great start!

To be honest `dyld: unsupported exported symbol kind in` seems as dead end on this way.

catap:
> From another hand we may introduce (relatively easy) support of 10.6 by introducing a few rust-legacy ports which is used to bootstrap it.
> 
> As far as I recall rust 0.5 can be build by clang/gcc :)
> 
> The 1.5x port may use 1.4x (or that ever it really need to be build) as stage0, and 1.4x is uses (for example) 1.2, and so on until it can't be build from scratch by clang.
> 
> It allows to rebuild everything on the target platform with macport's libraries.

I was wrong. Rust 0.1 is also needed its snapshot to build.

catap:
Well, I was surprised with rust but it defently need to be build with formula `1.x by 1.(x - 1)`. I not sure for the old one, but the last one which I've checked (1.57, 1.56, 1.55 and 1.54) are hold.

mascguy:
So just to clarify, are we ready to merge these changes, for 10.7 and 10.8? Or are there additional changes desired?

herbygillot:
> Well, I was surprised with rust but it defently need to be build with formula `1.x by 1.(x - 1)`. I not sure for the old one, but the last one which I've checked (1.57, 1.56, 1.55 and 1.54) are hold.

That's correct.  In Rust's Portfile, the previous version is tracked by the  `rustc_version` variable: https://github.com/macports/macports-ports/blob/3cb5fcb16aaebfffaf60db6263d2371a01c1d7c1/lang/rust/Portfile#L43

...which specifies the Rust version that immediately precedes the current version.

Pre-built versions of the Rust compiler,standard library and `cargo` for `rustc_version` are downloaded and used to bootstrap and build the compiler & standard library for the current Rust version (specified in `version`: https://github.com/macports/macports-ports/blob/3cb5fcb16aaebfffaf60db6263d2371a01c1d7c1/lang/rust/Portfile#L12).

catap:
@herbygillot I was surprised that they have so often releases, and usually all minor releases broke compatibility.

herbygillot:
> @herbygillot I was surprised that they have so often releases, and usually all minor releases broke compatibility.

I think broken compatibility with new minor releases is becoming less the case in recent releases, but that may have been the case before.

catap:
> > @herbygillot I was surprised that they have so often releases, and usually all minor releases broke compatibility.
> 
> I think broken compatibility with new minor releases is becoming less the case in recent releases, but that may have been the case before.

I've tested 1.57, 1.56, 1.55 and 1.54 and all of them needs -1 version :) For example 1.56 can't be build by 1.55 and 1.54 can't be build by 1.52

herbygillot:
> > > @herbygillot I was surprised that they have so often releases, and usually all minor releases broke compatibility.
> > 
> > 
> > I think broken compatibility with new minor releases is becoming less the case in recent releases, but that may have been the case before.
> 
> I've tested 1.57, 1.56, 1.55 and 1.54 and all of them needs -1 version :) For example 1.56 can't be build by 1.55 and 1.54 can't be build by 1.52

OK I see.  Well for what you're doing here, I would say you can just use `rustc_version` - that should always have the version of Rust that should be used to bootstrap and build the current version.

catap:
@herbygillot and new enough cargo also, because it compalins on unsupported features.... :) and hell is begin.

herbygillot:
> @herbygillot and new enough cargo also, because it compalins on unsupported features.... :) and hell is begin.

😅

mascguy:
So what else is needed for 10.7 and 10.8 at this point?

catap:
@mascguy nothing, I can build `bandwhich` for example on macOS 10.7

mascguy:
> @mascguy nothing, I can build `bandwhich` for example on macOS 10.7

Okay great, so we are ready to merge then!

Folks, is everyone ready to call this iteration done? Any last minute thoughts, concerns, etc?

kencu:
great effort! No downsides visible to me.

mascguy:
> great effort! No downsides visible to me.

Agreed, awesome work Kirill!

@cjones051073 and @herbygillot, are you folks comfortable with merging this?

herbygillot:
No objections.

mascguy:
@cjones051073 approved per his earlier comment 22-ish hours ago, so I think we're good-to-go:

https://github.com/macports/macports-ports/pull/13214#issuecomment-985054390

catap:
Yahoo! Waiting that CI says.

PR #13235, opened 2021-12-03
mint: fix build for macOS 10.14 and 10.15

klanchman:
#### Description

Closes: https://trac.macports.org/ticket/64099

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/mint/Portfile
+++ b/devel/mint/Portfile
@@ -2,9 +2,10 @@
 
 PortSystem              1.0
 PortGroup               github 1.0
+PortGroup               xcodeversion 1.0
 
 github.setup            yonaskolb Mint 0.17.1
-revision                0
+revision                1
 github.tarball_from     archive
 
 name                    mint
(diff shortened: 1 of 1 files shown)


klanchman:
I'm not sure if `universal_variant no` is still appropriate / if an actual variant should be added? I'm also assuming there's a better way to pass the archs to Swift than what I've done here.

But I think this is on the right track since the 10.14 build on Azure worked.

mascguy (approved):

mascguy:
> I'm not sure if `universal_variant no` is still appropriate / if an actual variant should be added? I'm also assuming there's a better way to pass the archs to Swift than what I've done here.
> 
> But I think this is on the right track since the 10.14 build on Azure worked.

Agreed, this looks great Kyle! Enough so that we can merge as-is.

@jmroot/@ryandesign/@kencu, relative to the Universal topic: Is the current approach - whereby Universal is auto-enabled by default when available - reasonable? (Particularly relative to existing MacPorts precedent/practices?)

Kyle can certainly support a formal Universal variant, on supported macOS releases. But overall, it's easier to simply enable when supported.

Relative to practical concerns:
* This port builds within a minute or two, even for the Universal case.
* Similarly, the resultant binary is tiny regardless.

So those aren't a problem.

But are there any technical gotchas that we should be aware of - even if rare - such that users should have the option to choose whether a Universal binary is built?

ryandesign:
> Is the current approach - whereby Universal is auto-enabled by default when available - reasonable?

No. As I've said elsewhere in recent days, do not lie to MacPorts and the user about what architectures you're installing.

If it is exceedingly difficult to prevent a port from building universal, then build universal and force the universal variant on. The pango port shows an example of how to force a variant on (in that case, the quartz variant). But this does not seem to apply here.

Otherwise, if you can support building universal, do so, via the universal variant.

If the universal variant is not selected, do not build universal.

klanchman:
> If the universal variant is not selected, do not build universal.
 
Sounds good, I'll make another update to fix that 👍 

Thanks for pointing out the pango port, reading that answered a question I had about making variants conditionally available by something like Xcode version.

PR #13242, opened 2021-12-04
rust: build scalpel.dylib from sources 

catap:
#### Description

Closes: https://trac.macports.org/ticket/64127

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/lang/rust/Portfile
+++ b/lang/rust/Portfile
@@ -418,31 +418,22 @@ if {${subport} ne "${ccwrap}" } {
     }
 
     if { ${os.platform} eq "darwin" && ${os.major} < 12 } {
-        # I don't need whole libcxx
-        # anyway, it is an easy way to get some missed symbols from compiler-rt
-        legacysupport.use_mp_libcxx yes
-
         # well, let me build a small library which contains only needed symbols
         set scalpel_dylib ${workpath}/scalpel.dylib
+
         pre-build {
-            set export_list_file ${workpath}/scalpel.export_list
-            # set scalpel_dylib ${workpath}/scalpel.dylib
-            set export_list [open ${export_list_file} w]
-
             # Since https://github.com/rust-lang/rust/commit/8a6ff90a3a41e6ace18aeb089ea0a0eb3726dd08
-            puts ${export_list} "___muloti4"
+            set scalpel_sources "${filespath}/llvm-13-builtins/muloti4.c"
 
-            close ${export_list}
-
-            system "${configure.cc} -dynamiclib -exported_symbols_list ${export_list_file} -o ${scalpel_dylib} ${prefix}/lib/libcxx/libc++.a"
+            # let's build the scalpel
+            system "${configure.cc} ${configure.cflags} [get_canonical_archflags cc] -dynamiclib -o ${scalpel_dylib} ${scalpel_sources}"
 
             build.env-append \
                         DYLD_FORCE_FLAT_NAMESPACE=1 \
                         DYLD_INSERT_LIBRARIES=${scalpel_dylib}
-
         }
 
-        # inside destroot it also calls cargo / rustc
+        # inside destroot it also calls prebuilt cargo
         destroot.env-append \
                         DYLD_FORCE_FLAT_NAMESPACE=1 \
                         DYLD_INSERT_LIBRARIES=${scalpel_dylib}
--- a/lang/rust/files/llvm-13-builtins/int_endianness.h
+++ b/lang/rust/files/llvm-13-builtins/int_endianness.h
@@ -1,0 +1,123 @@
+/*
+ * Source: https://github.com/llvm/llvm-project/blob/llvmorg-13.0.0/compiler-rt/lib/builtins/int_endianness.h
+ *
+ * The goal of this code is only allow to run binary
+ * which was linked against macOS 10.8 on macOS 10.7 and below.
+ *
+ * Such binaries are used to boostrap rust on stage0 only.
+ */
+
+//===-- int_endianness.h - configuration header for compiler-rt -----------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file is a configuration header for compiler-rt.
+// This file is not part of the interface of this library.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef INT_ENDIANNESS_H
+#define INT_ENDIANNESS_H
+
+#if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) &&                \
+    defined(__ORDER_LITTLE_ENDIAN__)
+
+// Clang and GCC provide built-in endianness definitions.
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+#define _YUGA_LITTLE_ENDIAN 0
+#define _YUGA_BIG_ENDIAN 1
+#elif __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#define _YUGA_LITTLE_ENDIAN 1
+#define _YUGA_BIG_ENDIAN 0
+#endif // __BYTE_ORDER__
+
+#else // Compilers other than Clang or GCC.
+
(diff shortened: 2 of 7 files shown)


catap:
@ryandesign here it is

ryandesign:
I haven't tried to build this, and CI won't help us since CI doesn't use OS versions that old, but i see you only added `${configure.cflags}` which doesn't ordinarily include the `-arch` flags. Usually you also have to add `[get_canonical_archflags cc]` to get those. If you have an older OS system that you're testing these builds on, you can check your logfile to see whether just adding `${configure.cflags}` resulted in the `-arch` flags getting added or not.

I also see a typo in the commit message: clfags => cflags

catap:
@ryandesign thanks, this makes things better I think.

kencu:
I have no idea what the cflags and the archflags might do if you are trying to reexport some of the symbols from a static lib into a dylib.

I guess the cflags will all be just ignored, as they would not appear to be able to do anything here as nothing is being built.

The archflags? I actually don't know. Perhaps you might just re-export one arch of a multi-arch static lib? Or generate an error if you try to re-export an arch that is not in the static lib? Hard to say.

Have to do some experimentation to see what might happen with this.

To get around all this, you could crib out the "c" file that builds the symbol needed from the compiler_rt upstream code, throw that into the ${filesdir}, and build that "c" file into the dylib with the needed symbol.

Then this all makes some sense, the cflags and arch would do the right things, and you no longer need a dependency on the libc++.a provider any more.

Probably easier, too, I would think, in the end.

catap:
@kencu good point. Special that for macOS 10.6 I've already started to use some code. Let me try to get some sources from LLVM-13 and see can I build it on 10.7.

kencu:
llvm-13 builds on 10.6

so does the rest of clang-13 etc (flang, etc) with a simple patch I added to a clang-13 ticket in trac, but you don't need those for rust...

catap:
@kencu here how it works right now :) I'm using my library only for building, and after that I don't need it.

I'm testing another way way without touching static library right now and as soon as it built and I confirm that it works (build cargo by this rust) => I'll update PR and it is good to review. ETA a few hours I guess.

mascguy (approved):

catap:
@mascguy / @kencu / @ryandesign feel free to review it.

mascguy (approved):

catap:
```
macOS-Lion:~ catap$ otool -L $(which rustc)
/opt/local/bin/rustc:
	@rpath/librustc_driver-7b2d2200d13688fe.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libstd-c6362a5f9528007e.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 159.1.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 28.4.0)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 46.1.0)
macOS-Lion:~ catap$ otool -L $(which cargo)
/opt/local/bin/cargo:
	/opt/local/lib/libMacportsLegacySupport.dylib (compatibility version 1.0.0, current version 1.0.4)
	/opt/local/lib/libgit2.1.3.dylib (compatibility version 1.3.0, current version 1.3.0)
	/opt/local/lib/libcurl.4.dylib (compatibility version 12.0.0, current version 12.0.0)
	/opt/local/libexec/openssl3/lib/libssl.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/opt/local/libexec/openssl3/lib/libcrypto.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 635.21.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 159.1.0)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 46.1.0)
macOS-Lion:~ catap$ 
```

quite nice!

kencu:
Yep, that should do it!

kencu:
@catap -- really good effort here with these rust issues. Well done, and appreciate the time you spent on this.

catap:
@kencu thanks! I glade that I can contribute to macports and have a lot of fun :)

catap:
@kencu meanwhile, if you have time and motivation, you may help me with macOS 10.6 support :)

Current code here: https://github.com/catap/macports-ports/commit/096d63253149787f0520f99a6faa53c166f70051

And it is failed deep inside.

kencu:
The error you see looks like either the function has address 0, or it's trying to access the variable and the variable is at address 0. But the variable has value 0, not address 0, and there is no function with that name. It is like the variable has not actually been allocated yet.

And there is some of the tricky stuff -- dyld (I believe) examines the headers of the code fragments and these code fragments are labelled (I believe) as using thread_local_variables if they do use them. And (I believe) when dyld sees that, it then allocates memory for the thread_local_variables as it loads the code fragments. And then tears down these variables when the code fragment goes out of scope.

Somebody who really knows (Jeremy or Iain are the only two people I know of who might know this for sure) might say everything I just told you is hogwash. But that is my understanding of how it works, at present.

kencu:
Here's an example of the kind of information that is available that gives me clues to the thoughts above:

https://github.com/golang/go/issues/17200

clearly, incomplete.

catap:
@kencu well, my way allows me to run on macOS 10.6 a code:
```rust
use std::cell::RefCell;
use std::thread;

thread_local! {
    static FOO: RefCell<f32> = RefCell::new(1.0);
}

fn main() {
    FOO.with(|foo| {
        println!("main: {}", *foo.borrow());
        *foo.borrow_mut() = 3.0;
        println!("main: {}", *foo.borrow());
    });

    thread::spawn(move|| {
        FOO.with(|foo| {
            println!("inner: {}", *foo.borrow());
            *foo.borrow_mut() = 4.0;
            println!("inner: {}", *foo.borrow());
        });
    }).join().unwrap();

    FOO.with(|foo| {
        println!("main: {}", *foo.borrow());
    });
}
```
as
```
macOS-Snow-Leopard:work catap$ env DYLD_FORCE_FLAT_NAMESPACE=1 DYLD_INSERT_LIBRARIES=scalpel.dylib /Volumes/SharedFolders/Home/src/rust/test
main: 1
main: 3
inner: 1
inner: 4
main: 1
macOS-Snow-Leopard:work catap$ 
```

anyway, rust distributive contains a few `__thread_ptrs` symbol which doesn't supported by macOS 10.6 :) I feel that this is the root cause of my issue.

kencu:
thread_local is already 100% supported on 10.6.8 using any gcc version > 4.8 or so, and using all our clang versions >= clang-5.0 ( in c++ if you have upgraded your libcxx to the thread_local-supporting version, as per the port notes).

However, if you compile something on 10.7, and you build it against a deployment target of 10.7, then clang will use the tlv_* functions instead. These do not exist on 10.6.8 at present. Whether having a library that implemented these 10.7 tlv_* functions would allow a 10.7 binary to be forced to run on 10.6.8 is a long shot, but it might work.

There are other constraints that could be put on the binary that would make it unrunnable on 10.6.8 beyond the missing tlv_* functions, but it might be possible, in some isolated cases, to force something to work.

It might also be possible to use some bootstrapped rust on 10.7 to build a version of rust using a modified llvm backend that would then use the emulated_tls functions we use in our llvm to make a rust that is compiled to run on 10.6 using emulated-thread_local_storage. There is nothing to much that stops clang and llvm on 10.7 from building such a Frankenstein monster. You'd have to fiddle with the libc++ a bit to make it like ours (our libc++abi has thread_local support in, specifically cxa_thread_atexit is compiled in).

kencu:
Oh, I see what you are up to -- that is a rust executable you built on 10.7 and moved to 10.6 that works with thread support. Yeppers, I get cha. On the path, then!

catap:
@kencu the challenge is run existed binary :) I've already achived that I can produce rust code on 10.7+ which is used thread local and which runs on 10.6.

Right now I'm trying to run the rust compiler itself.

catap:
@mascguy any chance to merge it? :)

mascguy:
@kencu Are you good with merging?

catap:
@mascguy I guess he confirm it before, and the last messages was about the next PR that I'm working on ;)

PR #13244, opened 2021-12-04
gnuradio: use python 3.9 as default

ra1nb0w:
#### Description

use python 3.9 as default

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.14.6 18G9323 x86_64
Xcode 10.3 10G8

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/gnuradio-1.0.tcl
+++ b/_resources/port1.0/group/gnuradio-1.0.tcl
@@ -1,0 +1,188 @@
+# -*- coding: utf-8; mode: tcl; c-basic-offset: 4; indent-tabs-mode: nil; tab-width: 4; truncate-lines: t -*- vim:fenc=utf-8:et:sw=4:ts=4:sts=4
+#
+# This PortGroup helps to maintain gnuradio dependencies coherent between
+# gnuradio and gnuradio modules.
+#
+# Usage:
+#
+# PortGroup           gnuradio 1.0
+#
+
+PortGroup           cmake 1.1
+PortGroup           boost 1.0
+
+categories          science comms
+platforms           darwin macosx
+
+# declare which gnuradio are you using
+options gnuradio.type
+default gnuradio.type "gnuradio"
+
+options gnuradio.python_version
+options gnuradio.default_python_variant
+
+if {[string first "37" $subport] > 1} {
+    default gnuradio.python_versions { 2.7 }
+    default gnuradio.default_python_variant +python27
+} else {
+    # the following versions use all the same python
+    default gnuradio.python_versions { 3.8 3.9 3.10 }
+    default gnuradio.default_python_variant +python39
+}
+
+# use C/C++11
+compiler.c_standard   2011
+compiler.cxx_standard 2011
+boost.version 1.71
+
+# see https://github.com/macports/macports-ports/pull/7805
+license_noconflict-append openssl
+
+# see https://github.com/macports/macports-ports/pull/7805
+license_noconflict-append graphviz
+
+# Define the available variants
+foreach py_ver ${gnuradio.python_versions} {
+    set py_ver_no_dot [join [split ${py_ver} "."] ""]
+    set variant_line {variant python${py_ver_no_dot} description "Build with python ${py_ver} support"}
+    foreach py_over ${gnuradio.python_versions} {
+        if { ${py_ver} == ${py_over} } { continue }
+        set py_over_no_dot [join [split ${py_over} "."] ""]
+        append variant_line " conflicts python${py_over_no_dot}"
+    }
+    append variant_line { { } }
+    eval $variant_line
+    if {[variant_isset python${py_ver_no_dot}]} {
+        if {${gnuradio.default_python_variant} != "+python${py_ver_no_dot}"} {
+            set gnuradio.default_python_variant ""
+        }
+    }
+}
+
+# set default python variant if not selected
+if {${gnuradio.default_python_variant} != ""} {
+    default_variants-append "${gnuradio.default_python_variant}"
+}
+
+# If a python variant is enabled, activate it
+set active_python_version ""
+set active_python_version_no_dot ""
+foreach py_ver ${gnuradio.python_versions} {
+    set py_ver_no_dot [join [split ${py_ver} "."] ""]
+    if {[variant_isset python${py_ver_no_dot}]} {
+        set active_python_version        ${py_ver}
+        set active_python_version_no_dot ${py_ver_no_dot}
+    }
+}
+
+depends_build-append \
+    port:pkgconfig
+
+depends_lib-append \
+    port:fftw-3-single \
+    path:lib/libmpir.dylib:mpir \
+    path:lib/libvolk.dylib:volk \
+    port:gmp \
+    port:python${active_python_version_no_dot}
+
+# need matplotlib for polar encoder/decoder, runtime only
+depends_run-append \
+    port:py${active_python_version_no_dot}-matplotlib \
+    port:py${active_python_version_no_dot}-opengl \
+    port:py${active_python_version_no_dot}-scipy \
+    port:py${active_python_version_no_dot}-numpy \
+    port:py${active_python_version_no_dot}-cheetah
+
+if {[string first "37" $subport] > 1} {
+    # still require cppunit for testing; NOTE: cppunit is checked for
+    # during configure, so we need it to be in depends_lib or
+    # depends_build to be used correctly. Choose the latter since it's
+    # not required for runtime; just for build/test.
+    depends_build-append \
+        port:cppunit
+} else {
+    # add dependencies for gnuradio >= 3.8
+}
+
+if {[string first "gr37-" $subport] >= 0} {
+    depends_lib-append \
+        port:gnuradio37
+
+    depends_build-append \
+        port:swig3-python
+
+    configure.args-append \
+        -DSWIG_EXECUTABLE=${prefix}/bin/swig3
+} elseif {[string first "gr-" $subport] >= 0} {
+    depends_lib-append \
+        path:lib/libgnuradio-runtime.dylib:gnuradio
+
+    depends_build-append \
+        port:swig-python
+
+    configure.args-append \
+        -DSWIG_EXECUTABLE=${prefix}/bin/swig
+}
+
+# set build type to release or debug, not "MacPorts" since the GR
+# CMake system checks for a valid value and this is not one of them
+if {[variant_isset debug]} {
+    cmake.build_type Debug
+} else {
+    cmake.build_type Release
+}
+
+# specify the Python version to use
+set python_framework_dir ${frameworks_dir}/Python.framework/Versions/${active_python_version}
+configure.args-append \
+    -DPYTHON_EXECUTABLE=${python_framework_dir}/bin/python${active_python_version} \
+    -DPYTHON_INCLUDE_DIR=${python_framework_dir}/Headers \
+    -DPYTHON_LIBRARY=${python_framework_dir}/Python \
+    -DGR_PYTHON_DIR=${python_framework_dir}/lib/python${active_python_version}/site-packages \
+    -DDOXYGEN_DOT_EXECUTABLE=${prefix}/bin/dot \
+    -DDOXYGEN_EXECUTABLE=${prefix}/bin/doxygen \
+    -DCMAKE_MODULES_DIR=${prefix}/share/cmake \
+    -DGMP_INCLUDE_DIR= \
+    -DGMP_LIBRARY= \
+    -DGMPXX_LIBRARY= \
+    -DMPIR_INCLUDE_DIR=${prefix}/include \
+    -DMPIR_LIBRARY=${prefix}/lib/libmpir.dylib \
+    -DMPIRXX_LIBRARY=${prefix}/lib/libmpirxx.dylib \
+    -DENABLE_DOXYGEN=OFF \
+    -DENABLE_SPHINX=OFF
+
+# remove top-level library and include paths, such that internal ones
+# are searched before any already-installed ones.
+configure.cppflags-delete -I${prefix}/include
+configure.ldflags-delete  -L${prefix}/lib
+
+# TODO check "macports-libstdc++" or "libstdc++" should not needed anymore, right?
+# disable C/CXX standard extensions (e.g., "gnu++11" and "gnu11"
+# rather than "c++11" and "c11"). GNU Radio itself does not
+# require these extensions, and when compiling with Clang they
+# don't really do anything anyway; so, just disable them.
+configure.args-append \
+    -DCMAKE_CXX_EXTENSIONS=OFF \
+    -DCMAKE_C_EXTENSIONS=OFF
+
+variant docs description "Install documentation" {
+    # texlive-latex not needed from 3.9.0.0
+    depends_build-append \
+        port:doxygen \
+        path:bin/dot:graphviz \
+        port:py${active_python_version_no_dot}-sphinx \
+        port:texlive-latex
+
+    configure.args-delete \
+        -DENABLE_DOXYGEN=OFF \
+        -DENABLE_SPHINX=OFF \
+        -DSPHINX_EXECUTABLE=
+
+    configure.args-append \
+        -DENABLE_DOXYGEN=ON \
+        -DENABLE_SPHINX=ON \
+        -DSPHINX_EXECUTABLE=${python_framework_dir}/bin/sphinx-build
+}
+
+default_variants-append \
+    +docs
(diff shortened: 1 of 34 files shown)


ra1nb0w:
what I don't understand is how we can force to move to the newer python version  during the upgrade.
I mean, if you have already installed gnuradio with python 3.7 and then you upgrade you still have gnuradio using python 3.7.

ra1nb0w:
I also notice that bladeRF is not building but I don't understand why.

mf2k:
> what I don't understand is how we can force to move to the newer python version during the upgrade. I mean, if you have already installed gnuradio with python 3.7 and then you upgrade you still have gnuradio using python 3.7.

There is no way to do that. It is by design that port maintains the currently selected set of variants when upgrading any port. Anyone can, however, use the `--enforce-variants` flag when upgrading. But that will normally fail in cases like this since the `+python37` variant conflicts with the `+python39` variant. So the user has two options in this case:

`sudo port upgrade --enforce-variants gnuradio -python37`

Or just manually reinstall it since the default variants will get used.

`sudo port install gnuradio`

If you want to actually force the upgrade, then remove all of the variants for python and always use python39.

ra1nb0w:
Thinking about this after this post I ended at the same conclusion. I mean, if I remove all other python variants the upgrade automatically pick up the new one. thank you very much @mf2k to clarify this aspect.

michaelld (approved):
I have no objections to moving to Python 3.9 otherwise, as it seems to work now across both GNU Radio and macOS versions.

In line with the discussion: I don't think the rev-bump is necessary for any of these changes, since it won't impact the Python version selected via upgrading. The rev-bump won't hurt either, so I'm approving and you can decide which way to go @ra1nb0w .

ra1nb0w:
I simplified. Permit only python 3.9. In the future, this permit to clear a lot of code.
What do you think?

science/PothosLiquidDSP/Portfile
```diff
@@ -19,37 +19,37 @@ version             0.0.0-git-20201204
...
 compiler.cxx_standard 2011
 
 # set Python version and variants
 set PythonVersionNoDot ""
 
 # select Python to use; default to 3.6
-variant python35 conflicts python36 python37 description {Build using Python 3.5} {}
-variant python36 conflicts python35 python37 description {Build using Python 3.6} {}
 variant python37 conflicts python35 python36 description {Build using Python 3.7} {}
+variant python38 conflicts python35 python36 description {Build using Python 3.8} {}
```
michaelld, line 31: Need to fix up conflicts with new variants

michaelld (changes requested):
I'm thinking about the reduction to just Py39 ... in the meantime here's something to fix

science/PothosLiquidDSP/Portfile
ra1nb0w: Oh right ;-) It was a propose. If you accept I will remove all variants since are useless (this simplify a lot the port).

cjones051073:
> > what I don't understand is how we can force to move to the newer python version during the upgrade. I mean, if you have already installed gnuradio with python 3.7 and then you upgrade you still have gnuradio using python 3.7.
> 
> 
> 
> There is no way to do that. It is by design that port maintains the currently selected set of variants when upgrading any port. Anyone can, however, use the `--enforce-variants` flag when upgrading. But that will normally fail in cases like this since the `+python37` variant conflicts with the `+python39` variant. So the user has two options in this case:
> 
> 
> 
> `sudo port upgrade --enforce-variants gnuradio -python37`
> 
> 
> 
> Or just manually reinstall it since the default variants will get used.
> 
> 
> 
> `sudo port install gnuradio`
> 
> 
> 
> If you want to actually force the upgrade, then remove all of the variants for python and always use python39. 
> 
> 

Actually, I think base now remembers if a variant was selected just because it was activated by default, or if the user explicitly requested it. So if the default moves from python 2.7 to 3.9 I think base will now switch the the new defaults automatically when the new port is installed.

ra1nb0w:
Correct @cjones051073

science/PothosLiquidDSP/Portfile
michaelld: hmmm ... I'm very mixed about whether to remove `+python*` variants or not. I would prefer to keep variants around, since end-users will want to select a specific Python version & just use that, which may or not be the one we've selected. Different ports will require different Python versions, which are guaranteed to not overlap with that we've selected. By providing `+python*` variants we can keep these various ports with the same Python version. So ... I guess I've talked myself into wanting to keep some variants, just change the default to Py39. Does this make sense to you? Please provide logic if you feel otherwise!

ra1nb0w: My considerations were:
- any gr module that works with python 3.7 it works fine with 3.9 (or at least ~99.9999%) and the performance are better (since also debian stable moved to 3.9);
- simplify a lot the ports without creating a new portgroup to manage the complexity and future changes;
- I don't think many users are selecting different version since point 1.
- many many python ports have moved to 3.9
- obviously, when we select a python version we should maintain it stables for a few years (at least)
- all gr-* modules, gnuradio and related packages will use the same python version (I already using 3.9 for everything)

At this point, I don't catch the real necessity of having multiple python versions except for performance comparisons.

michaelld: Very good arguments!

michaelld (changes requested):
Please go ahead & remove Python variants and move all to just Python39. If we get pushback from users with solid rationale then we can always move back to variants. I like the idea of forcing the use of the same Boost and Python for all GR ports, and Py39 is a reasonable selection.

cjones051073:
Whether or not providing varinats makes sense depends on whether or not the choice of python used is something that impacts end users. I.e. does gnuradio provide a python module that users can import into their preferred python session ? If it does then providing variants is correct, as you need to accommodate different choices. This does not mean you have to provide a lot a variants, but limiting things to just 3.9 is not acceptable really, given e.g. 3.10 is now latest release (and personally the one I prefer to use as far as possible). 

If python is just used as a build dep., then a single version is fine. Otherwise you should provide options.

cjones051073:
> Whether or not providing variants makes sense depends on whether or not the choice of python used is something that impacts end users. I.e. does gnuradio provide a python module that users can import into their preferred python session ? If it does then providing variants is correct, as you need to accommodate different choices. This does not mean you have to provide a lot a variants, but limiting things to just 3.9 is not acceptable really, given e.g. 3.10 is now latest release (and personally the one I prefer to use as far as possible).
> 
> If python is just used as a build dep., then a single version is fine. Otherwise you should provide options.

The above, btw. Is another way of say by all means keep. 3.9 as the default. But please also add 3.10 as a new option.

michaelld:
> > Whether or not providing variants makes sense depends on whether or not the choice of python used is something that impacts end users. I.e. does gnuradio provide a python module that users can import into their preferred python session ? If it does then providing variants is correct, as you need to accommodate different choices. This does not mean you have to provide a lot a variants, but limiting things to just 3.9 is not acceptable really, given e.g. 3.10 is now latest release (and personally the one I prefer to use as far as possible).
> > If python is just used as a build dep., then a single version is fine. Otherwise you should provide options.
> 
> The above, btw. Is another way of say by all means keep. 3.9 as the default. But please also add 3.10 as a new option.

Broadly speaking GR does provide Python importable modules. These are not required to use GR, but most users want them because the Python interface is robust and easier to use than C++.

So ... I guess @ra1nb0w ignore what I wrote about removing variants & let's move to, say, Py38, Py39, and Py310. We need to enforce whatever the GR Python variant is in all GR OOT ports too. That will take some time but it's cut-and-paste once working. Maybe create a PortGroup to do this?

ra1nb0w:
Now, I am confused ;-)
I still don't understand why a user (advanced that it is) want to use another python version except benchmarking.
From my point of view, scientific software should be provided with the best and stable environment possible no matter what.
But, as I said at the beginning and in the past, I can create a portgroup with all customization to simplify all ports.

ra1nb0w:
Just a side note: python 3.9 should fix arm64 support

cjones051073:
> Now, I am confused ;-)
> I still don't understand why a user (advanced that it is) want to use another python version except benchmarking.
> From my point of view, scientific software should be provided with the best and stable environment possible no matter what.
> But, as I said at the beginning and in the past, I can create a portgroup with all customization to simplify all ports.

GR should not force a user to use e.g. 3.9 if they have themselves decided they wish to use a different python version for their person usage. I for instance use 3.10 as my primary installation, and have a number of modules installed for it. What if I want to use them *and* GR at the same time ? I will need a single python session to import them into. 

If a port provides modules that can be imported into a specific python session, then that port should provide options for which python version the user can use. This is exactly why we have the python PG, to handle creating these subports. GR is not, and should not, think itself any different in this regard.

This is not to say you have to provide variants all the way back to 3.5 (for me that is too old now), but a 'reasonable' set. Only providing 3.9 is not acceptable, particularly when that is not even the most recent production python version at this time.

cjones051073:
In my option the following should be done

1. restore variants back to some reasonable cut off. say python 3.7
2. make 3.9 the default variant
3. add 3.10 as an additional possible variant.

michaelld:
@ra1nb0w what @cjones051073 writes is the MP way of doing things, and hence I strongly recommend we follow this:

> In my opinion the following should be done :-
> 
> 1. restore variants back to some reasonable cut off. say python 3.7
> 2. make 3.9 the default variant
> 3. add 3.10 as an additional possible variant.

4. enforce same Python variant for all GR OOT and any ports that use GR. [Same for Boost]

michaelld:
I'm fine with, and recommend, the oldest Python variant being 3.8 on Intel and 3.9 on ARM64 .. or, really, just go with 3.9 for both since that makes the code easier

cjones051073:
> I'm fine with, and recommend, the oldest Python variant being 3.8 on Intel and 3.9 on ARM64 .. or, really, just go with 3.9 for both since that makes the code easier

simply because 3.7 was the previous default variant, I think keeping it for now (at least on intel) to ease the transition is the way to go. You can then remove it at some reasonable time in the future (6,12 months), giving users a reason time to  migrate. but wiping it out right now is perhaps not that friendly...

_resources/port1.0/group/gnuradio-1.0.tcl
```diff
@@ -0,0 +1,201 @@
...
+        set active_python_version        ${py_ver}
+        set active_python_version_no_dot ${py_ver_no_dot}
+    }
+}
+
+depends_build-append \
+    port:pkgconfig
+
+depends_lib-append \
+    port:fftw-3-single \
```
michaelld, line 82: I think these next 4 dependencies (fftw though gmp) belong in the GR Portfile, not here, since they are not common to all GR OOT modules.

_resources/port1.0/group/gnuradio-1.0.tcl
```diff
@@ -0,0 +1,201 @@
...
+    path:lib/libmpir.dylib:mpir \
+    path:lib/libvolk.dylib:volk \
+    port:gmp \
+    port:python${active_python_version_no_dot}
+
+# need matplotlib for polar encoder/decoder, runtime only
+depends_run-append \
+    port:py${active_python_version_no_dot}-matplotlib \
```
michaelld, line 90: I think these dependencies belong in the GR Portfile, not here, since they are not common to all GR OOT modules.

_resources/port1.0/group/gnuradio-1.0.tcl
```diff
@@ -0,0 +1,201 @@
...
+    # still require cppunit for testing; NOTE: cppunit is checked for
+    # during configure, so we need it to be in depends_lib or
+    # depends_build to be used correctly. Choose the latter since it's
+    # not required for runtime; just for build/test.
+    depends_build-append \
+        port:cppunit
+} else {
+    # gr-vocoder removed in-tree libgsm, libcodec2, use system-wide libs
+    depends_lib-append \
+        port:codec2 \
```
michaelld, line 106: I think these dependencies (all in this `else` clause) belong in the GR Portfile, not here, since they are not common to all GR OOT modules.

_resources/port1.0/group/gnuradio-1.0.tcl
```diff
@@ -0,0 +1,201 @@
...
+set python_framework_dir ${frameworks_dir}/Python.framework/Versions/${active_python_version}
+configure.args-append \
+    -DPYTHON_EXECUTABLE=${python_framework_dir}/bin/python${active_python_version} \
+    -DPYTHON_INCLUDE_DIR=${python_framework_dir}/Headers \
+    -DPYTHON_LIBRARY=${python_framework_dir}/Python \
+    -DGR_PYTHON_DIR=${python_framework_dir}/lib/python${active_python_version}/site-packages \
+    -DDOXYGEN_DOT_EXECUTABLE=${prefix}/bin/dot \
+    -DDOXYGEN_EXECUTABLE=${prefix}/bin/doxygen \
+    -DCMAKE_MODULES_DIR=${prefix}/share/cmake \
+    -DGMP_INCLUDE_DIR= \
```
michaelld, line 154: Technically I think these CMake defines  (GMS and MPIR) belong in the GR Portfile, not here, since they are not common to all GR OOT modules. That said these won't hurt either since if CMake doesn't need them then they will just be ignored.

_resources/port1.0/group/gnuradio-1.0.tcl
```diff
@@ -0,0 +1,201 @@
...
+    -DENABLE_SPHINX=OFF
+
+# remove top-level library and include paths, such that internal ones
+# are searched before any already-installed ones.
+configure.cppflags-delete -I${prefix}/include
+configure.ldflags-delete  -L${prefix}/lib
+
+# disable using internal VOLK
+configure.args-append \
+    -DENABLE_INTERNAL_VOLK=OFF
```
michaelld, line 170: This is just used by GR itself, and thus should be in the GR Portfile

michaelld (changes requested):
Wow! Excellent work @ra1nb0w ! I think you went a touch overboard on moving dependencies from GR Portfile into the new GR PG -- which to be honest isn't the end of the world but also not the way MP does things. So let's get those moved back & then merge this & see how it does in the wild

_resources/port1.0/group/gnuradio-1.0.tcl
ra1nb0w: Yes I was quite lazy with dependencies since:
- I see that sometimes few of those (like ffwt, mpir and volk) was declared also in gr modules
- since are lib they don't create too much hassle

therefore, I prefer to leave as is since it doesn't create false dependency installation.

_resources/port1.0/group/gnuradio-1.0.tcl
ra1nb0w: Same consideration as before: scipy, numpy are used elsewhere.

_resources/port1.0/group/gnuradio-1.0.tcl
ra1nb0w: right. moved back. thank you

_resources/port1.0/group/gnuradio-1.0.tcl
ra1nb0w: right. back to gnuradio.

_resources/port1.0/group/gnuradio-1.0.tcl
ra1nb0w: I use the same considerations said before. Stay here. It generalizes the portgroup for every necessity.

ra1nb0w:
ping

ra1nb0w:
ok. I wait a bit. We can also move dependencies later on...

michaelld:
Sorry for the delay. Been away from work / on vacation. I'll get to this in the next few days.

ra1nb0w:
gnuradio updated to 3.8.5.0

michaelld (approved):
Thank you for the PR, as well as getting this updated with the just released GR 3.8.5.0. It's a huge amount of changes to lots of ports, and so this is difficult to test thoroughly! There's a lot of similar changes, and they read well, so I'm going ahead with the merge & we'll fix up issues afterwards. At least this way the PR won't go stale & we can start updating GR OOT ports as well as try to figure out support for GR39 and the forthcoming GR310.

ra1nb0w:
ok. thank you

ra1nb0w:
A few questions:
- I introduced `quartz` and `x11` variants to choose from gnuradio which graphical backend to use. This, however, create a fixed binary package only for x11 therefore if the user is using quartz he needs to build gnuradio from source. Removing these two variants fix the problem. Can be dangerous?
- What do you think to jump directly to gnuradio 3.10? It seems to have the same requirements of 3.9 for the modules?

have a great day

PR #13249, opened 2021-12-05
openvdb: update to 8.0.0

jasonliu--:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/graphics/openvdb/Portfile
+++ b/graphics/openvdb/Portfile
@@ -7,21 +7,24 @@ PortGroup               active_variants 1.1
 PortGroup               compiler_blacklist_versions 1.0
 PortGroup               boost 1.0
 
-github.setup            AcademySoftwareFoundation openvdb 7.2.3 v
-revision                1
-homepage                https://www.openvdb.org/
+github.setup            AcademySoftwareFoundation openvdb 8.0.0 v
+revision                0
+
 categories              graphics
+platforms               darwin
 license                 {MPL-2 LGPL-2.1+}
-maintainers             {mcalhoun @MarcusCalhoun-Lopez} openmaintainer
-description             sparse volume data structure and tools
+maintainers             {mcalhoun @MarcusCalhoun-Lopez} {@jasonliu--} openmaintainer
+
+description             sparse volumetric data structure and tools
 long_description        OpenVDB is an open source C++ library comprising a novel hierarchical data structure \
                         and a large suite of tools for the efficient storage and manipulation of \
                         sparse volumetric data discretized on three-dimensional grids.
-platforms               darwin
 
-checksums               rmd160  b85de208ed29ab1675956f78fd867eeaaecc591d \
-                        sha256  1e929a7614e7b47d9b51c56c5e76c89cb3fef572d6a6affb89b45af2c81a297c \
-                        size    2289366
+homepage                https://www.openvdb.org/
+
+checksums               rmd160  68e677f58b7fea6dc4bd8e6d55c531205195625c \
+                        sha256  28a811bae32b62a2080aad4776c91d4453a7d52e2edbe1ab4a21783ec4aaa6bb \
+                        size    2287123
 
 compiler.cxx_standard   2014
 # error: cannot refer to variable template 'Contains' without a template argument list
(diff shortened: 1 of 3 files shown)


graphics/openvdb/Portfile
```diff
@@ -7,28 +7,45 @@ PortGroup               active_variants 1.1
...
 
+depends_build-append    port:pkgconfig
+
+depends_lib-append      port:zlib \
+                        port:blosc \
+                        port:tbb \
+                        path:lib/pkgconfig/glfw3.pc:glfw \
+                        port:ilmbase \
+                        port:openexr
+
```
mascguy, line 48: Nitpick: Might be better to leave the patchfiles where they were before, prior to the dependencies, as the change makes it more difficult to discern the tangible changes in this PR.

jasonliu--: I tend to like to place the patchfiles in line with the MacPorts phases, i.e. `pre-patch`, `patchfiles`, `post-patch`, `pre-configure`, etc., but I see your point, relocating the patchfiles does cause the diff to become quite messy. So I've restored their original location, as suggested.

mascguy (approved):

mascguy:
LGTM. @MarcusCalhoun-Lopez ?

graphics/openvdb/Portfile
```diff
@@ -61,28 +70,30 @@ configure.args-append   -DUSE_CCACHE=OFF \
...
 
-foreach s ${pythons_suffixes} {
+foreach s $pythons_suffixes {
     set p python${s}
     set v [string index ${s} 0].[string index ${s} 1]
-    set i [lsearch -exact ${pythons_ports} ${p}]
-    set c [lreplace ${pythons_ports} ${i} ${i}]
-    variant ${p} description "Build the Python ${v} bindings" conflicts {*}${c} "
+    set i [lsearch -exact $pythons_ports ${p}]
+    set c [lreplace $pythons_ports ${i} ${i}]
```
mascguy, line 82: Hey Jason, this is another nitpicky-esque comment. But again, it comes down to tangible changes, vs. those that are aren't:

From a consistency perspective, why change the variable referencing convention from `${varname}`, to `$varname`? The former is used everywhere else in this port. Was this just a simple oversight when committing these changes...?

mascguy (changes requested):
Revert changes to variable references, to ensure all are surrounded by curly braces.

graphics/openvdb/Portfile
jasonliu--: This is purely a stylistic preference on my part, but I prefer to keep variables that have been declared elsewhere, like in PortGroups or global variables from base, as `${varname}`, and variables that have been declared by me in "this" Portfile, as `$varname` (obviously the exception is when the brackets are required, e.g. `${varname}.fileext`).

It lets me see at a glance whether the variable is local or not. Since it looks like I'm going to be taking over as maintainer of the port, I thought that I would take the opportunity to format things in a way that makes it easier for me to read. Also, it cuts down on file size, although this benefit is rather negligible.

jasonliu--: Oh, and as far as why I _do_ leave the single-letter variables as `${varname}`, it's because I'm a bit ambivalent about those. I have been bitten before where I have left the variables as `$i`, but then ended up doing something like `$iotherstuff` (when it should have been `${i}otherstuff`) and confused myself. 😝 

I might end up deciding to get rid of the brackets around the single-letter variables in a future PR... like I said, I'm ambivalent at the moment, lol.

mascguy: Well, this is another maintainer's port. And while folks are free to follow their own personal preferences for those that they maintain, we should remain consistent with the existing conventions for someone else's port.

As an example: If you submitted a PR for one of my ports, and you chose to make unnecessary changes like this, I'd ask you to promptly remove them.

And ultimately there's no way I'd merge it otherwise. So you'd either have to make the changes requested, or the PR would be rejected outright.

Ditto for this one: This port is maintained by Marcus, and there's a consistent convention in place. And there's simply no way that I'll merge this in it's current state.

If someone else is willing to - or Marcus chooses to - that's fine. But I can't, in good conscious, merge this as-is.

mascguy: Updated the commit, to limit the number of unnecessary changes.

mascguy (approved):
Approving based on the current state of this PR, contingent upon no unnecessary changes being pushed.

mascguy:
And thanks for updating this port Jason, much appreciated!

PR #13258, opened 2021-12-06
moreutils: update to 0.66

herbygillot:
###### Verification 
Have you

### DIFF
--- a/sysutils/moreutils/Portfile
+++ b/sysutils/moreutils/Portfile
@@ -5,7 +5,7 @@ PortGroup               legacysupport 1.0
 
 name                    moreutils
 conflicts               ts
-version                 0.65
+version                 0.66
 revision                0
 categories              sysutils
 platforms               darwin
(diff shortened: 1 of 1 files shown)


herbygillot:
Azure CI failure looks to be because of an expired SSL cert.

kurthindenburg:
I hadn't update since the freebsd mirrors didn't have the new version.  Is there really any reason to use the snapshot?   Does that only change w/ new versions?

herbygillot:
I was using the snapshot precisely because the FreeBSD mirrors didn't have the new version. I don't see an official release tarball - wouldn't the snapshot for the given version be a correct asset to use?

kurthindenburg:
I don't know how that snapshot link works.  If you're using the new version w/o issue, go ahead as I won't block it.

herbygillot:
> I don't know how that snapshot link works.

The snapshot link looks to work similarly to Github's archive links.  It downloads a tarball of the specified git tag from the author's original Git repository.

You can see everything here: https://git.joeyh.name/index.cgi/moreutils.git/

kurthindenburg (approved):
Thanks, this is better than waiting for freebsd to have it.

PR #13259, opened 2021-12-06
git: do not overwrite system-wide gitconfig, when adding configuration for the osxkeychain credential helper

herbygillot:
Do not overwrite the system-wide gitconfig when adding configuration for the osxkeychain credential helper

If the gitconfig does not exist, then it will be created. But if it *does*
exist, then the file will be checked to see whether it already contains the
desired config, and if it doesn't, the configuration will be appended to the
already existing file.

Fixes: https://trac.macports.org/ticket/63953

###### Verification 
Have you

### DIFF
--- a/devel/git/Portfile
+++ b/devel/git/Portfile
@@ -110,7 +110,8 @@ pre-destroot {
     xinstall -m 644 ${worksrcpath}/contrib/subtree/git-subtree.1 ${workpath}/man1
 }
 
-set docdestroot ${destroot}${prefix}/share/doc/git-doc
+set docdestroot         ${destroot}${prefix}/share/doc/git-doc
+set system_gitconfig    ${prefix}/etc/gitconfig
 
 post-destroot {
     foreach f {1 5 7} {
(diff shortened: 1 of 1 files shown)


devel/git/Portfile
```diff
@@ -227,35 +228,44 @@ variant credential_osxkeychain description {Install git credential-osxkeychain u
...
-        close ${f}
+    pre-activate {
+        set osxkc_config        "\[credential]\n\thelper = osxkeychain"
+
+        # Ensure the osxkeychain credential helper is configured in the
+        # system-wide gitconfig.  If the gitconfig file already exists,
+        # then append the configuration to the pre-existing file only
+        # if the file is currently missing the desired config.
+
+        if { ![file exists ${system_gitconfig}] } {
```
cjones051073, line 244: You gave a conceptual bug here. When you check the currently existing file you correctly check the *installed* version directly in ${prefix}, but you cannot directly write to this file, which is what you currently do. You have to write to the version in the DESTROOT ie at ${destroot}${prefix}, and if need be first clone the currently installed version before appending to it.

cjones051073 (changes requested):

mascguy (approved):

mascguy:
LGTM. And this will certainly be appreciated by users who rely on customization to the system config.

Great work Herby!

PR #13260, opened 2021-12-06
polyml: Fix build for arm64 and Mac OS <= 10.9

pclayton:
#### Description

Some MacPorts builders failed to build Poly/ML because
1. configure did not recognized 'arm64' (it was expecting 'arm')
https://github.com/polyml/polyml/issues/165

2. Mac OS <= 10.9 did not have the included header file 'mach-o/arm64/reloc.h'
https://github.com/polyml/polyml/issues/166

This branch addresses these issues with patches based on [upstream fixes](https://github.com/polyml/polyml/tree/MacPorts).

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.10.5 14F2511 x86_64
Xcode 6.2 6C131e

###### Verification 
Have you

- [ ] tried existing tests with `sudo port test`?  No, port doesn't have tests enabled.

### DIFF
--- a/lang/polyml/Portfile
+++ b/lang/polyml/Portfile
@@ -2,7 +2,7 @@ PortSystem 1.0
 PortGroup           github 1.0
 
 github.setup        polyml polyml 5.9 v
-revision            0
+revision            1
 categories          lang sml
 platforms           darwin
 maintainers         {veonix.com:phil.clayton @pclayton} openmaintainer
@@ -24,6 +24,10 @@ depends_lib         port:gmp
 
 configure.args      --mandir=${prefix}/share/man --build=${build_arch}-apple-darwin${os.major} \
                     --enable-shared
+
+patchfiles          patch-check-mach-o-reloc-h-file.diff \
+                    patch-update-config-for-arm64.diff
+patch.pre_args -p1
 
 post-destroot {
     xinstall -m 755 -d ${destroot}${prefix}/share/doc/${name}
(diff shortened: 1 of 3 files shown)


mascguy:
LGTM. But I'd like to hear @kencu's thoughts, given some of the tweaks.

pclayton:
@mascguy, thanks for looking at this.  Is there anything I could clarify relating to your concerns?

mascguy:
Seems reasonable. Let's merge, and we can always refine later if necessary.

And thanks for your contribution!

ryandesign:
We should avoid committing enormous patchfiles like the one that updates config.guess / config.sub. It would be simpler and more space-efficient to have the port use autoreconf. Enormous files stay in the git history forever so it is important to consider whether that is the best solution before they ever get merged to master.

The other patch patches both configure and configure.ac. You should either patch the generated files (configure) if you are not running auto(re)conf, or patch the source files (configure.ac et al) if you are running auto(re)conf. By patching both, and specifically by patching the source files after the generated files, a race condition can occur whereby if configure is patched in one second and configure.ac is not patched until the next second, the build system will think configure is out of date and will regenerate it, which might have unintended consequences.

pclayton:
Good point about autoreconf, that may well have resolved that issue.

I agree that only one of configure.ac and configure should have been patched.  As the port isn't running auto(re)conf and configure.ac isn't used, presumably there is no potential for the race condition to occur and the port doesn't need to be updated?

PR #13273, opened 2021-12-07
openvdb: rewrite Portfile, add new variants, enable features

jasonliu--:
#### Description

This commit is a fairly significant rewrite of the Portfile. After digging more deeply into the OpenVDB source code and documentation, it turns out that the Portfile is only performing a very minimal build. There are some features, such as support for hardware intrinsics, which are strongly recommended by the upstream authors, but are not enabled by default. I have added variants so that users can control some of these options from the 'port' command.

I have added the new jemalloc variant to the default variants, since using jemalloc is stronly recommended by the upstream devs. I have added the new docs variant to the default variants, as suggested by the Best Practices section of the MacPorts Guide.

In addition, I'm not sure why building the Python bindings is the default variant. Currently, the two ports that depend on openvdb are openimageio and blender. OpenImageIO doesn't need the Python module. I will need to check to see whether Blender somehow uses pyopenvdb (I'm doubtful: I believe that Blender links directly to libopenvdb.dylib at build time, which would likely be more performant than making calls to the Python module at run time), but I currently believe that the Python module is unused. I may remove python from the default variants list in the future, if it turns out that nothing is making use of it.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you


mascguy (approved):

mascguy:
AFAIK, Marcus is still unavailable for PR reviews. Plus Jason is now a co-maintainer of this port.

So let's go ahead and merge.

PR #13139, opened 2021-11-27
tkimg: update to 1.4.13

chrstphrchvz:
#### Description
From ANNOUNCE:
```
Changes compared to version 1.4.12:
  - Disable support of zstd in libtiff.

Changes compared to version 1.4.11:
  - Corrected bug in BMP and XBM parser.
  - Updated external libraries: JPEG 9d, PNG 1.6.37, TIFF 4.1.0.
  - Updated to latest TEA files.
  - Corrected nmake builds.
```

###### Type(s)

- [x] bugfix
- [x] enhancement

###### Tested on

macOS 10.15.7
Xcode 12.4 command line tools
(both +x11 and +quartz)

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [ ] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 
- [x] checked your Portfile with `port lint`?
- [ ] tried existing tests with `sudo port test`?
- [x] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?

PR #13332, opened 2021-12-13
qt5-qtwebengine: update to 5.15.8 + various other changes

chrstphrchvz:
Drop patch-qtwebengine_shrink_to_fit.diff in favor of upstream (Qt) approach
See https://github.com/qt/qtwebengine-chromium/commit/7ff159da128c4f249b468f3ff972f864d243c742

Use subsequent commit from qtwebengine-chromium 87-based branch to include bugfix

Do not extract ninja from qtwebengine-chromium distfile since build uses ninja from MacPorts

Disable `conflicts_build` for now

`os.platform` should be checked before `os.major`
See https://lists.macports.org/pipermail/macports-dev/2021-December/043976.html

Requires Xcode 10 or later (Build is expected to fail on macOS 10.13 builder, which has Xcode 9.4.1)
See: https://trac.macports.org/ticket/61921

nodejs16 (latest LTS) currently does not build on macOS 10.13
Fallback to installing nodejs14 (previous LTS) if needed
See: https://trac.macports.org/ticket/64130
Use `path:bin/node:…` instead of `bin:node:…` dependency
See https://lists.macports.org/pipermail/macports-dev/2021-December/043998.html

qt5-qtwebengine-gn: new subport (of gn-devel)
qt5-qtwebengine, qt5-qtwebengine-docs: build using qt5-qtwebengine-gn
Fixes arm64 build of qt5-qtwebengine-docs

`qt5-qtwebengine-docs`: fix fetch
Fixes: https://trac.macports.org/ticket/64348 

###### Verification 
Have you

- [ ] ~~tried a full install with `sudo port -vst install`?~~


chrstphrchvz:
Note that the port maintainer has been unavailable for review: https://github.com/macports/macports-ports/pull/13139#issuecomment-980638649

aqua/qt5/Portfile
```diff
@@ -1683,15 +1683,28 @@ foreach {module module_info} [array get modules] {
                 # see https://doc.qt.io/qt-5/qtwebengine-platform-notes.html
                 PortGroup           xcodeversion 1.0
 
-                minimum_xcodeversions   {16 8.3.3}
-                if { ${os.major} < 16 } {
+                # see https://trac.macports.org/ticket/61921
+                minimum_xcodeversions   {17 10.0.0}
+                if { ${os.major} < 17 } {
```
reneeotten, line 1688: this (and the ones below) need the additional `os.platform` check as well

aqua/qt5/Portfile
```diff
@@ -1683,15 +1683,28 @@ foreach {module module_info} [array get modules] {
...
+                        ui_error "${subport} requires macOS 10.13 or later"
                         return -code error "incompatible OS version"
                     }
                 }
 
+                # build requires nodejs as of 5.15.3
+                # nodejs12 (which chromium 87 uses by default) or later are known to work
+                if { ${os.major} >= 18 } {
+                    # prefer latest LTS nodejs as fallback
+                    depends_build-append bin:node:nodejs16
```
reneeotten, line 1700: this is probably not correct as it will allow any `node` binary in the path (also outside MacPorts) to satisfy the dependency. It should probably be `bin/node:nodejs16` (similar to as is done above for `ninja`, which will allow any `node` in the MacPorts' prefix to satisfy the dependency

aqua/qt5/Portfile
```diff
@@ -1683,15 +1683,28 @@ foreach {module module_info} [array get modules] {
...
+                } else {
+                    # nodejs16 (latest LTS) currently does not build on macOS 10.13
+                    # see https://trac.macports.org/ticket/64130
+                    # use nodejs14 (previous LTS) as fallback
+                    depends_build-append bin:node:nodejs14
```
reneeotten, line 1705: see above

reneeotten (changes requested):

aqua/qt5/Portfile
```diff
@@ -666,7 +666,7 @@ array set modules {
     }
     qtwebengine {
         {}
-        "port:python27 port:py27-ply path:bin/ninja:ninja port:gperf port:bison port:flex bin:node:nodejs16"
+        "port:python27 port:py27-ply path:bin/ninja:ninja port:gperf port:bison port:flex"
```
chrstphrchvz, line 669: nodejs dependency moved to special case section since comments are not allowed in the list here.

aqua/qt5/Portfile
chrstphrchvz: I should have at least left a comment explaining the `bin:…` rather than `path:…` dependency. While I am aware using non-MacPorts stuff can be undesirable, I may be inclined to leave this as `bin:…` rather than `path:…` unless it is confirmed that only MacPorts’ Node.js will be used for `path:…` dependency. I encountered issues building this port under trace mode (`too many open files` error, which I neglected to report), so trace mode may not be useful for enforcing MacPorts Node.js usage. And because Node.js is not a lightweight dependency (hence why I propose switching it to use known-working nodejs14 on 10.13, rather than unsuccessfully attempt building nodejs16), it seemed desirable to be lenient with how Node.js dependency is provided. See also https://github.com/macports/macports-ports/pull/12595#issuecomment-956960641

aqua/qt5/Portfile
chrstphrchvz: I would rather also correct all existing instances in the qt5 portfile first: d26e4d9ed050e467131512330922bc1aeb31f3c9. There remain many instances of this issue in other qt Portfiles; ideally they can be corrected simultaneously in a dedicated PR, and without exacerbating https://trac.macports.org/ticket/64126.

aqua/qt5/Portfile
reneeotten: I don't see an argument in your response to use `bin:node:nodejs16` instead of `bin/node:nodejs16`. Ryan just explained [here](https://github.com/macports/macports-ports/commit/afad77a86ba6be6572cf0aff35db0b13401196f1#commitcomment-61791005) that the former one shouldn't be used.

dliessi:
Is the `bin:` vs `path:` dependency the only issue blocking this PR? If @chrstphrchvz is not interested, I can take care of the change, although I would open a new PR, given that I cannot push to this branch.

chrstphrchvz:
Rebased and incorporated likely fix for https://trac.macports.org/ticket/64348 (**untested**—waiting for Xcode to reinstall before I can test…)

chrstphrchvz:
I was able to build and install `qt5-qtwebengine-docs @5.15.7_0` with Xcode 13.2.1 on macOS 12.1.

chrstphrchvz:
https://github.com/qt/qtwebengine/tree/v5.15.8-lts was tagged yesterday. I will probably revise this PR to incorporate the update.

chrstphrchvz:
I have successfully built `qt5-qtwebengine @5.15.8_0` and `qt5-qtwebengine-docs @5.15.8_0` on macOS 12.1 x86-64 using Xcode 13.2.1, `nodejs16 @16.13.1_0`. Running simplebrowser.py from PySide2:
Screen Shot 2022-01-05 at 3 59 49 AM

Would any volunteers (particularly those with an Apple Silicon Mac) like to verify that the 5.15.8 update builds for them? Pinging @abruto @amake @cgilles @TomDev234 @JulianGro @wrohv @abey79 @mf2k in case of interest.

abey79:
No problems installing `qt5-qtwebengine @5.15.8_0` on my side (Late 2021 M1 Max MBP / 12.1).
image

I did get an error for `qt5-qtwebengine-docs @5.15.8_0` though: [main.log.zip](https://github.com/macports/macports-ports/files/7815007/main.log.zip)

chrstphrchvz:
> 
> I did get an error for `qt5-qtwebengine-docs @5.15.8_0` though: [main.log.zip](https://github.com/macports/macports-ports/files/7815007/main.log.zip)

```
:info:build ERROR at //build/config/mac/BUILD.gn:44:5: Assertion failed.
:info:build     assert(false, "unknown current_cpu $current_cpu")
:info:build     ^-----
:info:build unknown current_cpu arm
:info:build See //build/config/BUILDCONFIG.gn:411:23: which caused the file to be included.
:info:build   _linker_configs = [ "//build/config/mac:strip_all" ]
:info:build                       ^-----------------------------
:info:build Project ERROR: GN run error!
```

Same error as https://github.com/macports/macports-ports/pull/12595#issuecomment-956007459, so it looks like patch-qtwebengine_gn_host_cpu_arm64.diff needs to also be applied to the docs build. Though I wonder if it would be better to use GN from the gn-devel port (or a new pinned port) instead of always building the bundled GN.

chrstphrchvz:
> …I wonder if it would be better to use GN from the gn-devel port (or a new pinned port) instead of always building the bundled GN.

The bundled GN has necessary Qt-specific additions. Using `configure.post_args-append -feature-webengine-system-gn` to force using external GN (since buildtools/configure.json has `"autoDetect": "false"` for feature `webengine-system-gn`) fails during build:
```
ERROR at /opt/local/var/macports/build/_opt_local_var_macports_sources_github.com_macports_macports-ports_aqua_qt5/qt5-qtwebengine-docs/work/qtwebengine-5.15.8-lts/src/pdf/release/BUILD.gn:156:21: Assignment had no effect.
  create_pri_file = true
                    ^---
You set the variable "create_pri_file" here and it was unused before it went
out of scope.
See /opt/local/var/macports/build/_opt_local_var_macports_sources_github.com_macports_macports-ports_aqua_qt5/qt5-qtwebengine-docs/work/qtwebengine-5.15.8-lts/src/pdf/release/BUILD.gn:155:1: whence it was called.
shared_library("QtPdf") {
^------------------------
See //BUILD.gn:1427:14: which caused the file to be included.
    deps = [ qtwebengine_target ]
            ^-----------------
Project ERROR: GN run error!
```

I’m not sure it how easy it is to create/maintain a separate port for the bundled GN, if only for avoiding spending several minutes rebuilding it from scratch (as ccache isn’t being used for building bundled GN) when building qtwebengine or qtwebengine-docs.

chrstphrchvz:
I intend to rebase this and update the qtwebengine-chromium ref to https://github.com/qt/qtwebengine-chromium/commit/d505a379d3b1bd92d15c5a7051c3ed7b83743424 which fixes a crash.

I successfully built the bundled GN as a separate port and patched qt5-qtwebengine/qt5-qtwebengine-docs to use it. I placed it in a subport of gn-devel for expediency. I had it install the binary to `${prefix}/libexec/qt5/bin/gn-qtwebengine5.15` so that it would not conflict with `gn` from gn-devel or elsewhere. I would like for #13614 to be merged first.

Regarding https://github.com/macports/macports-ports/commit/bd858ac96b4181de1d388f8e355b852577264dcd#commitcomment-61755245: I tried building 5.15.8 without `conflicts_build` and with 5.15.7 active, and did not notice any issue, although it is still not clear to me why `conflicts_build` this may have been needed.

chrstphrchvz:
This is ready for review again.

abey79:
Both `qt5-qtwebengine` and `qt5-qtwebengine-docs` appear to compile and install properly on my 2021 M1 MBP. I was somewhat surprised to see Python 2.7 installed as dependency. I believe it wasn't the case beforehand.

chrstphrchvz:
> Both `qt5-qtwebengine` and `qt5-qtwebengine-docs` appear to compile and install properly on my 2021 M1 MBP.

Thanks for testing.

> I was somewhat surprised to see Python 2.7 installed as dependency. I believe it wasn't the case beforehand.

The `python27` build dependency has been there since 6b182f3400f4c1e044edc9aab9e1ebef21a8b659. (I’m aware that upstream chromium reportedly has been struggling to migrate to Python 3, and that there is some pressure to complete this transition from other package managers—e.g. major Linux distributions—who are dropping Python 2; Arch Linux already patches qtwebengine to build using Python 3: https://github.com/archlinux/svntogit-packages/tree/packages/qt5-webengine/trunk.)

pmetzger:
Is this ready to merge?

chrstphrchvz:
If there are no other comments or objections, then I believe it is ready to merge.

PR #13344, opened 2021-12-14
lighttpd: update config; remove mod_cml

gstrauss:
#### Description

lighttpd config updates; remove mod_cml
- remove mod_cml
  long-deprecated module will be removed from upstream in Jan 2022
- add --with-lua to the core build; mod_magnet is better than mod_cml
- remove bzip2 dependency; seldom used with mod_deflate; brotli better

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Verification 
Have you

### DIFF
--- a/www/lighttpd/Portfile
+++ b/www/lighttpd/Portfile
@@ -5,7 +5,7 @@ PortGroup                   legacysupport 1.0
 
 name                        lighttpd
 version                     1.4.63
-revision                    1
+revision                    2
 checksums                   rmd160  f8b13b883825626ed04c3349ec5f2330c49a6a37 \
                             sha256  2aef7f0102ebf54a1241a1c3ea8976892f8684bfb21697c9fffb8de0e2d6eab9 \
                             size    1023568
@@ -31,7 +31,7 @@ use_xz                      yes
 depends_build-append        port:pkgconfig
 
 depends_lib                 port:brotli \
-                            port:bzip2 \
+                            port:lua \
                             port:pcre2 \
                             port:spawn-fcgi \
                             port:xxhashlib \
@@ -56,7 +56,7 @@ compiler.blacklist          *gcc-3.* *gcc-4.*
 configure.args-append       CC_FOR_BUILD="${configure.cc}" \
                             CFLAGS_FOR_BUILD="${configure.cflags}" \
                             --with-brotli \
-                            --with-bzip2 \
+                            --with-lua \
                             --with-pcre2 \
                             --with-xxhash \
                             --with-zlib \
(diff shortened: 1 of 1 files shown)


ryandesign:
> * remove mod_cml
>   long-deprecated module will be removed from upstream in Jan 2022

I suppose the mod_cml variant can be removed now, or we could wait until we update to the new version in Jan 2022.

> * add --with-lua to the core build; mod_magnet is better than mod_cml

I have no objection to that.

> * remove bzip2 dependency; seldom used with mod_deflate; brotli better

What would be the advantage of removing bzip2 support from lighttpd? lighttpd already depends on pcre2 which depends on bzip2, so the bzip2 port will already have to be installed for lighttpd.

gstrauss:
> I suppose the mod_cml variant can be removed now, or we could wait until we update to the new version in Jan 2022.

Up to you.  I was filing this more as a heads-up.  The scheduled removal of mod_cml has been announced in all recent lighttpd releases since the end of last year.

> What would be the advantage of removing bzip2 support from lighttpd? lighttpd already depends on pcre2 which depends on bzip2, so the bzip2 port will already have to be installed for lighttpd.

I didn't know that about pcre2.  The benefit to lighttpd removing bzip2 is marginal: slightly less code compiled in mod_deflate.  (The code is still in mod_deflate, but not compiled in by default, unless compiled --with-bzip2)  Related to the scheduled removal of other deprecated lighttpd modules, I was cleaning up the list of lighttpd direct dependencies.  bzip2 is a non-standard encoding which is seldom used (if at all anymore) in Accept-Encoding, especially since the introduction of zopfli, then brotli.

This is also up to you if you want to remove --with-bzip2 or not.

Please let me know if you'd like me to adjust this PR or if you would prefer to wait until lighttpd 1.4.64, currently scheduled for Jan 2022.

gstrauss:
Thanks, @ryandesign 

FYI: pcre2 source code may depend on bzip2 for a feature in `pcre2grep`, part of the (optional) pcre2-tools package.
The pcre2 *libraries* do not depend on bzip2.

ryandesign:
> The pcre2 _libraries_ do not depend on bzip2.

I had not investigated further than noting that the lighttpd port declares a library dependency on pcre2 and that declares a library dependency on bzip2. Yes, as you discovered, the pcre2grep program links with the bzip2 library; that's why the dependencies are declared the way they are. "Library dependency" in MacPorts means "it's needed at build time and at runtime".

Still, I agreed with you that bz2 encoding is probably not used, so I merged your PR as is.

PR #13345, opened 2021-12-14
hub: fix installation for zsh completion

laggardkernel:
Force pushed once to standardize the file permission for `_hub`, the zsh completion file, whose file perm was set 755 before.

---

#### Description

hub's zsh completion file is not installed correctly and completion doesn't work. Fix it.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14
xcode-select version 2354

###### Verification 
Have you

### DIFF
--- a/devel/hub/Portfile
+++ b/devel/hub/Portfile
@@ -4,6 +4,7 @@ PortSystem          1.0
 PortGroup           github 1.0
 
 github.setup        github hub 2.14.2 v
+revision            1
 categories          devel
 platforms           darwin
 license             MIT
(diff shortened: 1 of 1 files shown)


devel/hub/Portfile
```diff
@@ -39,18 +40,18 @@ post-extract {
 }
 
 destroot {
-    xinstall -W ${worksrcpath} bin/hub ${destroot}${prefix}/bin
+    xinstall -m 0755 -W ${worksrcpath} bin/hub ${destroot}${prefix}/bin
     set man-pages ${destroot}${prefix}/share/man/man1
     xinstall -d ${man-pages}
     foreach m [glob -directory ${worksrcpath}/share/man/man1 *.1] {
-        xinstall -W ${worksrcpath} ${m} ${man-pages}/
+        xinstall -m 0444 -W ${worksrcpath} ${m} ${man-pages}/
```
pmetzger, line 47: These should probably be 0644?

pmetzger (changes requested):
So other than the one little nit (man pages should probably be 0644) this looks good. Could you update it and we can then merge it?

devel/hub/Portfile
laggardkernel: @pmetzger Seems both 644 and 444 are okay. Have a check of files in `/usr/share/man/man*`. I remember I firstly set it 0644 and changed it to 0444 after seeing 0444 used in another Portfile.

pmetzger: It really should be 644 on any normal system. The other portfile was wrong.

laggardkernel: @pmetzger Okay, fixed.

devel/hub/Portfile
```diff
@@ -39,18 +40,18 @@ post-extract {
...
+        xinstall -m 0644 -W ${worksrcpath} ${m} ${man-pages}/
     }
     set bash-completions ${destroot}${prefix}/share/bash-completion/completions
     xinstall -d ${bash-completions}
-    xinstall -W ${worksrcpath} etc/hub.bash_completion.sh ${bash-completions}/hub
+    xinstall -m 0644 -W ${worksrcpath} etc/hub.bash_completion.sh ${bash-completions}/hub
     set site-functions ${destroot}${prefix}/share/zsh/site-functions
     xinstall -d ${site-functions}
-    xinstall -W ${worksrcpath} etc/hub.zsh_completion ${site-functions}/
+    xinstall -m 0644 -W ${worksrcpath} etc/hub.zsh_completion ${site-functions}/_hub
```
pmetzger, line 54: I only just noticed this is now being copied to `_hub` where before it didn't have its name changed. Is this correct?

laggardkernel: Yes. This is what I really want commit, not those `chmod`. A zsh completion file should starts with `_`, otherwise it's not used as completion file. The completion for `hub` in zsh didn't work before this commit.

[Quote from zsh-completions plugin](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command)
> Completion functions for commands are stored in files with names beginning with an underscore `_`, and these files should be placed in a directory listed in the `$fpath` variable.

laggardkernel: @pmetzger Sorry, seems I forgot to emphasize the name change in the description.

PR #13363, opened 2021-12-15
iTerm2: disable UBSan

i0ntempest:
Closes: https://trac.macports.org/ticket/62769

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2 13C90

###### Verification 
Have you

### DIFF
--- a/aqua/iTerm2/Portfile
+++ b/aqua/iTerm2/Portfile
@@ -7,7 +7,7 @@ PortGroup           xcodeversion 1.0
 
 if {${os.major} > 17} {
     version             3.4.15
-    revision            0
+    revision            1
     checksums           rmd160  a5319483e324f6408952d6db41a45860bd90cb20 \
                         sha256  2348fa198a36cd003a124573cb547f2da8266252479db70e60e17cdcfe32d89d \
                         size    29047663
@@ -16,7 +16,7 @@ if {${os.major} > 17} {
                         patch-nsur.diff
 } elseif {${os.major} > 16} {
     version             3.3.12
-    revision            1
+    revision            2
     checksums           rmd160  37d4204f97ac42431c5774a943ca0ab8f9a4c669 \
                         sha256  6eeeb3215d6725aa789d39108a8ebedbe1aa9f6d058dd33056d72471ecd02e14 \
                         size    20602868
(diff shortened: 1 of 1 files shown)


aqua/iTerm2/Portfile
```diff
@@ -55,6 +55,7 @@ github.livecheck.regex {(\d+(?:\.\d+)*)}
 
 post-patch {
     reinplace "s|CODE_SIGN_IDENTITY = \".*\";|CODE_SIGN_IDENTITY = \"\";|g" ${worksrcpath}/iTerm2.xcodeproj/project.pbxproj
+    reinplace "s|enableUBSanitizer = \"YES\"||g" ${worksrcpath}/iTerm2.xcodeproj/xcshareddata/xcschemes/iTerm2.xcscheme
```
gnachman, line 58: Using the Development build target is the root cause of this problem. Any reason you don't use Release instead? It will be quite a bit faster. Development tends to change and you'll likely have similar problems in the future (e.g., if ASAN gets enabled).

markemer: Let me give that a try - I'm not sure I knew that was the problem. I've been trying to move this to the XCode portgroup with not a lot of luck - but that could be a quick fix.

i0ntempest: The build runs `make prod` which I believe is using Release target? Through my experiment I've found it's the `-derivedDataPath ./DerivedData` arg added to fix permission errors causing the problem, even if you move this to Xcode pg this will still be needed, and will still trigger the link to UBSan.

i0ntempest: [Portfile.zip](https://github.com/macports/macports-ports/files/7724488/Portfile.zip)
This is my attempt to move to Xcode pg, in case you're interested.

gnachman: Yeah, `make prod` should be unaffected by that change. I don't see where in the Portfile it calls make, though. Do you have the output from `make`? That should make it very clear what build target is actually being used.

i0ntempest: If no other portgroups (cmake/xcode/meson etc.) is used then `./configure && make` is the default and we have configure script disabled in this port. Full build log is here https://build.macports.org/builders/ports-12_x86_64-builder/builds/5819/steps/install-port/logs/stdio

PR #13399, opened 2021-12-18
qemu: restore `+gtk3` and `+sdl2` variants

evanmiller:
#### Description

The old warnings no longer seem to apply. SDL and GTK do not conflict with each other, and can be specified at run-time with `-display sdl` or `-display gtk`.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -190,32 +190,15 @@ variant curses description {Use the curses text-only user interface} {
 
 # XXX: gtk/sdl need libepoxy for OpenGL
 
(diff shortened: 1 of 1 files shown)


raimue:
If I remember correctly, back then the thing was that the main loop for all of qemu was replaced based on a compile-time check condition "if on macOS" and that caused problems. However, it looks like this has changed. I just never tried again after I had them disabled for so long. Thanks for testing this! :+1:

PR #13402, opened 2021-12-18
gcc11: ensure all patches are applied

MarcusCalhoun-Lopez:
Patch file from
https://github.com/macports/macports-ports/commit/7f9e2fb69924a606ed00a7c09d0d07cfd9a0604b
is lost when `patchfile` command is later used.

See https://trac.macports.org/ticket/63677
No revbump since port either builds correctly or not at all.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

- [] tried existing tests with `sudo port test`?

### DIFF
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -69,7 +69,7 @@ if { ${os.arch} eq "arm" } {
 
 subport             libgcc11 { revision [ expr ${revision} + 0 ] }
 
-patchfiles          patch-Make-lang.in.diff
+patchfiles-append   patch-Make-lang.in.diff
 
 if { ${os.platform} eq "darwin" } {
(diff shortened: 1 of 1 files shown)


cjones051073:
You need to bump the revision as well.

MarcusCalhoun-Lopez:
> You need to bump the revision as well.

Sorry, could you explain why a revision bump is necessary?

As it states in the commit message, the port either builds correctly or not at all.
On macOS Monterey, the fact that patch-darwin21-support.diff is **not** applied means GCC does not build at all.
On older system where the patch is not needed, no change results in its application.

cjones051073:
fair enough.

PR #13426, opened 2021-12-21
virt-manager: prepend ${prefix}/bin and ${prefix}/sbin to app's PATH

sideeffect42:
#### Description

Makes virt-manager prefer MacPorts' binaries, e.g. /opt/local/bin/ssh.
Thus fixing errors that may occur because virt-manager is using macOS shipped binaries lacking some options present in the (mostly newer) MacPorts versions.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.1 20G224 x86_64
Xcode 13.2 13C90

###### Verification 
Have you

### DIFF
--- a/gnome/virt-manager/files/virt-manager.sh
+++ b/gnome/virt-manager/files/virt-manager.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+PATH=@PREFIX@/bin:@PREFIX@/sbin:${PATH:-$(/usr/bin/getconf PATH)}
 exec @PREFIX@/bin/virt-manager --no-fork "$@"
--- a/gnome/virt-manager/Portfile
+++ b/gnome/virt-manager/Portfile
@@ -7,7 +7,7 @@ PortGroup           active_variants 1.1
 PortGroup           app 1.0
 
 github.setup        virt-manager virt-manager 3.2.0 v
-revision            2
+revision            3
 categories          gnome emulators
 supported_archs     noarch
 maintainers         nomaintainer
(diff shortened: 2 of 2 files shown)


pmetzger:
Unfortunately it's not quite so simple. MacPorts may be installed in locations other than /opt/local and so you can't hard-code that path in a script. If the consensus is that this is a good idea, you will need to generate the appropriate path in code in the Portfile.

gnome/virt-manager/files/virt-manager.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+PATH=/opt/local/bin:/opt/local/sbin:${PATH:-$(/usr/bin/getconf PATH)}
```
mascguy, line 3: Per @pmetzger, you don't want to hard-code the path to `/opt/local`.

But the port is already setup to deal with this, and there's an easy fix: Simply use placeholder `@PREFIX@` instead.

mascguy: It looks like you'll also need to update the `reinplace` statement on line 87, as it's not currently performing a global replace. And this will prevent replacement of multiple occurrences on the same line.

So change the following on that line:

`s+@PREFIX@+${prefix}+`

to:

`"s|@PREFIX@|${prefix}|g"`

sideeffect42: Thanks for your comment @mascguy. Although @pmetzger's comment made sense to me I was a bit uncertain what I needed to do (I'm new to manipulating Portfiles).

I incorporated your suggestions in bbae678.

pmetzger:
LGTM though the commits will need to be squashed, either by @sideeffect42 or during the merge.

mascguy:
> LGTM though the commits will need to be squashed, either by @sideeffect42 or during the merge.

I took care of that, along with some minor dependency-related cleanup (the latter as a separate commit).

Test locally, and looks good. Merging.

mascguy:
And @sideeffect42, thank you so much for your contribution!

sideeffect42:
> And @sideeffect42, thank you so much for your contribution!

Thanks for your assistance, @mascguy.

PR #13437, opened 2021-12-22
wesnoth: update to 1.16.1

evanmiller:
#### Description

Update wesnoth to 1.16.1 and migrate it from Xcode to the CMake based build system. This allows for a simpler Portfile and building on systems that do not support Xcode 8. Patches are included to allow building all the way back to 10.6. (Note that 10.6 itself requires #13430.)

Game tested on 10.6.8 (x86_64).

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode x.y

###### Verification 
Have you


games/wesnoth/Portfile
```diff
@@ -96,31 +55,25 @@ depends_lib-append      port:boost \
...
+patchfiles-append       patch-wesnoth-sdl-main.diff
 
-build.dir               ${worksrcpath}/projectfiles/Xcode
+# src/global.hpp:28:2: error: "Compilation with NDEBUG defined isn't supported, Wesnoth depends on asserts."
+configure.cppflags-append -UNDEBUG
 
-xcode.configuration     Release
-xcode.target            {The Battle for Wesnoth}
+pre-destroot {
+    # Reproduce the logic from app 1.0 PG
```
ryandesign, line 67: Why is the app portgroup needed? Doesn't the build system already produce an app? If not, how does upstream produce the app for the Mac builds it distributes on its web site?

If the app portgroup is needed, why is reproducing its logic in this portfile necessary?

evanmiller: > Why is the app portgroup needed? Doesn't the build system already produce an app? If not, how does upstream produce the app for the Mac builds it distributes on its web site?

Wesnoth includes a CMake build and an Xcode build. (Also an scons build.) According to the docs, the Xcode build requires Xcode 8. Using the CMake build along with the app PG lets us target older systems that cannot run Xcode 8.

> If the app portgroup is needed, why is reproducing its logic in this portfile necessary?

Maybe I am mistaken but I thought that `pre-destroot` in a Portfile would override the `pre-destroot` in the app PG. I am open to suggestions on how to move the NIB file into Contents/Resources without interfering with the app PG logic.

ryandesign: > Maybe I am mistaken but I thought that `pre-destroot` in a Portfile would override the `pre-destroot` in the app PG.

You're mistaken! :) All `pre-` and `post-` blocks are cumulative and run in the order in which they were declared.

ryandesign: > Using the CMake build along with the app PG lets us target older systems that cannot run Xcode 8.

That's great! Did you check that the Info.plist that the app portgroup generated is sufficiently similar to the Info.plist they ship in their binary?

evanmiller: > You're mistaken! :) All pre- and post- blocks are cumulative and run in the order in which they were declared.

Good to know :-)

> That's great! Did you check that the Info.plist that the app portgroup generated is sufficiently similar to the Info.plist they ship in their binary?

There are some minor differences, e.g. copyright string and app category. I have just updated the Portfile to enable the Retina setting. I don't think any of the remaining differences affect the run-time.

cooljeanius:
Thanks for this PR! I didn't realize you had opened it when I filed a Trac ticket requesting the same thing...

evanmiller:
@cooljeanius I'll add a reference in the commit message so the ticket gets closed. Would be great if you could try building it locally.

mascguy (approved):

mascguy:
LGTM overall, but I'll defer to @ryandesign.

ryandesign:
Thanks!

> Patches are included to allow building all the way back to 10.6.

Could you submit your improvements to the developers? They look simple enough that hopefully they won't mind including them.

cooljeanius:
> Thanks!
> 
> > Patches are included to allow building all the way back to 10.6.
> 
> Could you submit your improvements to the developers? They look simple enough that hopefully they won't mind including them.

The official Mac packager is @hrubymar10 but @celticminstrel also seemed to be interested when I mentioned this PR on the Wesnoth Discord

evanmiller:
@cooljeanius @ryandesign I have opened a PR with the 10.8+ support. If it goes well we can submit the 10.6/10.7 code also.

PR #13463, opened 2021-12-24
postgresql96-server: Remove shell from `postgres` user

essandess:
* Fixes: https://trac.macports.org/ticket/64286

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/databases/postgresql96-server/Portfile
+++ b/databases/postgresql96-server/Portfile
@@ -5,24 +5,19 @@ PortSystem 1.0
 name                postgresql96-server
 version             9.6.24
 categories          databases
-platforms           darwin
 maintainers         {jwa @jyrkiwahlstedt}
 license             Permissive
+supported_archs     noarch
 
 set rname           postgresql96
+
 description         run ${rname} as server
 long_description    {*}${description}
-distfiles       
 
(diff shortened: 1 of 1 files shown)


ryandesign:
Yes, but it would be good to keep all the postgresql*-server ports as similar as possible.

essandess:
> Yes, but it would be good to keep all the postgresql*-server ports as similar as possible.

The only sane way to manage that is to combine them all as supports within a new stub port `postgresql-server`, then just set a variable `postgresql_version` to distinguish between the ports.

I'm not familiar with all the nuances and dependencies of all these ports. @jyrkiwahlstedt is something you could tackle? There are a few newer things in `postgresql96-server` that do not appear in the other ports.

ryandesign:
> The only sane way to manage that is to combine them all as subports within a new stub port `postgresql-server`, then just set a variable `postgresql_version` to distinguish between the subports.

That is one possible method, but it is not the method currently employed by these ports, and switching them to that method is not required to achieve the goal, and may not be desired by the ports' maintainer.

PR #13480, opened 2021-12-26
python27: fix support for virtualenv on macOS 12

buchi:
#### Description

On macOS 12 (Monterey) `NSLibraryNameForModule` returns the real path of the dynamic library instead of the path to the symlink in a virtualenv. This breaks support for virtualenv because then `sys.prefix` points to the path of the Python installation instead of the path of the virtualenv.

By applying the same patch as Apple does for its Python version, which no longer uses the deprecated `NSLibraryNameForModule` function, support for virtualenv is restored.

See also:
https://github.com/apple-oss-distributions/python/blob/python-136.120.2/2.7/fix/getpath.c.ed

Fixes building mozjs60 on x86_64 as well: https://trac.macports.org/ticket/61606#comment:2

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

- [x] tried existing tests with `sudo port test`? (4 tests fail, but these already fail before this change)

### DIFF
--- a/lang/python27/Portfile
+++ b/lang/python27/Portfile
@@ -9,7 +9,7 @@ name                python27
 epoch               2
 # Remember to keep py27-tkinter and py27-gdbm's versions sync'd with this
 version             2.7.18
-revision            6
+revision            7
 
 set major           [lindex [split $version .] 0]
 set branch          [join [lrange [split ${version} .] 0 1] .]
@@ -45,8 +45,11 @@ patchfiles          patch-Makefile.pre.in.diff \
                     darwin20.diff \
                     arm.patch \
                     implicit.patch \
-                    openssl_ver.patch \
-                    patch-getpath.diff
+                    openssl_ver.patch
+
+if {${os.platform} eq "darwin" && ${os.major} >= 21} {
+    patchfiles-append patch-getpath.diff
+}
 
 depends_build       port:pkgconfig
 depends_lib         port:bzip2 \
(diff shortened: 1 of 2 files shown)


mascguy:
Thomas, welcome to MacPorts, and thank you for your first contribution!

This also seems like a worthwhile change. But I'll defer to the `python27` maintainer, @jmroot, for thoughts/feedback.

jmroot:
I'll probably need some time to grok what this is doing and why it helps. Please don't merge in the meantime.

kencu:
of some interest, the noted repo:

https://github.com/apple-oss-distributions

seems to have nicely collected many or all of Jeremy's (and other folks, presumably) patches that are used by Apple when including open-source software in macOS.

Some of that might be useful...there are other dumps of apple's open source mods available, but this one seems quite recent and well organized.

jmroot:
That appears to just be the new place Apple keeps the files that were previously only hosted at https://opensource.apple.com/source/. The front page now only links to https://opensource.apple.com/releases/ which points  you to the GitHub repo.

kencu:
I see more there…like this python patch here. 

But perhaps if you dug hard enough you’d find the patches on the old site we all have used for many many years now. I didn’t spend hours comparing.

Anyway, interesting they made a whole new dump available…

jmroot:
> But perhaps if you dug hard enough you’d find the patches on the old site we all have used for many many years now.

Yes, you would. https://opensource.apple.com/source/python/python-136.120.2/2.7/fix/getpath.c.ed

kencu:
Hey Josh, thanks very much for going to the trouble to find that example. Good to know -- I guess that Apple patch has been sitting there all these years after all.

I still like the organization of the new site -- I think it will prove very useful, to me at least, and likely others.

jmroot:
Yes, it's certainly nice that they've finally given that site an update.

PR #13492, opened 2021-12-26
hadoop, mahout: delete neglected ports

chrstphrchvz:
`hadoop`: Many upstream releases published since port was last updated in 2013. Build issues were never resolved.
Closes: https://trac.macports.org/ticket/40393
Closes: https://trac.macports.org/ticket/46667
Closes: https://trac.macports.org/ticket/49113
Closes: https://trac.macports.org/ticket/53595

`mahout`: Port last updated in 2014; outdated since release of 0.10.0 in 2015. Depends on hadoop which is outdated and broken.

These ports have no other dependents.

###### 
###### Verification 
Have you


pmetzger:
I agree with this. There's no point in keeping this stuff in when it's ancient and broken, and clearly no one wanted these enough to fix them to work after many years.

mascguy:
Given that Hadoop still has plenty of industry momentum, I'd prefer that we keep it, and update to the latest release. And if there aren't any takers, I'm also willing to update and fix the issues.

CCing @essandess, in case Steve has any interest in this port/space...

mascguy:
As for `mahout`, lets keep as well. Once Hadoop is fixed, that'll be tackled too.

pmetzger:
How is someone supposed to run Hadoop on a single mac? I doubt there are actually any users. If there were, we would have gotten significant pushback in bug reporsts before now. People who use it have it installed in huge fleets of machines and have people dedicated to its care and feeding.

I've found from personal experience that maintaining a port you yourself don't use is also hard; you won't notice problems with the usability, and it's hard to stay motivated. 

I won't tell you not to make it work; it's harmless and you get to pick what you do with your time. But it's also not likely to get any users.

mascguy:
> How is someone supposed to run Hadoop on a single mac? I doubt there are actually any users. If there were, we would have gotten significant pushback in bug reporsts before now. People who use it have it installed in huge fleets of machines and have people dedicated to its care and feeding.
> 
> I've found from personal experience that maintaining a port you yourself don't use is also hard; you won't notice problems with the usability, and it's hard to stay motivated.
> 
> I won't tell you not to make it work; it's harmless and you get to pick what you do with your time. But it's also not likely to get any users.

All true. However, the port is still quite useful for folks interested in Hadoop development.

As for the time investment: Since I've been developing heavily in the Java space for years - and already have a lot of familiarity with the various build tools involved - I don't expect this to require much effort.

PR #13503, opened 2021-12-27
haskell_stack-1.0 PG: Fix architecture checks on M1

essandess:
* Fixes: https://trac.macports.org/ticket/63944

stack builds x86_64 binaries on both arm64 and x86_64 architectures 

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 arm64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/haskell_stack-1.0.tcl
+++ b/_resources/port1.0/group/haskell_stack-1.0.tcl
@@ -70,8 +70,8 @@ post-extract {
     xinstall -m 0755 -d "[option haskell_stack.stack_root]"
 }
 
-# stack builds x86_64 binaries
-supported_archs     x86_64
+# stack builds x86_64 binaries on both arm64 and x86_64 architectures
+supported_archs     arm64 x86_64
 
 # libHSbase shipped with GHC links against system libiconv, which provides the
 # 'iconv' symbol, but not the 'libiconv' symbol. Because the compilation


kencu:
I can't exactly guess, so when the arm mac builds a haskell port using this addition, what do they look like? 

Does macports think it installed an arm port (and registered it as such)?

But is it really arch x86_64?

essandess:
> I can't exactly guess, so when the arm mac builds a haskell port using this addition, what do they look like?
> 
> Does macports think it installed an arm port (and registered it as such)?
> 
> But is it really arch x86_64?

This is stated both in the code and above:
> stack builds x86_64 binaries on both arm64 and x86_64 architectures

The binaries it builds run on both arm64 and x86_64 architectures

kencu:
does port think it installed an arm arch? Is it it really x86_64?

If so, no bueno.

kencu:
nice try tho, I tried to sell this idea myself, but shot down 100% by the macports managers.

essandess:
> does port think it installed an arm arch? Is it it really x86_64?
> 
> If so, no bueno.

Yes, bueno. x86_64 binaries run on arm; therefore, haskell-built ports supports both arm64 and x86_64. Otherwise there’s a bug in MacPorts identified in the trac ticket.

kencu:
sell it to Ryan, then. He turned this idea down 10 times already :)

kencu:
as you know.

essandess:
I do not know what you are referring to.

The only fixes I'm aware for the haskell-related arm/intel install issues is to:
1. Add the `supported_archs` for both Arm and Intel because both work.
2. Remove the `supported_archs` directive completely.
3. Fix whatever MacPorts issue causes https://trac.macports.org/ticket/63944

The first is the obvious, simplest and I believe correct approach.

kencu:
> I do not know what you are referring to.

I am simply referring to any "fix" whereby macports thinks that it has installed a port with architecture "arm64" where in the final analysis the files on the disk are of architecture "x86_64".

I also proposed such a solution, "for now". In fact, I implemented it a year ago, and you reverted it!

https://github.com/macports/macports-ports/commit/3ca592edfd14722597e29f8f18c8c781a5b52b56#diff-c925442da5b6f614ff77da74d25253ccb45e29e1b855fdc91ab58ed3c4ede6be

I have since proposed again, and Ryan has forcefully come out against the plan.

So -- unless things have changed in a significant way -- which they might well have -- a proposal that implements installing an x86_64 port under the guise of being an arm64 port has been disallowed.

kencu:
as you know.

essandess:
@ryandesign May we please get some of your cycles on this PR?

As it stands on M1, `haskell_stack` port builds fail with an architecture mismatch error, e.g.
```bash
sudo port -pNc destroot adblock2privoxy
Error: Cannot install adblock2privoxy for the arch 'x86_64' because
Error: its dependency stack is only installed for the arch 'arm64'
Error: and does not have a universal variant.
Error: Unable to execute port: architecture mismatch
```

I'm happy to kick this upstairs to a general MacPorts bug if necessary, but something is definitely broken in the logic of building with `stack` on M1, and this PR fixes it.

essandess:
Fixed by https://github.com/macports/macports-ports/commit/b1e5f733d4076a9ae947ccfc71dd91856ebf39fb

PR #13512, opened 2021-12-28
go-tools: update to 0.1.8

herbygillot:
CC: @mascguy - this PR also _restores_ the `bundle` binary.  Instead of wholesale _deleting_ a binary from a port to prevent conflict, we should seek to just rename it instead. In this way users of the port will still have all the binaries that they expect would be there, while at the same time avoiding path conflicts. 

###### Verification 
Have you

### DIFF
--- a/devel/go-tools/Portfile
+++ b/devel/go-tools/Portfile
@@ -3,22 +3,21 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/golang/tools 0.1.7 v
+go.setup            github.com/golang/tools 0.1.8 v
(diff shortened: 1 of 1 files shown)


mascguy (approved):

mascguy:
In this particular case, I believe the consensus was that `bundle` is a component only used internally. So that's why it was deleted.

But I absolutely agree with you in-principle: We should rename binaries to avoid conflicts, rather than removing them.

mascguy:
For anyone interested, see comments on previous PR, for reasoning behind deletion:

https://github.com/macports/macports-ports/pull/12501

PR #13526, opened 2021-12-29
lporg: new port

tsln1998:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.1
Xcode 13.1

###### Verification 
Have you

### DIFF
--- a/sysutils/lporg/Portfile
+++ b/sysutils/lporg/Portfile
@@ -1,0 +1,224 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/blacktop/lporg 20.4.7 v
+categories          sysutils
+maintainers         {@tsln1998 gmail.com:tsln1998} openmaintainer
+license             MIT
+platforms           darwin
+revision            0
+
+description         Organize Your macOS Launchpad Apps
+
+long_description    This project is meant to help people setting up a brand new Mac \
+                    or to keep all of their Launchpad Folders in sync across devices.
+
+checksums           ${distname}${extract.suffix} \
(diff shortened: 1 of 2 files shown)


sysutils/lporg/Portfile
```diff
@@ -0,0 +1,224 @@
...
+                    github.com/AlecAivazis/survey \
+                        lock    v2.3.2 \
+                        rmd160  4d4ffa6f9a7088f295ced08f60ba3a0bbbd39cb4 \
+                        sha256  5420c336cc99a5a92df0b24742f1217ab2e57ddeca497ca9ea689a2d2c5d4ff2 \
+                        size    1239289
+
+set rfc3339 [exec /bin/sh -c "/bin/date +%Y-%m-%dT%T%z | /usr/bin/sed 's/..$/:&/'"]
+
+build.env-append    CGO_ENABLED=1
+build.post_args     -ldflags '-s -w -X main.Version=${version} -X main.BuildTime=${rfc3339}'
```
herbygillot, line 220: Although we do have some ports doing so, it's generally encouraged not to include `BuildTime`.  This is because it makes the build less reproducible; if we build the exact same port with the exact same dependencies one or more times, the build result will be slightly different because the `BuildTime` will be different each time.  So it would be good to consider removing it or setting it to always blank.

tsln1998:
The source code has been patched to delete BuildTime 😶

herbygillot:
> The source code has been patched to delete BuildTime 😶

OK, I didn't expect that you were going to patch the actual Go code - I thought you could accomplish the same by just doing something like `'-X main.BuildTime='` in the Portfile.

But since you are actively patching the BuildTime parameter out, then you no longer need the `rfc3339` variable in the Portfile, and you also don't need `'-X main.BuildTime=${rfc3339}'`. So I think let's remove those and this should be good to go.

tsln1998:
BuildTime is referenced by other unused variables in the source code, so I want to keep it.

However, it has been determined that this variable has no use in the current version.

herbygillot:
OK, this looks good. Merging.  Thank you for your contribution to MacPorts, @tsln1998.

PR #13534, opened 2021-12-29
par: add universal support

casr:
#### Description

  - use makefile port group (for universal support)
  - add livecheck

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/textproc/par/Portfile
+++ b/textproc/par/Portfile
@@ -35,7 +35,7 @@ variant i18n description adds support for multibyte characters {
 makefile.override   CC CFLAGS LDFLAGS
 
 build.target
-build.pre_args-append    -f protoMakefile
+build.args          -f protoMakefile
 
 destroot {
     xinstall "${worksrcpath}/${name}" "${destroot}${prefix}/bin/"
--- a/textproc/par/Portfile
+++ b/textproc/par/Portfile
@@ -10,7 +10,7 @@ version             1.52
 revision            3
 categories          textproc
 platforms           darwin
-maintainers         {gmail.com:qbarnes @qbarnes}
+maintainers         {gmail.com:qbarnes @qbarnes} openmaintainer
 description         paragraph reflow for email
 long_description    ${description}
 license             Permissive
--- a/textproc/par/Portfile
+++ b/textproc/par/Portfile
@@ -35,10 +35,7 @@ variant i18n description adds support for multibyte characters {
 makefile.override   CC CFLAGS LDFLAGS
 
 build.target
-
-post-patch {
-    file copy "${worksrcpath}/protoMakefile" "${worksrcpath}/Makefile"
-}
+build.pre_args-append    -f protoMakefile
 
 destroot {
     xinstall "${worksrcpath}/${name}" "${destroot}${prefix}/bin/"
@@ -46,4 +43,4 @@ destroot {
 }
 
 livecheck.type      regex
-livecheck.regex     "Par-(\\d+(?:\\.\\d+)(?:\\.\\d+)).tar.gz"
+livecheck.regex     {Par-(\d+(?:\.\d+)*)\.tar\.gz}
--- a/textproc/par/Portfile
+++ b/textproc/par/Portfile
@@ -3,13 +3,14 @@
 # Based on: http://www.openbsd.org/cgi-bin/cvsweb/ports/textproc/par/Makefile
 
 PortSystem          1.0
+PortGroup           makefile 1.0
 
 name                par
 version             1.52
-revision            2
+revision            3
 categories          textproc
 platforms           darwin
-maintainers         gmail.com:qbarnes
+maintainers         {gmail.com:qbarnes @qbarnes}
 description         paragraph reflow for email
 long_description    ${description}
 license             Permissive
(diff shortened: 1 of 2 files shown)


textproc/par/Portfile
```diff
@@ -18,26 +19,31 @@ master_sites        ${homepage}
...
-                        -c ${configure.optflags} [get_canonical_archflags cc]" \
-                    LINK1="${configure.cc} [get_canonical_archflags ld]"
+makefile.override   CC CFLAGS LDFLAGS PREFIX MANPREFIX
+
 build.target
 
+post-patch {
+    file copy "${worksrcpath}/protoMakefile" "${worksrcpath}/Makefile"
+}
+
```
casr, line 42: I realise this was previously covered by `build.args   -f protoMakefile`, however, when I was doing the upgrade to 1.53.0 it had a test phase that would also need this. Therefore it seemed less awkward to rename than to reference it.

Happy to do it more like we have here too though: https://github.com/macports/macports-ports/blob/28a01613446ef5e25dbdf51b5fd65084e2dfbb65/sysutils/entr/Portfile#L32-L34

pmetzger:
@qbarnes

qbarnes:
The changes to `protoMakefile` look great!  Nice to bring it up to a more standard use.

A lot of the changes to `Portfile` I don't know how to read since I haven't kept up on necessary macports changes, but as long as they pass the live build process, they must be good to go too.

casr:
Thanks for taking a look.

> The changes to `protoMakefile` look great! Nice to bring it up to a more standard use.

I'll see if we can upstream these changes. I bet it makes it awkward for many other package managers too.



> A lot of the changes to `Portfile` I don't know how to read since I haven't kept up on necessary macports changes, but as long as they pass the live build process, they must be good to go too.

Yeah the [makefile port group](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/makefile-1.0.tcl) was a new one for me too. It doesn't appear to have documentation so I resorted to reading the source but does seem to make ports like this a little less easier to work with.

textproc/par/Portfile
```diff
@@ -3,13 +3,14 @@
...
+PortGroup           makefile 1.0
 
 name                par
 version             1.52
-revision            2
+revision            3
 categories          textproc
 platforms           darwin
-maintainers         gmail.com:qbarnes
+maintainers         {gmail.com:qbarnes @qbarnes}
```
pmetzger, line 13: Can `openmaintainer` be added?

casr: Added.

textproc/par/Portfile
ryandesign: Since the port doesn't enable the test phase (i.e. doesn't say `test.run yes`) and overrides the destroot phase, it only needs the `-f` flag in `build.args` so that would be preferable to copying the Makefile needlessly.

Even if the port did need the flag in several phases, it takes but a moment to write the code that copies e.g. `build.args` to `destroot.args` (which would be `destroot.args {*}${build.args}`).

textproc/par/Portfile
```diff
@@ -18,26 +19,31 @@ master_sites        ${homepage}
...
+    file copy "${worksrcpath}/protoMakefile" "${worksrcpath}/Makefile"
+}
+
 destroot {
     xinstall "${worksrcpath}/${name}" "${destroot}${prefix}/bin/"
     xinstall -m 444 "${worksrcpath}/${name}.1" "${destroot}${prefix}/share/man/man1/"
 }
+
+livecheck.type      regex
+livecheck.regex     "Par-(\\d+(?:\\.\\d+)(?:\\.\\d+)).tar.gz"
```
ryandesign, line 49: You can save yourself a lot of escaping here by using curly braces instead of quotation marks, since you aren't using variable substitution. (And if you were using variable substitution, then the quotation marks would be superfluous, as they are in many other places in this Portfile.)

Unless there's something special about the way this project does version numbers, I would think there isn't a reason to limit livecheck to just version numbers with one, two, or three components. Removing that limit simplifies the regex too.

```suggestion
livecheck.regex     {Par-(\d+(?:\.\d+)*)\.tar\.gz}
```

casr: Ah, that's great! Thanks for that. I've added in 49aaeb5

textproc/par/Portfile
casr: Added in 61c6e2f.

For future reference, what does the the `{*}` do at the beginning of `{*}${build.args}`?

pmetzger:
Is this ready to merge?

casr:
@pmetzger think it is good from my end

pmetzger:
Gah. Apologies to all. Instead of doing a squash and merge I did a rebase and merge so all four commits are now in history. This was not intended.

neverpanic:
Came here to complain, you already saw it. Shit happens ¯\\\_(ツ)\_/¯

textproc/par/Portfile
ryandesign: It is the [expand operator](https://wiki.tcl-lang.org/page/%7B%2A%7D).

`destroot.args`, like all MacPorts options, is a list.

1. `destroot.args a b` sets it to a two-item list: a b.
2. `set mylist [list a b]` `destroot.args $mylist` would not be the same as (1): it would set `destroot.args` to a one-item list, and that one item is a two-item list a b.
3. To accomplish the same thing as (1) with a variable you would need to expand the list `mylist` into each of its elements using the expand operator: `destroot.args {*}$mylist`. Now `destroot.args` is a two-item list a b.

PR #13550, opened 2022-01-01
qgis: obsolete port

mascguy:
#### Description

Obsolete port, per discussion on ticket: https://trac.macports.org/ticket/63836


pmetzger:
Merged. We should probably close the associated ticket.

PR #13551, opened 2022-01-01
mrustc: new port

evanmiller:
#### Description

[mrustc](https://github.com/thepowersgang/mrustc) is an alternative Rust compiler suitable for bootstrapping Rust. The immediate motivation is to provide a basic Rust compiler on Snow Leopard, which isn't supported by the main `rust` port.

In the future `mrustc` might be used as a stage0 compiler for `rust`, but that's beyond the scope of this work.

`mrustc` requires a run-time C compiler. Since the system GCC on 10.6 cannot compile `mrustc`'s generated C files (which use thread-local storage), I have patched things to use `clang-mp-9.0` as the default run-time compiler on all platforms. This default can be overridden at run-time with `CC`. The choice of default compiler will be a good candidate for port variants in the future.

Tested through compiling and running a basic "Hello, world!" Rust program, so I think it's ready for others to play around with.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/lang/mrustc/Portfile
+++ b/lang/mrustc/Portfile
@@ -1,0 +1,104 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem      1.0
+PortGroup       github 1.0
+
+github.setup    thepowersgang mrustc 4bee8893e8f471d315c1d2d9535216b116cdfe23
+
+version         20220111
+revision        0
+categories      lang devel
+license         MIT
+maintainers     nomaintainer
+description     Alternative implementation of Rust
+supported_archs x86_64
+
+long_description \
+                In-progress alternative rust compiler. Capable of building a \
+                fully-working copy of rustc, but not suitable for everyday use \
+                (due to inadequate error messages).
+
+master_sites-append   https://static.rust-lang.org/dist/
+
+set rust_version 1.39.0
+set rust_version_major [join [lrange [split ${rust_version} .-] 0 1] .]
+set rust_version_major_underscore [join [lrange [split ${rust_version} .-] 0 1] _]
+distfiles-append rustc-${rust_version}-src.tar.gz
+
+checksums       mrustc-${github.version}.tar.gz \
+                rmd160  c16b5b1ceb5f79f61712c22488662a0c68464ea9 \
+                sha256  28f53fca94ef4edec4f1e37d0d9d3c12bca5b4b867dbb2a62e80692afaee141e \
+                size    1173691 \
+                rustc-${rust_version}-src.tar.gz \
+                rmd160  d5b04b87fc336e3be7d592f70de0363aa66622aa \
+                sha256  b4a1f6b6a93931f270691aba4fc85eee032fecda973e6b9c774cd06857609357 \
+                size    152803201
+
+use_configure   no
+
+compiler.cxx_standard 2014
+compiler.c_standard 2011
+
+# Thread-local storage is not needed to compile mrustc, but is required
+# for compiling Rust files (e.g. the standard library)
+compiler.thread_local_storage yes
+
+# A C compiler is needed at run-time; it doesn't have to be the same version as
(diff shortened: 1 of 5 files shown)


evanmiller:
> Installing dependency (9 of 25) 'xz' with variants '' ... [FAIL]

Anyone know why this fails on CI?

herbygillot:
> Anyone know why this fails on CI?

I've triggered the CI jobs to re-run.

mascguy (approved):

evanmiller:
Updated the code and added a test target (which passes locally).

mascguy:
@catap Kirill, you might be interested in this too!

catap:
@mascguy thanks, it is interesting. I haven't got a time to finish 10.6 :) maybe someone else can do it.

@evanmiller FYI I've backported LLVM changes: https://reviews.llvm.org/D115250?id=392731 maybe you may suggest to use this patch at upstream of this tool

kencu:
My thread_local hack for llvm for <10.7 requires other changes in the built libc++abi to actually work, and should be considered "macports-only".

Upstream llvm will never accept them, as they have long dropped support for older darwin systems (the only reason older darwin works on macports is other patches and config hacks I made, and the stuff in legacysupport).

We should not be trying to upstream these hacks anywhere --just embarassing to do so, when people try them they won't work -- and consider them our mess to maintain while we do.

If you still try to upstream my tlv patches somewhere like llvm, please don't put my name on them!!!!

lang/mrustc/Portfile
```diff
@@ -0,0 +1,87 @@
...
+
+use_configure   no
+
+compiler.cxx_standard 2014
+compiler.c_standard 2011
+
+# This will be the default compiler, which can be overridden
+# at run-time with the CC environment variable.
+# Support variants in the future, but this will do for now
+depends_run-append port:clang-9.0
```
cjones051073, line 45: It would be better to not hardcode a version here, but to tweak the compiler backlisting etc. to ensure a compatible compiler is picked, and then to depend on that. This would allow newer compilers to be used when, e.g., clang-9.0 is not available (such as on newer systems which inevitably when one day stop supporting this compiler).

Note to add the correct runtime dep you can use the same logic as in rust.

https://github.com/macports/macports-ports/blob/a34b9f304f3a13bb9239e2ae9917ee3a41de3e9f/lang/rust/Portfile#L210

lang/mrustc/Portfile
```diff
@@ -0,0 +1,87 @@
...
+
+# Set the default compiler
+patchfiles-append patch-mrustc-compiler.diff
+
+# Set the default Rust version to match our target
+patchfiles-append patch-mrustc-version.diff
+post-patch {
+    reinplace "s|@MACPORTS_MRUSTC_LIBDIR@|${prefix}/lib/mrustc-${rust_version_major}|" ${worksrcpath}/src/main.cpp
+    reinplace "s|@MACPORTS_MRUSTC_TARGET_VER@|${rust_version_major_underscore}|" ${worksrcpath}/src/main.cpp
+    reinplace "s|@MACPORTS_MRUSTC_CC@|clang-mp-9.0|" ${worksrcpath}/src/trans/codegen_c.cpp
```
cjones051073, line 67: once you have correctly set the compiler backlisting, please use ${configure.cxx} here instead of hardcoding the compiler name.

cjones051073 (changes requested):

evanmiller:
Updated the code and adopted @cjones051073's suggestions regarding the run-time compiler.

This build may fail due to some recent changes in mrustc, reported here: https://github.com/thepowersgang/mrustc/issues/210

evanmiller:
I have isolated and fixed the recent local build failures described in the linked issue.

Updated the code and Portfile to consolidate patches and have better GCC compatibility.

I have a slight preference to wait until https://github.com/thepowersgang/mrustc/pull/192 is merged, but if CI passes this PR should be ready as-is.

evanmiller:
@pmetzger Will do. Still working through a GCC-related issue over here: https://github.com/thepowersgang/mrustc/issues/214

evanmiller:
This is OK to commit; I will address any remaining issues as they are resolved upstream.

PR #13556, opened 2022-01-03
jreleaser: cleanup

ryandesign:
Use github portgroup

Don't use `foreach` where it's unnecessary

Remove `distname ${name}-${version}` since that'd the default

### DIFF
--- a/devel/jreleaser/Portfile
+++ b/devel/jreleaser/Portfile
@@ -46,14 +46,10 @@ destroot {
     xinstall -m 755 -d ${target}
 
     # Copy over the needed elements of our directory tree
-    foreach f [glob -dir ${worksrcpath} *] {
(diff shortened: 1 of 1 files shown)


aalmiray:
LGTM. Thank you! 😁

mascguy:
Andres, since this portfile is generated, will these changes be lost on the next update? Or are you able to incorporate these improvements in your template, so that they'll be included?

aalmiray:
@mascguy I've added the modifications to the template used to generate this specific Portfile https://github.com/jreleaser/jreleaser/commit/35bfd781a86c28b2f6be895aea73f82e62c30c26

Will also add an option for anyone else using JReleaser and following the GH conventions so that they also can benefit form the github port group.

PR #13560, opened 2022-01-03
cdrtools: fix arm64 build by explicitly disabling stack scanning

markmentovai:
#### Description

avoffset, a helper program run during the cdrtools build, was crashing on arm64 when built with optimization, with `EXC_BREAKPOINT` (`SIGTRAP`), triggered by a `brk` instruction inserted by the compiler in a code path that was provably attempting to dereference a null pointer. Under the same conditions on x86_64, the compiler inserted a `ud2` instruction resulting in `EXC_BAD_INSTRUCTION` (`SIGILL`). On either architecture, when not under optimization, the null pointer dereference resulted in `EXC_BAD_ACCESS` (`SIGSEGV`). avoffset has signal handlers for both `SIGILL` and `SIGSEGV`, and recovered “gracefully” by configuring the build to disable stack scanning and exiting cleanly. Without a `SIGTRAP` handler, the avoffset crash results in a failed build on arm64 under optimization.

It would be easy to add a `SIGTRAP` handler, but a more direct approach is to explicitly disable stack scanning by defining `NO_SCANSTACK` during the build. The stack scanning concept and implementation are both ill-advised, and none of it was working on any architecture or any build configuration on macOS anyway. Even with avoffset “working” (not crashing), all it did on macOS was detect the failure and disable stack scanning for the rest of the build.

Closes: https://trac.macports.org/ticket/63062

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

MacBookPro18,2
macOS 12.1 21C52 arm64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/sysutils/cdrtools/Portfile
+++ b/sysutils/cdrtools/Portfile
@@ -52,6 +52,7 @@ use_configure       no
 compiler.blacklist  llvm-gcc-4.2 macports-llvm-gcc-4.2 {clang < 300}
 
 configure.cflags-append \
+                    -DNO_SCANSTACK \
                     -Wno-error=implicit-function-declaration
 
 build.cmd           smake


ryandesign:
> defining `NO_SCANSTACK` during the build

I see you did not increase the port's `revision`, so that means this change only affects avoffset and other tools used only during the build? It does not change the contents of any files that get installed?

markmentovai:
> > defining `NO_SCANSTACK` during the build
> 
> I see you did not increase the port's `revision`, so that means this change only affects avoffset and other tools used only during the build? It does not change the contents of any files that get installed?

Yes. The installed output is bit-for-bit identical (assuming `ZERO_AR_DATE` is set to ensure static libraries don’t wind up with non-deterministic timestamps, as expected). I did verify this on x86_64, the only architecture for which there are existing successful builds of the cdrtools port.

Detailed analysis:

The use of stack scanning is disabled in any event: nothing in the build sets `USE_SCANSTACK` _except_ for the build of avoffset, which determines stack characteristics when `HAVE_SCANSTACK` is defined (which it normally is). But because `USE_SCANSTACK` is undefined elsewhere, avoffset’s output is irrelevant to the rest of the build. Even if `USE_SCANSTACK` was defined elsewhere, every part of the source code that considers `HAVE_SCANSTACK` also checks the macros that avoffset defines, namely, `FP_INDIR` and `AV_OFFSET`, neither of which will have been defined by an avoffset that determined that stack scanning was impossible, as it always does on macOS. With either of those macros undefined, `HAVE_SCANSTACK` would also be undefined.

The larger `NO_SCANSTACK` switch undefines `HAVE_SCANSTACK` so that avoffset does not even attempt to make this irrelevant determination.

markmentovai:
@ryandesign per the analysis above in https://github.com/macports/macports-ports/pull/13560#issuecomment-1006672722, are we good with not bumping `revision`?

pmetzger:
LGTM. I will merge this.

PR #13590, opened 2022-01-07
vscode: new port

sainnhe:
#### Description

Add new port [vscode](https://github.com/microsoft/vscode).

The official vscode release contains tracks/telemetry. This port however, compiles from source code and doesn't contain any tracks/telemetry binaries.

There are two available variants:

- `stable` (default): patching product.json to use vscode marketplace, enable settings sync, update extension proposed api to unblock some features in stable release.
- `insiders`: like stable but update extension proposed api to unblock some features in insider release, also the sync url is set to insider channel.

###### Tested on

macOS 11.6.2 20G314 arm64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/editors/vscode/Portfile
+++ b/editors/vscode/Portfile
@@ -1,0 +1,78 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+
+github.setup            microsoft vscode 1.63.2
+revision                0
+categories              editors
+maintainers             {@sainnhe gmail.com:sainnhe} \
+                        openmaintainer
+license                 MIT
+supported_archs         x86_64 arm64
+
+description             Visual Studio Code - Open Source (Code - OSS)
+
+long_description \
+    Visual Studio Code combines the simplicity of a code editor with what developers \
+    need for their core edit-build-debug cycle. It provides comprehensive code editing, \
+    navigation, and understanding support along with lightweight debugging, a rich \
+    extensibility model, and lightweight integration with existing tools.
+
+github.tarball_from     archive
+
+checksums               rmd160  179bf6d4b2738022d8fa7ccedf509ee83898bbf9 \
+                        sha256  21fc9bc17ba4cf480b1e006f298363d86215c339c480f8d781cabcfedad2d624 \
+                        size    15411738
+
+depends_build-append    port:nodejs16 \
+                        port:npm8 \
+                        port:yarn
+
+use_configure           no
+
+patchfiles              patch-postinstall-fix.diff
+
+variant stable conflicts insiders description {Patch with features in VSCode stable release} {
+    patchfiles-append   patch-features-stable.diff
+}
+
+variant insiders conflicts stable description {Patch with features in VSCode insiders release} {
+    patchfiles-append   patch-features-insiders.diff
+}
+
+default_variants        +stable
+universal_variant       no
+
+if {${build_arch} eq "arm64"} {
+    set vscode-arch "arm64"
+} else {
+    set vscode-arch "x64"
+}
+
+build {
+    if {[variant_isset stable]} {
+        system "/usr/bin/curl -sL https://github.com/dhanishgajjar/vscode-icons/raw/master/icns/default.icns -o ${worksrcpath}/resources/darwin/code.icns"
+    } elseif {[variant_isset insiders]} {
+        system "/usr/bin/curl -sL https://github.com/dhanishgajjar/vscode-icons/raw/master/icns/default_insider_light.icns -o ${worksrcpath}/resources/darwin/code.icns"
+    }
+    system "export PYTHON=/usr/bin/python3 && cd ${worksrcpath} && yarn && yarn gulp vscode-darwin-${vscode-arch}"
+}
+
+destroot {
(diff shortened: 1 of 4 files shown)


sainnhe:
Download ripgrep failed because the rate limit on github (unauthenticated) api requests is only 60 request per hour.

Related issue:

- https://github.com/microsoft/vscode-ripgrep/issues/9
- https://github.com/microsoft/vscode/issues/28434

The recommended workaround is to add a `GITHUB_TOKEN` environment variable. Should this environment variable be added to this repository or directly to the Portfile?

sainnhe:
> would it be possible to add it as a dependency (depends_build or depends_lib) and convince the build process to use the MacPorts-provided ripgrep?

It's possible to create a patch to do this, but it's still very hard to stop it from downloading prebuilt ripgrep, because vscode-ripgrep is marked as dependency of some other node packages.

> I don't think that's something that makes sense to do.

In [postinstall.js](https://github.com/microsoft/vscode-ripgrep/blob/8651265884cd9733b1bca26e8475ac498ad632d0/lib/postinstall.js#L77), the token is read via `process.env['GITHUB_TOKEN']`.

IMO it should work if `GITHUB_TOKEN` is provided.

I don't mind to create a new token though. I've added it to Portfile, the CI should be able to run successfully.

This is a simply test token, feel free to change it or remove it.

sainnhe:
Looks like the github token can't be committed. Github will scan valid tokens in commits and they will be revoked automatically when github found them.

sainnhe:
IMO the best solution might be generating a github token and add it to `/.github/workflows/main.yml`. Do you have any other ideas?

aeiouaeiouaeiouaeiouaeiouaeiou:
Maybe porting [vscodium](https://github.com/VSCodium/vscodium) would be a better alternative.

editors/vscode/files/patch-postinstall-fix.diff
```diff
@@ -0,0 +1,10 @@
+--- build/npm/postinstall.js.orig	2022-01-06 22:30:54.000000000 +0800
++++ build/npm/postinstall.js	2022-01-07 17:28:45.000000000 +0800
+@@ -90,5 +90,5 @@
+ 	yarnInstall(watchPath);
+ }
+ 
+-cp.execSync('git config pull.rebase merges');
+-cp.execSync('git config blame.ignoreRevsFile .git-blame-ignore');
++// cp.execSync('git config pull.rebase merges');
++// cp.execSync('git config blame.ignoreRevsFile .git-blame-ignore');
```
ryandesign, line 10: Isn't it better, simpler, shorter, easier to read for the patch to remove these lines rather than to comment them out?

editors/vscode/Portfile
```diff
@@ -0,0 +1,78 @@
...
+
+if {${build_arch} eq "arm64"} {
+    set vscode-arch "arm64"
+} else {
+    set vscode-arch "x64"
+}
+
+build {
+    if {[variant_isset stable]} {
+        system "/usr/bin/curl -sL https://github.com/dhanishgajjar/vscode-icons/raw/master/icns/default.icns -o ${worksrcpath}/resources/darwin/code.icns"
```
ryandesign, line 55: Don't download things manually with `curl`. Instead, add these files to `distfiles`. You may want to use [the `?dummy=` hack](https://trac.macports.org/wiki/PortfileRecipes#fetchwithgetparams) to rename the distfile so that it is not subject to the problems associated with [unversioned distfiles](https://trac.macports.org/wiki/PortfileRecipes#unversioned-distfiles).

editors/vscode/Portfile
```diff
@@ -0,0 +1,78 @@
...
+    } elseif {[variant_isset insiders]} {
+        system "/usr/bin/curl -sL https://github.com/dhanishgajjar/vscode-icons/raw/master/icns/default_insider_light.icns -o ${worksrcpath}/resources/darwin/code.icns"
+    }
+    system "export PYTHON=/usr/bin/python3 && cd ${worksrcpath} && yarn && yarn gulp vscode-darwin-${vscode-arch}"
```
ryandesign, line 59: Don't `cd` within `system`; use the `-W` argument instead.
```suggestion
    system -W ${worksrcpath} "export PYTHON=/usr/bin/python3 && yarn && yarn gulp vscode-darwin-${vscode-arch}"
```

editors/vscode/Portfile
```diff
@@ -0,0 +1,78 @@
...
+
+variant stable conflicts insiders description {Patch with features in VSCode stable release} {
+    patchfiles-append   patch-features-stable.diff
+}
+
+variant insiders conflicts stable description {Patch with features in VSCode insiders release} {
+    patchfiles-append   patch-features-insiders.diff
+}
+
+default_variants        +stable
```
ryandesign, line 44: Since stable and insiders conflict, stable [should not be the default](https://trac.macports.org/wiki/PortfileRecipes#default_variants) if the user selected insiders.
```suggestion
if {![variant_isset insiders]} {
    default_variants    +stable
}
```

ryandesign (commented):
You're discovering some of the reasons why we don't like ports that download things during the build. If the developers of this software don't already offer a complete source code download file that can build their software, you should suggest to them that they offer this. It is a standard thing for open source software developers to do.

ryandesign:
With complex applications like this one that contain lots of other libraries that they want to download and build, the best solution may be for the user to download a binary from its developer and for it not to be in MacPorts.

sainnhe:
OK, fair enough. I'm going to close this PR.

PR #13516, opened 2021-12-28
discord-rpc: new port

jasonliu--:
#### Description

Originally, I had intended to package the [Discord GameSDK](https://discord.com/developers/docs/game-sdk/sdk-starter-guide), but after stumbling upon discord/discord-rpc#290 and seeing the questions regarding the Discord GameSDK's unclear licensing terms, I went ahead and packaged [discord-rpc](https://github.com/discord/discord-rpc) instead, even though the authors are calling it "deprecated".

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [ ] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 
- [x] checked your Portfile with `port lint`?
- [x] tried a full install with `sudo port -vst install`?
- [x] tested basic functionality of all binary files?

PR #13600, opened 2022-01-07
armagetronad: new port

jasonliu--:
#### Description

[Armagetron Advanced](http://www.armagetronad.org) is a 3-D snake game, based on the light cycle sequence from the movie Tron.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/games/armagetronad/Portfile
+++ b/games/armagetronad/Portfile
@@ -1,0 +1,286 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           boost 1.0
+
+github.setup        ArmagetronAd armagetronad 0.2.9.1.0 v
+github.tarball_from releases
+extract.suffix      .tbz
+
+set pretty_name     "Armagetron Advanced"
+# 'year_released' is the year that the tarball was published
+# on GitHub, not the year that we are currently in.
+set year_released   2021
+
+categories          games
+license             GPL-2+
+maintainers         @jasonliu-- openmaintainer
+                    #  v Note: https site has an expired certificate
+homepage            http://www.${name}.org
+description         $pretty_name, a 3-D Tron-like light cycle game
+long_description    $pretty_name is a 3-D snake game, based on the \
+                    light cycle sequence from the movie Tron. You can \
+                    play against AI opponents, against other human \
+                    users over the network, or a mixture of both.
+
+subport ${name}-common {
+    supported_archs noarch
+    description     common files for the $pretty_name packages
+    long_description-append \
+                    \n\nThis package contains files that are common to \
+                    both the ${name} and ${name}-dedicated packages.
+}
+subport ${name}-dedicated {
+    description     dedicated game server for $pretty_name
+    long_description-append \
+                    \n\nThis package installs the dedicated network \
+                    server for the $pretty_name game. It allows you to \
+                    host games that other online players can connect \
+                    to, and play multi-player matches. You can also \
+                    host online multi-player matches directly from \
+                    inside the game, but the dedicated server allows \
+                    you to continue hosting matches while not actively \
+                    running the game.
+}
+
+distfiles           ${distname}${extract.suffix}:${name}
+checksums           ${distname}${extract.suffix} \
+                    rmd160  76eed9dd9fe6d888eba0fae67696040ec1eccb7d \
+                    sha256  59b6c7c01ce3f8cca5437e33f974a637529541a11aa4f52c1a5c17499e26f6a1 \
+                    size    1943758
+
+if {$subport ne "${name}-common"} {
+    master_sites-append http://deb.debian.org/debian/pool/main/a/${name}:debian_pkg
+    distfiles-append    ${name}_${version}-2.debian.tar.xz:debian_pkg
+    checksums-append    ${name}_${version}-2.debian.tar.xz \
+                        rmd160  cc1151617b372d61b6b02b647adb699fdfba3ea0 \
+                        sha256  28b778440ee2d7a8657fec2a8a468723017645a98259a7afac9fa67841a6c36f \
+                        size    29852
+}
+
+depends_build-append    port:autoconf \
+                        port:automake \
+                        port:pkgconfig
+
+if {$subport eq "${name}-common"} {
+    set depends_type    build
+    boost.depends_type  $depends_type
+} else {
+    set depends_type    lib
+}
+depends_$depends_type-append \
+                        port:libxml2
+if {$subport ne "${name}-dedicated"} {
+    depends_$depends_type-append \
+                        port:freetype \
+                        port:libsdl \
+                        port:libsdl_image \
+                        port:libsdl_mixer \
+                        port:libpng
+}
+if {$subport ne "${name}-common"} {
+    depends_run-append  port:${name}-common
+}
+
+post-patch {
+    if {$subport ne "${name}-dedicated"} {
+        reinplace "/LIBS -framework GLU/a\\
+            \\        LIBS=\"\$LIBS -framework OpenGL\"\\
+        " \
+            ${worksrcpath}/configure.ac
+    }
+
+    # Avoid "error: use of undeclared identifier 'SOCK_CLOEXEC'".
+    # SOCK_CLOEXEC doesn't exist in macOS (Reference:
+    # https://www.mail-archive.com/libguestfs@redhat.com/msg22188.html)
+    reinplace \
+        {/socktype = socktype_/,/socktype |= SOCK_CLOEXEC/s/ndef WIN32/def SOCK_CLOEXEC/} \
+        ${worksrcpath}/src/network/nSocket.cpp
+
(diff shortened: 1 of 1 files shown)


games/armagetronad/Portfile
```diff
@@ -0,0 +1,291 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           boost 1.0
+
+github.setup        ArmagetronAd armagetronad 0.2.9.1.0 v
+github.tarball_from releases
+extract.suffix      .tbz
```
ryandesign, line 9: So it's a bzip2-compressed tarball? Then you need to specify `use_bzip2 yes` so that MacPorts uses the right extraction command. The "wrong" extraction command (the one for gzip-compressed tarballs) happens to work on OS X 10.9 and later, but earlier systems require the correct command.

games/armagetronad/Portfile
```diff
@@ -0,0 +1,291 @@
...
+
+distfiles           ${distname}${extract.suffix}:${name}
+checksums           ${distname}${extract.suffix} \
+                    rmd160  76eed9dd9fe6d888eba0fae67696040ec1eccb7d \
+                    sha256  59b6c7c01ce3f8cca5437e33f974a637529541a11aa4f52c1a5c17499e26f6a1 \
+                    size    1943758
+
+if {$subport ne "${name}-common"} {
+    master_sites-append http://deb.debian.org/debian/pool/main/a/${name}:debian_pkg
+    distfiles-append    ${name}_${version}-2.debian.tar.xz:debian_pkg
```
ryandesign, line 54: Ah, and this is an xz-compressed tarball? MacPorts does not support automatically decompressing multiple distfiles of different types. (I haven't gotten around to adding this capability yet, though I would like to.) You should let MacPorts extract one of the files (perhaps the main .tbz file), list that file in `extract.only`, and extract any additional distfiles with differing compression formats manually in a `post-extract` block with `system`. Don't forget to add `depends_extract bin:xz:xz` if you need to extract xz files.

games/armagetronad/Portfile
```diff
@@ -0,0 +1,291 @@
...
+}
+if {$subport ne "${name}-common"} {
+    depends_run-append  port:${name}-common
+}
+
+post-patch {
+    if {$subport ne "${name}-dedicated"} {
+        reinplace "/LIBS -framework GLU/a\\
+            \\        LIBS=\"\$LIBS -framework OpenGL\"\\
+        " \
```
ryandesign, line 89: As you know from feedback in previous PRs, patchfiles should be used instead of reinplaces unless the reinplace makes the same replacement a prohibitively large number of times or a variable substitution is needed.

games/armagetronad/Portfile
```diff
@@ -0,0 +1,291 @@
...
+        set infoplist_strings \
+            $app_resources/English.lproj/InfoPlist.strings
+        move ${destroot}${infoplist_strings}.in \
+             ${destroot}${infoplist_strings}
+        foreach re [list \
+            "s/@version@/${version}/" \
+            "s/@year@/$year_released/" \
+        ] {
+            reinplace $re ${destroot}$infoplist_strings
+        }
```
ryandesign, line 191: This can be done as a one-liner... and usually when we replace placeholders we use the `g` flag to ensure all occurrences are replaced.
```suggestion
        reinplace "s/@version@/${version}/g;s/@year@/$year_released/g" ${destroot}$infoplist_strings
```

games/armagetronad/Portfile
```diff
@@ -0,0 +1,291 @@
...
+            ] {
+                reinplace -E $re ${workpath}/debian/man/${subport}.6
+            }
+        } elseif {$subport eq ${name}} {
+            reinplace -E \
+                "s|/usr\(/share/doc/\)${name}-common|${prefix}\\1${name}|" \
+                ${workpath}/debian/man/${subport}.6
+        }
+        system -W ${workpath}/debian/man [join [list \
+            "gzip -9c ${subport}.6" \
```
ryandesign, line 278: Don't gzip manpage manually; MacPorts will do it for you. If you gzip them, MacPorts will just ungzip them and regzip them.

ryandesign (commented):
Wooow it's been a long time since I played Armagetron.

games/armagetronad/Portfile
jasonliu--: > MacPorts does not support automatically decompressing multiple distfiles of different types. (I haven't gotten around to adding this capability yet, though I would like to.)

Interesting... it seems to work fine for this Portfile, at least when I tested it on macOS 10.11. I suspect it might have something to do with what you mention in your above comment, about the "wrong" extraction command happening to work on OS X 10.9 and later. Does that mean that I only need to add that stuff for the sake of macOS <= 10.9? Because adding an `extract.only`, `post-extract`, `depends_extract`, etc. seems like a lot of extra work just for the sake of extracting a couple of distfiles, and obviously I wouldn't have submitted the PR if the entire Portfile hadn't worked, and it does seem to work flawlessly, even with `port -vst` turned on.

games/armagetronad/Portfile
jasonliu--: Indeed, and my other PRs have been using more patchfiles, for example in #13516. I was hoping that since this reinplace is only adding a single line, it would be ok? :sweat_smile:

games/armagetronad/Portfile
jasonliu--: Ah nice catch. It's always good to be able to eliminate a for-loop. 👍

games/armagetronad/Portfile
jasonliu--: Oh, I didn't know that. I'll just copy the manpage into destroot, then. 👍

games/armagetronad/Portfile
jasonliu--: After thinking it over, how about this? Since the current form of the portfile seems to "just work", both locally on my 10.11 machine, as well as the 10.15 and 11.0 CI builds, why don't we merge the portfile as-is, and then I can see which macOS versions might be having a problem by looking at the buildbot logs, and tailor a PR for those (probably older) macOSes?

kencu: FYI, I have used this to extract multiple different distfiles of different types in my own ports for some years:
```
extract {
    foreach f ${distfiles} {
        set f [lindex [split ${f} :] 0]
        ui_msg "Extracting: ${f}"
        if {[string match "*bz2*" ${f}]} {
            ui_msg "${workpath} \"/usr/bin/tar xjf ${distpath}/${f}\""
            system -W ${workpath} "/usr/bin/tar xjf ${distpath}/${f}"
        }
        if {[string match "*gz*" ${f}]}  {
            ui_msg "${workpath} \"/usr/bin/tar xzf ${distpath}/${f}\""
            system -W ${workpath} "/usr/bin/tar xzf ${distpath}/${f}"
        }
    }
}
```

add types as needed

jasonliu--:
@pmetzger, before you merge this PR, I have a (hopefully) quick question, which I have asked on the dev mailing list, and I will repeat here.

Should I be adding `supported_archs noarch` to the game data files, similar to the `redeclipse` and `warzone2100` subports (#13644)? I'm asking for this port because the situation here is a bit more unique. The question I should be asking really should be: Is `noarch` meant to specify that the build itself isn't supposed to be architecture-specific, or does it mean that the final output products are not architecture-specific?

Here for `armagetronad`, the subport for the game data files basically runs through the entire build, including compiling the source code, but then in the end simply throws away those compiled objects, and the only files that are retained (i.e. staged into destroot) are the game data files, which are basically just a bunch of text files and images (which are not architecture-specific). It seems that the upstream developers wrote the makefiles this way because some of the game data files get generated during the compile process, and they never bothered to cleanly split off the generating of the game data files into a separate makefile that can be run by itself.

pmetzger:
> The question I should be asking really should be: Is noarch meant to specify that the build itself isn't supposed to be architecture-specific, or does it mean that the final output products are not architecture-specific?

I think the latter, but I'm not sure, this isn't something I think I ever would have known.

jasonliu--:
> > The question I should be asking really should be: Is noarch meant to specify that the build itself isn't supposed to be architecture-specific, or does it mean that the final output products are not architecture-specific?
> 
> I think the latter, but I'm not sure, this isn't something I think I ever would have known.

Well, based on feedback, both from you and from Josh [on the dev mailing list](https://lists.macports.org/pipermail/macports-dev/2022-January/044067.html), I've gone ahead and added a `supported_archs noarch` to the game data files subport. Adding it seemed to make no difference to the build on my local machine. It if dies horribly on the builders, I guess I'll just have to submit another PR to remove it, lol.

I made sure to remember to squash this time! Ok, let's merge and see what happens!

games/armagetronad/Portfile
ryandesign: Only adding a single line is an ideal time to use a patchfile!

games/armagetronad/Portfile
ryandesign: Yes, the problem only affects OS X < 10.9.

PR #13607, opened 2022-01-08
zk: New port

judaew:
#### Description

[zk](https://github.com/mickael-menu/zk) is a command-line tool helping you to maintain a plain text Zettelkasten or personal wiki.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/office/zk/Portfile
+++ b/office/zk/Portfile
@@ -1,0 +1,42 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/mickael-menu/zk 0.9.0 v
+github.tarball_from archive
+revision            0
+
+categories          office
+license             MIT
+maintainers         {judaew @judaew} openmaintainer
+
+description         A plain text note-taking assistant
+long_description    zk is a command-line tool helping you to maintain a plain \
+                    text Zettelkasten or personal wiki.
+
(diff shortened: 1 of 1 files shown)


office/zk/Portfile
```diff
@@ -0,0 +1,39 @@
...
+long_description    zk is a command-line tool helping you to maintain a plain \
+                    text Zettelkasten or personal wiki.
+
+checksums           rmd160  66acc66142193c2830a444f295e1408a9a29b245 \
+                    sha256  102754f2722e7e4ff0a83715085c8bfdac090b440f89020123481a95b566730f \
+                    size    626070
+
+depends_lib-append  port:icu
+
+set date_build [clock format [clock seconds] -format %Y-%m-%d]
```
harens, line 24: As always, thanks @judaew for your work. ⭐  Minor comment from me.

For the purpose of [reproducible builds](https://reproducible-builds.org/), it might be best to suppress the timestamp. You could do something like the following:

https://github.com/macports/macports-ports/blob/82f2ed1c041914bba1c3f8e6df0a5a9077e7aaf5/devel/glab/Portfile#L25-L29

See the [reproducible builds docs on timestamps](https://reproducible-builds.org/docs/timestamps/) for more details. Otherwise, LGTM. 👍

herbygillot (commented):
Looks good, @judaew. I agree with @harens on the subject of the build date. You can do as @harens recommends or leave it blank/unset, but we want to lean towards not setting a build date when we build ports so that they are more reproducible.  I think we actually do have some ports doing that today, but we'll need to fix those going forward.

office/zk/Portfile
judaew: Nice solution, thanks @harens 🌟

PR #13611, opened 2022-01-09
squish: fix dylib version number

jasonliu--:
#### Description

The version-numbered dylib that gets generated is currently hardcoded to `0.0`, which results in a versioned dylib of `libsquish.0.0.dylib`. This commit fixes the incorrect usage of the dylib version number; it will now start generating the versioned dylib as `libsquish.1.15.dylib`. I have left the `SOVERSION 0.0` in place in the `CMakeLists.txt` file, so that `libsquish.0.0.dylib` will still get generated as a (symlink) compatibility version; this means that revbumps to dependent ports are not needed.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/graphics/squish/Portfile
+++ b/graphics/squish/Portfile
@@ -6,7 +6,7 @@ PortGroup           muniversal 1.0
 
 name                squish
 version             1.15
-revision            1
+revision            2
 checksums           rmd160  01a380439426f5a40ad8f42ada58a09d2f1d3fbb \
                     sha256  628796eeba608866183a61d080d46967c9dda6723bc0a3ec52324c85d2147269 \
                     size    59199
@@ -29,6 +29,11 @@ extract.suffix      .tgz
 extract.mkdir       yes
 
 patchfiles-append   add-pkgconfig-file.diff
+
+post-patch {
+    reinplace -E "s/(^\[\[:space:]]+VERSION) 0.0/\\1 ${version}/" \
+        ${worksrcpath}/CMakeLists.txt
+}
 
 configure.args      -DBUILD_SHARED_LIBS=ON \
                     -DBUILD_SQUISH_WITH_OPENMP=OFF


ryandesign:
Developers are free to use any version numbering scheme they want for their libraries, even no version numbering at all. If you think squish should change its library version number, you should suggest that to the developers so they can consider it and include it in a future release. Until they decide to do that, I'm not sure we should second-guess them about this.

You are proposing that the library major version number should be equal to the project version number, which is not necessarily correct. Only the developers of this library know if and when they change the library in such a way that its major version needs to be increased. Many projects are designed to be able to release new versions without needing to increase their library major version. Consider gettext which last changed its library's major version to 8 (libintl.8.dylib) in version 0.18 in 2010 but has had several releases since then.

jasonliu--:
On the one hand, I can see your point about developers being free to use any numbering scheme they want. But on the other hand, it seemed to me that using a numbering scheme of `0.0` was more of a mistake or oversight than a choice. Even your example of gettext still uses a number other than `0` for the library number.

Part of the reason why I believed that the use of `0.0` for the library number was most likely a mistake or oversight, was the fact that in the previous squish-related PR (#13459), I had to [reinplace the version number](https://github.com/macports/macports-ports/pull/13459#discussion_r775101798) in the `libsquish.pc` pkg-config file, because even though the tarball was tagged as version 1.15, the `libsquish.pc` file was still hard-coded as 1.14. Seems like to me that they aren't managing their version numbering very well, no?

I will try to get in contact with the developer to make changes upstream when I have some time, as discussed in #13459, but I've been busy with work from my clients recently, so I don't know when that will be.

ryandesign:
> Seems like to me that they aren't managing their version numbering very well, no?

That may be, but it's for them to decide, especially when it comes to dylib version numbers since those get baked into anything using the library. Looking at my installed ports, I have 104 dylibs with major version "0"; 3 including libsquish with major version "0.0", 22 with major version "0.0.0", and around 300 with no major version at all. So this alone does not necessarily indicate a problem.

PR #13614, opened 2022-01-10
gn-devel: build tests only during test phase

chrstphrchvz:
#### Description
No revision bump because this does not affect installed files.

###### Tested on

macOS 12.1
Xcode 13.2.1

###### Verification 
Have you

### DIFF
--- a/devel/gn-devel/Portfile
+++ b/devel/gn-devel/Portfile
@@ -34,7 +34,7 @@ configure.pre_args
 
 build.dir           ${worksrcpath}/out
 build.cmd           ${prefix}/bin/ninja
-build.target
+build.target        gn
 # Build verbosely
 build.args          -v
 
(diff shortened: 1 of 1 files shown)


chrstphrchvz:
@pmetzger does this look okay to merge?

devel/gn-devel/Portfile
```diff
@@ -34,7 +34,7 @@ configure.pre_args
 
 build.dir           ${worksrcpath}/out
 build.cmd           ${prefix}/bin/ninja
-build.target
+build.target        gn
```
pmetzger, line 37: I presume changing the build target has no impact on installed files?

pmetzger:
Looks generally okay. I left one tiny question above.

devel/gn-devel/Portfile
chrstphrchvz: Correct, the only installed file is the `gn` binary, and the `gn` target is enough to build it. So unless someone is running `port test`, the `gn_unittest` target is needlessly built as well. I guess the unusual thing here is that if a target isn’t specified, then ninja builds all targets, unlike make which would only build some default target.

PR #13618, opened 2022-01-10
py-bpython: Remove Python 3.6 and add Python 3.10

judaew:
#### Description

See https://lists.macports.org/pipermail/macports-dev/2021-December/044043.html

- py-backports.cached-property: Remove python 3.6
- py-{curtsies,bpython}: Remove Python 3.6

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/python/py-argh/Portfile
+++ b/python/py-argh/Portfile
@@ -21,7 +21,7 @@ checksums           rmd160  42da2aa1dfae654419acfc62db73d0fb27ecd23f \
                     sha256  e9535b8c84dc9571a48999094fda7f33e63c3f1b74f3e5f3ac0105a58405bb65 \
                     size    32913
 
-python.versions     27 35 36 37 38 39 310
+python.versions     27 37 38 39 310
 
 if {${name} ne ${subport}} {
     depends_build-append \
--- a/python/py-argh/Portfile
+++ b/python/py-argh/Portfile
@@ -21,11 +21,28 @@ checksums           rmd160  42da2aa1dfae654419acfc62db73d0fb27ecd23f \
                     sha256  e9535b8c84dc9571a48999094fda7f33e63c3f1b74f3e5f3ac0105a58405bb65 \
                     size    32913
 
-python.versions     27 35 36 37 38 39
+python.versions     27 35 36 37 38 39 310
 
 if {${name} ne ${subport}} {
     depends_build-append \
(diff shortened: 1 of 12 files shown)


pmetzger:
LGTM. Arguably 3.10 might be added (where appropriate, if it passes tests).

judaew:
@pmetzger Adding a new version of Python is like cancer, especially with tests. It takes a lot of time.

python/py-argh/Portfile
```diff
@@ -21,11 +21,28 @@ checksums           rmd160  42da2aa1dfae654419acfc62db73d0fb27ecd23f \
                     sha256  e9535b8c84dc9571a48999094fda7f33e63c3f1b74f3e5f3ac0105a58405bb65 \
                     size    32913
 
-python.versions     27 35 36 37 38 39
+python.versions     27 35 36 37 38 39 310
```
pmetzger, line 24: Presumably we want to get rid of 35/36 here?

pmetzger (commented):
I see what you mean about metastasis here. But it was good work!

python/py-argh/Portfile
lpsinger: I agree. 36 is past end of life. If you remove this, then you can also remove the `if {${python.version} >= 37}` conditional.

lpsinger (changes requested):
Shouldn't we also be dropping support for 2.7?

lpsinger:
> > Shouldn't we also be dropping support for 2.7?
> 
> As I know, all package managers try to drop Python 2.7. But here the [digital_rf](https://ports.macports.org/port/digital_rf/details/) port depends on `py27-watchdog`, which in turn depends on `py27-argh`. I am not sure if this port can work with the Python 3.* version. Anyway, it is better to leave it to another PR.

According to the README file of digital_rf, it supports Python >= 3.5. I'd feel better about updating digital_rf first (yes, in a separate PR) so that the diff of the present PR is much smaller.

pmetzger:
We can do this in two steps then, getting rid of the 2.7 support here after taking this initial step.

judaew:
@pmetzger Can I merge it?

pmetzger:
@judaew Go for it!

PR #13634, opened 2022-01-12
bazelisk: new port

0x1DA117:
#### Description

This adds a port for [bazelisk](https://github.com/bazelbuild/bazelisk). I'm not quite sure if it is a good idea to set the `GOPROXY` environment variable back to it's default value.

Since bazelisk is a wrapper for bazel, it builds itself using bazel. If I should avoid to set `GOPROXY`, than I would have to create a different way to build bazelisk.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.6 20G165
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/devel/bazelisk/Portfile
+++ b/devel/bazelisk/Portfile
@@ -1,0 +1,41 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/bazelbuild/bazelisk 1.11.0 v
+revision            0
+categories          devel
+maintainers         {@0xIDANT danieltrautmann.com:me} openmaintainer
+license             Apache-2
+
+description         A user-friendly launcher for Bazel
+
+long_description    Bazelisk is a wrapper for Bazel written in Go. \
+                    It automatically picks a good version of Bazel given your \
+                    current working directory, downloads it from the official \
+                    server (if required) and then transparently passes through \
+                    all command-line arguments to the real Bazel binary. You \
+                    can call it just like you would call Bazel.
+
+conflicts           bazel
+
+build.cmd           ./build.sh
+build.env-append    BAZELISK_VERSION=v${version}
+build.env-delete    GO111MODULE=off GOPROXY=off
+
+patchfiles          patch-bazelisk-version.diff \
+                    patch-clean-and-stop-bazel-after-build.diff
+
+destroot {
+    xinstall -m 0755 \
+        ${worksrcpath}/bin/${name}-${goos}-${goarch} \
+        ${destroot}${prefix}/bin/${name}
+
+    ln -s ${prefix}/bin/${name} ${destroot}${prefix}/bin/bazel
+}
+
+checksums           ${distname}${extract.suffix} \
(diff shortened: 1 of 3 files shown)


ryandesign:
Seems to fail in CI:
```
--->  Checksumming mitchellh-go-homedir-v1.1.0.tar.gz
Error: Failed to checksum bazelisk: mitchellh-go-homedir-v1.1.0.tar.gz does not exist in /opt/local/var/macports/distfiles/go
```

devel/bazelisk/Portfile
```diff
@@ -0,0 +1,54 @@
...
+        ${worksrcpath}/bin/${name}-${goos}-${goarch} \
+        ${destroot}${prefix}/bin/${name}
+
+    ln -s ${prefix}/bin/${name} ${destroot}${prefix}/bin/bazel
+}
+
+checksums           ${distname}${extract.suffix} \
+                        rmd160  c27d46e18702ee16735d31818f931b93189c1d6c \
+                        sha256  0842cb7c0e828839e75c80c253bb0ba7a1e95120e21567e86f5bfb60ee59e467 \
+                        size    152999
```
ryandesign, line 38: You specify checksums for a main distfile, but you also specify `fetch.type git` which means a main distfile won't be used.

Why do you need `fetch.type git`?

0xIDANT: I used `fetch.type git` because the build process needs a Git repository to get information about the version. I guess I could circumvent this by patching some files.

devel/bazelisk/Portfile
```diff
@@ -0,0 +1,57 @@
...
+                    current working directory, downloads it from the official \
+                    server (if required) and then transparently passes through \
+                    all command-line arguments to the real Bazel binary. You \
+                    can call it just like you would call Bazel.
+
+conflicts           bazel
+
+build.cmd           ./build.sh
+build.env-append    BAZELISK_VERSION=v${version}
+build.env-append    GOPROXY=https://proxy.golang.org,direct
```
herbygillot, line 25: > I'm not quite sure if it is a good idea to set the GOPROXY environment variable back to it's default value.

That's fine, plenty of Go ports do so already.  But instead of this, use:
```
```

...to simply remove the environment variables that the `golang` portgroup sets by defalt.

0xIDANT: Thanks for the hint

0xIDANT:
@ryandesign @herbygillot now it's ready and working

devel/bazelisk/Portfile
herbygillot: One more thing.  Since you are now using `build.env-delete    GO111MODULE=off GOPROXY=off`, you no longer are using `go.vendors`.

Remove the entire `go.vendors` section, and your `checksums` can just look like this:
```
checksums               rmd160  c27d46e18702ee16735d31818f931b93189c1d6c \
                        sha256  0842cb7c0e828839e75c80c253bb0ba7a1e95120e21567e86f5bfb60ee59e467 \
                        size    152999
```

herbygillot:
Looks good - just one more comment (above).

devel/bazelisk/Portfile
0xIDANT: Thanks, done

0x1DA117:
@herbygillot @ryandesign can you review again please?

herbygillot:
Thanks @0x1DA117

PR #13638, opened 2022-01-12
Edit:- Openjdk-java new subport

usersxx:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you


catap:
@usersxx I see that you may install JDK: 8, 11 and 17 from a page: https://www.oracle.com/java/technologies/downloads/

Maybe add all of them to the port?

The second thought: I feel that calling port `-oracle` is the right things.

Also, as far as I recall the forking of OpenJDK is started when Oracle changed the license.  So, at least `NoMirror` and some license should be defined in your update.

usersxx:
Ok, But these downloads are from jdk.java.net 
Open source builds of oracle jdk

breun:
@catap The big confusion is that there is a difference between Oracle JDK and Oracle OpenJDK. 

Oracle OpenJDK is available from https://jdk.java.net under GPL2 with Classpath Exception license, and these builds don't get LTS (Long Term Support) updates.

Oracle JDK is available from oracle.com and was recently switched to the 'Oracle No-Fee Terms and Conditions' (before it required a license for running in production). 

To avoid this confusion I always only provided Eclipse Temurin (formerly AdoptOpenJDK) and Azul Zulu OpenJDK builds in MacPorts, because those both do not require a license, but do get LTS updates.

I personally don't see any benefit in using Oracle OpenJDK, because you just get less updates than from other OpenJDK vendors, but if there is a need for Oracle OpenJDK in MacPorts, then I'm also not going to block adding it.

breun:
@usersxx Could you maybe put your changes on a separate branch so not every update to `master` also ends up in this pull request? Also please squash your changes to a single commit.

usersxx:
See this openjdk page in wikipedia (https://en.wikipedia.org/wiki/OpenJDK#OpenJDK_versions)

usersxx:
If you click Openjdk GA/EA buils in sidebar of openjdk(openjdk.java.net), you will be redirected to jdk.java.net and thus the jdk on jdk.java.net are built officially by oracle.

usersxx:
Also, official creator of java was designed by James Gosling and its development was succeeded to sun which was acquired by oracle so java by oracle (All jdk by Oracle:- Oracle OpenJDK, GraalVM, Oracle Java SE, Oracle GraalVM Enterprise edition) are official and this I prefer these only but Other jdk are just alternative too so they need to be used too.

usersxx:
I am closing this pull request and opening a new one because openjdk ports can't be added here

catap:
@breun honestly? I doubt that anyone notice differences between adopt, zulu and orcale JDK. It may have differences, but it is extreamly small and you must solve quite a challenge to measure it :)

usersxx:
(see this:- https://i.stack.imgur.com/P4Gmt.png)

catap:
@usersxx do you know any difference not in support or license, but in the platform itself between zulu and orcale openjdks?

breun:
@usersxx It's fine to prefer one OpenJDK distribution over another, but James Gosling works for Amazon on Amazon's OpenJDK (Coretto) now. :)

breun:
@catap Did I say users would experience differences? The differences between OpenJDK vendors is mostly what versions they provide and for how long they support these versions. Using a build from Oracle mostly makes sense if you pay for support, otherwise you're just getting shorter support lifetimes compared to other vendors. Oracle JDK used to have some components that were not available in OpenJDK, but these days I believe the functional difference is basically zero indeed.

catap:
> but these days I believe the functional difference is basically zero indeed.

Well... I know some differences between vendors.

For example RH's build of openjdk8 includes shenandoah GC :) Why? Because RH pays salary to the team who build it.

But I'm really using shenandoah in production, and I have no idea about another differences.

breun:
That example might be one of the biggest differences indeed. Shenandoah is also included in other OpenJDK 12+ distributions though. It's time to move on from Java 8, Java 18 will be out next week. :)

catap:
@breun a project which is used jdk8 and shenandoah enjoys some perofrmance benefits that aren't available on jdk11 and after that.

To be honest I've found a few bugs inside shenandoah, and the last one was missed bariers inside atomics :)

PR #13643, opened 2022-01-12
json-mode.el: Submission

essandess:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/editors/json-mode.el/Portfile
+++ b/editors/json-mode.el/Portfile
@@ -1,0 +1,51 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem 1.0
+PortGroup           github 1.0
+
+github.setup        joshwnj json-mode 1.8.0 v
+name                ${github.project}.el
+categories          editors
+maintainers         nomaintainer
+supported_archs     noarch
+license             GPL-3+
+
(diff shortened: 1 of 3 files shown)


pmetzger:
I don't get why one would use this over using the built in Emacs package manager for the same functionality. I use both Emacs and MacPorts, and as much as I love MacPorts, it is not the right tool for every job.

essandess:
Any constructive comment on this PR, rather than opinions on personal package manger choices?

mascguy:
> TBH, I haven't looked much at the details because, as I said, I don't think it's a good idea. There are many thousands of Emacs packages. If we package this, how many more are we going to have to package? Why package json-mode and not (say) helm or slime or avy or the thousands of others?
> 
> Emacs packages are designed at this point to be managed and updated by the Emacs package infrastructure using Elpa/Melpa/etc. Alternative sources of Emacs packages are part of the mechanism so you're not beholden to the repositories they manage, and indeed, Melpa is one such mechanism already. It's a good system, well designed for Emacs, and integrates in to Emacs itself via the package management tools. I don't see why we would compete with it. Are we signing up to duplicate the whole infrastructure for thousands of packages?

Unfortunately I can appreciate both sides here, so I can't provide a tie-breaking vote.

Anyone else... thoughts?

essandess:
I also see @pmetzger's side, but also appreciate having a few Emacs options available to install within MacPorts. I'm still on the fence.

PR #13653, opened 2022-01-13
redeclipse: fix licenses

jasonliu--:
#### Description

The source code of Red Eclipse is licensed zlib and the game data is licensed CC, so this commit splits up the `license` line so that they apply to the individual subports.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.11.6 15G22010 x86_64
Xcode 8.2.1 8C1002

###### Verification 
Have you

### DIFF
--- a/games/redeclipse/Portfile
+++ b/games/redeclipse/Portfile
@@ -7,7 +7,6 @@ name                redeclipse
 set pretty_name     "Red Eclipse"
 version             1.6.0
 categories          games
-license             zlib CC-BY-SA-3
 maintainers         @jasonliu--
 
 homepage            https://www.${name}.net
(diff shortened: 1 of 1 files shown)


games/redeclipse/Portfile
```diff
@@ -30,6 +29,8 @@ if {$subport eq ${name}} {
     PortGroup       app 1.0
 
     github.setup    ${name} base ${version} v
+    revision        1
```
ryandesign, line 32: Why does the revision need to be increased?

jasonliu--: The previous revision successfully built on the builders, so I was under the impression that they wouldn't run through the build and upload the binaries to packages.macports.org unless I revbumped. Is that not the case? If it's not, I can remove the revbump.

ryandesign: That's not the case. All commits notify the builders, and they decide if a build is necessary. A change in distributability due to a change of license does trigger a build.

jasonliu--: Oh, I didn't know that. Ok, I'll remove the revbump, then.

ryandesign: See the documentation for `revision` at https://guide.macports.org/#reference.keywords.

PR #13660, opened 2022-01-14
armagetronad: Use the right extract commands

ryandesign:
#### Description

Extract bzip2 files with bzip2; extract xz files with xz. Fixes extract failure on OS X 10.8 and earlier.

Specify tag for primary distfile's master_sites.

Use debian fetch group.

###### Type(s)

- [X] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H1519 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/games/armagetronad/Portfile
+++ b/games/armagetronad/Portfile
@@ -6,6 +6,7 @@ PortGroup           boost 1.0
 
 github.setup        ArmagetronAd armagetronad 0.2.9.1.0 v
 github.tarball_from releases
+use_bzip2           yes
 extract.suffix      .tbz
 
 set pretty_name     "Armagetron Advanced"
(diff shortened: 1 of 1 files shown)


jasonliu--:
Looks good to me, lol. Seems like it's making things awfully super-specific for something that's only a problem for macOS <= 10.8! 😆

It's a good thing you're the one who did this! I had no idea there was such a thing as a "debian fetch group", as well as a `${portutil::autoconf::tar_command}`.

ryandesign:
> Seems like it's making things awfully super-specific for something that's only a problem for macOS <= 10.8! 😆

Perhaps the reason for my insistence on doing it "the right way" is that I used MacPorts for several years before the release of OS X 10.9, so "the right way" was the only way that would work. And I would hate for this simple issue to be the only reason why the port could not be used on 10.8 and earlier.

I agree that it is a little more convoluted this way, and welcome the day when MacPorts base properly supports extracting archives of any type without needing to be told what commands to use for each.

> I had no idea there was such a thing as a "debian fetch group",

It's a good idea to inform yourself about which fetch groups exist so that you can use them when appropriate. Read https://github.com/macports/macports-ports/blob/master/_resources/port1.0/fetch/mirror_sites.tcl.

> as well as a `${portutil::autoconf::tar_command}`.

This is how you get access to values discovered when the MacPorts base configure script was run. There certainly aren't many occasions when accessing this is necessary, but in this case I thought it was a good idea since which tar command gets used by default did change at some point (from /usr/bin/gnutar in OS X 10.8 to /usr/bin/tar in OS X 10.9).

PR #13671, opened 2022-01-15
*-w64-mingw32-gcc: Add dependency zstd

Gcenx:
Opportunistic links `zstd` for LTO byte code compression, so might as well add it.

#### Description
After https://github.com/macports/macports-ports/commit/52cf3d239d60402dcd3843f927810e9b57a7c63f the rebuilt packages now link `zstd` but doesn't get installed so *-w64-mingw32-gcc packages are currently broken without manually installing `zstd`

```
dyld: Library not loaded: /opt/local/lib/libzstd.1.dylib
  Referenced from: /opt/local/libexec/gcc/x86_64-w64-mingw32/11.2.0/cc1
  Reason: image not found
x86_64-w64-mingw32-gcc: internal compiler error: Abort trap: 6 signal terminated program cc1
```

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505

###### Verification 

### DIFF
--- a/cross/i686-w64-mingw32-gcc/Portfile
+++ b/cross/i686-w64-mingw32-gcc/Portfile
@@ -11,7 +11,7 @@ set mingw_target    ${mingw_arch}-${mingw_name}
 crossgcc.setup      ${mingw_target} 11.2.0
 crossgcc.languages  {c c++ fortran objc obj-c++}
 dist_subdir         gcc[lindex [split ${version} .] 0]
-revision            1
+revision            2
 
 maintainers         {mojca @mojca} openmaintainer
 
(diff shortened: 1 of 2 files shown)


mojca (approved):

mojca:
Do we need to add this dependency to the crossgcc portgroup as well?

Gcenx:
> Do we need to add this dependency to the crossgcc portgroup as well?

It might be a good idea to add it into crossgcc portgroup for gcc10 and later since this was originally added in gcc10.

kencu:
fyi opportunistic use of zstd can also cause troubles if you try to use a macports-gcc compiler to bootstrap a gcc build outside macports.

PR #13712, opened 2022-01-20
terraform-1.1: update to 1.1.4

herbygillot:
- add ARM64 support for Terraform 1.1 & 1.0

Fixes: https://trac.macports.org/ticket/64222

###### Verification 
Have you

### DIFF
--- a/sysutils/terraform/Portfile
+++ b/sysutils/terraform/Portfile
@@ -17,18 +17,51 @@ maintainers             {emcrisostomo @emcrisostomo} \
 # *NOTE* Remember to update `latestVersion` on a version upgrade.
 set latestVersion       terraform-1.1
 
+set armAvailable        no
+
+proc terraformBaseVersion {} {
+    global subport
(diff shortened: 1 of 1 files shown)


jpeeler:
Is there a way to stick with the x86 version for now? I'm pretty sure I'm in a situation where not all the providers have been updated to provide an arm release (but this also may just be because they are pinned to old versions and I need to get on it and update everything).

herbygillot:
> Is there a way to stick with the x86 version for now? I'm pretty sure I'm in a situation where not all the providers have been updated to provide an arm release (but this also may just be because they are pinned to old versions and I need to get on it and update everything).

Perhaps we can use a variant to do this. It could look something like this: `sudo port install terraform-1.1 +x86only`

Open a ticket in Trac ( https://trac.macports.org/ ) and I'll try to take a look soon.

jpeeler:
Thanks! https://trac.macports.org/ticket/64535

PR #13719, opened 2022-01-20
crystal: update to 1.3.2

mohd-akram:
###### Tested on
macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification
Have you

### DIFF
--- a/lang/crystal/Portfile
+++ b/lang/crystal/Portfile
@@ -2,10 +2,9 @@
 
 PortSystem          1.0
 PortGroup           github 1.0
-PortGroup           openssl 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
-github.setup        crystal-lang crystal 1.3.1
+github.setup        crystal-lang crystal 1.3.2
 revision            0
 categories          lang
 platforms           darwin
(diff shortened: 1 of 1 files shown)


lang/crystal/Portfile
```diff
@@ -2,10 +2,9 @@
 
 PortSystem          1.0
 PortGroup           github 1.0
-PortGroup           openssl 1.0
```
herbygillot, line 5: Does Crystal no longer use OpenSSL?

mohd-akram: It does, good catch, I forgot to add the `depends` line. No need for the PortGroup.

herbygillot:
So the `openssl` portgroup is the preferred way to add OpenSSL support.  With it, you can control whether your project uses OpenSSL 1.1 or OpenSSL 3 (ie., `openssl.branch 1.1`).  Just linking straight to the `libssl.dylib` in the lib directory will give you OpenSSL 3.  I'm curious, is there a reason why you're removing the `openssl` portgroup?

mohd-akram:
I added it previously because I thought I needed it to get the path to static OpenSSL libs, but there are symlinks for that in `/opt/local/lib`. This is how it was before, and it allows using libre SSL.

herbygillot:
> I added it previously because I thought I needed it to get the path to static OpenSSL libs, but there are symlinks for that in `/opt/local/lib`. This is how it was before, and it allows using libre SSL.

Got it, good point about LibreSSL.

PR #13720, opened 2022-01-20
abcde: switch to Perl 5.34

chrstphrchvz:
###### Tested on

macOS 10.15.7
Perl 5.34.0

###### Verification 
Have you

### DIFF
--- a/audio/abcde/Portfile
+++ b/audio/abcde/Portfile
@@ -5,7 +5,7 @@ PortGroup               perl5 1.0
 
 name                    abcde
 version                 2.9.3
-revision                0
+revision                1
 categories              audio
 platforms               darwin
 maintainers             {eclipsed.net:gr @0x6772} {gmx.us:chrischavez @chrstphrchvz} openmaintainer
@@ -25,7 +25,7 @@ checksums               rmd160  530c875121a7e541ffb60336565774a6d0b04d6e \
                         sha256  046cd0bba78dd4bbdcbcf82fe625865c60df35a005482de13a6699c5a3b83124 \
                         size    160876
 
-perl5.branches          5.28
+perl5.branches          5.34
 
 depends_lib             port:vorbis-tools \
                         port:lame \


0x6772:
I confess I'm relatively inexperienced with this port maintainer process. Shouldn't we be lint/building against macos-12 too?

chrstphrchvz:
> Shouldn't we be lint/building against macos-12 too?

Ideally, yes, but the CI provided by GitHub Actions does not yet publicly offer macOS 12: https://github.com/actions/virtual-environments/issues/3649

I’ve not personally checked this port on macOS 12, but it currently has no known failures according to https://ports.macports.org/port/abcde/builds/ so I expect it to work after switching it to Perl 5.34 including on macOS 12.

0x6772:
> Ideally, yes, but the CI provided by GitHub Actions does not yet publicly offer macOS 12: [actions/virtual-environments#3649](https://github.com/actions/virtual-environments/issues/3649)

Gotcha.

> I’ve not personally checked this port on macOS 12, but it currently has no known failures according to https://ports.macports.org/port/abcde/builds/ so I expect it to work after switching it to Perl 5.34 including on macOS 12.

I'm largely ignorant here. Have you found that "build successful" is a sufficient test case for "abcde actually works to rip CDs and encode audio files" in the past?

That is, the "build" is trivial for abcde (if the ports it depends on build, so will it), but I don't think that actually validates that the software will actually recognize a CD in a device, read from it successfully, and executes the various encoding libraries with the correct flags for the current versions of those libraries.

Maybe I'm overthinking it?

0x6772:
Btw, around a few other things, I'm in the process of doing that "actually works" test on macOS 11.6.2 (Darwin Kernel Version 20.6.0), which I think should be plenty sufficient, I just didn't have an alt-root environment where I'd already pulled the various dependent ports up to the Perl 5.34 bits to test with, so that's taking a few minutes.

0x6772 (approved):
Changes make sense to me, and I was able to rip a CD successfully after upgrading the several dependent ports.

ryandesign:
> switch to Perl 5.34

Why? Ports should use perl 5.28 (to match the perl5 portgroup) unless there's a good reason not to.

https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/perl5-1.0.tcl#L17

jmroot:
> > switch to Perl 5.34
> 
> Why? Ports should use perl 5.28 (to match the perl5 portgroup) unless there's a good reason not to.
> 
> https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/perl5-1.0.tcl#L17

Though that default should change, since 5.28 is no longer getting security fixes.

PR #13744, opened 2022-01-22
libpano13: update to 2.9.21

olupton:
#### Description
Update libpano13 from 2.9.19 to 2.9.21. Migrate to using CMake to build.

###### Type(s)
- [x] enhancement

###### Tested on
macOS 12.1 21C52 x86_64
Xcode not installed, I just have the latest commandline tools.

###### Verification 
Have you

### DIFF
--- a/graphics/libpano13/Portfile
+++ b/graphics/libpano13/Portfile
@@ -1,34 +1,45 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
-PortSystem          1.0
+PortSystem            1.0
+PortGroup             cmake 1.1
 
-name                libpano13
-version             2.9.19
-revision            1
-categories          graphics
-platforms           darwin
-maintainers         {gmail.com:freespace @freespace}
-license             GPL-2+
+name                  libpano13
+version               2.9.21
+revision              0
+categories            graphics
+platforms             darwin
+maintainers           {gmail.com:freespace @freespace}
+license               GPL-2+
 
-description         The cross-platform library behind Panorama Tools
+description           The cross-platform library behind Panorama Tools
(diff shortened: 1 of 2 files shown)


pmetzger:
You removed `configure.args` which means this now may or may not use Java depending on whether it happens to be installed. I'd suggest leaving the `configure.args` etc.

olupton:
The CMake configuration disables Java by default: https://sourceforge.net/p/panotools/libpano13/ci/libpano13-2.9.21/tree/CMakeLists.txt#l100 -- if you feel strongly, I can explicitly set it to `OFF`.

pmetzger:
The minimal diff from the existing portfile will be easiest to merge. Even if something isn't built by default, it's easier for a maintainer with only casual acquaintance to the port to understand what options are in use if the configuration is explicit.

olupton:
OK, I added explicit options as you requested.

graphics/libpano13/Portfile
```diff
@@ -1,34 +1,45 @@
...
+                      -DSUPPORT_JAVA_PROGRAMS=OFF \
+                      -DTIFF_INCLUDE_DIR=${prefix}/include \
+                      -DZLIB_ROOT=${prefix}
+patchfiles-append     patch-endian-detection.diff
 
-configure.args      --with-jpeg=${prefix} \
-                    --with-tiff=${prefix} \
-                    --with-png=${prefix} \
-                    --with-zlib=${prefix} \
-                    --without-java
```
pmetzger, line 32: This was the thing you removed that I was referring to. I don't think you need the `-D` stuff you did above, just to leave this stuff in place.

olupton: The new version only supports building via CMake, which is why I changed this. The `-D` stuff is an attempt at a literal translation from `./configure`. I did note this in the commit message, but maybe I should have given it more prominence in the PR description.

pmetzger: Ah, I really didn't understand this.

PR #13761, opened 2022-01-24
libbiosig: Update to 1.8.6, fix implicit function declarations, fix license

ryandesign:
#### Description

Update to 1.8.6. There are newer versions available including 1.9.x and 2.x but I did not know whether those newer versions would be compatible with the ports that depend on this port so I'm leaving further updates to you for now.

Fix implicit declaration of function. I don't see this [build failure](https://build.macports.org/builders/ports-11_arm64-builder/builds/44996/steps/install-port/logs/stdio) on my system so while I'm fairly sure this fix is correct to fix this implicit declaration of function I don't know whether there are additional implicit declaration of functions remaining to be fixed.

Fix license. The DESCRIPTION file says it is GPL 3 or later, not GPL 3 only.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H1615 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/science/libbiosig/Portfile
+++ b/science/libbiosig/Portfile
@@ -31,6 +31,8 @@ depends_lib         port:SuiteSparse_CHOLMOD \
                     port:zlib \
                     port:libiconv
 
+patchfiles          implicit.patch
+
 cmake.out_of_source yes
 
 configure.args-append \
--- a/science/libbiosig/Portfile
+++ b/science/libbiosig/Portfile
@@ -4,11 +4,14 @@ PortSystem          1.0
 PortGroup           cmake 1.0
 
 name                libbiosig
-version             1.8.4
-revision            1
+version             1.8.6
+revision            0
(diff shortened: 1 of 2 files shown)


neurodroid:
@ryandesign Here's a patch that would bring it to 2.3.1:
https://github.com/macports/macports-ports/compare/master...neurodroid:patch-1

ryandesign:
Do you know whether the other ports that depend on libbiosig are compatible with 2.3.1?

Your patch does not incorporate all of the fixes I submitted here. Perhaps if you agree we could merge this PR and then you could submit another to update it further, if desired.

neurodroid:
> Do you know whether the other ports that depend on libbiosig are compatible with 2.3.1?
> 
> Your patch does not incorporate all of the fixes I submitted here. Perhaps if you agree we could merge this PR and then you could submit another to update it further, if desired.

OK to merge this PR

PR #13764, opened 2022-01-25
git: require C99

evanmiller:
#### Description

C99 is now "experimentally required" by git, erroring out if not detected:

https://github.com/git/git/blob/master/git-compat-util.h#L14

Adding `-std=gnu99` fixes the build on Tiger/Leopard. I have also cleaned up the flags being passed to the build.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/git/Portfile
+++ b/devel/git/Portfile
@@ -62,10 +62,12 @@ extract.only        git-${version}${extract.suffix} \
 
 use_configure       no
 
+compiler.c_standard 1999
+
 variant universal   {}
 
-set CFLAGS          "-I. -Wall -O2 -I${prefix}/include [get_canonical_archflags cc]"
-set LDFLAGS         "-L${prefix}/lib [get_canonical_archflags ld]"
+set CFLAGS          "${configure.cflags} -std=gnu99 [get_canonical_archflags cc]"
+set LDFLAGS         "${configure.ldflags} [get_canonical_archflags ld]"
 
 build.args          CFLAGS="${CFLAGS}" \
                     LDFLAGS="${LDFLAGS}" \


ryandesign:
In this PR you've made some substantial changes to the way the port builds:

* `-I.` was in CFLAGS; now it's not. Is that ok? If the build system now adds `-I.` on its own, then that's fine.
* `-O2` was in CFLAGS; now `-Os` is. This changes how the port builds, so the revision must be increased.
* `-I${prefix}/include` was in CFLAGS; now it's not. Is that ok? If the build system now uses CPPFLAGS, where MacPorts puts `-I${prefix}/include` by default, then that's fine.
* Now `-headerpad_max_install_names` is in LDFLAGS when it wasn't before. This too necessitates a revision increase.

There's no reason why this port needs to define new CFLAGS and LDFLAGS variables when MacPorts already provides configure.cflags and configure.ldflags variables for that purpose. The port should just use those.

evanmiller:
> In this PR you've made some substantial changes to the way the port builds:
> 
> * `-I.` was in CFLAGS; now it's not. Is that ok? If the build system now adds `-I.` on its own, then that's fine.

The build system adds `-I.` on its own

> * `-O2` was in CFLAGS; now `-Os` is. This changes how the port builds, so the revision must be increased.

OK

> * `-I${prefix}/include` was in CFLAGS; now it's not. Is that ok? If the build system now uses CPPFLAGS, where MacPorts puts `-I${prefix}/include` by default, then that's fine.

git's Makefile uses CPPFLAGS:

https://github.com/git/git/blob/master/Makefile#L1255

> * Now `-headerpad_max_install_names` is in LDFLAGS when it wasn't before. This too necessitates a revision increase.

OK

> 
> There's no reason why this port needs to define new CFLAGS and LDFLAGS variables when MacPorts already provides configure.cflags and configure.ldflags variables for that purpose. The port should just use those.

CFLAGS and LDFLAGS have to be provided on the command line rather than through the environment:

https://github.com/git/git/blob/5d01301f2b865aa8dba1654d3f447ce9d21db0b5/Makefile#L1206-L1211

herbygillot:
`git` has just seen another version update: https://github.com/macports/macports-ports/commit/01c395279d33d9b6f83000900fb49c3c34c3e582

So that will qualify as a "bump".

ryandesign:
> git's Makefile uses CPPFLAGS:

But I see that the portfile doesn't actually supply any CPPFLAGS to the makefile.

But I see that `-I/opt/local/include` is nevertheless making its way into the compile commands via some other means so that's fine.

> CFLAGS and LDFLAGS have to be provided on the command line rather than through the environment:

Which has nothing to do with the creation of new MacPorts variables to store those flags, which remains unnecessary to do.

In other words, this would be more straightforward:

```
configure.cflags-append \
                    -std=gnu99

build.args          CFLAGS="${configure.cflags} [get_canonical_archflags cc]" \
                    LDFLAGS="${configure.ldflags} [get_canonical_archflags ld]" \
```

Or the makefile 1.0 portgroup could be used which does some of these things for you.

> `git` has just seen another version update: [01c3952](https://github.com/macports/macports-ports/commit/01c395279d33d9b6f83000900fb49c3c34c3e582)
> 
> So that will qualify as a "bump".

Yup!

PR #13789, opened 2022-01-27
policykit: Update to 0.120 + 3 additional commits; duktape: new port, dependency of updated polkit

neverpanic:
#### Description

policykit: Update to 0.120 + 3 additional commits

Add rudimentary support for the meson build system to the gobject_introspection PortGroup, which is used by polkit. This may need further adaption in case other build systems do not name their gobject introspection configure flag 'introspection'.

Switch polkit to the meson build system, drop the autoconf related files and sections from the Portfile. Re-enable the livecheck that seems to have been disabled because of warning in upstream's NEWS file that has been also present in previous versions and was not a good reason to keep polkit at an older version.

Fix CVE-2021-4034, a local privilege escalation in pkexec – although I have not verified whether this is exploitable on macOS.

The updated polkit would require a newer mozjs – however, polkit upstream is working on replacing polkit with the much smaller and simpler to build duktape JavaScript engine, which will ship in 0.121. Instead of packaging the newer version of mozjs, I packaged duktape and pulled in the commit the enables using it. See https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/97 for further discussion that suggests that pretty much all distributions will switch to duktape as soon as it is available.

CVE: CVE-2021-4034
Closes: https://trac.macports.org/ticket/63277

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [x] security fix

###### Tested on
macOS 11.6.2 20G314 x86_64

###### Verification 
Have you


evanmiller:
Note that this PR conflicts with #13652. The `duktape` approach here certainly sounds promising; I will try to test it on some legacy systems where mozjs78 is currently a no-go.

evanmiller:
`duktape` builds just fine on 10.4.11 / PPC and the "Hello, world" test works 👍

Amazingly simple compared to the `mozjs60` build...

neverpanic:
I guess that means it's the better option since mozjs78 needs rust?

evanmiller:
> I guess that means it's the better option since mozjs78 needs rust?

The Rust requirement rules out mozjs78 on 10.6 and earlier, so I personally prefer this PR for that reason. However, since the `duktape` integration is newer and uses backported patches, it may be buggier.

If `policykit` includes a test suite, I think it would be worth including it as a test phase in the Portfile. That way I can confirm everything is working as expected on these older systems.

evanmiller:
Some local testing updates:

* `duktape` builds on 10.6 x86_64 and 10.4 PPC
* `duktape` "Hello, world" builds and runs
* `policykit` builds on 10.6 x86_64 and 10.4 PPC
* `policykit` tests (adding `-Dtests=true`) do not build on 10.6 or 10.4 due to function signatures in `mocklibc` conflicting with the system declarations, e.g.

```
../polkit-a2bf5c9c83b6ae46cbd5c779d3055bff81ded683/subprojects/mocklibc-1.0/src/grp.c:30: error: conflicting types for 'setgrent'
 /usr/include/grp.h:105: error: previous declaration of 'setgrent' was here
```

So maybe don't add the policykit test target just yet.

evanmiller:
One other note: it looks like policykit dependencies (including `gconf`) will need a revbump.

neverpanic:
Thanks for catching that – I don't actually have polkit installed by default on by system, I'm just patching it because of the pkexec security issue, so I missed that.

Let's wait for @dbevans to chime in maybe, and then I can quickly fix those missing revbumps. It should be pretty quick to do that:

```
$ port echo depends:policykit | wc -l
9
```

neverpanic:
Here's the revbump of the dependents. Fixing the tests is probably not worth it – I patched the first issue, but further issues popped up where the tests needed fgetgrent, which doesn't exist on macOS.

neverpanic:
Ping @dbevans. Any opinions on this?

neverpanic:
Merging this to fix the open CVE now, plus it's a significant reduction in dependencies for polkit.

evanmiller:
@neverpanic Can you take a look at #13652? This was waiting on an updated policykit (in this case for M1 compatibility).

PR #13801, opened 2022-01-28
grass: update to GRASS 8.0 (revitalize obsolete port) 

nilason:
#### Description

GRASS GIS has now released a new major version 8.0.0. Efforts have been made to move from "versioned" executable, e.g. starting version 7.8.x was made with `grass78`, now the default is `grass`.

This PR suggests to revitalize the old `grass` port, for coming releases. The transition from 6 to 7 (up till 7.8) was a troublesome period (integration of Python as API and GUI, and then Python 2 to 3 etc.) which necessitated to a degree distinction of version. Now the situation is more stable.

Update: Just to clarify, the intention of this PR is to replace `grass7` with `grass`, for any upcoming versions (GRASS 8+).

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you


nilason:
@Veence Do you have any opinion or thoughts on this?

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           wxWidgets 1.0
+PortGroup           github 1.0
+PortGroup           debug 1.0
 
+github.setup        OSGeo grass 8.0.0
 name                grass
-epoch               1
```
ryandesign, line 6: The epoch can never be removed or decreased.

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+                    port:libpng \
+                    port:liblas\
+                    path:lib/pkgconfig/cairo.pc:cairo \
+                    port:zlib \
+                    port:readline
+
+patchfiles          800-patch.diff
+
+configure.cflags    -DGL_SILENCE_DEPRECATION
+configure.ldflags   -Wl,-rpath,${prefix}/lib
```
ryandesign, line 47: You normally should not overwrite the flags MacPorts provides, but you can append to them.
```suggestion
configure.cflags-append -DGL_SILENCE_DEPRECATION
configure.ldflags-append -Wl,-rpath,${prefix}/lib
```
Why is rpath needed? Usually we try to avoid rpath in MacPorts.

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+
+configure.args-append \
+    --prefix=${prefix}/share \
+    --exec-prefix=${prefix} \
```
ryandesign, line 51: MacPorts uses `configure.pre_args` for the prefix argument. If you're changing the prefix, that's the place to do it.

```suggestion
configure.pre_args \
      --prefix=${prefix}/share
configure.args-append \
    --exec-prefix=${prefix} \
```

${prefix}/share is a strange prefix. Why do this?

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+    --with-zstd \
+    --with-zstd-includes=${prefix}/include \
+    --with-zstd-libs=${prefix}/lib \
+    --with-readline-includes=${prefix}/include/readline \
+    --with-readline-libs=${prefix}/lib
+
+build.target            default
+
+# Python 3 variants
+set python_suffixes {36 37 38 39}
```
ryandesign, line 88: 3.10 is the default version of python now. Does this software support it? If so, add 310 to python_suffixes here.

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+        reinplace "s|@DEPLOYMENT_TARGET@|${macosx_deployment_target}|" ${info_plist}
+    }
+
+    post-destroot {
+        copy ${worksrcpath}/${app_name} ${destroot}${applications_dir}
+    }
+}
+
+variant mysql5 description {Add MySQL 5 support} {
+    depends_lib-append      path:bin/mysql_config5:mysql5
```
ryandesign, line 229: It's strange to add a mysql5 variant at this point, with the mysql5 port haven't been obsolete for so many years already. Variants for newer mysql software (mysql8, mysql57, mysql56, mysql55, mariadb-10.7, mariadb-10.6, mariadb-10.5) would be more reasonable, if this software supports that.

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+post-configure {
+    reinplace -E "s|\\\$\\\(CPPFLAGS)| \$(CPPFLAGS) \
+        \\\\\"-Drestrict=\\\\\" \\\\\"-D__attribute__(x)=\\\\\" \\\\\"-D_Nonnull=\\\\\" \
+        \\\\\"-Dint8_t=char\\\\\" \\\\\"-DCF_INLINE=\\\\\" \\\\\"-D_Null_unspecified=\\\\\" \
+        \\\\\"-D__DARWIN_OS_INLINE=\\\\\"|" ${worksrcpath}/python/libgrass_interface_generator/Makefile
+}
+
+livecheck.type      regex
+livecheck.url       ${master_sites}
+livecheck.regex     grass-(8.\[0-9.\]+)${extract.suffix}
```
ryandesign, line 300:
```suggestion
livecheck.regex     grass-(8\\.\[0-9.\]+)[quotemeta ${extract.suffix}]
```
However, this isn't the "grass8" port; this is the "grass" port. Why limit livecheck to version 8.x? Wouldn't we want to upgrade to 9.x when that becomes available?

gis/grass/Portfile
```diff
@@ -1,11 +1,300 @@
...
+    configure.args-append   --with-libblas=${prefix}/lib/libopenblas.dylib
+    configure.args-append   --with-lapack-includes=${prefix}/include
+    configure.args-append   --with-liblapack=${prefix}/lib/libopenblas.dylib
+}
+
+post-configure {
+    reinplace -E "s|\\\$\\\(CPPFLAGS)| \$(CPPFLAGS) \
+        \\\\\"-Drestrict=\\\\\" \\\\\"-D__attribute__(x)=\\\\\" \\\\\"-D_Nonnull=\\\\\" \
+        \\\\\"-Dint8_t=char\\\\\" \\\\\"-DCF_INLINE=\\\\\" \\\\\"-D_Null_unspecified=\\\\\" \
+        \\\\\"-D__DARWIN_OS_INLINE=\\\\\"|" ${worksrcpath}/python/libgrass_interface_generator/Makefile
```
ryandesign, line 295: This reinplace is difficult to comprehend and there's no indication here of why it's happening. If you're fixing some bug, add a comment with the URL to the bug report.

Since you're not using any MacPorts variables in the reinplace, you could save yourself a lot of escaping by using `{}` string delimiters instead of `""`.

ryandesign (commented):
Is the idea that this new grass 8 port will replace the grass7 port?

Every time the grass7 port was updated, it would change the directory into which it installed things, thus breaking any other port which used grass7. See https://trac.macports.org/ticket/64411. If that can be avoided with this new grass 8 port that would be great.

gis/grass/Portfile
ryandesign: It looks like what this reinplace is trying to is, for each occurrence of `$(CPPFLAGS)`, add some stuff after that. The better way to do that would be to find where `CPPFLAGS` is set and add those things there. That would be only one change needed, vs. however many occurrences of `$(CPPFLAGS)` there are. And then instead of a reinplace it could be implemented as a patchfile which is highly preferable.

gis/grass/Portfile
nilason: fixed

gis/grass/Portfile
nilason: By default grass installs:

1. executable at `$prefix/bin/grass`
2. package/directory at `$prefix/grass80`

Setting `--prefix` and `--exec-prefix` installs these to more proper places (I now updated to install grass dir into $prefix/lib instead of $prefix/share, which seems more adequate).
I realise this is bit of a hack, I will look at if/how this can be fixed upstreams, but for the time being I'd suggest to leave this as it has been for grass7.

Content of grass directory:
```
grass80
├── AUTHORS
├── bin
│   └── ...
├── CHANGES
├── CITING
├── contributors_extra.csv
├── contributors.csv
├── COPYING
├── demolocation
│   └── ...
├── docs
│   └── ...
├── driver
│   └── ...
├── etc
│   └── ...
├── fonts
│   └── ...
├── GPL.TXT
├── gui
│   └── ...
├── include
│   └── ...
├── INSTALL
├── lib
│   └── ...
├── REQUIREMENTS.html
├── scripts
├── share
│   └── ...
├── translators.csv
└── utils
    └── ...
```

gis/grass/Portfile
nilason: GRASS v8 support Python 3.10, but wxPython 4.1.1. does not. So for now, no Python 3.10 for GRASS.

gis/grass/Portfile
nilason: Right, changed to support mysql8 and mysql57, reflecting gdal (minus mysql5).

gis/grass/Portfile
nilason: Indeed, updated. Please check if I got the regex right (eg. `grass-X.Y.Z.tar.gz`).

gis/grass/Portfile
nilason: These are for passing to the Python package ctypesgen, which generates python interface of C code. Ctypesgen in turn send these to compiler preprocessing (eg. `clang -E ...`), which result is used for further parsing. The ctypesgen parser does not recognise all C code yet (such as keywords like `restrict` or `__attribute__(x)` statements) and throws several thousands lines of error messages. These are now mainly (probably only) noise, but I'm working upstreams to at least reduce them to a minimum. Meanwhile, I suggest to leave it as it has been in grass7.

nilason:
> Is the idea that this new grass 8 port will replace the grass7 port?

Yes.
 
> Every time the grass7 port was updated, it would change the directory into which it installed things, thus breaking any other port which used grass7. See https://trac.macports.org/ticket/64411. If that can be avoided with this new grass 8 port that would be great.

Unfortunately the GRASS directory (see https://github.com/macports/macports-ports/pull/13801#discussion_r796556724) is not yet renamed, i.e. is still versioned like `grass80`. This is a problem acknowledged by the GRASS dev team upon bumping major version (lot of hard coded stuff). This *may* be patched, but I'm afraid it could lead to a unsustainable patch.
There is a way to query the grass executable, with:
```sh
grass --config path       # = ${prefix}/bin/grass
/opt/local/lib/grass80    # = ${prefix}/lib/grass80
```
if there is a way to check that in Portfiles of ports that depends on GRASS, then that could be the solution.

gis/grass/Portfile
nilason: The configure settings for shared library linking are hard coded per platform [(aclocal.m4)](https://github.com/OSGeo/grass/blob/eba61a1ca122758c9aae4ff0c28bab895609d9f6/aclocal.m4#L506-L512). The change to rpath was made for the official dist of the GRASS.app bundle.
Temporary solution is a provided  patch. I have planned to make these settings at least overridable during configuration.

nilason:
The GRASS 8 release is [announced](https://grass.osgeo.org/news/2022_01_30_grass_gis_8_0_0_released/) with link to the port `grass`. Would greatly appreciate if we could move this forward.

nilason:
Just bumped this to GRASS version 8.0.1.

@ryandesign I addressed all your issues which I possibly could do (at this point). Do you have any advice on how to proceed?

ryandesign:
@Veence added a new grass8 port today in faaccbf62e59951907baab862e23ec128604b1b7. Where does that leave this PR?

nilason:
I’d say it’s very unfortunate. This have been up for about 2 months. Could at least discussed this issue.

nilason:
I’m really disappointed, tell me what should I have done differently.

Veence:
Oops, I’m sorry, I overlooked that thread. I hope I didn't offend anyone. If I did, I dearly apologise.
I also apologise for being very sluggish lately. Professionally, I have exited the GIS world, so I’m not as involved as I was six months ago. I also had a ton of work to do lately, and that didn't help things either.
Frankly, if someone wants to take over the maintenance of the GIS ports (qgis, grass, gdal, and so on), be my guest. I feel both slobbish and clumsy as a maintainer, so the overall quality of the packages should not suffer at all from the transition.

nilason:
@Veence Thanks for your frank and honest words, did really defuse my state of mind!
I did actually seriously consider offering taking maintenance of GRASS (having the advantage being an active grass developer), but I have only a fraction of experience with the others compared to you. The latter is even more true concerning general MacPorts knowledge. I will try activate myself with the other GIS ports to relieve the workload, if that is okay with you.

@ryandesign Regarding revitalising the `grass` port and replacing the versioned ports. What do you think, could it be possible to quickly make the new grass8 obsolete and replace it with this one? Possibly updated with e.g. the OpenML support which was a good addition I've been having in mind (but not come to yet).

Veence:
Apologies again.

I remember meeting some of the Grass developers at a Foss4G conference that took place near Paris, at ENSG Marne-la-Vallée. They were Czech, IIRC. 

You’re welcome to help with the GIS ports anytime. I don’t consider them to be my turf, so please tread on my toes as often as you want. :)

I think we can still rename grass8 into grass, change the dependencies and increase the revisions of each dependent. Since the only port that depends on Grass is QGis, I think this is manageable, but I’d still defer to Ryan’s superior wisdom on that matter.

Also, I added an openmp variant for grass8

port installed grass8
Warning: port definitions are more than two weeks old, consider updating them by running 'port selfupdate'.
  grass8 @8.0.1_0+openblas+openmpi+postgresql14+proj8+python310 (active)

Though I made a mistake in the published Portfile, and openmpi doesn’t appear as an existing variant. Also, the openmpi support is primitive, requiring the use of clang-13. That can probably be vastly improved, and maybe that would be a good opportunity to change the name of the port.

ryandesign:
> could it be possible to quickly make the new grass8 obsolete and replace it with this one?

Many things are possible! :) I'm not really involved in these ports, I just offered some quick feedback on this PR when I saw it. I usually associate Vincent with these ports.

> There is a way to query the grass executable, with:
> 
> ```shell
> grass --config path       # = ${prefix}/bin/grass
> /opt/local/lib/grass80    # = ${prefix}/lib/grass80
> ```
> 
> if there is a way to check that in Portfiles of ports that depends on GRASS, then that could be the solution.

Sure, for example:

```
pre-configure {
    configure.args-append --with-grass=[exec ${prefix}/bin/grass --config path]
}
```

(It has to be in `pre-configure`, not out in the main part of the Portfile, because you can't be sure until right before the configure phase that the dependencies (grass) have been installed.)

Though if some update of the grass port changes the install name of the grass library (which includes its major version number) then everything linking with the grass library needs to have its revision increased to rebuild it.

PR #13817, opened 2022-01-29
mrustc: update to 20220125

catap:
#### Description

It also:
 - blacklisted unsupported compillers;
 - added full rust and cargo bootsrap;
 - fixed build on macOS 10.6;
 - claim myself as maintainer.

Yeah, it brings rust to macOS 10.6.

This rust is installed under `/opt/local/libexec/mrustc-1.39`.

See: https://trac.macports.org/ticket/64146

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

macOS 12.1 21C52 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you


catap:
And I'd love to share the picture:

Screen shot 2022-01-29 at 11 58 14 PM

Cc: @evanmiller @kencu

catap:
@g5pw / @herbygillot the next step will be integrate it into existed rust port. Because it requires some patches to rust and rust's llvm.... well.. I suggest to install `mrustc` as dependency and add a few symlinks for `cargo` and `rust` port :)

Do you have anything in mind?

kencu:
Great work - lots of nice effort here. Will be important if MacPorts is to not soon lose 10.6 support soon, for sure.

BTW, I found something interesting here for you, on a separate but somewhat related topic. One of the fellows who is around MacPorts sometimes has written up a very detailed, exactly what happens when and in what code, description of how the darwin 10.7+ thread_local_storage program is implemented. Some details here I had not seen summarized before:

https://github.com/iains/gcc-darwin-arm64/issues/1#issue-681508669

kencu:
Now that you have a rust that works on 10.6, perhaps that could be used to step-by-step move towards a current rustc manually, at least. Then that could be used either directly, or as a bootstrap for rust instead of the one usually pulled by rustup.

I have done something very similar for ghc, and I use this to build all the current ports in macports that use ghc (shellcheck, pandoc, etc) on 10.6.8, and as a bootstrap compiler for newer ghc versions (every once in a while, I update them to the current releases):

https://github.com/kencu/ghc-for-older-darwin-systems/releases

catap:
@kencu it uses `mrustc` as the way to compile it and upstream is working to support `1.54`.

To be honest it still needs some patches to LLVM and rust to make it works, but this port includes full bootstrapped rust :)

kencu:
yep, I get it. But now that you have a working rustc, you could forget all about mrustc now, and use your working rust 1.39 to step up to the current rust, in a few steps, with whatever rust patches are needed of curse.

This is exactly what I did with ghc.

catap:
@kencu I'm too lazy to do it to be honest. To build rust `1.58` I must use rust `1.57+`; it is `1.39` that means that I need to iterate the loop for 19 times.

This also includes to update patches to LLVM and rust. I do have everything ready for `1.54`. Or at least as far as I can test.

kencu:
if mrustc is likely get you to 1.54, sure, faster then.

I haven’t looked at what the reach is for rust… for ghc it’s 2 jumps.

catap:
@kencu a few jumps is simpler :) but rust supports quite strange policy and can't be build but rust compiler that is more than a moth old.

kencu:
I see you’re right about that… they can only build with the one before… yeah, that is a lot of steps.

https://news.ycombinator.com/item?id=18686720

catap:
@kencu the issue that I need bootstrap not only rust, but also cargo... seems like insanity, isn't it?

lang/mrustc/Portfile
```diff
@@ -2,14 +2,16 @@
...
 PortSystem      1.0
 PortGroup       github 1.0
+PortGroup       openssl 1.0
+PortGroup       compiler_blacklist_versions 1.0
 
-github.setup    thepowersgang mrustc 4bee8893e8f471d315c1d2d9535216b116cdfe23
+github.setup    thepowersgang mrustc 3ec8cd8b1a3431124d278baefc8d0e34dc69b930
 
-version         20220111
+version         20220115
```
evanmiller, line 10: Doesn't really matter but the title says `20220125` (which I think is the correct value)

lang/mrustc/Portfile
```diff
@@ -2,14 +2,16 @@
...
 revision        0
 categories      lang devel
 license         MIT
-maintainers     nomaintainer
+maintainers     {@catap korins.ky:kirill} openmaintainer
```
evanmiller, line 14: Thank you for adopting my poor orphan child :-)

lang/mrustc/Portfile
```diff
@@ -59,46 +62,120 @@ compiler.blacklist *gcc*
 #   depends_run-append port:[string map {"macports-gcc-" "gcc"} ${configure.compiler}]
 #}
 
-# https://github.com/thepowersgang/mrustc/pull/192
-patchfiles-append patch-mrustc-macos-bootstrap.diff
+# Patches avaialble as git branch: https://github.com/catap/mrustc/tree/macports
```
evanmiller, line 65: `avaialble` => `available`

lang/mrustc/Portfile
```diff
@@ -59,46 +62,120 @@ compiler.blacklist *gcc*
...
+                port:libgit2 \
+                port:nghttp2 \
+                port:zlib
+
+# use the system python27 if present
+if {${os.platform} eq "darwin" && ${os.major} >= 11} {
+    set pythonfullpath   /usr/bin/python2.7
+} else {
+    set pythonfullpath   ${prefix}/bin/python2.7
+    depends_lib-append   port:python27
```
evanmiller, line 107: Should this be `depends_run`? (`depends_run` is used if there is no library linkage.)

lang/mrustc/Portfile
```diff
@@ -59,46 +62,120 @@ compiler.blacklist *gcc*
...
+        xinstall -m 0644 -W ${worksrcpath}/output-${rust_version} $f \
+            ${destroot}${prefix}/lib/mrustc-${rust_version_major}
+    }
+
+    # keep boostraped rust
+    xinstall -d ${destroot}${prefix}/libexec/mrustc-${rust_version_major}/bin
+    foreach f {rustc cargo} {
+        xinstall -m 0755 -W ${worksrcpath}/output-${rust_version} $f \
+            ${destroot}${prefix}/libexec/mrustc-${rust_version_major}/bin
+        system -W ${destroot}${prefix}/bin/ "ln -s ../libexec/mrustc-${rust_version_major}/bin/$f ./mrustc-${f}"
```
evanmiller, line 161: Can't wait to build some things with it :-)

lang/mrustc/Portfile
```diff
@@ -59,46 +62,120 @@ compiler.blacklist *gcc*
...
+
+    system -W ${destroot}${prefix}/libexec/mrustc-${rust_version_major}/lib/rustlib/${configure.build_arch}-apple-darwin/lib \
+        "install_name_tool -id ${prefix}/libexec/mrustc-${rust_version_major}/lib/rustlib/${configure.build_arch}-apple-darwin/lib/libtest.dylib ./libtest.dylib"
+
+    system -W ${destroot}${prefix}/libexec/mrustc-${rust_version_major}/lib/rustlib/${configure.build_arch}-apple-darwin/lib \
+        "install_name_tool -change ${worksrcpath}/run_rustc/output-${rust_version}/build-std2/${configure.build_arch}-apple-darwin/release/deps/libstd.dylib ${prefix}/libexec/mrustc-${rust_version_major}/lib/rustlib/${configure.build_arch}-apple-darwin/lib/libstd.dylib ./libtest.dylib"
 }
 
 # Run a Hello, world! test
-test.env            RUSTC_VERSION=${rust_version} MRUSTC_LIBDIR=output-${rust_version}
```
evanmiller, line 102: `MRUSTC_LIBDIR` was intentional because it needs to know where the built libraries are before they are installed. (Otherwise the `test` phase will look for libraries in `${prefix}`.)

lang/mrustc/files/0009-macOS-10.6-1.39.0-fix-build-issues-pre-Lion.patch
```diff
@@ -0,0 +1,137 @@
...
++ 
++ MachOYAML::LoadCommand::~LoadCommand() = default;
++--- src/llvm-project/llvm/tools/obj2yaml/macho2yaml.cpp
+++++ src/llvm-project/llvm/tools/obj2yaml/macho2yaml.cpp
++@@ -14,7 +14,21 @@
++ #include "llvm/Support/ErrorHandling.h"
++ #include "llvm/Support/LEB128.h"
++ 
++-#include <string.h> // for memcpy
+++#include <string.h> // for memcpy abd strnlen
```
evanmiller, line 48: `abd` => `and`

evanmiller (commented):
Very cool! Just a few minor comments from me

lang/mrustc/Portfile
kencu: In llvm/clang, it is listed as depends_lib, because you need it for both build and run (depends_build + depends_run = depends_lib). Here you only need it for the build I believe, and not at runtime, so it could be depends_build.

lang/mrustc/Portfile
catap: Indeed!

lang/mrustc/Portfile
catap: @evanmiller nope, the only reason to use python2.7 is make clang's cmake happy. After it is built, python2.7 is useless.

kencu: depends_build, as I mentioned

Gcenx:
I’ll be keeping an eye on your mrustc fork as mrustc seems to be the only sane way to bootstrap for i686-apple-darwin

evanmiller:
I am seeing some build failures locally (10.6.8) that look like upstream bugs – something about non-UTF-8 C sources and then an unrecognized symbol in the generated C code. I will try again with your latest patches and report back.

catap:
@evanmiller I've pushed disabling parallel builds, it should help.

The root cause that `mrustc` generates huge `.c` files and if it runs in paralells it might be corrupted :)

Can you try one more time?

evanmiller:
> Can you try one more time?

@catap Will do

kencu:
I've also started it running, and will report back later on too. Thanks again!

catap:
I just pushed a two more commits that introduced `rustc` and `cargo` via symlink. The good news it allows to build some old software. I've tried `racer` and it works.

Seems that partially support of `rust` for `10.6` is here.

kencu:
So - are we using mrustc as the compiler, or the rustc that is bootstrapped with mrustc as the compiler? I would hope the latter, but still looking...

catap:
@kencu I'm using `mrustc` as compiller to bootstrap supported rust and cargo (with some patches) which is used as usual rust compiler.

catap:
To be honest: I'm not sure about last two commits and would like to have some discussion / suggestion.

kencu:
what is "mrustc-rustc", exactly? rustc?

```
system -W ${destroot}${prefix}/bin/ "ln -s ./mrustc-rustc ./rustc"
```

catap:
@kencu yes, it is `rustc` which was bootstrapped by `mrustc`.

catap:
@kencu / @evanmiller / @Gcenx  any chance that you have access to i686? :) If yes, can you try to update supported arch? I feel that this can work here.

kencu:
I have tried it four times now, and I keep getting an error during the build that looks like this:

```
Process:         mrustc [21360]
Path:            /opt/local/var/macports/build/_opt_kencu-ports_lang_mrustc/mrustc/work/mrustc-3ec8cd8b1a3431124d278baefc8d0e34dc69b930/bin/mrustc
Identifier:      mrustc
Version:         ??? (???)
Code Type:       X86-64 (Native)
Parent Process:  minicargo [21304]

Date/Time:       2022-01-30 16:39:18.061 -0800
OS Version:      Mac OS X 10.6.8 (10K549)
Report Version:  6

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Application Specific Information:
abort() called

Thread 0 Crashed:  Dispatch queue: com.apple.main-thread
0   libSystem.B.dylib             	0x00007fff80cbb0b6 __kill + 10
1   libSystem.B.dylib             	0x00007fff80d5b9f6 abort + 83
2   mrustc                        	0x00000001dc041a02 Span::bug(std::__1::function<void ()(std::__1::basic_ostream<char, std::__1::char_traits<char> >&)>) const + 114
3   mrustc                        	0x00000001dda58c74 HirDeserialiser::deserialise_mir_terminator_() + 1764
4   mrustc                        	0x00000001dda54649 HirDeserialiser::deserialise_mir_terminator() + 313
5   mrustc                        	0x00000001dda543d6 HirDeserialiser::deserialise_mir_basicblock() + 182
6   mrustc                        	0x00000001ddaef9a3 D<MIR::BasicBlock>::des(HirDeserialiser&) + 35
7   mrustc                        	0x00000001ddaef976 _ZZN15HirDeserialiser15deserialise_vecIN3MIR10BasicBlockEEENSt3__16vectorIT_NS3_9allocatorIS5_EEEEvENKUlvE_clEv + 38
8   mrustc                        	0x00000001ddaef936 _ZNSt3__18__invokeIRZN15HirDeserialiser15deserialise_vecIN3MIR10BasicBlockEEENS_6vectorIT_NS_9allocatorIS6_EEEEvEUlvE_JEEEDTclclsr3std3__1E7forwardIS6_Efp_Espclsr3std3__1E7forwardIT0_Efp0_EEEOS6_DpOSC_ + 54
9   mrustc                        	0x00000001ddaef8d6 _ZNSt3__128__invoke_void_return_wrapperIN3MIR10BasicBlockEE6__callIJRZN15HirDeserialiser15deserialise_vecIS2_EENS_6vectorIT_NS_9allocatorIS8_EEEEvEUlvE_EEES2_DpOT_ + 54
10  mrustc                        	0x00000001ddaef896 _ZNSt3__110__function12__alloc_funcIZN15HirDeserialiser15deserialise_vecIN3MIR10BasicBlockEEENS_6vectorIT_NS_9allocatorIS7_EEEEvEUlvE_NS8_ISB_EEFS5_vEEclEv + 54
11  mrustc                        	0x00000001ddaee59d _ZNSt3__110__function6__funcIZN15HirDeserialiser15deserialise_vecIN3MIR10BasicBlockEEENS_6vectorIT_NS_9allocatorIS7_EEEEvEUlvE_NS8_ISB_EEFS5_vEEclEv + 45
12  mrustc                        	0x00000001ddaedf7b std::__1::__function::__value_func<MIR::BasicBlock ()()>::operator()() const + 75
13  mrustc                        	0x00000001ddaeaba3 std::__1::function<MIR::BasicBlock ()()>::operator()() const + 35
14  mrustc                        	0x00000001ddaea9d1 std::__1::vector<MIR::BasicBlock, std::__1::allocator<MIR::BasicBlock> > HirDeserialiser::deserialise_vec_c<MIR::BasicBlock>(std::__1::function<MIR::BasicBlock ()()>) + 769
15  mrustc                        	0x00000001dda542c3 std::__1::vector<MIR::BasicBlock, std::__1::allocator<MIR::BasicBlock> > HirDeserialiser::deserialise_vec<MIR::BasicBlock>() + 99
16  mrustc                        	0x00000001dda54039 HirDeserialiser::deserialise_mir() + 313
17  mrustc                        	0x00000001dda5391e HirDeserialiser::deserialise_exprptr() + 126
18  mrustc                        	0x00000001ddad7a68 HirDeserialiser::deserialise_constant() + 312
19  mrustc                        	0x00000001ddb336ea HirDeserialiser::deserialise_valueitem() + 458
20  mrustc                        	0x00000001ddb33513 D<HIR::ValueItem>::des(HirDeserialiser&) + 35
21  mrustc                        	0x00000001ddb334c6 HIR::VisEnt<HIR::ValueItem> HirDeserialiser::deserialise_visent<HIR::ValueItem>() + 70
22  mrustc                        	0x00000001ddb33473 D<HIR::VisEnt<HIR::ValueItem> >::des(HirDeserialiser&) + 35
23  mrustc                        	0x00000001ddb333de std::__1::unique_ptr<HIR::VisEnt<HIR::ValueItem>, std::__1::default_delete<HIR::VisEnt<HIR::ValueItem> > > HirDeserialiser::deserialise_ptr<HIR::VisEnt<HIR::ValueItem> >() + 62
24  mrustc                        	0x00000001ddb319d3 D<std::__1::unique_ptr<HIR::VisEnt<HIR::ValueItem>, std::__1::default_delete<HIR::VisEnt<HIR::ValueItem> > > >::des(HirDeserialiser&) + 35
25  mrustc                        	0x00000001dda5ab73 std::__1::unordered_map<RcString, std::__1::unique_ptr<HIR::VisEnt<HIR::ValueItem>, std::__1::default_delete<HIR::VisEnt<HIR::ValueItem> > >, std::__1::hash<RcString>, std::__1::equal_to<RcString>, std::__1::allocator<std::__1::pair<RcString const, std::__1::unique_ptr<HIR::VisEnt<HIR::ValueItem>, std::__1::default_delete<HIR::VisEnt<HIR::ValueItem> > > > > > HirDeserialiser::deserialise_istrumap<std::__1::unique_ptr<HIR::VisEnt<HIR::ValueItem>, std::__1::default_delete<HIR::VisEnt<HIR::ValueItem> > > >() + 675
26  mrustc                        	0x00000001dda5a74d HirDeserialiser::deserialise_module() + 205
27  mrustc                        	0x00000001ddb371bb HirDeserialiser::deserialise_typeitem() + 491
28  mrustc                        	0x00000001ddb36fc3 D<HIR::TypeItem>::des(HirDeserialiser&) + 35
29  mrustc                        	0x00000001ddb36f76 HIR::VisEnt<HIR::TypeItem> HirDeserialiser::deserialise_visent<HIR::TypeItem>() + 70
30  mrustc                        	0x00000001ddb36f23 D<HIR::VisEnt<HIR::TypeItem> >::des(HirDeserialiser&) + 35
31  mrustc                        	0x00000001ddb36e8e std::__1::unique_ptr<HIR::VisEnt<HIR::TypeItem>, std::__1::default_delete<HIR::VisEnt<HIR::TypeItem> > > HirDeserialiser::deserialise_ptr<HIR::VisEnt<HIR::TypeItem> >() + 62
32  mrustc                        	0x00000001ddb35483 D<std::__1::unique_ptr<HIR::VisEnt<HIR::TypeItem>, std::__1::default_delete<HIR::VisEnt<HIR::TypeItem> > > >::des(HirDeserialiser&) + 35
33  mrustc                        	0x00000001dda5af83 std::__1::unordered_map<RcString, std::__1::unique_ptr<HIR::VisEnt<HIR::TypeItem>, std::__1::default_delete<HIR::VisEnt<HIR::TypeItem> > >, std::__1::hash<RcString>, std::__1::equal_to<RcString>, std::__1::allocator<std::__1::pair<RcString const, std::__1::unique_ptr<HIR::VisEnt<HIR::TypeItem>, std::__1::default_delete<HIR::VisEnt<HIR::TypeItem> > > > > > HirDeserialiser::deserialise_istrumap<std::__1::unique_ptr<HIR::VisEnt<HIR::TypeItem>, std::__1::default_delete<HIR::VisEnt<HIR::TypeItem> > > >() + 675
34  mrustc                        	0x00000001dda5a79c HirDeserialiser::deserialise_module() + 284
35  mrustc                        	0x00000001dda5b6ad HirDeserialiser::deserialise_crate() + 365
36  mrustc                        	0x00000001dda5ca1c HIR_Deserialise(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 140
37  mrustc                        	0x00000001dc0949ef AST::ExternCrate::ExternCrate(RcString const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 351
38  mrustc                        	0x00000001dc0944f5 AST::ExternCrate::ExternCrate(RcString const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 37
39  mrustc                        	0x00000001dc0934ea AST::Crate::load_extern_crate(Span, RcString const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 6266
40  mrustc                        	0x00000001dc0938b4 AST::Crate::load_extern_crate(Span, RcString const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 7236
41  mrustc                        	0x00000001dc09172d AST::Crate::load_externs() + 1533
42  mrustc                        	0x00000001dc027451 main::$_3::operator()() const + 257 (main.cpp:304)
43  mrustc                        	0x00000001dc00085a void CompilePhaseV<main::$_3>(char const*, main::$_3) + 42 (main.cpp:110)
44  mrustc                        	0x00000001dbffe53c main + 780 (main.cpp:297)
45  mrustc                        	0x00000001dbffe174 start + 52

Thread 0 crashed with X86 Thread State (64-bit):
  rax: 0x0000000000000000  rbx: 0x0000000000000000  rcx: 0x00007fff5fbf6b48  rdx: 0x0000000000000000
  rdi: 0x0000000000005370  rsi: 0x0000000000000006  rbp: 0x00007fff5fbf6b60  rsp: 0x00007fff5fbf6b48
   r8: 0x00007fff5fbf6910   r9: 0x0000000000000020  r10: 0x00007fff80cb70fa  r11: 0x0000000000000202
  r12: 0x0000000000000000  r13: 0x0000000000000000  r14: 0x0000000000000000  r15: 0x0000000000000000
  rip: 0x00007fff80cbb0b6  rfl: 0x0000000000000202  cr2: 0x00000001ddb1f010

```

but I have done a slight modification to things that might make my system different. Let's see if Evan gets a different error. I tried with several different linkers and three different bootstrap clangs, and reset the optflags to "-O0" and still it happens, so it could be me.

catap:
@kencu this is so-called poor log :(

Can you share end of debug log? It allows to guess which file you need to read to understand the real issue.

evanmiller:
Still building on x86_64, will report back on that later tonight / tomorrow...

> To be honest: I'm not sure about last two commits and would like to have some discussion / suggestion.

I think `mrustc` should *not* provide rustc and cargo, only `mrustc` and `minicargo`. Then in the main `rust` port, use `mrustc` to provide a fully bootstrapped 1.39 compiler. That is, the `rust` compilation should look like

`mrustc` (stage0) => `rustc` (stage1) => `rustc` (stage2) => `cargo` etc

mrustc is the "alternative Rust compiler" so I think it should provide only itself – and not "real" Rust.

catap:
@evanmiller the issue that `mrustc` requires to patch sources of rust to be able to build it. Plus it requires additional patches for LLVM which is ships with rust.

From another hand bootstrap rust by mrustc is a tricky and will make rust port much more complicated. To be honest it is right now too complicated.

I thought about your way a lot, and current way with symlink and huge `mrust` port seems as the easy way to do it :)

evanmiller:
@catap Makes sense, I know you've put a lot of work into this so far!

In that case I would just suggest a rename of the binaries:

`${prefix}/bin/mrustc` (main mrustc, no version attached)
`${prefix}/bin/minicargo` (same)
`${prefix}/bin/rustc-1.39` (rustc compiled by mrustc)
`${prefix}/bin/cargo-1.39` (cargo compiled by mrustc)

Then on 10.6 the `rust` port symlinks `rustc` => `rustc-1.39` and `cargo` => `cargo-1.39`. Revbump when we upgrade to 1.54 :-)

kencu:
I deactivated the older mrustc that I had installed, and now the new one is building away, much further along.

It's the 5th try, and AFAICT the only different thing is the deactivated older mrustc, so I guess that is the difference, although I haven't done a retest to prove it, etc.

kencu:
failed because it found my MacOSX10.7.sdk in the SDKs folder and errored out. That is probably because we're not using the cmake 1.1. PortGroup here, and so the SDK is not be specified. Will see if I can send that in.

catap:
@kencu it haven't used cmake PortGroup because it calls cmake from series of make files.

catap:
BTW I've decided to split this PR into three indepened PRs:
 1. delivering mrustc
 2. switching rust to use mrustc on macOS <10.7
 3. switching cargo.

catap:
@evanmiller good point with naming. Much beater that I've used.

Just pushed an update.

evanmiller:
OK, the complete build is not finished (computer went to sleep) but it has successfully built `rustc` and `cargo` and is currently rebuilding the standard library using `rustc`. There were a few warnings in the generated C code but these can be addressed upstream.

I personally would prefer to keep the `mrustc` port "lean and mean" (`mrustc` + stdlib) and use a separate port to provide `rustc-1.39` and `cargo-1.39`. This will keep the `mrustc` build time <1 hr and make it easier to experiment with `mrustc` on new platforms (i686, PPC, ...).

catap:
@evanmiller well... from another hand until `mrustc` is ready, bootstrap of rust is very tricky and seems more like a side effect of this repo.

I can add a subport, but it makes things more complicated and longer in term of build because each subport is building separately.

evanmiller:
@catap I don't feel too strongly about it so do whatever is most productive for you. But I think you can supply a path to `mrustc` to prevent a re-build:

https://github.com/thepowersgang/mrustc/blob/master/minicargo.mk#L36

catap:
@evanmiller perhabs you're right to split it into 4 ports: `mrustc`, `minicargo`, `mrustc-rustc` and `mrustc-cargo`.

It requires some additional efforts and for now I haven't time to it. I'll do it in near future, but let keep it as draft for a while.

From another hand anyone can test current approach and report if something very bad :)

evanmiller:
@catap Sounds good. Perhaps the extra time will allow upstream to sort out the 1.54 issues...

evanmiller:
Everything builds here (`mrustc` => `rustc` => stdlib) on 10.6.8 👍

catap:
@evanmiller from another hand spliting isn't something huge. Maybe I'll do it today :)

evanmiller:
@catap I would just suggest keeping `minicargo` packaged with `mrustc` since a casual user will expect this to be included. So maybe the ports will be `mrustc`, `mrustc-rustc`, and `mrustc-cargo` (or `mrustc-rust` can include both `rustc` and `cargo`...).

catap:
@evanmiller I'm testing right now:
```
mrustc @20220125 (lang, devel)
Sub-ports:            minicargo, mrustc-rustc, mrustc-cargo, rustc-1.39
```

and `rustc-1.39` contains `rustc` and its `libstd`. It allows to make bootstrap of `cargo` quite native, from one hand, and keep `rust` port simpler.

Gcenx:
@catap I'll retry building the `i686-apple-darwin` target after the split, opened a PR to your fork adding `i686-apple-darwin` I'm sure more changes will be needed but the current mrustc port takes forever to compile so tracking down issues.

catap:
Ok, pushed the split:
```
mrustc @20220125 (lang, devel)
Sub-ports:            minicargo, cargo-1.39, mrustc-rustc, rustc-1.39
```

An idea of split:
 - `mrustc` contains `mrustc` compiler and its libraries;
 - `minicargo` contains `minicargo` which is needed to build everything;
 - `cargo-1.39` usual `cargo` which was compiled by `mrusct`;
 - `rustc-1.39` usual `rustc` which was boostraped with stage0 compiler `mrustc`.

lang/mrustc/Portfile
```diff
@@ -38,6 +40,7 @@ use_configure   no
 
 compiler.cxx_standard 2014
 compiler.c_standard 2011
+compiler.blacklist-append {clang < 900} {macports-clang-3.[0-9]} macports-clang-4.0
```
evanmiller, line 43: I think this line is canceled out by `compiler.blacklist *gcc*` below (should be `blacklist-append`, or else fix the underlying GCC issue... :-)

lang/mrustc/Portfile
```diff
@@ -59,46 +62,276 @@ compiler.blacklist *gcc*
...
+                OPENSSL_INCLUDE_DIR=[openssl::include_dir] \
+                OPENSSL_LIB_DIR=[openssl::lib_dir] \
+                RUSTC_VERSION=${rust_version} \
+                MRUSTC_TARGET_VER=${rust_version_major} \
+                "MINICARGO_FLAGS=--features curl-sys/force-system-lib-on-osx${legacy_features}" \
+                LIBGIT2_SYS_USE_PKG_CONFIG=1 \
+                OUTDIR_SUF=-${rust_version} \
+                "LLVM_CMAKE_OPTS_EXTRA=PYTHON_EXECUTABLE=${pythonfullpath}${legacy_cmake_opts}"
+
+# Step 0: let start from building mrustc and its libds
```
evanmiller, line 155: `libds` => `libs`

lang/mrustc/Portfile
```diff
@@ -59,46 +62,276 @@ compiler.blacklist *gcc*
...
 }
 
-# Run a Hello, world! test
-test.env            RUSTC_VERSION=${rust_version} MRUSTC_LIBDIR=output-${rust_version}
-test.run            yes
-test.target         test
+# Step 1: now I need minicargo
+subport minicargo {
+    description     Alternative implementation of Rust cargo
+    long_description {*}${description}
```
evanmiller, line 172: Should mention in the long description that the command line interface is not compatible with regular cargo (unlike mrustc which mostly accepts the same flags as rustc).

lang/mrustc/Portfile
```diff
@@ -59,46 +62,276 @@ compiler.blacklist *gcc*
...
+        xinstall -m 0755 ${worksrcpath}/output-${rust_version}/cargo ${destroot}${prefix}/bin/cargo-${rust_version_major}
+    }
+
+    livecheck.type  none
+}
+
+# Step 3: build a rustc by mrustc
+subport mrustc-rustc {
+    description     Rust which was compiled by mrustc.
+    long_description {*}${description}
```
evanmiller, line 233: Maybe mention that this port is intended for MacPorts developers (end users will want either `mrustc` or `rustc-<version>`.)

lang/mrustc/Portfile
```diff
@@ -59,46 +62,276 @@ compiler.blacklist *gcc*
...
+    build.target    -f minicargo.mk output-${rust_version}/rustc
+
+    destroot {
+        xinstall -m 0755 ${worksrcpath}/output-${rust_version}/rustc ${destroot}${prefix}/bin/mrustc-rustc-${rust_version_major}
+   }
+
+    livecheck.type  none
+}
+
+# Step 4: build rustc by mrusc-rustc
```
evanmiller, line 260: `mrusc-rustc` => `mrustc-rustc`

lang/mrustc/Portfile
```diff
@@ -59,46 +62,276 @@ compiler.blacklist *gcc*
...
+subport rustc-${rust_version_major} {
+    version         ${rust_version}
+    revision        0
+
+    description     Rustc and libstd which was bootstraped via mrustc.
```
evanmiller, line 265: `was bootstraped` => `were bootstrapped`

lang/mrustc/Portfile
```diff
@@ -59,46 +62,276 @@ compiler.blacklist *gcc*
...
+
+    depends_build-append \
+                port:libtool
+
+    set legacy_cmake_opts " CMAKE_LIBTOOL=${prefix}/bin/libtool"
+
+}
+
+# #See: https://github.com/thepowersgang/mrustc/issues/225
+# use_parallel_build no
```
evanmiller, line 137: Parallel builds are back on?

evanmiller (commented):
A few minor comments. I will try to build it again tonight.

evanmiller:
One issue that will need addressing with this subport approach is what happens when we switch to rust 1.54? Are the `cargo-1.39` and `rust-1.39` subports still available? Are they deprecated? Will users be automatically upgraded or do they need to install separately? Will `mrustc` need variants (or even subports) for each Rust version?

I think we need to decide this now, before putting the Rust version into the subport names.

evanmiller:
> I think we need to decide this now, before putting the Rust version into the subport names.

The simple solution would be to call the ports `rustc-legacy` and `cargo-legacy` and then not worry about keeping around old versions.

lang/mrustc/Portfile
Gcenx: Seem that minicargo is only really used to bootstrap mrustc-cargo

Instead change the description to explain minicargo is simply used to bootstrap mrustc-cargo

lang/mrustc/Portfile
catap: @evanmiller yay!

lang/mrustc/Portfile
catap: @evanmiller / @Gcenx nope, it is also used to build rustc. Let me rephase this a bit.

lang/mrustc/Portfile
catap: @evanmiller not sure hot to say it short :) Can you suggest anything?

lang/mrustc/Portfile
catap: @evanmiller nope, thanks.

catap:
@evanmiller I not sure that `-legacy` is good name. But I haven't got better one => let use it when.

lang/mrustc/Portfile
catap: It may works and depends on number of used CPU. But usually it fails ;)

catap:
Thus, maybe call binaries `rustc-legacy` and `cargo-legacy`?

lang/mrustc/Portfile
evanmiller: @catap Maybe rename it `rustc-bootstrap`? Since it really provides `rustc` and not `mrustc`.

"A stage-1 Rust compiler used only to compile rustc-legacy"

evanmiller:
> Thus, maybe call binaries `rustc-legacy` and `cargo-legacy`?

I think the version numbers should remain in the binary names (unless you decide to make `rust` conflict with `rustc-legacy` instead of symlinking it).

evanmiller:
OK – it is building the last package locally now, but so far so good. I'm not sure if there is a way to configure this, but some of the progress bars don't make sense (they get to 100% but then keep building for a long time). I think this is MacPorts reading the output of CMake even though CMake doesn't "own" the build.

catap:
@evanmiller progress bar is getting from `cmake` as result it isn't well aliened :(

catap:
@evanmiller I've just pushed the update. Right now it has:
 - `mrustc` contains `mrustc` compiler and its libraries;
 - `minicargo` contains `minicargo`;
 - `mrustc-rustc` contains `rustc` which was compiled by `mrustc`;
 - `mrustc-cargo` contains `cargo` which was compiled by `mrustc`;
 - `rustc-legacy` usual `rustc-1.39` which was boostraped with stage0 compiler `mrustc`;
 - `cargo-legacy` usual `cargo-1.39` which was compiled by `minicargo` with `rustc-1.39`.

It rebuilds LLVM twice to each time when it builds `rustc`. I can cache it by introducing one more port to cache llvm but it seems like an overkill.

Regarding `use_parallel_build no`. Now it may work, or might not. Anyway, build is quite frigiel and it may crashes at random place with parallel building => I've decided to switch it off.

lang/mrustc/Portfile
```diff
@@ -38,67 +40,325 @@ use_configure   no
...
+                port:pkgconfig
+
+depends_lib-append \
+                port:curl \
+                port:libgit2 \
+                port:nghttp2 \
+                port:zlib
+
+# use the system python27 if present
+if {${os.platform} eq "darwin" && ${os.major} >= 11} {
```
Gcenx, line 101: The latest version of macOS Monterey removed python, not only python3 gets made available when command line tools are installed.

catap:
@evanmiller / @Gcenx if you was able to build mrustc on macOS 10.6 can I ask you which compiler have you used?

Details: https://trac.macports.org/ticket/64585

Seems that something was updated on my system and it doesn't work anymore.

Gcenx:
@catap I’ve not attempted to compile this on 10.6, Ive been attempting to get this to compile target `i686-apple-darwin`.

Since this was split into subports it might be easier to deal with but the compile time is annoying.

kencu:
The right way to rebuild libcxx should be in the libcxx port notes, if you need it.

I put it there to show up if you don't have the emulated_tls variant enabled.

catap:
@evanmiller / @kencu the wired things that it works without any rebuilding...

catap:
@kencu any reason why `libcxx` should be rebuilded?

kencu:
libcxx is currently a two-stage build on 10.6.8, because clang-3.4 can't build the parts of libcxx that are needed for thread_local storage to work (cxa_thread_atexit, in particular).

To get around this:

1. you build a partially capable version of libcxx with clang-3.4, then build clang-5.0
2. you use clang-5.0 to rebuild a fully capable libcxx with thread_local storage enabled

there are ways this can be automated, but it's a huge PITA, and I never got around to it. I was willing to make a prebuilt libcxx with thread_local enabled available for download, but it is preferred to build it from bare metal of course.

If you or Evan or anyone else feels like taking on the libcxx thread_local bootstrap problem, go nutz! 

My thought as to how this might be done was to use an isolated gcc7 (or gcc10) to build a bootstrap clang-5.0 against libstdc++ (that works), and then use that isolated, fully capable clang-5.0 to build libcxx with thread_local enabled off the bat. I started working on this, but got distracted by the wonders of Ubuntu and Debian.

catap:
@kencu my point here is simple: without rebuilding `libcxx` with `+emulated_tls` this port can't be built => means no buildbot for all rust ports :(

And I have no idea how to enforce dependency for port's variant.

lang/mrustc/Portfile
catap: @Gcenx I'm using macOS 12.1 on my laptop and it has `python2.7`:
```
√ macports-ports % ls -l /usr/bin/python2.7
lrwxr-xr-x  1 root  wheel  75 Dec  8 00:39 /usr/bin/python2.7 -> ../../System/Library/Frameworks/Python.framework/Versions/2.7/bin/python2.7
√ macports-ports %
```

Gcenx:
@catap as this is a draft it might be a good idea to resolve the libcxx issue fully, @kencu already provided a mostly complete gcc7-bootstrap port in a now closed PR a couple of additional patches and it’s basically good to go.

Next would be a clang5-bootstrap that would be used to build libcxx, then modify libcxx and bump accordingly so libcxx is complied with +emulated_tls by default

lang/mrustc/Portfile
Gcenx: @catap once you update your system that will be gone, it’s gotten a lot of people complaining online.

kencu:
> And I have no idea how to enforce dependency for port's variant.

Well, there is a macports PortGroup for this, active_variants.

catap:
@kencu thanks, this makes things much better!

Gcenx:
Quick example;
```
PortGroup active_variant 1.1

if {${os.platform} ne "darwin" && ${os.major} < 11} {
   require_active_variants port:libcxx emulated_tls
}
```

Though as @kencu already pointed out that doesn’t really resolve the issue.

The real issue is ensuring libcxx is built with +enunlated_tls from the start.

catap:
@Gcenx yes, it doesn't resolve an issue. Anyway, it fails near at start with nice message. And not "something can't be linked" after an hour.

kencu:
If you folks are redoing the bootstrapping of libcxx, then there is one more thing to ponder.

Back in 2016 when I set this up, I enabled cxa_thread_atexit in libcxx and it needed emutls.c. So I copied that into libcxxabi.dylib along with cxa_thread_atexit and moved on.

But -- emutls.c really should not primarily be in libxccabi.dylib. It should exist instead in the compiler_rt.a library for each version of clang from clang-5.0 up to the newest one that matters on 10.6.8 (currently I build clang-13 there). Right now, because emutls.c only exists in libcxxabi.dylib, there is this weird thing that happens where if you want to build a c program that uses Thread_local (or whatever it is in c) it requires linking to libcxxabi.dylib to get the emutls symbols.

That is frickin' stupid, and I should have put the emutls.c in compiler_rt.a in the first place, but never did. 

Now we have a number of programs that use thread_local that look for emutls in libcxxabi.dylib, but none on the buildbots at least (that never had this working). 

So perhaps this is the time to put emultls.c where it belongs, get it out of libcxxabi.dylib, make Thread_local for c programs work properly, and all will back to as it should be again.

catap:
@kencu I agree with you and I feel that the really right place to put `emutls.c` into LLVM code :)

catap:
Regarding question why it is soo long to compile:
```
macOS-Snow-Leopard:rustc-build catap$ ls -lh librustc.rlib.*
-rw-r--r--  1 macports  wheel   862M Feb  2 02:31 librustc.rlib.c
-rw-r--r--  1 macports  wheel    18K Feb  2 02:26 librustc.rlib.d
-rw-r--r--  1 macports  wheel   1.6M Feb  2 02:30 librustc.rlib.hir
macOS-Snow-Leopard:rustc-build catap$ 
```

yeah, it compiles near of 1Gb files.

kencu:
> @kencu I agree with you and I feel that the really right place to put `emutls.c` into LLVM code base :)

It actually already lives in compiler_rt, just disabled on Apple. Just have to tweak the cmake scripts to get it in there on < 10.7.

Gcenx:
> It actually already lives in compiler_rt, just disabled on Apple. Just have to tweak the cmake scripts to get it in there on < 10.7.

@kencu so that's why you were mentioning needing a clang bootstrap compiler to compile libcxx and why gcc7-bootstrap wouldn't be used to compile libcxx.

evanmiller:
@catap / others Will 10.7+ users have any need for `rust-legacy` and `cargo-legacy`? If not then maybe eliminate these subports and roll them into the main ports / Portfiles. So then from the user's perspective 10.6 has real Rust and cargo, just a few (or many) versions behind, no symlinks. Then the complete Rust toolchain is:
* `mrustc`
* `mrustc-rustc`
* `mrustc-cargo`
* `mrustc-llvm`
* `rust`
* `cargo`

kencu:
No, that’s not why you would need clang-5.0 bootstrap.

You can build libcxx with gcc7, in fact. There are issues in the Portfile and build to be overcome, and it winds up either statically or dynamically linked to parts of libgcc forever after, but it can be done. I ran it built like that on macports for several years on 10.5 PPC.

going through the gcc7 -> clang-5.0 bootstrap phase, all built ideally against system roots, gives you a clean bootstrap compiler that can be used to build all existing toolchain parts without altering any existing toolchain Portfiles, all of which assume clang is the compiler. (You also need to build it with cmake-bootstrap, that builds with system tools too.)

So you can then use that isolated bootstrap clang-5.0 to build cctools, libcxx with emulated_tls, a current ld64 and libtapi, a current cmake, etc. — a full toolchain.

OTOH, it’s a big project, and would anyone really mind installing a prebuilt downloaded fully-capable libcxx???? It’s not the NSA we’re supporting here………

catap:
@evanmiller I'm thinking a bit more about that and would like to make full (but long) bootstrap via a few (20?) subports.

Anyway, it will be separated PR and separated port.

It also allows to remove a lot of my hacks from rust port which I've used to bootstrap it on 10.7 :)

=> let me keep it for everyone ;)

P.S. I'm almost finished to test it and soon should push the new version and be ready for review.

Gcenx:
@kencu thanks that mostly confirmed what I had in mind.

@evanmiller best to keep these isolated from the main rust/cargo ports.

@catap that sounds like a good idea but I’d hold off until mrustc can compile a functional i686 target and can compile 1.5.4 to shrink the amount of jumps required.

Each rust/cargo bump should ideally be installed into the same location, use 1.5.4 to compile 1.5.5 then before installation deactivate the current 1.5.4, something similar is done for for mingw-gcc ports.

Gcenx:
> x86_64 rust is supposed to be able to build i386-darwin target software. It was still an option last time I looked.

I don’t see this explained anywhere and Macports also doesn’t allow this.

> So the current x86_64 rust you download should be able to build a current i386 rust directly.

I remember linking a lot of documentation on doing this a while back and it sucked hard lightly why the rust/cargo Portfiles don’t already handle this.

> If it can’t, upstream might even fix that for you…

Upstream no longer supports this target so I doubt they would fix things, they might accept PRs that fix any issues though.

kencu:
what can currently be done inside macports depends on lots of things (as you and I know well!).

As mentioned, last time I looked, i386-darwin was still an available target, as llvm supports it.

If it is not any longer an available target, then you’ve got much larger problems than building an i386-arch rustc…

Gcenx:
> what can currently be done inside macports depends on lots of things (as you and I know well!)

Agreed.

> As mentioned, last time I looked, i386-darwin was still an available target, as llvm supports it.
> 
> If it is not any longer an available target, then you’ve got much larger problems than building an i386-arch rustc…

I was thinking you meant compiling software directly using x86_64 rust/cargo that I believe isn’t possible.

The target still exists upstream and lightly won’t be removed. It’s just compiling stage0 is a PITA that’s where I believe mrustc can be into play. This will also be the case once upstream no longer provides x86_86 bootstrap compilers for Darwin.

kencu:
it is supposed to be possible to compile i386-apple-darwin directly from an x86_64 rust (just like we do with gcc in macports).

It looks like you are also supposed to be able to build an i386-apple-darwin rustc too. Both are Tier3, of course.

https://doc.rust-lang.org/nightly/rustc/platform-support.html

MacPorts dropped i386 rust 2 years ago when upstream stopped making an i386-darwin prebuilt available. That was not actually really necessary, TBH, but the rust Portfile was a mess of spagetti, and I think frustration came in, and nobody was prepared to sort that out.

catap:
@kencu yeah, and my attempt to bootstrap it via a few sub-ports is just a way to simplified rust port. It is very messy :(

Gcenx:
@kencu and the fallout from not dealing it that then is after something moves to rust each Portfile becomes a mess due to needing to force an older version just for i386 support.

@catap I can see the provided ports becoming the way to bootstrap current rust/cargo for all targets Macports needs.

catap:
@Gcenx yep. My plan is:
 - deliver `rust-legacy` and `cargo-legacy` as part of this PR;
 - introduce `rust-XX` and `cargo-XX` ports to bootstrap whole chain from legacy to current one as new port `rust-bootstrap`;
 -  move current rust and cargo port to something like `rust-vanilla` and `cargo-vanilla`, and use it on unsupported architecures.

But, let move one step on the time :)

catap:
About the last push. It contains:
 - `mrustc` as compiler and its cargo;
 - `mrustc-llvm` LLVM which was used to build both rustc that allows to save a lot of time
 - `mrustc-rustc` and `mrustc-cargo` contain `rustc` and `cargo` which was compiled by `mrustc`. Slow, huge and unoptimized;
  - `rustc-legacy` and `cargo-legacy` contain old `rustc` and `cargo` which was compiled rustc compiler.

So, the first one and the last ones is useful, everything in the middle is optional and mainly was added to make rebuilding and development simpler.

Gcenx:
> @Gcenx yep. My plan is:
> 
> * deliver `rust-legacy` and `cargo-legacy` as part of this PR;
> * introduce `rust-XX` and `cargo-XX` ports to bootstrap whole chain from legacy to current one as new port `rust-bootstrap`;
> * move current rust and cargo port to something like `rust-vanilla` and `cargo-vanilla`, and use it on unsupported architecures.
> 
> But, let move one step on the time :)

Sounds good to me.

I have a though, wouldn’t it make more sense to treat each target as it’s own support?

That would mean we could simply patch everything to force i686-apple-darwin and not need to worry.

Then the main rust/cargo ports could use our bootstrapped version over upstreams provided stage0.

catap:
> I have a though, wouldn’t it make more sense to treat each target as it’s own support?

Well, you just need libs to be honest. So, it might be good idea, indeed.

kencu:
> @Gcenx  and the fallout from not dealing it that then is after something moves to rust each Portfile becomes a mess 

So Dean I don't want to pollute this PR too much, and TBH I don't know much about rust at all. But I spent five minutes and cross-compiled and installed the i686-apple-darwin rust toolchain like this on 10.14, and it went without a hitch (once I set the SDK):

```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
source $HOME/.cargo/env
rustup target list

rustup +nightly component add rust-src
cargo new foo
cd foo
export SDKROOT=/
cargo +nightly build -Z build-std --target i686-apple-darwin
```

I have not as yet actually build anything with rust as i686 using this installed target toolchain, but this might get you close to where you want to be.

Adding this to the rust port might not be all that hard? Hope you found this helpful.

kencu:
yes, cross-compiling to i686 works:
```
$ cargo new kentest1
$ export SDKROOT=/
$ cd kentest1
$ cargo +nightly build -Z build-std --target i686-apple-darwin

$ file ./target/i686-apple-darwin/debug/kentest1
./target/i686-apple-darwin/debug/kentest1: Mach-O executable i386

$ ./target/i686-apple-darwin/debug/kentest1
Hello, world!

```

Gcenx:
@kencu guess I’ll give rustup another go it’s been a while maybe there was a reason I didn’t use it?

catap:
Regarding naming: I've renamed subports back to `rust-1.39` and `cargo-1.39`. Why? Because I've started to write port to bootstrap all chain :) Now I have `1.40` and testing `1.41`...

It allows to bootstrap near the last possible rust for any macOS.

As soon as `mrustc` introduced support of `1.54.0` I can short the tree to bootstrap and introduce support of `arm64`.

evanmiller:
Still building OK here on 10.6 as of yesterday's pushes... looks like you have a plan in motion so I'm going to unsubscribe from updates here. Just @ me if you need anything :-)

catap:
@evanmiller to be honest I won't plan to touch this PR if it passes CI. Anyway, I'll ping directly everyone who is involve when it happens.

catap:
@evanmiller / @kencu / @Gcenx the PR is ready and passed all CI. I haven't planned to touch it anymore. May I ask you for review?

evanmiller:
I haven't built the latest-latest but it looks OK to me.

catap:
@kencu because you are a person who touched macports clang I'd love to ask you a question. Maybe you know it or can suggest where to ask.

I've noticed a small difference with `llvm` which is ships with macports and which I build in by rust. Difference related to `llvm/IR` or something near here. Let me show:
MacPorts' clang-9.0:
```
macOS-Snow-Leopard:x86_64-apple-darwin catap$ nm -gU /opt/local/libexec/llvm-9.0/lib/libLLVM.dylib | grep 4llvm10CallbackVHE
0000000002ddfc80 S __ZTIN4llvm10CallbackVHE
0000000001b08d30 S __ZTSN4llvm10CallbackVHE
0000000002ddfc50 S __ZTVN4llvm10CallbackVHE
macOS-Snow-Leopard:x86_64-apple-darwin catap$ 
```
and rust's LLVM:
```
macOS-Snow-Leopard:x86_64-apple-darwin catap$ nm -gU llvm/lib/libLLVM.dylib | grep 4llvm10CallbackVHE
0000000002f89b98 S __ZTVN4llvm10CallbackVHE
macOS-Snow-Leopard:x86_64-apple-darwin catap$ 
```

kencu:
```
/opt/local/bin/llvm-cxxfilt-mp-11 __ZTIN4llvm10CallbackVHE
typeinfo for llvm::CallbackVH
```
So that is what we're looking for.

Presumably that type CallbackVH crept in somewhere between llvm/clang-9.0 and whatever version of llm (with their own patches, I might add) they are using.

Let's see where ....

kencu:
so it seems that llvm-9.0 on MacPorts (I'm on 10.14 today) does not recognize that symbol, but llvm-10 and up do.
```
Macpro2010:~$ /opt/local/bin/llvm-cxxfilt-mp-9.0 __ZTIN4llvm10CallbackVHE
__ZTIN4llvm10CallbackVHE
Macpro2010:~$ /opt/local/bin/llvm-cxxfilt-mp-10 __ZTIN4llvm10CallbackVHE
typeinfo for llvm::CallbackVH
Macpro2010:~$ /opt/local/bin/llvm-cxxfilt-mp-11 __ZTIN4llvm10CallbackVHE
typeinfo for llvm::CallbackVH
Macpro2010:~$ /opt/local/bin/llvm-cxxfilt-mp-12 __ZTIN4llvm10CallbackVHE
typeinfo for llvm::CallbackVH
Macpro2010:~$ /opt/local/bin/llvm-cxxfilt-mp-13 __ZTIN4llvm10CallbackVHE
```

the llvm's used by rust do not correlate exactly with the llvm's released by the llvm project. the rust guys have some extra patches here and there.

but perhaps this starts a clue as to what is going on?

catap:
@kencu I guess I figured it out: https://github.com/rust-lang/rust/pull/93640

catap:
@kencu yep, this patch help to move forward. Good.

catap:
One more force push: excluded unneeded patch

catap:
To be honest: I don't like an idea with tons of `install_name_tool` and some bash wrappers. I feel that no rush here and I can think a bit longer to make better (and simpler) approach.

evanmiller:
As a small data point, using `rust-1.39` I was finally able to build `mozjs68` on Snow Leopard. Looking forward to the final release.

catap:
A brand new layout is here :)

 - `mrustc` a port that brings `mrustc` and `minicargo` which can be used as dedicated compiler;
 - `mrustc-rust` a rust which was bootstraped via `mrustc`, a bit ugly yeah;
 - `rust-X` a specified version of `rust` which was bootstraped by `rust-(X-1)` or `mrustc-rust` at begining;
 - `rust-legacy-bootstrap` a port that contains patches that introduce support macOS 10.6 :)

I'm working to port my patch for macOS 10.6 to newest version of rust but it time consuming task.

Yes, latest rust on 10.6 seems quite possible.

To be honest: a patch for rust is surprisingly small:
```
 31 files changed, 190 insertions(+), 15 deletions(-)
```

Right now I've pushed only first step: `mrustc` and `mrustc-rust`. CI time limit is near of 6 hours and I don't thing that it is possible to build rust with llvm twice for this time.


 The next PR will be `rust-1.39`.  And so on, until I've reached the current state.
 
 Cc: @kencu, @evanmiller, @Gcenx

catap:
Seems that 1.54.0 is here. Let me update when.

kencu:
If you can now go clang -> mrustc -> rustc 1.54 then that is a real leap forward!

Thanks, will try.

catap:
@kencu I'll ping you when it's done. Right now I'm polish it :)

catap:
@kencu anyway, I have strange question. Right now CI fails because of wrong checksum and the issue that port fetches both distfiles from github.

Do you know any nice way to overstep this issue?

catap:
@kencu, @evanmiller, @Gcenx it's ready for review.

Feel free to install `mrustc-rust` and enjoy usual `rust` and `cargo` in `/op/local/libexec/mrustc-rust`. I haven't pushed the next steps because building time of this is incredible (~5 hours) and I hope to have a green mark from CI.

Anyway, if your machine has enough RAM (enough means 8G) feel free to edit this port file and uncomment lines 180-181.

The root cause of such memory consumption is paralell run of `clang` which is compiling:
```
-rw-r--r--  1 macports  wheel   834M Feb 10 15:15 rustc-build/librustc_query_impl.rlib.c
-rw-r--r--  1 macports  wheel   431M Feb 10 16:00 rustc-build/librustc_mir.rlib.c
-rw-r--r--  1 macports  wheel   399M Feb 10 15:04 rustc-build/librustc_middle.rlib.c
```

evanmiller:
Have you tested GCC support since the Core2 patches were merged? Would be nice to support it at least with the base `mrustc`.

kencu:
Last night I installed these:

```
  mrustc @20220209_0 (active) requested_variants='' platform='darwin 10' archs='x86_64' date='2022-02-09T21:39:41-0800'
  mrustc-rust @20220209_0 (active) requested_variants='' platform='darwin 10' archs='x86_64' date='2022-02-10T03:04:57-0800'
  rust-legacy-bootstrap @20220209_0 (active) requested_variants='' platform='darwin 10' archs='noarch' date='2022-02-09T21:26:22-0800'
```
and they seem to provide a working rustc and cargo, versions 1.54.
```
SnowLeopardServerVM:mrustc $ /opt/local/libexec/mrustc-rust/bin/cargo  -V
cargo 1.54.0
SnowLeopardServerVM:mrustc $ /opt/local/libexec/mrustc-rust/bin/rustc  -V
rustc 1.54.0-stable-mrustc
```

I'm not sure what you changed after that, but these seem to work. 

Now have to figure out how to get MacPorts to accept them as replacements for rust/cargo, until such time as they are used to bootstrap to current rust versions in the real ports.

Looking good.

catap:
@kencu I've updated used version of `mrustc` which allows to run without any additional patches to it :)

I also fixed a build of cargo which produces unworking cargo if `libgit2` was installed.

Regarding integration: my plan is backport my legacy patch to current master and introduce a few `rust-...` ports. As soon as it merged I start to push them one-by-one.

After that I plan to rename current rust port to `rust-upstream` which allows to remove near of all hacks that it has :)

And when add a simple wrapper-port `rust`  which installs `rust-upstream`  for modern OS and some legacy port for old one.

To be honest everything is ready and just wait to be opened :)

evanmiller:
Lint says:

```
port lint rust-legacy-bootstrap:
Warning: Line 27 should say "use_configure no" instead of declaring an empty configure phase
```

I think at this point I wouldn't worry about CI and I would just let the build bots tell you about any platform issues.

catap:
@evanmiller well, when I can add one more ready to use commit: `rust-1.54` :)

kencu:
Do you think you really need the rust-legacy-bootstrap port? It will cause blowback from the MacPorts managers to have that, just providing patches for other ports (at least that looks like what it does to me).

If you just need a couple of big patches, the usual way is to put them somewhere accessible (see the apple-gcc42 port's gpl variant, for example) and download the big patch during the build. Homebrew hosts about 10,000 patches on github that are downloaded during their builds.

In the big picture, this works. It has a lot of hacks at present, which is fine by me, but will cause big maintenance headaches for MacPorts going forward if you get busy with something else or have a stroke or something. I won't be fixing such issues if/when they arise.

So it would be good to get one of the people who will be on the hook to fix such issues should they arise to hit the commit button and move this along. That way, nobody can say afterwards they didn't like this or that about how it was done.

PR #13822, opened 2022-01-30
emacs-mac-app: use self-contained build

amake:
This unifies the build process with emacs-mac-app-devel. The patch for `lib-src/Makefile.in` is only needed for the latter.

I'm opening this PR to take advantage of the CI checks.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.2 21D49 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you


pmetzger:
Do any of your recent fixes (such as for building a self contained .app bundle) apply to the other Emacs port?

amake:
The NS port is already self-contained (unlike `--with-mac`, `--with-ns` implies self-contained) and uses basically the same strategy for `make install` and destroot. So I don't think so.

The only thing I noticed that could be applied to the regular Emacs port is that the build arg `BYTE_COMPILE_EXTRA_FLAGS='--eval "(setq comp-speed 2)"'` is probably unneeded: `comp-speed` was renamed to `native-comp-speed` and 2 is its default value.

PR #13829, opened 2022-01-30
binwalk: update python dependency from 3.9 to 3.10

jrabinow:
#### Description

binwalk: bump python dependency from 3.9 to 3.10

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.1 21C52 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/cross/binwalk/Portfile
+++ b/cross/binwalk/Portfile
@@ -6,7 +6,7 @@ PortGroup           github 1.0
 
 github.setup        ReFirmLabs binwalk 2.3.3 v
 github.tarball_from archive
-revision            0
+revision            1
 
 categories          cross
 platforms           darwin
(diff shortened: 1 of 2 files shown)


ryandesign:
Because this changes binwalk's library dependencies and the contents of the files it installs, its revision must be increased.

The commit message only mentions binwalk but you also changed py-pylzma. Maybe the py-pylzma change should be in its own commit.

jrabinow:
Done!
I've separated out 2 commits as suggested - please let me know if there should be a separate PR.

Previously, when updating dependencies, I did things in a single commit without trouble - should I adopt a "one commit per port" approach going forward?

ryandesign:
> I've separated out 2 commits as suggested - please let me know if there should be a separate PR.

Thanks. Since the second commit depends on the first, and they're both small changes, and both ports are maintained by the same person, they're fine being in the same PR.

> Previously, when updating dependencies, I did things in a single commit without trouble - should I adopt a "one commit per port" approach going forward?

Not necessarily. If you're doing the same thing in a dozen ports, like updating your maintainer email address, then doing it in one commit is fine. But here, where you're doing two different things, I think it's clearer to have separate commits. Consider the situation where someone may be looking at [a log of commits only for the py-pylzma port](https://github.com/macports/macports-ports/commits/master/python/py-pylzma). It would be confusing to see a commit message there which doesn't mention how it affected the py-pylzma port.

jrabinow:
Got it, makes sense. Thanks for the explanation

PR #13848, opened 2022-01-31
oui: new port

thatmattlove:
#### Description

oui is a MAC Address CLI toolkit written in go

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.1 21C52 arm64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/net/oui/Portfile
+++ b/net/oui/Portfile
@@ -1,0 +1,39 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+revision            0
+
+go.setup            github.com/thatmattlove/oui 0.1.8 v
+
+homepage            https://github.com/thatmattlove/oui
+
+categories          net
+
+description         MAC Address CLI Toolkit
+
+long_description    Look up MAC Address vendors and convert MAC Addresses \
+                    to other formats — offline and at the console.
+
+maintainers         {stunninglyclear.com:matt @thatmattlove} openmaintainer
+
+license             BSD
+
+depends_build-append \
+                    port:goreleaser
+
+fetch.type          git
+
+build.env-delete    GO111MODULE=off GOPROXY=off
+build.cmd           goreleaser
+build.pre_args      build
+build.args          --single-target
+
+github.livecheck.regex  {([0-9.]+)}
+
+destroot {
+    xinstall -m 0755 \
+        ${worksrcpath}/dist/${name}_${goos}_${goarch}/${name} \
+        ${destroot}${prefix}/bin/
+}


net/oui/Portfile
```diff
@@ -0,0 +1,44 @@
...
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+revision            0
+
+version             0.1.8
+name                oui
+
+homepage            https://github.com/thatmattlove/oui
```
ryandesign, line 11: Don't set `version`, `name`, `homepage` manually; let `github.setup` do it for you.

net/oui/Portfile
```diff
@@ -0,0 +1,44 @@
...
+categories          net
+
+description         MAC Address CLI Toolkit
+
+long_description    Look up MAC Address vendors and convert MAC Addresses \
+                    to other formats — offline and at the console.
+
+maintainers         {stunninglyclear.com:matt @thatmattlove} openmaintainer
+
+license             BSD-3-Clause-Clear
```
ryandesign, line 22: This is not a license name we recognize. (We do not use SPDX license identifiers yet.) We use the license name "BSD" for this.

net/oui/Portfile
```diff
@@ -0,0 +1,44 @@
...
+distname            ${name}_${version}_${os.platform}_${os.arch}64
+
+checksums               ${name}_${version}_darwin_amd64.tar.gz \
+                        sha256  cd26e80c45569ce5aced00892f7226b84dbe25b4a372c74fcd45df5427264fba \
+                        rmd160  997ded2bbb555f3a2d7baedb5ba83c331607f08b \
+                        size    5086101 \
+                        ${name}_${version}_darwin_arm64.tar.gz \
+                        sha256  45cf5925aa6d51d5edf8a1915f3d58b841fa97de3b0ed18273326080c959dc59 \
+                        rmd160 2a3f3c440182b70623cc510764bea709541ff93b \
+                        size    4924616
```
ryandesign, line 36: You've only provided checksums for arm64 and x86_64 binaries, so you'll need to add the line `supported_archs arm64 x86_64` somewhere so that users on other architectures get a better error message.

The arm64 rmd160 line is not indented the same as the others.

net/oui/Portfile
```diff
@@ -0,0 +1,44 @@
...
+
+worksrcdir
```
ryandesign, line 38: Instead of clearing `worksrcdir`, we prefer you set `extract.mkdir yes` and then use `${worksrcpath}` instead of `${workpath}` in the `destroot` block below.

ryandesign (commented):
I see that this port is installing a binary. MacPorts is typically a build-from-source system. Can this be built from source? @herbygillot You've dealt with lots of go-based ports so maybe you can weigh in on whether this portfile is doing things in the recommended way.

thatmattlove:
Thanks for your review, I'll address those with another commit.

Regarding building from source, it's certainly possible, but I ran into two hurdles with this:

1. Using `go.vendor` to specify Go dependencies and checksums (using [amake/go2port](https://github.com/amake/go2port)). This _almost_ worked, except that one nested indirect dependency failed the checksum validation, which prevented the build from succeeding.
2. Trying to hijack `configure.cmd` to run `go install` prior to `go build` (`${build.cmd}`). This failed for some reason I couldn't quite figure out (and unfortunately did not record) — I recall it being related to not being able to resolve the dependencies at build, as if `go install` did not put the dependencies in the expected location, even though `$GOPATH` seemed to be set properly.

Ultimately, it seemed much more straightforward to package the binaries. However, if you have any guidance on how to handle either of the above scenarios, I'm certainly willing to give it a shot.

herbygillot:
Hey @thatmattlove , @ryandesign:

So 100% agreed that `go2port` can be _very_ fidgety.  It has a few issues, the the primary one is that it is not doing the proper normalization for modules whose URLs have changed.  For example, `go2port` will see `github.com/jtolds/gls` from `go.mod` and add that in exactly that way to `go.vendors`, but the Github URL has changed to `github.com/jtolio/gls`. There are a number of other issues as well, such as `go2port` not doing the right thing for module paths that are deeper than 3 levels, such as `github.com/author/module/foo/bar`.

MacPorts is setting `GO111MODULE=off` and `GOPROXY=off` so that it can force the use of vendored modules with `go.vendor`.  You can remove those environment variables and Go can build as it normally does, without the use of `go.vendor`.  Here's a minimal Portfile that does that:

```
PortSystem          1.0

go.setup            github.com/thatmattlove/oui 0.1.8 v
github.tarball_from archive
revision            0

# Re-enable Go modules and allow Go to fetch dependencies itself

checksums           ${distname}${extract.suffix} \
                        rmd160  fc17b9f3afbc6f357c59fc1d6f0f98273f4ff7a0 \
                        sha256  b3cb775c8ea6bf48b3b2bebcae7bb6c072620f11cdcf1b1092bae8fa15989e82 \
                        size    22507
```

`oui` builds just fine with the default `build.cmd` (`go build`), but if you want to run a `go install` before the `go build`, you can stuff both into the `build.cmd`:

```
build.cmd           {go install && go build}
```

As an aside, I see `oui` uses `goreleaser`.  If one does go the `goreleaser` route, you need to use the `git` fetch method, instead of the checksummed source archive that we normally prefer in MacPorts.  You can see an example of this in the [Portfile for `cue`](https://github.com/macports/macports-ports/blob/master/devel/cue/Portfile). But again, MacPorts prefers that ports use explicit checksummed source archives by default.

thatmattlove:
@herbygillot — thank you, this was _tremendously_ helpful! I've updated the Portfile to be more similar to the Portfile for `cue`, and confirmed it builds perfectly.

thatmattlove:
Wonderful, done.

herbygillot:
Apologies @thatmattlove, one last nitpick: let's amend the commit message. As it is right now, there's a lot in there that doesn't match the current commit's contents.  After that, I think we should be good to go.

herbygillot:
Thanks for contributing to MacPorts, @thatmattlove.

PR #13902, opened 2022-02-06
mozjs68: new port

evanmiller:
#### Description

gjs is out of date; newer versions require an updated mozjs. A gjs update can be prepared after this and #13897 are merged.

One major improvement over the mozjs60 Portfile is that mozglue is now compiled statically. This eliminates the need for two of the patches, and also avoids a file conflict over `libmozglue.dylib` with mozjs60.

After #13817 is merged, this mozjs68 port can be made available on 10.6+ (already compiled and tested locally on 10.6.8 / x86_64). mozjs78 and mozjs91 will be explored later. 

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/lang/mozjs68/Portfile
+++ b/lang/mozjs68/Portfile
@@ -1,0 +1,124 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim: fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           muniversal 1.0
+PortGroup           legacysupport 1.1
+
+# clock_gettime, TARGET_OS_SIMULATOR
+legacysupport.newest_darwin_requires_legacy 15
+
+name                mozjs68
+version             68.4.2
+set version_major   68
+categories          lang
+platforms           darwin
+license             {MPL-2 LGPL-2.1+}
+maintainers         nomaintainer
+description         JavaScript-C Engine
+long_description    SpiderMonkey is Mozilla's JavaScript engine written in C/C++. \
+                    It is used in various Mozilla products, including Firefox, \
+                    and is available under the MPL2.
+
+homepage            https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey
+# build from GNOME releng tarball
+master_sites        https://ftp.gnome.org/pub/GNOME/teams/releng/tarballs-needing-help/mozjs/
+
+# For Rust
+supported_archs     x86_64 arm64
+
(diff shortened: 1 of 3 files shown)


evanmiller:
CI on macOS 11 says

```
checking that libclang is new enough... 
 ERROR: The libclang located at None is too old (need at least 4.0).
  
 Please make sure to update it or point to a newer libclang using
 --with-libclang-path.
```

Will dig into it later.

evanmiller:
> Will dig into it later.

The Rust (or bindgen) components need an LLVM. I've set the Portfile up to require a MacPorts Clang, and then point mozjs68 to the corresponding `llvm-config` via an environment variable. It's not clear to me if there will be an issue if this LLVM version does not match the LLVM of the installed Rust. Reading the source comments, it looks like Rust is actually not used at all in the JS interpreter; so if there is some kind of issue here, I think it will only affect the exposed Rust bindings, which at the moment is a fairly niche use case.

mascguy (approved):

evanmiller:
Updated to support Snow Leopard (x86_64 only).

kencu:
I installed this, and it works. I didn't run the full test suite though -- should have I guess.

It supports javascript async/await structures nicely (as do 52 and 60) and I tried to use it to all full async/await support back into TenFourFox, by symlinking this js into the TenFourFox bundle instead of the js45-era js that is in there now, but it wouldn't work to allow TFF to support GitHub normally, for example. Maybe someday!

I put that question here, BTW:

https://github.com/classilla/tenfourfox/issues/653

evanmiller:
I would guess with the deep connection between the JS runtime and the DOM that Firefox links into libmozjs rather than running the js binary in a separate process. So my uninformed expectation is that at the very least a full rebuild would be necessary (and likely additional glue code).

To add to the "Maybe someday" wish-list, if we get a bootstrapped Rust rolling on 10.6 then perhaps forking a newer version of Firefox will become a possibility, at least on x86_64.

After this is merged I'll see about tricking the build into working on PPC. As I said above I believe that Rust is not actually used for any of the core code – the story changes with mozjs78 though.

mascguy (approved):

evanmiller:
@mascguy Once this is merged, maybe time to move the `gobject_introspection` update out of -devel? Then we can get a gjs update rolling.

mascguy:
> @mascguy Once this is merged, maybe time to move the `gobject_introspection` update out of -devel? Then we can get a gjs update rolling.

Done: https://github.com/macports/macports-ports/commit/3eec87eed585dfa681320de08bf4c71421152e00

PR #13926, opened 2022-02-08
sqlcipher: disable tcl to fix build on old macOS

z7z8th:
Build on macOS 10.12 and older, will install some tcl libs to
/System/Library/Frameworks, so disable tcl in sqlcipher,
and disable sqlcipher-tools on macOS 10.12 and older.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.2 20G314 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/databases/sqlcipher/Portfile
+++ b/databases/sqlcipher/Portfile
@@ -56,6 +56,10 @@ if {${subport} eq ${name}} {
                         --enable-tempstore=yes \
                         AWK=/usr/bin/awk
 
+    if {${os.major} <= 16} {
+        configure.args-prepend  --disable-tcl
+    }
+
     # search in worksrcpath for sqlcipher/sqlite3.h first -- don't pick up an installed one!
     configure.cppflags-prepend  -DSQLITE_DISABLE_INTRINSIC \
                                 -DSQLITE_ENABLE_COLUMN_METADATA \


z7z8th:
Hi @i0ntempest , I disabled tcl for sqlcipher since it's not used by default on new macOS, and disabled sqlcipher-tools on old macOS, please help review, thanks.

i0ntempest:
Are you sure there's no way to change where whose files are installed? It would be better to change the install location rather than outright disabling tcl. Otherwise, lgtm.

z7z8th:
The problem is I don't have an old macOS to test on it. Otherwise there could be some ways.

i0ntempest:
Check here[https://www.macworld.co.uk/how-to/download-old-os-x-3629363/] for how to get old macOS installers, and I believe VMware Fusion is now free for personal use now if you want to make a VM. 
I see you also disabled the whole `sqlcipher-tools` subport for old systems, does it require tcl to be enabled?

z7z8th:
Thanks, will try on old OS. That's because I don't know how to check if `sqlcipher-tools` build ok on old OS, can you tell how to find build log for specific port on the console?

i0ntempest:
For merged commits you may go to commits view of this repo and cleck on the check status (either a check for success or a cross if some failed). `sqlcipher-tools` does installs fine on all systems, so I think you could just use conditionals to add the arg for disabling tcl on old systems.

z7z8th:
Thanks, a change to disable tcl for only sqlcipher is also pushed, and I will try on old macOS.

i0ntempest (approved):

PR #13933, opened 2022-02-09
gdb: update to 11.2, provide patch to avoid lock-ups

jrabinow:
#### Description

Upgrade GDB to version 11.2

Workaround for this bug https://sourceware.org/bugzilla/show_bug.cgi?id=24069 which causes GDB to hang when running a binary.
While the root cause is still unknown and the patch included here is only a workaround, the patch has reportedly a near-100% success rate in preventing lock-ups so it's still an improvement on the status quo.

Let's keep an eye on the bug and revert this commit once one of 2 things
happen:
  - this patch gets merged into master
  - the root cause is debugged and the proper fix is shipped from upstream

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.2 21D49 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/gdb/Portfile
+++ b/devel/gdb/Portfile
@@ -90,6 +90,14 @@ post-destroot {
     }
 }
 
+# These patches should be removed once they get merged into mainstream.
+# Fingers crossed they get merged in for gdb 11.3
+# https://sourceware.org/bugzilla/show_bug.cgi?id=24069
+platform darwin {
+    patchfiles-append \
+        patch-darwin-nat.c.diff
+}
+
 set pythons_suffixes {27 35 36 37 38 39 310}
 
 set pythons_ports {}
--- a/devel/gdb/Portfile
+++ b/devel/gdb/Portfile
@@ -5,7 +5,7 @@ PortGroup       muniversal 1.0
 PortGroup       compiler_blacklist_versions 1.0
 
 name            gdb
-version         11.1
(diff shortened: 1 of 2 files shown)


devel/gdb/Portfile
```diff
@@ -90,6 +90,14 @@ post-destroot {
     }
 }
 
+# These patches should be removed once they get merged into mainstream.
+# Fingers crossed they get merged in for gdb 11.3
+# https://sourceware.org/bugzilla/show_bug.cgi?id=24069
+if {${os.platform} eq "darwin" } {
```
ryandesign, line 96: Note that this can be written more succinctly:
```suggestion
platform darwin {
```

jrabinow:
```
$ port lint --nitpick devel/gdb
--->  Verifying Portfile for gdb
Error: Failed to lint gdb: can't read "platform_arch": no such variable
Error: Processing of port gdb failed
[Exit 1]
```

I'm afraid I'm not sure why this syntax isn't working, if you have any pointers I could look into, I'd be happy to change things to match your suggestion. Either way, thanks for the protip!

jrabinow: Seems like I had a umask problem. Now that that's fixed, I was able to use the syntax you suggested

PR #13936, opened 2022-02-09
google-cloud-sdk: bump python dependency version from 3.9 to 3.10

jrabinow:
#### Description

google-cloud-sdk: bump python from 3.9 to 3.10

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.2 21D49 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/devel/google-cloud-sdk/Portfile
+++ b/devel/google-cloud-sdk/Portfile
@@ -5,7 +5,7 @@ PortGroup           python 1.0
 
 name                google-cloud-sdk
 version             372.0.0
-revision            0
+revision            1
 categories          devel python
 license             Apache-2
 maintainers         {breun.nl:nils @breun} openmaintainer
@@ -41,7 +41,7 @@ master_sites        https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/
 
 worksrcdir          ${name}
 
-python.default_version 39
+python.default_version 310
 
 post-patch {
     reinplace "s|\"disable_updater\": false|\"disable_updater\": true|" ${worksrcpath}/lib/googlecloudsdk/core/config.json


jrabinow:
My bad, I hadn't seen that. Thank you for flagging this issue. It looks like this package already has python3.9 declared as dependency so we're already in incompatible territory today.

As far as I can tell, the `gcloud` command completely ignores the dependency version in the portfile and invokes `/opt/local/bin/python3` directly, meaning that we could put python3.12 in the portfile and the `gcloud` command would still work as long as there's a python binary (or a symlink to the python binary) at the correct location on the filesystem.

Except that we have a bug [here](https://issuetracker.google.com/issues/205005959) which I ran into yesterday. I had switched the python version back to 3.9 in the port file and noticed the bug was still there, so I thought it didn't make a difference. It turns out that sticking with python 3.9 DOES avoid the bug, but the version is not set in the portfile definition, it's set using `port select --set python3 python39`

Basically, the python version in the portfile doesn't make a difference as far as I can tell, but I still think the best course of action here is to wait until Google fixes the bug before bumping python versions. We can close this PR in the meantime. What do you think?

ryandesign:
> this package already has python3.9 declared as dependency so we're already in incompatible territory today

According to https://issuetracker.google.com/issues/170125513 it's working fine with python 3.9 as of version 319.0.0.

> `gcloud` command completely ignores the dependency version in the portfile and invokes `/opt/local/bin/python3` directly

The port must be fixed so that it runs the python that's declared as a dependency in the portfile.

> we have a bug [here](https://issuetracker.google.com/issues/205005959)

That does seem to say that it's incompatible with python 3.10. As such, this PR should be closed until that's resolved.

jrabinow:
@ryandesign:

> The port must be fixed so that it runs the python that's declared as a dependency in the portfile.

That may be an interesting task, it looks like the `gcloud` command is a shellscript wrapper that attempts to find all python versions installed on the system and uses the one it likes best (not quite true but close enough) to launch the main python codebase. What you're suggesting may be doable with a patch, and I understand the want for actually using the declared dependencies and nothing else, but I also suspect this is going to turn into an uphill battle against Google, potentially at every release.

> > we have a bug [here](https://issuetracker.google.com/issues/205005959)
> 
> That does seem to say that it's incompatible with python 3.10. As such, this PR should be closed until that's resolved.

Will do, @breun I will let you close this to make sure you get the chance to share your thoughts on the subject.

PR #14002, opened 2022-02-15
chromedriver: Update to version 98.0.4758.80, add arm64, undetected variant

essandess:
* Update to version 98.0.4758.80
* Add arm64 support
* Add "undetected" variant

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.2.1 21D62 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/python/py-undetected-chromedriver/Portfile
+++ b/python/py-undetected-chromedriver/Portfile
@@ -1,0 +1,88 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           python 1.0
+
+name                py-undetected-chromedriver
+version             3.1.3
+categories-append   python www
+maintainers         nomaintainer
+license             GPL-3
+supported_archs     noarch
+
+description         Custom Selenium Chromedriver
+long_description    {*}${description} | Zero-Config | \
+                    Passes ALL bot mitigation systems \
+                    (like Distil / Imperva/ Datadadome / CloudFlare IUAM)
+
+homepage            https://github.com/ultrafunkamsterdam/undetected-chromedriver
+
+checksums           rmd160  487c1f61efb09340b7db6bd2cd201a0a6c01ba26 \
+                    sha256  d8884a9d2a65fbe272c184d4c8f92fc6c8e686aa49bf441dd63bf92ff58007e8 \
+                    size    39157
+
+python.versions     37 38 39 310
+
+if {${name} ne ${subport}} {
+    depends_build-append \
+                    port:python${python.version} \
+                    port:py${python.version}-requests \
+                    port:py${python.version}-setuptools \
+                    port:py${python.version}-websockets
+
+    depends_lib-append \
+                    port:py${python.version}-selenium
+
+    depends_test-append \
+                    port:py${python.version}-pytest
+
+    post-patch {
+        fs-traverse f ${worksrcpath}/undetected_chromedriver {
+            if {[file isfile ${f}] && [string match "*.py" ${f}]} {
+                reinplace -E -q \
+                    "1s|^(#!\[\[:space:]]*)/usr/bin/env\[\[:space:]]+python3|\\1${python.bin}|" \
+                    ${f}
+            }
+        }
+    }
+
+    test.run        yes
+    test.env-append PYTHONPATH=${worksrcpath}/build/lib
+    test.cmd        py.test-${python.branch}
+    test.target     tests/v2/test_uc.py
+}
+
+variant port_chromedriver_binary \
+    description {Patch the chromedriver port binary.} {
+    if {${name} ne ${subport}} {
+        # diff -Naurd ./uc-orig/undetected_chromedriver/__init__.py ./uc-new/undetected_chromedriver/__init__.py | sed -E -e 's/\.\/uc-(orig|new)\//.\//g' | sed -E -e 's|/opt/local|@PREFIX@|g' > ~/Downloads/patch-undetected_chromedriver-__init__-py.diff
+        patchfiles-append \
+                    patch-undetected_chromedriver-__init__-py.diff \
+                    patch-undetected_chromedriver-patcher-py.diff
+        post-patch {
+            reinplace "s|@PREFIX@|${prefix}|g" \
+                    ${worksrcpath}/undetected_chromedriver/__init__.py
+            reinplace "s|@NAME@|chromedriver|g" \
+                    ${worksrcpath}/undetected_chromedriver/__init__.py
+        }
+    }
+
+    notes-append \
+        "This package is intended to be used with the port
+
+    chromedriver +undetected
+
+and the Python code:
+
+    import undetected_chromedriver as uc
+    driver = uc.Chrome()
+
+To use separate chromedriver binaries with distinct signatures, use:
+
+    driver = uc.Chrome(executable_path='${prefix}/bin/chromedriver-original', executable_path_mkdtemp=True)
+    driver.delete_executable_path_mkdtemp()
+"
+}
+
+default_variants-append \
+                    +port_chromedriver_binary
--- a/www/chromedriver/Portfile
+++ b/www/chromedriver/Portfile
@@ -3,12 +3,12 @@
 PortSystem          1.0
 
 name                chromedriver
-version             97.0.4692.71
+version             98.0.4758.80
 categories          www
 platforms           darwin
 maintainers         {@catap korins.ky:kirill} openmaintainer
 license             Apache-2
-supported_archs     x86_64
+supported_archs     arm64 x86_64
 
 description         ChromeDriver - WebDriver for Chrome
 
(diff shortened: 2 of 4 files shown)


catap:
@essandess maybe keep old version of chromedriver as subport?

essandess:
> @essandess maybe keep old version of chromedriver as subport?

If someone posts an explicit ask or issue, we should include it. But this Portfile is getting cluttered with quite old versions as subports that are never removed. Would you really like the previous version kept around?

catap:
@essandess well, at least you have removed subport `chromedriver-97` whcih may make migration for someone who had installed it quite painful.

I agree to start remove with MacPorts policy some old and unused subports.

catap:
@essandess and one more point: I suggest to add dependency as the first commit.

www/chromedriver/Portfile
```diff
@@ -3,12 +3,12 @@
...
 
 name                chromedriver
-version             97.0.4692.71
+version             98.0.4758.80
 categories          www
 platforms           darwin
 maintainers         {@catap korins.ky:kirill} openmaintainer
 license             Apache-2
-supported_archs     x86_64
+supported_archs     arm64 x86_64
```
catap, line 11: I guess that `arm64` should be also restricted since version 98?

essandess:
> Your new variant should be restricted for 96+ version / subport only.
> 
> See: https://github.com/ultrafunkamsterdam/undetected-chromedriver/blob/77509e6da2aa14ae8f6f5ff49f8af6dba27e05e9/undetected_chromedriver/devtool.py#L180

To the best of my knowledge, the patch code will work with any of the `chromedriver` binary versions:

https://github.com/ultrafunkamsterdam/undetected-chromedriver/blob/77509e6da2aa14ae8f6f5ff49f8af6dba27e05e9/undetected_chromedriver/patcher.py#L216-L233

The version above is simply a default that I believe was the latest available at the time.

Therefore, I believe the `undetected` variant is relevant to all subports.

catap:
@essandess sorry, I've missread the code. I've read it as `min` instead of `main` version :)

www/chromedriver/Portfile
essandess: No, Google has provided M1 code for all available versions. Please see https://sites.google.com/chromium.org/driver/downloads.

essandess:
> @essandess and one more point: I suggest to add dependency as the first commit.

Erg, that's beyond my git abilities, sorry.

catap:
> > @essandess and one more point: I suggest to add dependency as the first commit.
> 
> Erg, that's beyond my git abilities, sorry.

`git rebase -i origin/master` and re-order lines inside your text editor :)

catap:
BTW google hasn't provide `arm64` driver for the last subport: https://chromedriver.storage.googleapis.com/index.html?path=79.0.3945.36/

Also, I see that they release 99.0.4844.17 — any reason to not update to the last version?

essandess:
> `git rebase -i origin/master` and re-order lines inside your text editor :)

Done, thanks!

essandess:
> BTW google hasn't provide `arm64` driver for the last subport: https://chromedriver.storage.googleapis.com/index.html?path=79.0.3945.36/
> 
> Also, I see that they release 99.0.4844.17 — any reason to not update to the last version?

@catap There are some larger issues here:
* Google has deprecated the previous distribution site I've updated to https://sites.google.com/chromium.org/driver/downloads.
* **All** of the previous `checksum`'s for the new site are now incorrect. This means updating everything—for both `x86_64` **and** `arm64`—or just obsoleting these previous versions. I propose and have pushed an implementation of the later approach. If previous versions are necessary or requested, we can add as needed.
* Version 98 is the latest "stable" release, and works with the latest release of Chrome, so that's the one we want.

essandess:
> 1. all of old ports are cached inside MacPorts distfile => you may see that your commit is passed CI and it has access to all subports ;)

This doesn’t address the M1 builds. And unnecessarily complicates the distribution site logic. Google moved the page and we need to move with it.

> 2. I suggest to add something like:
> 
> ```
> if {[vercmp $version 98.0.4758.80] >= 0} {
>   supported_archs arm64 x86_64
> } else {
>   supported_archs x86_64
> }
> ```

But there are M1 builds for nearly everything. This isn’t necessary.

The bottom line question is why keep all those previous version subports around, unless there’s demand for a specific one. Especially if there’s no maintainer cycles to fix everything. I know I can’t support these.

I believe we should just obsolete all previous versions, as effectively happens with version updates.

catap:
@essandess are you sure that replacing this port with tensorflow's metadata is the right idea?

essandess:
Groan maybe one shouldn’t issue git commands on a mobile device. I’ll clean up my mess later, sorry.

essandess:
@catap Any more comments on this PR? I’ve tested on several machines and believe it’s ready to commit.

essandess:
> @essandess because you're on it I'd like to suggest to obsolete all subports and do not add a subport for new chrome driver.
> 
> Let make in one year simple portfile without any subports.
> 
> I'm thinking about it for some time and honestly haven't see any use of old chrome driver.

I've removed the `latest_major_version` port code so we don't keep adding old versions.

essandess:
> <s>believe it’s ready to commit.</s>

We should think about this comment for a couple days and how it might be implemented on MacPorts.

My first impulse is to create the option for a separate patched binary in `/tmp` or `tempfile.mkdtmp`.

* https://github.com/ultrafunkamsterdam/undetected-chromedriver/issues/496#issuecomment-1046023309
* https://github.com/ultrafunkamsterdam/undetected-chromedriver/issues/479

catap:
@essandess maybe split your PR into two? The first one with upgrade version and make subports obsolete. The second one adds undetected variant.

www/chromedriver/Portfile
```diff
@@ -32,129 +56,159 @@ use_configure       no
 build {}
 
 destroot {
-    xinstall -m 755 \
-        ${worksrcpath}/chromedriver \
+    copy \
```
catap, line 59: Any reason to not use `xinstall`?

www/chromedriver/Portfile
```diff
@@ -32,129 +56,159 @@ use_configure       no
...
+    depends_lib-append \
+                    port:python${python_version} \
+                    port:py${python_version}-undetected-chromedriver
+
+    require_active_variants \
+                    py${python_version}-undetected-chromedriver \
+                    port_chromedriver_binary
+
+    post-patch {
+        copy \
```
catap, line 84: Same here

python/py-undetected-chromedriver/Portfile
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           python 1.0
+
+name                py-undetected-chromedriver
+version             3.1.3
+categories-append   python www
+maintainers         nomaintainer
```
catap, line 9: Two comments:
1. I feel that commit message should be: `py-undetected-chromedriver: new port`
2. Maybe you would like to take care of this port? You may put openmaintainer policy.

www/chromedriver/Portfile
essandess: Doesn’t work for symlinks.

www/chromedriver/Portfile
```diff
@@ -32,129 +56,159 @@ use_configure       no
...
+            ${worksrcpath}/${subport}-undetected
+        ln -s \
+            ${prefix}/bin/${subport}-undetected \
+            ${worksrcpath}/${name}
+    }
 
-subport ${name}-${latest_major_version} {
-    conflicts       ${name}
     post-destroot {
+        copy \
```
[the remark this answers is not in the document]
essandess: Need symlinks.

www/chromedriver/Portfile
```diff
@@ -32,129 +56,159 @@ use_configure       no
...
+            ${worksrcpath}/${subport}-${major_version} \
+            ${destroot}${prefix}/bin
         ln -s \
-            ${prefix}/bin/${name}-${latest_major_version} \
-            ${destroot}${prefix}/bin/chromedriver
+            ${prefix}/bin/${subport}-${major_version} \
+            ${destroot}${prefix}/bin/${subport}-original
+        copy \
```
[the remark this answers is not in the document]
essandess: Need symlinks.

python/py-undetected-chromedriver/Portfile
essandess: Submission is standard practice.

I will update the port when I’m able but do not have the cycles to commit to being a maintainer.

www/chromedriver/Portfile
catap: maybe when `file move`? It is much faster and should be much safer for symlinks.

catap (approved):

essandess:
I’ve tested the latest push and too the best of my knowledge this PR is ready to commit.

www/chromedriver/Portfile
essandess: Not sure what you mean. This just copies a symlink to `destroot`, which is what we need here.

catap:
@essandess I mean to use `move` instead of `copy` which is 1) faster and 2) safer for symlinks :)

www/chromedriver/Portfile
essandess: Also fwiw it copies the actual file without the `undetected` variant. This code works for both.

essandess:
> @essandess I mean to use `move` instead of `copy` which is 1) faster and 2) safer for symlinks :)

What is the safety issue? Portfile `copy` just copies the symlink, which is what we want.

catap:
> What is the safety issue? Portfile `copy` just copies the symlink, which is what we want.

Move is atomic operation when it is happened on the same filesystem. Otherwise recursive copy is more tricky one.

Anyway, at this case it doesn't make any differences => feel free to ignore this remark.

catap:
@essandess any news here?

essandess:
> @essandess any news here?

As [above](https://github.com/macports/macports-ports/pull/14002#issuecomment-1048939255):
> I’ve tested the latest push and too the best of my knowledge this PR is ready to commit.

catap:
@mascguy may I ask you to merge it? :)

mascguy (approved):

PR #14035, opened 2022-02-17
glib2: patch universal only

michaelld:
#### Description

glib2: move part of universal.patch causing issues on ARM64 to non-ARM patch
gtk3: rev-bump for glib2 patch change

Ref (among others): https://trac.macports.org/ticket/64377

NOTE: this doesn't fix the underlying issue, but since most folks do not install as +universal this will help most users.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

Intel:
macOS 11.6.4 20G507
Xcode 12.3 12C33

ARM64:
macOS 11.6.4 20G507
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/devel/glib2/Portfile
+++ b/devel/glib2/Portfile
@@ -12,7 +12,7 @@ name                        glib2
 conflicts                   glib2-devel
 set my_name                 glib
 version                     2.64.6
-revision                    1
+revision                    2
 checksums                   rmd160  27938fd75bdcd4ca0afc8ebce6b137ed3c1fe6a5 \
                             sha256  c36ee07a70164c71f046016fe6aaacd6368333c42590bc0cba47c344ffb853f1 \
                             size    4781576
--- a/devel/glib2/files/universal.patch
+++ b/devel/glib2/files/universal.patch
@@ -114,7 +114,7 @@ here and in config.h.ed.
  @glib_os@
  
 -@glib_vacopy@
-+#ifdef __LP64__
++#if defined(__ppc64__) || defined(__x86_64__)
 +#define G_VA_COPY_AS_ARRAY 1
 +#endif
 +
--- a/gnome/gtk3/Portfile
+++ b/gnome/gtk3/Portfile
@@ -11,7 +11,7 @@ name                gtk3
 conflicts           gtk3-devel
 set my_name         gtk3
 version             3.24.31
-revision            0
+revision            1
 epoch               1
 
 set proj_name       gtk+


michaelld:
This is a convenience PR for the port owners, as this is a critical issue on ARM64 and since the ports are openmaintainer, and the tweaks are minimal, I could just push these directly ... but it would be wiser to give the port owners their due first.

ryandesign:
I think use of the universal variant on arm64 systems is not as uncommon as you think. I think there are still quite a few ports that exclude arm64 from their supported_archs thus requiring their dependencies, which might include glib2, to be built universal.

I would prefer to continue applying universal.patch always, but fix it so that it works for arm64. Would this change do it?

```diff
--- a/devel/glib2/files/universal.patch
+++ b/devel/glib2/files/universal.patch
@@ -114,7 +114,7 @@ here and in config.h.ed.
  @glib_os@

 -@glib_vacopy@
-+#ifdef __LP64__
++#if defined(__ppc64__) || defined(__x86_64__)
 +#define G_VA_COPY_AS_ARRAY 1
 +#endif
 +
```

About increasing the gtk3 revision to rebuild it: is that really necessary? If so, it raises the question of how many other ports that depend on glib2 also need their revisions increased.

michaelld:
In my testing, removing the `VA_COPY` lines did the trick, resolving the `glib2`/`gtk3` issue on ARM64. And, yes, it required rebuilding `gtk3` for the ports I was interested in (GNU Radio related).

That said, if one _isn't_ building as `+universal` I don't see the point in applying the patch. Without the `universal.patch`, the resulting file looks to very accurately represent the correct state of system parameters for the native build.

If you want to do a "both and" I'm good with that: remove the `VA_COPY` portion of the patch as well as only apply the patch when doing a universal build. I think this would be the ideal solution.

ryandesign:
> Without the `universal.patch`, the resulting file looks to very accurately represent the correct state of system parameters for the native build.

Naturally. However, I want to continue to apply the patch unconditionally. This has several desirable outcomes:

1. Someone who is updating the port to a newer version but who is not using the universal variant cannot fail to notice when the patch needs to be updated.
2. It is impossible for there to be unintended differences between the universal and non-universal builds (with regard to what is being patched).

michaelld:
@ryandesign I don't agree with you here, but in the spirit of compromise and getting things working more expediently I've pushed fixes that just remove the section of the patch that seems to be causing the issue

ryandesign:
I don't think we can just remove that part of the patch. `G_VA_COPY_AS_ARRAY` needs to be defined to `1` on some architectures, at least that was the case when the patch was designed. I have not re-verified that since the switch to the meson build system.

michaelld:
@ryandesign OK so we have an issue to get resolved with `glib2` and the `universal.patch` patch. The `G_VA_COPY_AS_ARRAY` section is causing issues on ARM64/M1 with `glib2`/`gtk3`/`py-gobject3`, and needs to be removed or fixed. I have read and understand your points on why to include the `universal.patch` regardless of whether building universal or not, and we are agreeing to disagree on this, but as this is your port and you wish to keep applying this patch for all builds that's your prerogative. I strongly doubt I can track down whether the COPY is fixed or not on all platforms; in my testing thus far it looks valid on macOS 11 and 12, but that leaves out all of OSX of course. So ... what next?

kencu:
Would an arm or intel arch test added into the patch help to apply it only when needed?

michaelld:
What @kencu notes might work, since I don't think the VA_COPY stuff is needed for ARM64/M1 at all. I think it's just older Intel OSX that need it. If this is an option @ryandesign then I'll split the current patch into 2 parts, and the VA_COPY would be applied only to Intel. That would address the current ARM64/M1 issue and should allow the older OSX with whatever issue(s) they have with the VA_COPY stuff to still work. What do you think?

michaelld:
@ryandesign I just pushed the changes @kencu  proposed: splitting the part of universal.patch that's causing the issue on ARM64 into its own patch that is applied for non-ARM builds. Still need to test / verify this on my ARM Mac. If this change does indeed work, I hope you would consider it. I don't know of a better way to go this with the current glib2 version. Updating to a newer version might help, but that has its own issues.

michaelld:
Confirmed that this latest PR changeset works on my ARM64 Mac. After updating `glib2` and then `gtk3`, `gnuradio-companion` runs successfully. There might be other issues with actual runtime for GRC, but any such will be in a different PR.

mascguy (approved):
LGTM. @ryandesign and @kencu ?

kencu:
I know Ryan was not too keen on a conditional patch.

I thought there was a compiler define for ARM that we could test for, in a use-it-always patch eg:

```
# defined (ARMTHINGY)
do the stuff
# else
don't
# endif
```

(had to add a space after the # as github goes wild if you don't)

mascguy:
@ryandesign/@michaelld/@evanmiller, given the changes to `glib2-devel` via [PR 14300 - glib2-devel: update to 2.66.8; improve stability for +quartz on ARM](https://github.com/macports/macports-ports/pull/14300), should this PR be updated with those changes? Or...?

mascguy:
@ryandesign / @michaelld / @evanmiller, given the changes to `glib2-devel` via [PR 14300 - glib2-devel: update to 2.66.8; improve stability for +quartz on ARM](https://github.com/macports/macports-ports/pull/14300), should this PR be updated with those changes? Or...?

michaelld:
@mascguy thanks for the info! let me try glib2-devel on my ARM64 to see if it works as desired in this regard. If so, then yes I think backporting that patch change should make folks here happy. It's what ken is noting, and, really, is the only version left that I haven't tried!

michaelld:
Yes that patch does the trick. I will update this PR to do that for the non-devel ports ... sometime soonish ...

michaelld:
Done! This set of changes (glib2 -> gtk3) does the trick for me on ARM64. I think it will fix a bunch of PyXY GTK3 issues without having to rebuild any of those PyXY ports :)

mascguy (approved):

PR #14046, opened 2022-02-18
bind9: fix build on older systems

jmroot:
#### Description

Make sure a compiler with atomic builtins is used, and fix issues with detecting and using them.

###### Type(s)
- [x] bugfix

###### Tested on
macOS 10.7, 10.9, 10.13
Xcode 4.6.3, 6.2, 10.1


danielluke:
LGTM - thanks.

PR #14071, opened 2022-02-21
MoltenVK: Bump min requirement

Gcenx:
Buildbot was unable to mount the dmg on 10.11 to bump back to 10.12.

###### Type(s)

- [x] bugfix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 

### DIFF
--- a/graphics/MoltenVK/Portfile
+++ b/graphics/MoltenVK/Portfile
@@ -3,7 +3,8 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        KhronosGroup MoltenVK 1.1.17 v
+github.setup        KhronosGroup MoltenVK 1.1.7 v
+epoch               1
 
 set sdkversion      1.3.204.0
 categories          graphics
@@ -59,13 +60,13 @@ destroot {
 }
 
 platform darwin {
-    if {${os.major} < 15} {
+    if {${os.major} < 16} {
         archive_sites
         distfiles
         depends_build
         known_fail  yes
         pre-fetch {
-            ui_error "${subport} @${version} requires OS X El Capitan or later"
+            ui_error "${subport} @${version} requires macOS Sierra or later"
             return -code error "incompatible OS X version"
         }
     }


chrstphrchvz:
Is this in response to the following error?

```
hdiutil: attach failed - no mountable file systems
hdiutil: detach failed - No such file or directory
```

I encountered the same error locally on 10.15 Catalina; I haven’t yet opened a ticket for this.

Gcenx:
@chrstphrchvz yes that’s the error I’d seen on the buildbot for 10.11 so assumed the dmg was in fine newer format that didn’t work on that version.

However it’s strange you’ve seen this on macOS Catalina when there prebuilt packages for 10.12 and later see http://packages.macports.org/MoltenVK/

chrstphrchvz:
I had tried `sudo port -s built MoltenVK` because I was curious what else was in installer.dat which the package from the builder isn’t installing. I haven’t found what might be causing this, but I wonder if it might be an issue in MacPorts’ `use_dmg yes` handling/how MacPorts is running `hdiutil`. I don’t think it has anything to do with a corrupt image (as some online reports of similar errors suggest) since the checksum phase is fine, and I can manually `hdiutil attach` the dmg without errors.

Gcenx:
Agreed this sounds more like an issue with how Macports is handling dmg files.

In going to close this PR since ether issue now seems more to be an issue with Macports dmg file handling.

Gcenx:
> I had tried sudo port -s build MoltenVK because I was curious what else was in installer.dat which the package from the builder isn’t installing.

@chrstphrchvz it contains all the usual files that the VulkanSDK usually installs. It’s possible to prefer a manual install of the VulkanSDK using this method but the Vulkan-loader would need to be compiled from source so it knows of Macports prefix.

As LunarG is so dam slow pushing SDK releases I’d opted to instead fork MoltenVK repository and directly attach the archives built via GitHub actions. But as that’s not really the Macports way I’d opted to share the VulkanSDK unpacking method.

Gcenx:
@pmetzger a response to your tag, I’d already opened this PR however according to the findings from @chrstphrchvz the issue isn’t the dmg itself but more lightly with how Macports handles mounting/un mounting dmg files.

chrstphrchvz:
Another thought: I haven’t ruled out that this could instead be due to flakiness in `hdiutil` itself.

Gcenx:
The VulkanSDK packages are now using APFS so is method won’t work on on. OS X 10.11

Gcenx:
@ryandesign are you looking into an alternative or just didn’t get prompted to this PR?

pmetzger:
Is there a proposed solution to this? Maybe we just accept that older versions can't get the newest code?

Gcenx:
> Is there a proposed solution to this? Maybe we just accept that older versions can't get the newest code?
@pmetzger the only possible _solutions_ would be as follows:
- convince LunarG to change there packaging
- add an APFS drive to 10.11 buildbot
- repackage a package.macports build
- have a buildbot build MoltenVK and uploaded the custom archive

Not sure if any Macports member would be ok with any of the solutions outside of convincing LunarG to change how they’ve packaging the VulkanSDK

ryandesign:
> * convince LunarG to change there packaging

Sounds like a good idea. Has anybody asked them to do this? What is the URL of that request?

> * add an APFS drive to 10.11 buildbot

Not possible. APFS was introduced in macOS 10.12.

> * have a buildbot build MoltenVK and uploaded the custom archive

Isn't that what I had the MolkenVK port doing before you changed it in a03f281b3f8740477d426cd462ca1d9a701efed6?

Gcenx:
> Sounds like a good idea. Has anybody asked them to do this? What is the URL of that request?

https://vulkan.lunarg.com/issue/home?limit=10;q=;mine=false;org=false;khronos=false;lunarg=false;indie=false;status=new,open

> Not possible. APFS was introduced in macOS 10.12.

There was a free third-party driver but the vendor had since removed it.

> Isn't that what I had the MolkenVK port doing before you changed it in [a03f281](https://github.com/macports/macports-ports/commit/a03f281b3f8740477d426cd462ca1d9a701efed6)?

No I’m referring to having a builtbot that makes the VulkanSDK that can then be repackaged for the supported macOS versions.

The other easier alternative is repackaging a 1.1.9 archive from https://packages.macports.org/MoltenVK/

As it would be using something built by Macports builtbot it would be a little more acceptable than using my mirror that provides archives.

Gcenx:
I never asked them to change the archive format, I had asked upstream to provide there GitHub actions artifacts but that was denied.

PR #13922, opened 2022-02-08
vim: bump patchlevel, add python310 variant

jrabinow:
#### Description

This PR:
 - updates vim to patch level 4324: needed as current vim patchlevel (2683) [has a bug when compiling with python3.10]( https://github.com/vim/vim/issues/6183)
 - refactors the python variants code in the portfile
 - adds a python3.10 variant

There is an issue with the `vim` port:
 - `sudo port install -vs ./editors/vim` [works](https://github.com/macports/macports-ports/files/8019818/success_vim8.2.4324_build_log.txt), `sudo port install -vst ./editors/vim` does not, both for the [existing patchlevel](https://github.com/macports/macports-ports/files/8019801/failure_vim8.2.2683_build_log.txt) and for the [new patchlevel](https://github.com/macports/macports-ports/files/8019811/failure_vim8.2.4324_build_log.txt) introduced in this PR. I have tried finding out is going on, I'm afraid debugging trace mode is beyond me ([failure_vim8.2.4324_main.log](https://github.com/macports/macports-ports/files/8019815/failure_vim8.2.4324_main.log) and [failure_vim8.2.4324_manual_configure_cmd_run.txt](https://github.com/macports/macports-ports/files/8019817/failure_vim8.2.4324_manual_configure_cmd_run.txt))
 - `port variants vim` lists python variants in alphabetical order rather than numeric order. I'm not sure if anything can be done about that.
 - tests fail both for patchlevels both current (2683) and updated (4324)

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.2 21D49 x86_64
Xcode 13.1 13A1030d

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] checked your Portfile with `port lint --nitpick`?
- [ ] tried existing tests with `sudo port test`? FAIL
- [ ] tried a full install with `sudo port -vst install`? FAIL
- [x] tested basic functionality of all binary files?

PR #14089, opened 2022-02-22
vim, MacVim: add comment to keep patchlevels in sync

jrabinow:
#### Description

vim, MacVim: add comment to keep patchlevels in sync

Following my recent PRs #13922 and #14084 to bump vim patchlevels, I was made aware that [vim and MacVim patchlevels should be kept in sync](https://github.com/macports/macports-ports/pull/12530#issuecomment-939512070). Since #13922 was already merged into master, I wanted to update MacVim to the same patchlevel, but it looks like latest MacVim is based on vim 8.2.3455 as of today, so syncing patchlevels isn't possible unless we downgrade vim.

Even if we can't fix things today, the comments that this PR adds will hopefully help prevent this issue from reoccurring.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/editors/MacVim/Portfile
+++ b/editors/MacVim/Portfile
@@ -4,6 +4,9 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
+# The vim port should always follow the patch level used in the MacVim port. This
+# is because they both share the same set of configuration files and ensures
+# that the user's configuration files will always work with both.
 set vim_version     8.2
 set snapshot        171
 github.setup        macvim-dev macvim ${snapshot} snapshot-
--- a/editors/vim/Portfile
+++ b/editors/vim/Portfile
@@ -3,6 +3,9 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
+# The vim port should always follow the patch level used in the MacVim port. This
+# is because they both share the same set of configuration files and ensures
+# that the user's configuration files will always work with both.
 set vim_version     8.2
 set vim_patchlevel  4324
 github.setup        vim vim ${vim_version}.${vim_patchlevel} v


editors/vim/Portfile
```diff
@@ -3,6 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
+# The vim port should always follow the patch level used in the MacVim port.This
+# is because they both share the same set of configuration files and ensures
+# that the user's configuration files will always work with both.
+# https://github.com/macvim-dev/macvim/releases
```
raimue, line 9: I think we don't need the URL here, because whoever wants to update is supposed to look at the MacVim port.

editors/MacVim/Portfile
```diff
@@ -4,6 +4,9 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
+# The vim port should always follow the patch level used in the MacVim port.This
```
raimue, line 7: MIssing space in first line: `port. This`

editors/vim/Portfile
```diff
@@ -3,6 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
+# The vim port should always follow the patch level used in the MacVim port.This
```
raimue, line 6: MIssing space in first line: `port. This`

raimue (commented):
Thank you for providing this! Totally makes sense to spell it out in the Portfile to make it clear.

I am sorry that at the moment I cannot be as active with regard to maintaining the vim ports as I used to be. Your port updates and patches are appreciated! Keep up the good work.

jrabinow:
No worries on the review time, thanks for being a maintainer in the first place! :-)

I've updated the changeset to account for your comments.

PR #14100, opened 2022-02-23
xz-bootstrap: cleanup install path

catap:
#### Description

As part of https://github.com/macports/macports-ports/pull/13995 I've removed `libexec/libcxx-bootstrap`.

Here simple move to where `xz-bootstrap` is installed.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y
Xcode x.y

###### Verification 
Have you


jmroot:
Why would this be desirable? Doesn't just removing dependencies mean that things will be used opportunistically?

catap:
@jmroot all this bootstrap ports are required to build cmake and clang on very old system which allows to build libcxx.

All this dependencies are optional, and it can be build and use without when on clean system.

The goal for `-bootstrap` build is built something with as little dependencies as possible.

jmroot:
Users don't generally have clean systems. In any case that only explains removing some of the dependencies, not all the other changes.

catap:
@jmroot here a few changes.

1. Moving `xz-bootstrap` and `python27-bootstrap` from `libexec/libcxx-bootsrap`.

Why? To keep things clean. Right now `cmake-bootsrap` is installed to `libexec/cmake-bootsrap`; brand new `gcc10-bootsrap` is installed to `libexec/gcc10-bootstrap` and the same happens with `clang-11-boostrap` which hasn't been yet merged.

2. Remove dependencies from `python27-bootstrap`.

`python27-bootstrap` has a few dependencies and one of them is openssl. Python27 is old enough and can be built with old openssl from macOS 10.6 and early.

I feel that this dependency was added as a sort of side effect when openssl PG was introduced.

Thus, this port is *build* requirements for clang and cmake, this port shouldn't be used in normal life. Only to bootstrap them.

So, here, I simple removed all this unnecessary dependency and if it picks something from installed system, well, it won't make any harm. is it?

3. Mark some ports obsolete.

All this ports was used via `python27-bootstrap` and as soon as I've cleaned up this dependency... well... we don't need them anymore.

Do you see any use case of `gperf-bootstrap` for example?

kencu:
I'd just like add a short note of support for what @catap is doing here, and all the work and attention he is putting into this issue of late. 

Trimming down the dep tree for the bootstrap ports if possible, especially python27, does make the bootstrap setup considerably simpler. Certainly we have to then make sure that somehow the deps aren't just found opportunistically instead, as that would just cause havoc.

Obsoleting some of these bootstrap ports at this point in time is very neat-and-tidy but might be a bit premature -- we could leave them sit for a bit to be sure they aren't needed.

jmroot:
> 1. Moving `xz-bootstrap` and `python27-bootstrap` from `libexec/libcxx-bootsrap`.
> 
> Why? To keep things clean. Right now `cmake-bootsrap` is installed to `libexec/cmake-bootsrap`; brand new `gcc10-bootsrap` is installed to `libexec/gcc10-bootstrap` and the same happens with `clang-11-boostrap` which hasn't been yet merged.

I don't think that's a good enough reason to make everyone reinstall these ports.

> 2. Remove dependencies from `python27-bootstrap`.
> 
> `python27-bootstrap` has a few dependencies and one of them is openssl. Python27 is old enough and can be built with old openssl from macOS 10.6 and early.
> 
> I feel that this dependency was added as a sort of side effect when openssl PG was introduced.
> 
> Thus, this port is _build_ requirements for clang and cmake, this port shouldn't be used in normal life. Only to bootstrap them.
> 
> So, here, I simple removed all this unnecessary dependency and if it picks something from installed system, well, it won't make any harm. is it?

In my experience, undeclared dependencies being picked up causes many problems. So I will only accept this if you can show that the system libs will always be used.

> 3. Mark some ports obsolete.
> 
> All this ports was used via `python27-bootstrap` and as soon as I've cleaned up this dependency... well... we don't need them anymore.

Of course, obsoleting them if they have no dependents makes sense.

devel/gperf/Portfile
```diff
@@ -27,18 +26,14 @@ configure.args  --infodir=${prefix}/share/info
...
-        known_fail          yes
-    }
-}
-# Also needed by later clangs.
-if {${os.platform} eq "darwin" && ${os.major} < 11 && ${cxx_stdlib} eq "libc++"} {
-    clang_dependency.extra_versions 3.7
+    PortGroup   obsolete 1.0
+    revision    [ expr ${revision} + 1 ]
+
+    #------------------------------------------------------------------------------
```
jmroot, line 32: Ascii art to highlight this comment isn't necessary.

devel/ncurses/Portfile
```diff
@@ -1,5 +1,6 @@
-PortSystem 1.0
```
jmroot, line 1: clang_dependency portgroup again

devel/pkgconfig/Portfile
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           clang_dependency 1.0
```
jmroot, line 4: and again

lang/llvm-3.4/Portfile
```diff
@@ -77,15 +75,12 @@ if {${subport} eq "llvm-${llvm_version}"} {
...
+    revision                    [ expr ${revision} + 1 ]
+    configure.cxx_stdlib        libstdc++
     # Have to also use bootstrap versions of deps that use libstdc++ in
     # order to be able to build libc++.
-    depends_lib-replace         port:python27 port:python27-bootstrap \
-                                port:ncurses port:ncurses-bootstrap
-
-    configure.cppflags-prepend  -I${bootstrap_prefix}/include
-    configure.ldflags-prepend   -L${bootstrap_prefix}/lib
-    configure.env-append        PATH=${bootstrap_prefix}/bin:$::env(PATH)
```
jmroot, line 88: I assume you checked that this works right without the PATH change?

lang/python27/Portfile
```diff
@@ -68,33 +68,30 @@ openssl.branch      1.1
...
-    depends_build-replace   port:pkgconfig port:pkgconfig-bootstrap
-    depends_lib-replace     port:ncurses port:ncurses-bootstrap
-    depends_lib-delete      port:db48 \
-                            port:gettext-runtime \
-                            port:libedit \
-                            port:sqlite3
-    # no need to be 'port select'able
+
+    depends_build-delete    port:pkgconfig
+    # OpenSSL PG adds dependency via callback, clena it the same way :)
```
jmroot, line 84: Better to avoid including the portgroup in the first place if it shouldn't be used.

textproc/libiconv/Portfile
```diff
@@ -2,7 +2,6 @@
 
 PortSystem              1.0
 PortGroup               muniversal 1.0
-PortGroup               clang_dependency 1.0
```
jmroot, line 5: clang_dependency again

devel/gperf/Portfile
```diff
@@ -1,7 +1,6 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
-PortSystem 1.0
-PortGroup  clang_dependency 1.0
```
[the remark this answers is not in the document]
kencu: the idea is that it is about to disappear, and be replaced by something new.

system roots -> gcc10-bootstrap -> all deps needed to build clang11-bootstrap -> everything

the order of operations here has to match, however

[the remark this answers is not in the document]
kencu: perhaps the something new will still need to be in a revised clang dependency portgroup, however. I haven’t looked deeply into this part at all yet, as right now the process is still stuck at getting gcc10-bootstrap built right.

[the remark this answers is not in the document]
kencu: right now they do -- but the idea will be to move all of them (all possible anyway) to build with clang11-bootstrap.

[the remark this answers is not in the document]
kencu: and someone could possibly (possibly) dump all the older ones, as none of them would be needed any longer.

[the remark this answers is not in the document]
kencu: That is all to be sorted out once this other stuff (gcc10-bootstrap and clang11-bootstrap work). 

The basic idea is that all the clang deps will build with clang11-bootstrap, and the clang_dependency tangle that makes them all build with ancient clang versions will disappear (as will, hopefully, all the ancient clang versions).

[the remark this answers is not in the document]
kencu: @catap -- I'm not sure where this PR is headed. You might just close it? We have to get the other stuff sorted out and come up with a coherent plan first.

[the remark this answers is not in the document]
jmroot: > The basic idea is that all the clang deps will build with clang11-bootstrap

That is exactly what the portgroup ensures, just with a different clang port. So I maintain that removing it is not appropriate in this PR. (And most likely it should never be removed, just modified.)

catap:
@kencu, @jmroot I'm back. Sorry, I have quite a mess in my real life which required a lot of attention.

I just made the force push to this PR which should address all question.

Let me summarise it here to make things simpler.

1. Removing clang_dependency PG was a mistake. I've tried to delivere everything in one huge PR but @kencu raised a point that it should be split and I've forgot to clean it up here. Should be better now.

2. Dependencies. I've implemented the same logic that I've used inside gcc10-bootstrap port to sterilise everything to not depends on anything from macports. I've also tested it with installing all dependencies from non bootstrap version, build from scratch bootstrap port, remove everything and run rev-dep.

3. Prefix was set correctly and ascii art was also removed.

devel/gperf/Portfile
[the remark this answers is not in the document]
catap: yeah, it was defently too early. Too early.

lang/llvm-3.4/Portfile
catap: @jmroot yes, I've build it it.

lang/python27/Portfile
catap: Good idea, indeed.

devel/gperf/Portfile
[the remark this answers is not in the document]
jmroot: > I'm not sure where this PR is headed. You might just close it? We have to get the other stuff sorted out and come up with a coherent plan first.

If you need to rethink things that's fine. But please don't interpret criticism of the specifics of a proposal as outright rejection of the entire idea.

devel/pkgconfig/Portfile
```diff
@@ -38,21 +38,11 @@ set docdir          ${prefix}/share/doc/${name}
 
 # This port is used by clang-3.4 to bootstrap libcxx
 subport ${name}-bootstrap {
-    set myprefix ${prefix}/libexec/libcxx-bootstrap
```
jmroot, line 41: Similar to xz, all use of `myprefix` should be removed.

lang/llvm-3.4/Portfile
```diff
@@ -77,15 +75,12 @@ if {${subport} eq "llvm-${llvm_version}"} {
 }
 
 if {${os.platform} eq "darwin" && ${os.major} < 11 && ${cxx_stdlib} eq "libc++"} {
-    configure.cxx_stdlib    libstdc++
+    revision                    [ expr ${revision} + 1 ]
```
jmroot, line 78: Expressions should be brace-quoted. Having a different revision depending on the platform is a bit problematic anyway though.

lang/python27/Portfile
```diff
@@ -64,40 +63,65 @@ depends_lib         port:bzip2 \
...
+    openssl.branch  1.1
+}
 
 # This port is used by clang-3.4 to bootstrap libcxx
 subport ${name}-bootstrap {
-    set stdprefix ${prefix}
-    prefix ${prefix}/libexec/libcxx-bootstrap
+    revision                [ expr ${revision} + 1 ]
+
+    use_xz                  no
```
jmroot, line 77: Unfortunately I don't think this works the way you would expect.

lang/python27/Portfile
```diff
@@ -64,40 +63,65 @@ depends_lib         port:bzip2 \
...
+    configure.cxxflags
+    configure.cppflags
+    configure.optflags
+    configure.ldflags
+
+    configure.universal_cflags
+    configure.universal_cxxflags
+    configure.universal_cppflags
+    configure.universal_ldflags
+    configure.universal_args
```
jmroot, line 108: Same comment as for xz.

lang/python27/Portfile
```diff
@@ -64,40 +63,65 @@ depends_lib         port:bzip2 \
...
-        # libstdc++ can't be used on Mojave and later
-        known_fail          yes
-    }
-    configure.args      ac_cv_header_libintl_h=no \
-                        ac_cv_lib_intl_textdomain=no \
-                        ac_cv_search_bind_textdomain_codeset=no
+
+    configure.args          ac_cv_header_libintl_h=no \
+                            ac_cv_lib_intl_textdomain=no \
+                            ac_cv_search_bind_textdomain_codeset=no
```
jmroot, line 124: Is this still necessary and appropriate? Shouldn't using the system gettext be OK?

textproc/libiconv/Portfile
```diff
@@ -39,18 +39,11 @@ set myprefix ${prefix}
 
 # This port is used by clang-3.4 to bootstrap libcxx
 subport ${name}-bootstrap {
-    set myprefix ${prefix}/libexec/libcxx-bootstrap
```
jmroot, line 42: Use of `myprefix` needs to be cleaned up in this Portfile too.

jmroot (changes requested):

devel/gperf/Portfile
```diff
@@ -27,18 +27,11 @@ configure.args  --infodir=${prefix}/share/info
...
-    }
-}
-# Also needed by later clangs.
-if {${os.platform} eq "darwin" && ${os.major} < 11 && ${cxx_stdlib} eq "libc++"} {
-    clang_dependency.extra_versions 3.7
+    PortGroup   obsolete 1.0
+    revision    1
+
+    # Obsolete Date: 2022-01-17
+    # This was used only by libiconv-bootstrap, no replacment here.
```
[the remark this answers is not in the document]
catap: got it

devel/gperf/Portfile
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
-PortSystem 1.0
-PortGroup  clang_dependency 1.0
+PortSystem      1.0
+PortGroup       clang_dependency 1.0
```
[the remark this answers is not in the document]
catap: reverted

archivers/xz/Portfile
```diff
@@ -47,10 +47,41 @@ if {${subport} eq ${name}} {
...
+    configure.distcc    no
+
+    # sterilize PATH
+    configure.env-append    PATH=/usr/bin:/bin:/usr/sbin:/sbin
+    build.env-append        PATH=/usr/bin:/bin:/usr/sbin:/sbin
+
+    # sterilize dependencies
+    depends_build
+    depends_lib
+    depends_run
```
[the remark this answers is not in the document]
catap: @jmroot that adds a few if. Right now it keeps everything related to bootstrap in bootstrap subport.

lang/llvm-3.4/Portfile
catap: Let bump whole revision when.

lang/python27/Portfile
catap: I expect that it doesn't try to install `xz` by dependencies for extract phase.

Instead it uses `xz-bootstrap` and provided command to extract an archive.

Also, as it works. Have I missed something?

archivers/xz/Portfile
```diff
@@ -47,10 +47,41 @@ if {${subport} eq ${name}} {
 
 # This port is used by clang-3.4 to bootstrap libcxx
 subport ${name}-bootstrap {
-    revision                1
-    set my_prefix ${prefix}/libexec/libcxx-bootstrap
```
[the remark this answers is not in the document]
catap: done

devel/pkgconfig/Portfile
catap: done

lang/python27/Portfile
catap: @jmroot it was here before, and I can try to remove it, but it may broke the port somewhere. I prefer to keep it here. Do you suggest to remove it?

textproc/libiconv/Portfile
catap: Done

lang/python27/Portfile
jmroot: In any port that currently sets `use_xz yes` (autoconf for example), try adding `use_xz no` after that. You'll see that xz is still listed in the extract dependencies.

So this line does nothing, and it's the overwriting of `depends_extract` that removes the xz dep.

lang/python27/Portfile
jmroot: If you're worried about changes breaking the port, you probably shouldn't make all these other changes. :)

This was only here to prevent gettext from being found in `$prefix` (b68b114), so it should no longer be needed with the flags cleared of references to `$prefix`.

archivers/xz/Portfile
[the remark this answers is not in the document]
jmroot: Fair enough.

jmroot:
The adding of flags for db48 in the python27 Portfile needs to be prevented for the bootstrap port. (This was a pre-existing oversight.)

archivers/xz/Portfile
```diff
@@ -47,10 +47,41 @@ if {${subport} eq ${name}} {
...
+    configure.cxxflags
+    configure.cppflags
+    configure.optflags
+    configure.ldflags
+
+    configure.universal_cflags
+    configure.universal_cxxflags
+    configure.universal_cppflags
+    configure.universal_ldflags
+    configure.universal_args
```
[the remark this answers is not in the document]
catap: cleaned up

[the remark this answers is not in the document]
kencu: that's on me. I am never 100% sure what is going to slip in where, and went a bit overboard to purify the build I guess.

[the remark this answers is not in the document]
catap: @kencu to be honest I agree with you paranoia and personally I'm also a bit paranoid and tried to except side effects here :)

catap:
@jmroot if I not mistaken I've addressed all your comments. May I ask you to re-review it?

archivers/xz/Portfile
```diff
@@ -47,10 +47,29 @@ if {${subport} eq ${name}} {
...
+    # Avoid macports-clang dep (doesn't use C++ anyway)
+    configure.cxx_stdlib
+    compiler.whitelist      clang llvm-gcc-4.2 gcc-4.2 apple-gcc-4.2
+
+    # sterilize MacPorts build environment; we want nothing picked up from MP prefix
+    compiler.cpath
+    compiler.library_path
+
+    configure.cflags
+    configure.cxxflags
```
jmroot, line 61: `cflags` and `cxxflags` shouldn't be cleared.

devel/pkgconfig/Portfile
```diff
@@ -67,7 +52,7 @@ configure.args      --disable-silent-rules \
 
 if {$subport eq "${name}-bootstrap"} {
     # search bootstrap prefix and then normal prefix
-    configure.args-append   --with-pc-path=${myprefix}/lib/pkgconfig:${myprefix}/share/pkgconfig:${prefix}/lib/pkgconfig:${prefix}/share/pkgconfig
+    configure.args-append   --with-pc-path=${prefix}/lib/pkgconfig:${prefix}/share/pkgconfig:${prefix}/lib/pkgconfig:${prefix}/share/pkgconfig
```
jmroot, line 55: This branch of the if statement will never be taken, since pkgconfig-bootstrap no longer exists.

lang/python27/Portfile
```diff
@@ -61,43 +63,48 @@ depends_lib         port:bzip2 \
...
-openssl.branch      1.1
+
+    openssl.branch  1.1
+}
 
 # This port is used by clang-3.4 to bootstrap libcxx
 subport ${name}-bootstrap {
-    set stdprefix ${prefix}
-    prefix ${prefix}/libexec/libcxx-bootstrap
+    revision                [ expr ${revision} + 1 ]
```
jmroot, line 75: Remember to [brace your expressions](https://wiki.tcl-lang.org/page/Brace+your+expr%2Dessions). :)

jmroot (changes requested):
Getting close.

catap:
@jmroot here it is

archivers/xz/Portfile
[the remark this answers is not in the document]
jmroot: There are still a few `my_prefix` occurrences.

jmroot (changes requested):

kencu (commented):
Thanks for asking for review, will leave this with @jmroot

PR #14119, opened 2022-02-23
mpv: Set MACOS_SDK_VERSION and use latest SDK

ryandesign:
#### Description

mpv: Set MACOS_SDK_VERSION and use latest SDK

Fixes build on Mac OS X 10.7 and on Xcode versions where the SDK version does not match the OS version.

Closes: https://trac.macports.org/ticket/61733
Closes: https://trac.macports.org/ticket/61647
Closes: https://trac.macports.org/ticket/62889
Closes: https://trac.macports.org/ticket/63694

###### Type(s)

- [X] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
Mac OS X 10.7.5 11G63 x86_64
Xcode 4.6.3 4H1503

###### Verification 
Have you

### DIFF
--- a/multimedia/mpv/Portfile
+++ b/multimedia/mpv/Portfile
@@ -262,14 +262,26 @@ platform darwin {
         }
     }
 
-    # Fix for https://trac.macports.org/ticket/61319
-    configure.env-append   MACOS_SDK=${configure.sdkroot}
-
     post-extract {
         xinstall -m 0644 -W "${filespath}" config-maintainer "${worksrcpath}/TOOLS/"
     }
 
     pre-configure {
+        if {${use_xcode}} {
+            configure.sdkroot ${developer_dir}/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
+        }
+        if {${configure.sdkroot} ne ""} {
+            set sdkv [exec /usr/libexec/PlistBuddy -c Print:ProductVersion ${configure.sdkroot}/System/Library/CoreServices/SystemVersion.plist]
+        } else {
+            set sdkv ${configure.sdk_version}
+        }
+        if {[llength [split ${sdkv} .]] < 2} {
+            set sdkv ${sdkv}.0
+        }
+        configure.env-append \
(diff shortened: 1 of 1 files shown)


ryandesign:
I see now that the port used to set `MACOS_SDK_VERSION` but it was removed in 0f60885f6cac1b9e2b10415dc5ebb749172fc37e and the commit message doesn't say why.

Setting `MACOS_SDK_VERSION` is also said to fix https://trac.macports.org/ticket/61647.

ryandesign:
I suspect this will also fix build failure on macOS 12, since upstream does not handle macOS 12 and later correctly. https://github.com/mpv-player/mpv/issues/9907

I see the PR run failed on macOS 11 because the script assumes the SDK version will have two components (like `11.0`) and we are only passing one (`11`). Let me push a fix for that...

ryandesign:
> I see the PR run failed on macOS 11 because the script assumes the SDK version will have two components (like `11.0`) and we are only passing one (`11`).

Looks like this was the reason why MACOS_SDK_VERSION was removed: https://trac.macports.org/ticket/62889. So now that that's fixed it should be fine to continue setting MACOS_SDK_VERSION, yes?

ryandesign:
https://trac.macports.org/ticket/63694 says on macOS 11 with Xcode 13.1 (which contains the macOS 12 SDK) it was necessary to set the SDK version to 12. So I'll push a change that uses the actual SDK version found rather than the SDK version MacPorts requested.

i0ntempest:
Thank you. This complicated port has been in need of some vetting for some time, hopefully this will fix some of the problems.

ryandesign:
Well... the CI build still fails on macOS 11, and I see that it is using the SDK from the CLT even though `use_xcode` is set to `yes`. I'm trying to figure out why that is.

ryandesign:
Ah, right. MacPorts has determined that Xcode doesn't have the macOS 11 SDK it wants to use and then falls back to the CLT SDK. So that's inconvenient.

PR #14133, opened 2022-02-25
flyctl: update to 0.0.301

herbygillot:
- use Golang portgroup
- switch from pre-built binaries to building from source using goreleaser
- remove arch restrictions (supported_archs) now that we build from source
- add @herbygillot as co-maintainer

###### Verification 
Have you

### DIFF
--- a/devel/flyctl/Portfile
+++ b/devel/flyctl/Portfile
@@ -1,15 +1,17 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           github 1.0
+PortGroup           golang 1.0
 
-github.setup        superfly flyctl 0.0.246 v
+go.setup            github.com/superfly/flyctl 0.0.301 v
+github.tarball_from archive
 revision            0
+
 categories          devel
-platforms           darwin
-supported_archs     arm64 x86_64
+installs_libs       no
 license             Apache-2
 maintainers         {netinertia.co.uk:jamesog @jamesog} \
+                    {gmail.com:herby.gillot @herbygillot} \
                     openmaintainer
 
 description         Command line tools for fly.io services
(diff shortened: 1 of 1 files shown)


jamesog (approved):
Thanks @herbygillot, this is great! I've been trying for a while to get this port to play nicely with the MacPorts build system.

Does the fact that goreleaser uses `go mod download` do the right thing for the build system's caching then, or is this just to get it building from source?

herbygillot:
> Thanks @herbygillot, this is great! I've been trying for a while to get this port to play nicely with the MacPorts build system.
> 
> Does the fact that goreleaser uses `go mod download` do the right thing for the build system's caching then, or is this just to get it building from source?

I'm using `goreleaser` to build it from source in exactly the method that they fly.io team is building it.

`goreleaser` actually does things in the way _least_ preferred by MacPorts: it requires a Git checkout, instead of checksummed archives, which means the build assets are not being cached by MacPorts mirrors.

I think what might have been stopping you from getting this ti build when you were looking at this yourself is that the `golang` portgroup sets both the `GO111MODULE` and `GOPROXY` environment variables to `OFF`, which will disable Go from using modules proper and downloading dependencies proper.  We use `build.env-delete` to get rid of those so that Go (and `goreleaser`) can do what they need to do.

I suppose I can take another crack at getting this to build from source using archives, mimicking the flags that `goreleaser` is using to build...

mascguy (approved):

mascguy:
> I suppose I can take another crack at getting this to build from source using archives, mimicking the flags that `goreleaser` is using to build...

Herby, do you want to keep this open, and investigate that option further?

jamesog:
>  think what might have been stopping you from getting this to build when you were looking at this yourself is that the golang portgroup sets both the GO111MODULE and GOPROXY environment variables to OFF, which will disable Go from using modules proper and downloading dependencies proper. We use build.env-delete to get rid of those so that Go (and goreleaser) can do what they need to do.

I've tried patching the portgroup to accept downloading archives from the various sources of the dependencies of flyctl, but e.g. wireguard-go is complex as it's on self-hosted cgit and doesn't support the same archive type as e.g. GitHub. Supporting custom domains is very complex with the way the portgroup works now :-(

herbygillot:
> > I suppose I can take another crack at getting this to build from source using archives, mimicking the flags that `goreleaser` is using to build...
> 
> Herby, do you want to keep this open, and investigate that option further?

Yes, I think let me take a day or so to look at that.

> I've tried patching the portgroup to accept downloading archives from the various sources of the dependencies of flyctl, but e.g. wireguard-go is complex as it's on self-hosted cgit and doesn't support the same archive type as e.g. GitHub. Supporting custom domains is very complex with the way the portgroup works now :-(

Yeah, unfortunately languages that download and manage their own dependencies are going to be at odds with MacPorts, which ideally wants to be the one fetching dependencies.  I agree that the `golang` portgroup is unwieldy.  At some point we do have to stop disabling `GO111MODULE` as that is now a very standard feature for Go projects these days, and most build processes expect to see that enabled.  There was a PR that was suggesting an alternative vendoring approach for Go ports: https://github.com/macports/macports-ports/pull/10381 It would be good to see if there are other approaches that could make sense as well.

jamesog:
#10381 does indeed look promising. I'd be interested in helping improve the Go situation if I can as Go's my primary language and I know the ecosystem reasonably well, but not so much how the MacPorts architecture works, or indeed Tcl.

herbygillot:
> #10381 does indeed look promising. I'd be interested in helping improve the Go situation if I can as Go's my primary language and I know the ecosystem reasonably well, but not so much how the MacPorts architecture works, or indeed Tcl.

Agreed that Tcl and MacPorts can be somewhat opaque.

The gist of the problem is determining a way in which MacPorts can download the dependencies listed in a `go.mod`, and provide them in such a way that `go build` sees and uses them when it comes time to build, and at the same time prevent `go build` from trying to reach out and fetch the dependencies itself.

The way MacPorts' `golang` portgroup does that today is by disabling Go's network access with `GOPROXY=off`, downloading the dependencies listed in `go.vendors` and installing them in a similar directory structure to how Go might do it within the port's working `GOPATH`, then disabling the modern Go modules system with `GO111MODULES=off` to force Go to use that directory structure.  It's not optimal (and can be somewhat kludgy as we've said before), but it mostly achieves MacPorts' goals around having all dependencies needed to build a port. However we do need to reexamine this as disabling Go modules isn't ideal at all.

jamesog:
Thanks, Herby!

PR #14148, opened 2022-02-28
gcc10-bootstrap: fix build on Big Sur

catap:
It also remvoes `bin/gcc` which makes a false impression that `gcc` supports `-arch` flag like `clang`.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

macOS 11.6.1 20G224 x86_64
Xcode 12.5.1 12E507

###### Verification 
Have you

### DIFF
--- a/lang/gcc10-bootstrap/Portfile
+++ b/lang/gcc10-bootstrap/Portfile
@@ -1,6 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           cltversion 1.0
 PortGroup           muniversal 1.0
 
 # ideally for this bootstrap port we want no deps on any MacPorts port. This port
@@ -10,7 +11,7 @@ name                gcc10-bootstrap
 
 # Note, this is the last version of gcc which haven't required c++11 for bootstrap
 version             10.3.0
-revision            3
+revision            4
 epoch               0
 
 platforms           darwin
@@ -157,9 +158,10 @@ platform darwin 8 {
                             PATH=${workpath}/bins:/usr/bin:/bin:/usr/sbin:/sbin
 }
 
+prefix              ${prefix}/libexec/${name}
+
 configure.cmd       ${worksrcpath}/configure
 configure.dir       ${workpath}/build
-configure.pre_args  --prefix=${prefix}/libexec/${name}
 configure.args      --enable-languages=c,c++,objc,obj-c++ \
                     --enable-bootstrap \
                     --with-system-zlib \
@@ -174,30 +176,17 @@ if {${os.major} >= 18 && ${configure.sdkroot} ne ""} {
     configure.args-append --with-sysroot="[regsub {MacOSX1[0-9]\.[0-9]+\.sdk} ${configure.sdkroot} {MacOSX.sdk}]"
 }
 
-set universal_targets ""
-foreach arch ${configure.universal_archs} {
-    switch -- ${arch} {
-        arm64   {set universal_targets "${universal_targets},aarch64-apple-${os.platform}${os.version}"}
-        x86_64  {set universal_targets "${universal_targets},x86_64-apple-${os.platform}${os.version}"}
-        i386    {set universal_targets "${universal_targets},i686-apple-${os.platform}${os.version}"}
-        ppc     {set universal_targets "${universal_targets},powerpc-apple-${os.platform}${os.version}"}
-        ppc64   {set universal_targets "${universal_targets},powerpc64-apple-${os.platform}${os.version}"}
+# clang (as) from Xcode 12.5 has various problems with gcc build
+if { ${os.platform} eq "darwin" && \
+         ( [ vercmp ${xcodeversion} 12.5 ] >= 0 || [ vercmp ${cltversion} 12.5 ] >= 0 ) } {
+    pre-configure {
+        ui_warn "Applying '--without-build-config' workaround to Xcode ${xcodeversion} / CLT ${cltversion}"
+        ui_warn "If versions > 12.5 please check if it is still required"
     }
-}
-
-variant universal {
-    configure.args-replace \
-                    --disable-multilib \
-                    --enable-multilib
-
-    configure.args-replace \
-                    --disable-multiarch \
-                    --enable-multiarch
-
-    if { [string length ${universal_targets}] > 0 } {
-        configure.args-append \
-            --enable-targets=[string replace ${universal_targets} 0 0]
-    }
+    # gcc has build issues on macOS 11.3 with the use of Xcode clang as 'as'
+    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100340
+    # https://trac.macports.org/ticket/62775
+    configure.args-append  --without-build-config
 }
 
 default_variants    +universal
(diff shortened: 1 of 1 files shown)


catap:
@kencu FYI

kencu:
To be honest, this is not really the way I see this going at the moment.

First, before anything, can someone demonstrate that this can be used to build something universal using the muniversal PortGroup? Try libffi, mfpr, isl, gmp -- anything.

Gcenx:
@catap you’ve bot updated the notes section explaining how to use the provided compilers, it’s still saying to use gcc.

catap:
@Gcenx thanks, I'll forget about it.

kencu:
Still have to see if this works, first of all.

Even then -- because this compiler would now be different than every other compiler currently in macports regarding how to use it, we have to expect this will not be accepted. 

But at least, if it can work, that is one thing.

kencu:
(btw -- gcc does support the arch flag like clang, but just not multiple arch flags at this moment in time.)

Gcenx:
The thing is the way this bootstrap compiler is setup each `${arch}-apple-${os.platform}${os.major}-gcc` binary only supports a single arch.

To use this for universal compiles does requires setting the correct target compiler for the selected arch.

As this works differently I did express the desire to add the shell wrapper, but it seems everyone confused this for driver-driver so I’d dropped it.

> ie. as above, have it build something universal using the muniversal PortGroup.

I did provided main.log of my universal2 compile for clang-11-bootstrap in the other thread, I personally won’t have time to compile anything else until I’ve done the wine-devel/wine-staging-7.3 packages.

kencu:
No, I was not confused by the shell script idea.

Can't speak for everyone else.

the clang11-bootstrap thing is not really relevant, because it uses a cmake trick that is not useful for the general bunch of ports out here, as per the (other other other) PR. It's nice to see that worked, but insufficient for this to be accepted.

Gcenx:
@catap could you compile something as pre @kencu and provide a complete main.log

Gcenx:
@catap I’d still think it’s best to provide a shell wrapper that provides gcc g++ etc that points to the desired target compiler that might help make this more usable by more Ports

catap:
@Gcenx  / @kencu I've made the force push and introduced simple shell script that allows to use this port file to build something universal.

Will update my clang port to this new mechanism ASAP.

kencu:
I’ll have to sit down with this and see what it actually can do. I really appreciate all the work here…this simple simple bootstrap gcc10 was never meant to be the project that solves multi-arch universal building on MacPorts with gcc.

That is a big project, that will require driverdriver.c to be updated and two or more full separate gcc toolchains merged, probably with a huge shell script as in apple-gcc42.

My apologies if I had any rolecin leading you down this rabbithole here. I judt wanted what we had near the beginning of this project, like gcc7-bootstrap was, with the i386 SnowLeopard issue sorted somehow out…

Gcenx:
This is getting rather confusing _again_

I’d personally keep providing target compilers that only support a single arch.

The shell wrapper doesn’t really make this easier to use, from my view the current implementation makes this worse than setting the compiler per arch. The only way to really make this seemless is reading the appropriate flags from the environment and setting the needed target compiler accordingly.

<br>

The way I’d think would make the shell wrapper viable is doing the following.

Having gcc check cflags for arch and setting the compiler accordingly, this would be the same for g++ etc. Could make use of basename to keep this a single template and simply copy into place for each binary. However using a shell wrapper will probably impact compile times.

kencu:
> The only way to really make this seamless is reading the appropriate flags from the environment and setting the needed target compiler accordingly.

that is precisely what I was wondering if the muniversal PG would accept and might (or might not) do.

Nobody has ever demonstrated that works (setting the compiler to a different compiler on a _per-arch_ basis) and I think it won't, but I would be delighted to be wrong about that.

catap:
@Gcenx the issue here that you can use quite limited number of value inside `configure.compiler` and to change it, you must update the base :(

Gcenx:
@catap then we simply set CC/CXX for each arch for the desired compiler.

catap:
@kencu thanks for multilib part. It was enabled globally by mistake.

@Gcenx let me try this idea, maybe it seems simpler

lang/gcc10-bootstrap/Portfile
```diff
@@ -250,16 +239,50 @@ build.target        bootstrap-lean
...
 
+post-destroot {
+    foreach f [glob -nocomplain -d ${destroot}${prefix}/bin c++ g++ gcc gcc-ar gcc-nm gcc-ranlib] {
+        file delete $f
+    }
+
+    set cc [open "${destroot}${prefix}/bin/cc" w 0755]
+    puts ${cc} "#!/bin/sh"
+    puts ${cc} "GCC_TARGET=\"\$\{GCC_TARGET:-$merger_host(${build_arch})\}\""
+    puts ${cc} "PROG=\$(basename \"\$0\")"
```
Gcenx, line 250: From memory `$(basename $0)` won’t work for via symlinks, instead just add the script for each binary

catap:
@Gcenx / @kencu now it should be much better.

Gcenx:
@catap still might want to provide a build log showing something being compiled as universal to put @kencu mind at ease.

kencu:
that is looking pretty interesting! Looking forward to kicking the tires and seeing what we get.

catap:
@kencu / @Gcenx I've took a random port which is used muniversal PG: `libpostal`. Applied a diff:
```
diff --git a/textproc/libpostal/Portfile b/textproc/libpostal/Portfile
index b8fe4cfd46a..35c0ab3fdd6 100644
--- a/textproc/libpostal/Portfile
+++ b/textproc/libpostal/Portfile
@@ -43,4 +43,18 @@ if {(!${universal_possible} || ![variant_isset universal]) && [info exists merge
     configure.args-append   $merger_configure_args(${configure.build_arch})
 }
 
+if {${universal_possible} && [variant_isset universal]} {
+    array set cpu_arch_map {arm64 aarch64 i386 x86 ppc powerpc ppc64  powerpc64 x86_64 x86_64}
+
+    configure.cc    {}
+    configure.cxx   {}
+
+    foreach {arch target} [array get cpu_arch_map] {
+        lappend merger_configure_env(${arch}) \
+            CC=${prefix}/libexec/gcc10-bootstrap/bin/${target}-apple-${os.platform}${os.major}-gcc
+        lappend merger_configure_env(${arch}) \
+            CXX=${prefix}/libexec/gcc10-bootstrap/bin/${target}-apple-${os.platform}${os.major}-g++
+    }
+}
+
 github.livecheck.regex {([0-9.]+)}
```

and tested to build it with universal variant via gcc10-bootstrap as `sudo port -sk install libpostal +universal`. It produces:
```
catap@Kirills-mini-m1 logs % file /opt/local/lib/libpostal.1.dylib     
/opt/local/lib/libpostal.1.dylib: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
/opt/local/lib/libpostal.1.dylib (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
/opt/local/lib/libpostal.1.dylib (for architecture arm64):	Mach-O 64-bit dynamically linked shared library arm64
catap@Kirills-mini-m1 logs %  
```

The build log: [main.log.gz](https://github.com/macports/macports-ports/files/8192552/main.log.gz)


P.S. I've used m1 as the faster machine, anyway,

catap:
@kencu just an idea: I may move my code from notes to muniversal for everything before 10.7 for example.

kencu:
I think you've done some interesting stuff here that leverages macports tech to make gcc build universal for us, despite there being no driverdriver.c update, etc.

I think -- you can expect pushback, as it is different. For example, the whole compiler blacklisting plan, etc, etc, etc ...

If it has a discrete role, ie in building a fixed set of bootstrap ports that know exactly how this is set up, it should be OK. 

the muniversal portgroup goes back to times when Apple was still involved in this I believe. Although there have been some additions and revisions over the years, remarkably little to be honest, considering everything that has happened in the past 10 years.

People will be quite cautious to touch how it works too much, because it frankly works so good ...

Gcenx:
Think it’s better to keep the notes within the port, this would keep it inline with other special ports like the other bootstrap ports.

catap:
@kencu / @Gcenx I've agree with you. I just put some ideas on the table.

I think this PR is ready on my side.

kencu:
I'm going to have to look at this a bit more closely. I don't have an arm64 machine, so can you show me what gcc libraries /opt/local/lib/libpostal.1.dylib is linked to, exactly? 

I'm not sure what we have, exactly, in terms of libraries in gcc to link against right now -- both mutlilib and multiarch have been added. We've never had a gcc compiler build as multiarch -- what do you get with that? Is it one big fat libgcc with all the arches in it? 

How are universal things linked? against one library in gcc, or against several, one for each arch?

Still some questions here, and you can expect (once this gets looked at carefully after it's been finally committed) lots of criticisms about how this is being done, and quite likely -- a total redo.

catap:
@kencu
```
catap@Kirills-mini-m1 ~ % otool -L /opt/local/lib/libpostal.1.dylib
/opt/local/lib/libpostal.1.dylib (architecture x86_64):
	/opt/local/lib/libpostal.1.dylib (compatibility version 2.0.0, current version 2.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
/opt/local/lib/libpostal.1.dylib (architecture arm64):
	/opt/local/lib/libpostal.1.dylib (compatibility version 2.0.0, current version 2.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
catap@Kirills-mini-m1 ~ % 
```

I've used m1 because it is the faster machine that I have and it rebuilds gcc for less than a 2 hours :)

My goal here is simple: let delivery it as is. Current changes introduce a nice side effect that allows to use this gcc as something that can produce universal variant on system which doesn't support by clang.

For me it is nice side effect that costs near of nothing for me to implement (after all work that I did here).

As soon as it merged, I've update `clang-11-bootstrap` and `libcxx` and may move forward with `mrustc` which should allow to bring `rust` on 10.6.

kencu:
So good (or bad) luck I guess, that particular port uses no libraries in gcc.

Tried to build on mojave last night, error noted, will look into it:
[gcc10-bootstrap-20220307-mojave-error.txt.zip](https://github.com/macports/macports-ports/files/8207263/gcc10-bootstrap-20220307-mojave-error.txt.zip)

kencu:
So -- this is a lot of work indeed. We need a few beta testers here to demonstrate this work well.

At the moment, the only systems I (we) care about this working on are 10.6 through 10.11 Intel, as those are the only ones that need to use this to bootstrap to clang11 bootstrap. 

All the rest, for now, are totally irrelevant and a distraction.

So if anyone would care to pitch in on some of those systems, install this, kick the tires, build a few things universal with it, especially maybe something that links into the gcc libraries, show how it works OK.

For example: with this current process, if something links into one of the gcc libraries like libstdc++.dylib or libgcc_s.1.dylib, does each arch link to a different dylib? I am not clear on how this turned out, with two parallel compilers, one for i386 and one for x86_64, each one having it's own libgcc and each one having multilib (and multiarch!) turned on.

What have we got?

I'll try building it on 10.6 next I guess, as universal, and see.

catap:
@kencu I can build universal clang-11-bootstrap and after that libcxx on 10.6 for example.

Gcenx:
@catap could you try this another way instead?

Use gcc10-bootstrap to compile libcxx and then compile clang-11-bootstrap, that could avoid the need to later depend on gcc10-bootstrap.

kencu:
but then libcxx will depend on gcc10-bootstrap I believe :>

catap:
@kencu thanks for merge it, I'll move with clang / libcxx on this week.

catap:
and it haven't help with Big Sur :(

Gcenx:
@catap I applied the upstream patch worked without a rebase even.

catap:
@Gcenx which patch? :)

Gcenx:
@catap I’ll link you later I believe I’d mentioned the issue in the libcxx thread

kencu:
This BigSur fix didn't fix the build on BigSur? Oh, crap.

Gcenx:
@catap the following patch fixed the compile issues on my system https://gcc.gnu.org/git/?p=gcc.git;a=patch;h=8b333df9484c1697f3a80530a47aa90b1859e970

PR #14153, opened 2022-02-28
hugo: update to 0.93.0

herbygillot:
###### Verification 
Have you

### DIFF
--- a/www/hugo/Portfile
+++ b/www/hugo/Portfile
@@ -3,29 +3,35 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/gohugoio/hugo 0.92.2 v
+go.setup            github.com/gohugoio/hugo 0.93.1 v
+github.tarball_from archive
 revision            0
-checksums           rmd160  f5953f9829be8b48c5fcf398819931f21ca8e671 \
-                    sha256  73d20ce55298f809a8f735c33a4c90e8e36fa56dbc94a6a36ba23fb19eb9ce07 \
-                    size    28559747
+checksums           rmd160  eb21a34a97b522240cf87733dc41158620597292 \
+                    sha256  b21d83c1011a785ab28716227849d661671e00ed9c43b195789835d439d0d60e \
+                    size    28562897
 
 categories          www
-platforms           darwin
 license             Apache-2
 maintainers         {isi.edu:calvin @cardi} openmaintainer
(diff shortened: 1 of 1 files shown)


herbygillot:
@cardi - I think it would make sense to have Bash completion & tags support be built-in by default instead of optional variants.  Thoughts?

herbygillot:
> I agree. I think that'd be useful, and wouldn't impose an undue overhead for those that don't use those features.
> 
> There is also an existing ticket to update how bash completion is generated: https://trac.macports.org/ticket/64735 that I was building a patch for.
> 
> Would you prefer to have hugo be updated to 0.93.0 first before moving the variants to default in another PR, or should we fold all of those changes into this PR?

I don't have a strong preference either way, but I suppose it probably makes more sense to just fix everything in one PR.

I (_think_) when we remove the variant (because we're consolidating the feature into the main build), whoever who currently has `hugo` installed with the variant enabled might run into an error state (I _think_).  So we will have to mention this in the port notes.

cardi:
I've attached a patch based off of this PR that:
- updates hugo to 0.93.1
- builds and installs bash completions and extended support by default
- removes the `+bash_completion` and `+extended` variants

I've verified the install works with `sudo port -vst install`, and the bash completion scripts are generated and installed properly.

Patch: [0001-hugo-update-to-0.93.1.patch.txt](https://github.com/macports/macports-ports/files/8174174/0001-hugo-update-to-0.93.1.patch.txt)

I think all you'd have to do on your branch is apply the patch and squash the commits (feel free to use or modify my commit message):
```shell
curl -LO https://github.com/macports/macports-ports/files/8174174/0001-hugo-update-to-0.93.1.patch.txt
git am 0001-hugo-update-to-0.93.1.patch.txt
git rebase -i HEAD~2
```

> I (think) when we remove the variant (because we're consolidating the feature into the main build), whoever who currently has hugo installed with the variant enabled might run into an error state (I think). So we will have to mention this in the port notes

I installed this on my own setup with hugo and didn't run into any issues. (I think `port` ignores any variant that doesn't exist, e.g., `sudo port install hugo +foobar` where `foobar` doesn't exist.)

pmetzger:
I think the general consensus for almost a year or two has been to include completion files by default since they are so small compared to the rest of a package, so I agree that should happen.

cardi:
LGTM. Thank you, @herbygillot!

herbygillot:
Thank you @cardi, and also agreed, @pmetzger.

PR #14189, opened 2022-03-04
foundationdb: update to 6.3.24

usersxx:
#### Description
Updated foundationdb to 6.3.4, Changed `description` and `long_description`

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G65 x86_64
 

###### Verification 
Have you

### DIFF
--- a/databases/foundationdb/Portfile
+++ b/databases/foundationdb/Portfile
@@ -5,27 +5,30 @@ PortGroup           github 1.0
 PortGroup           cmake 1.1
 PortGroup           compiler_blacklist_versions 1.0
 
-github.setup        apple foundationdb 6.3.12
-revision            1
+github.setup        apple foundationdb 6.3.24
+revision            0
 
 categories          databases
 license             Apache-2
 maintainers         {me.com:spam_brian @brianwells} openmaintainer
 platforms           darwin
-description         FoundationDB is a distributed database providing ACID transactions.
-long_description    FoundationDB is a scalable, fault-tolerant, ordered \
-                    key-value store with full ACID transactions.  This port \
-                    will install version ${version} of the FoundationDB \
-                    binaries suitable for development on macOS.
+description         FoundationDB gives you the power of ACID transactions in a distributed database.
+long_description    FoundationDB is a distributed database designed to handle large volumes of structured data \
+                    across clusters of commodity servers. It organizes data as an ordered key-value store and \
+                    employs ACID transactions for all operating systems. It is especially well-suited for read/write workloads \
+                    but also has excellent performance for write-intensive workloads. Users interact with the \
+                    database using API language binding. FoundationDB is Open source, is Ready for production, \
+                    provides Industry-leading performance, is Fault tolerant and is Multi-model. This Port provides \
+                    Latest FoundationDB Release to macOS with full support for all these features.
 
 homepage            https://www.foundationdb.org/
 
-checksums           rmd160  2bf9a258d5db140d232fb6355fc810010824c014 \
-                    sha256  034f69d4ddf7faccd7cae2a6438d64643ac98a3a9c7b3e6ed2a03105c480df95 \
-                    size    9694306
+checksums           rmd160  470b145f9974cc0898a176425d72f9bc43704248 \
+                    sha256  e5ab9925d1a1f2f3ceb48ce4d3c5211176aedf3f04df7d3945b578f7646b9fb2 \
+                    size    10654924
 
 # Required python version
-set py_ver          3.9
+set py_ver          3.10
 set py_ver_nodot    [string map {. {}} ${py_ver}]
 
 depends_lib-append  port:mono


usersxx:
@herbygillot Can you merge this pr?

herbygillot:
> @herbygillot Can you merge this pr?

This port has a maintainer.  In _most_ instances, the port will be left open for up to 36 hours to give the maintainer the opportunity to review and approve.  That may not always be followed, but that's the normal operating procedure.

My only comment would be that I would add the Ruby and Java bindings as variants.

databases/foundationdb/Portfile
```diff
@@ -5,31 +5,40 @@ PortGroup           github 1.0
 PortGroup           cmake 1.1
 PortGroup           compiler_blacklist_versions 1.0
 
-github.setup        apple foundationdb 6.3.12
+github.setup        apple foundationdb 6.3.24
 revision            1
```
herbygillot, line 9: `revision` should reset to 0 when the version gets bumped.

databases/foundationdb/Portfile
```diff
@@ -5,28 +5,44 @@ PortGroup           github 1.0
...
 # Required python version
-set py_ver          3.9
+set py_ver          3.10
 set py_ver_nodot    [string map {. {}} ${py_ver}]
+set jdk_ver         8
+set ruby_ver        3.1
+set ruby_ver_nodot  [string map {. {}} ${ruby_ver}]
+default_variants    +python${py_ver_nodot}
+
+variant openjdk${jdk_ver} description {add java binding support with openjdk meta port} {
```
herbygillot, line 38: Inside the braces for each variant is where you add the dependencies and build arguments that would enable that variant in the build process:

```
variant openjdk${jdk_ver} description {add java binding support with openjdk meta port} {
    depends_build-append port:openjdk-...
    build.args-append -DENABLE_JAVA.... # whatever build args specifically enable Java
}
```

brianwells:
Thanks for the PR, @usersxx.  I agree with @herbygillot that the the Ruby and Java bindings should be variants and appreciate you making these adjustments to your PR.

databases/foundationdb/Portfile
```diff
@@ -5,28 +5,51 @@ PortGroup           github 1.0
 PortGroup           cmake 1.1
 PortGroup           compiler_blacklist_versions 1.0
 
-github.setup        apple foundationdb 6.3.12
-revision            1
+github.setup        apple foundationdb 6.3.24
+revision            0
```
brianwells, line 9: I believe the `revision` should be incremented with each change to the Portfile.

brianwells: On second thought, the version increase should be sufficient and thus the change of `revision` to `0` is fine :)

databases/foundationdb/Portfile
```diff
@@ -5,28 +5,51 @@ PortGroup           github 1.0
...
 set py_ver_nodot    [string map {. {}} ${py_ver}]
+set jdk_ver         8
+set ruby_ver        3.1
+set ruby_ver_nodot  [string map {. {}} ${ruby_ver}]
+default_variants    +python${py_ver_nodot}
+
+variant openjdk${jdk_ver} description {add java binding support with openjdk meta port} {
+    depends_build-append port:openjdk8
+    if {${configure.build_arch} eq "x86_64"} {
+        configure.env-append  PYTHON_ROOT_DIR=${prefix}/bin/python${py_ver} JAVA_HOME=/Library/Java/JavaVirtualMachine/openjdk8-temurin/Contents/Home
```
brianwells, line 41: We shouldn't need to append `PYTHON_ROOT_DEV=...` to the configure env here since it is already done so at line 58

databases/foundationdb/Portfile
```diff
@@ -5,28 +5,51 @@ PortGroup           github 1.0
...
+    }
+}
+
+variant ruby${ruby_ver_nodot} description {add ruby binding support with ruby} {
+    depends_build-append port:ruby${ruby_ver_nodot}
+    configure.env-append  PYTHON_ROOT_DIR=${prefix}/bin/python${py_ver} RUBY_ROOT_DIR=${prefix}/local/bin/ruby${ruby_ver}
```
brianwells, line 47: We shouldn't need to append `PYTHON_ROOT_DEV=...` to the configure env here since it is already done so at line 58. Just the `RUBY_ROOT_DIR=...` should be sufficient.

herbygillot:
Are there no distinct build flags that explicitly enable or disable each of these add-ons?

The way this seems to be, if you just happened to have Python installed while building this port, you would get Python add-ons whether you wanted them or not.

brianwells:
According to the FoundationDB [docs](https://github.com/apple/foundationdb#compiling-from-source), the build process will "build all language bindings for which it can find all necessary dependencies." So there are no build flags for any of the language bindings. The C binding depends on the Python binding, so the Portfile included a dependency on Python. All the other bindings depend on the C binding (and Python binding) so they can be optional. From the [cmake](https://github.com/apple/foundationdb/blob/96dd86ebf8c63d5650b8b8c6ea8943ab9e2c6733/cmake/FDBComponents.cmake) file it looks like there are build flags to enable/disable the bindings and they default to ON if the appropriate executable is found.

herbygillot:
> From the [cmake](https://github.com/apple/foundationdb/blob/96dd86ebf8c63d5650b8b8c6ea8943ab9e2c6733/cmake/FDBComponents.cmake) file it looks like there are build flags to enable/disable the bindings and they default to ON if the appropriate executable is found. We would have to find a way to set the build flags OFF for non-variant and then ON for variants.

Nice, that looks like it would do the trick.  In the standard non-variant build, you'd have something like:
```
build.args-append -DBUILD_JAVA_BINDING=OFF \
                  -DBUILD_RUBY_BINDING=OFF
```

Then in each variant, you would add the build dependencies needed, replace the relevant build option to enable it (`build.args-replace -DDBUILD_JAVA_BINDING=OFF -DDBUILD_JAVA_BINDING=ON`), and enable other dependent build flags if that's necessary (like [`BUILD_RUBY_BINDING` also needs `BUILD_C_BINDING`](https://github.com/apple/foundationdb/blob/96dd86ebf8c63d5650b8b8c6ea8943ab9e2c6733/cmake/FDBComponents.cmake#L149) for example)

databases/foundationdb/Portfile
herbygillot: This path looks incorrect: `RUBY_ROOT_DIR=${prefix}/local/bin/ruby${ruby_ver}`

The standard MacPorts prefix is `/opt/local`, so this would set `RUBY_ROOT_DIR` to `/opt/local/local/bin/ruby${ruby_ver}`

herbygillot:
Ah, the build failed because this is using `ninja`, not `cmake`.  With `cmake`, you can directly control build options using `-D`:

```
ports_ports_databases_foundationdb/foundationdb/work/build" && ninja -j1 all -DBUILD_JAVA_BINDING=OFF -DBUILD_RUBY_BINDING=OFF -j1 -v 
  ninja: invalid option -- D
```

brianwells:
> Ah, the build failed because this is using `ninja`, not `cmake`. 

@usersxx: There must be some way to pass those build options to `ninja`. I'll have some time to investigate later this evening but perhaps you may figure something out as well.

brianwells:
@usersxx: After some digging into other ports that also use Ninja+cmake, it appears that the correct way to handle the flags is through `configure.args-append`. So they get passed to `cmake` instead of `ninja`.

In the non-variant build it would be:
```
configure.args-append -DBUILD_JAVA_BINDING=OFF \
                 -DBUILD_RUBY_BINDING=OFF
```

Then in a variant something like:
```
configure.args-replace -DBUILD_JAVA_BINDING=OFF -DBUILD_JAVA_BINDING=ON
```

usersxx:
Yep, looks good to me

herbygillot:
> @usersxx: After some digging into other ports that also use Ninja+cmake, it appears that the correct way to handle the flags is through `configure.args-append`. So they get passed to `cmake` instead of `ninja`.
> 
> In the non-variant build it would be:
> 
> ```
> configure.args-append -DBUILD_JAVA_BINDING=OFF \
>                       -DBUILD_RUBY_BINDING=OFF
> ```
> 
> Then in a variant something like:
> 
> ```
> configure.args-replace -DBUILD_JAVA_BINDING=OFF -DBUILD_JAVA_BINDING=ON
> ```

My fault, it should have been `configure` in my example, and not `build`.  🤦🏾

databases/foundationdb/Portfile
```diff
@@ -40,11 +40,13 @@ variant openjdk${jdk_ver} description {add java binding support with openjdk met
     if {${configure.build_arch} eq "x86_64"} {
         configure.env-append  PYTHON_ROOT_DIR=${prefix}/bin/python${py_ver} JAVA_HOME=/Library/Java/JavaVirtualMachine/openjdk8-temurin/Contents/Home
     }
+    configure.args-replace -DBUILD_JAVA_BINDING=OFF -DBUILD_JAVA_BINDING=ON
 }
 
 variant ruby${ruby_ver_nodot} description {add ruby binding support with ruby} {
     depends_build-append port:ruby${ruby_ver_nodot}
-    configure.env-append  PYTHON_ROOT_DIR=${prefix}/bin/python${py_ver} RUBY_ROOT_DIR=${prefix}/bin/ruby${ruby_ver}
+    configure.env-append  PYTHON_ROOT_DIR=${prefix}/bin/python${py_ver} RUBY_ROOT_DIR=${prefix}/bin/ruby${ruby_ver
```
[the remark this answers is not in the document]
usersxx: done

usersxx:
Is this port ready to merge

brianwells:
This looks good.

brianwells (approved):
This update to FoundationDB version 6.3.24 looks good. The addition of variants for go, ruby, openjdk are very helpful.

usersxx:
> This update to FoundationDB version 6.3.24 looks good. The addition of variants for go, ruby, openjdk are very helpful.

@herbygillot you can merge this pr as @brianwells approved these changes

herbygillot:
So I took a look at the most recent build output, and it turns out the build options are not working as we expected they would:

```
  -- =========================================
  --    Components Build Overview 
  -- =========================================
  -- Build Java Bindings:                  ON
  -- Build with TLS support:               ON
  -- Build Go bindings:                    OFF
  -- Build Ruby bindings:                  ON
  -- Build Python sdist (make package):    ON
  -- Build Documentation (make html):      OFF
  -- Build Bindings (depends on Python):   ON
  -- Configure CTest (depends on Python):  ON
  -- Build with RocksDB:                   OFF
  -- =========================================
  -- CPACK_COMPONENTS_ALL clients-deb;clients-el7;clients-tgz;clients-versioned;server-deb;server-el7;server-tgz;server-versioned
  -- Configuring done
  -- Generating done
  CMake Warning:
    Manually-specified variables were not used by the project:
  
      BUILD_JAVA_BINDING
      BUILD_RUBY_BINDING
      CMAKE_COLOR_MAKEFILE
      CMAKE_OBJCXX_COMPILER
      CMAKE_OBJC_COMPILER
      CMAKE_POLICY_DEFAULT_CMP0025
      CMAKE_POLICY_DEFAULT_CMP0060
```
-- from https://github.com/macports/macports-ports/runs/5463881287?check_suite_focus=true#step:8:504

Java and Ruby bindings are still on even though we expected them off, and the reason is because CMake is saying it doesn't recognize those options ("Manually-specified variables were not used by the project").

If we're done fighting this dragon, then we can remove the variants as they don't work as we're expecting them to, or we could dig deeper.

usersxx:
> If we're done fighting this dragon, then we can remove the variants as they don't work as we're expecting them to, or we could dig deeper.

Yep I agree

usersxx:
@herbygillot You can merge this pull request, I have removed all variants.

herbygillot:
Merged.  Thanks for your efforts on this @usersxx (and thanks for taking the time to review and assist, @brianwells).

PR #14201, opened 2022-03-05
submission: php-xmlrpc: new port (from PECL)

stronk7:
#### Description

Before PHP 8.0, the xmlrpc extension was bundled with PHP, but
now it's moved to PECL.

https://wiki.php.net/rfc/unbundle_xmlprc
https://php.watch/versions/8.0/xmlrpc

Current version (1.0.0RC3) works ok for PHP 8.0 and 8.1. While it's
really using old libraries and the recommendation is to move away
from it to newer/better alternatives, there is still some code out
there using it.

Hence, this provides the extension, built from PECL, for PHP 8.0 and up.
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.4 20G417 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/php/php-xmlrpc/Portfile
+++ b/php/php-xmlrpc/Portfile
@@ -1,0 +1,22 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               php 1.1
+
+name                    php-xmlrpc
+version                 1.0.0RC3
+categories-append       devel net
+maintainers             {ryandesign @ryandesign} openmaintainer
+license                 PHP-3.01
+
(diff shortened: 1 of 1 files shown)


stronk7:
Wow, thanks @ryandesign !

For sure I wouldn't have found how to make the trick ever! Curiously it was working locally.

Thanks!

ryandesign:
The reason for the CI failure was the missing pkgconfig dependency. You probably already had it installed, but build machines deliberately do not have anything installed that the port does not specify.

The libxml2 dependency would have already been there due to the main php port but since this port uses it directly it should be specified here directly too.

The `vercmp` block is unnecessary here at the moment but is a style I use in many of the other php module ports. It is common for new versions of modules to drop support for old PHP versions; this style makes it possible to continue to offer the older module versions for older PHP versions.

stronk7:
Thanks!

PR #14240, opened 2022-03-13
nodejs8: require OpenSSL 1.1

evanmiller:
#### Description

Builds are failing across the board, see https://ports.macports.org/port/nodejs8/builds/

Closes: https://trac.macports.org/ticket/64095

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/nodejs8/Portfile
+++ b/devel/nodejs8/Portfile
@@ -39,8 +39,10 @@ distname                node-v${version}
 
 depends_build-append    port:pkgconfig
 
+# Require OpenSSL 1.1 and not generic OpenSSL
+# https://trac.macports.org/ticket/64095
 depends_lib-append      port:python27 \
-                        path:lib/libssl.dylib:openssl
+                        port:openssl11
 
 proc rec_glob {basedir pattern} {
     set files [glob -directory $basedir -nocomplain -type f $pattern]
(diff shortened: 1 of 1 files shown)


herbygillot:
Ports have been moving to using the `openssl` portgroup, so it may make sense to use it here.

We can remove the `depends_lib` reference to ` port:openssl11`, as well as the `configure` arguments configuring against the OpenSSL library and replace them with:

1. Adding the `openssl` portgroup:
  ```
  Portgroup openssl 1.0
  ``` 
2. And since OpenSSL 1.1 is required here:
  ```
  openssl.branch 1.1
  ```

This is already being done in the Portfile for `nodejs13`:
https://github.com/macports/macports-ports/blob/7903be22b83491ae8f565be5707c759ae8c98053/devel/nodejs13/Portfile#L14-L15

evanmiller:
@herbygillot Interesting, thanks. I did hit an additional snag needing to pass in `--shared-openssl-libname=crypto.1.1,ssl.1.1` in order to avoid a conflict when the plain `openssl` port is installed. Not sure if that case would be covered by the PG.

The `nodejs8` port is now deprecated, so unless there's specific desirable functionality provided by the PG, I'd personally prefer to keep the change list to a minimum here.

PR #14265, opened 2022-03-15
cask: update to 0.8.7

wyuenho:
cask: update to 0.8.7
* update default python version to 3.10

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3
Xcode 13.3

###### Verification 
Have you

### DIFF
--- a/devel/cask/Portfile
+++ b/devel/cask/Portfile
@@ -5,7 +5,7 @@ PortGroup           python 1.0
 PortGroup           github 1.0
 PortGroup           elisp 1.0
 
-github.setup        cask cask 0.8.6 v
+github.setup        cask cask 0.8.7 v
 revision            0
 description         Project management tool for Emacs
 long_description    Cask is a project management tool for Emacs that helps \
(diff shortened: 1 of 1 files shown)


amake:
cask 0.8.7 doesn't work on macOS properly because it lacks [this patch](https://github.com/cask/cask/commit/1c329de12cb2fba1552df5f1fb65a5613aa4db50). So either someone should prompt upstream to tag a new release and update to that (preferable), or the patch should be incorporated into the port.

Note also that the above patch introduces an ad hoc invocation of `python` (though there are probably more elsewhere as well). That needs to be modified to use the MacPorts `${python.bin}`.

wyuenho:
Alternatively, I can submit a patch upstream to remove the dependency on Python altogether and depend on coreutils on older macOSes, if we can find out when macOS started supporting `readlink -f`.

amake:
I dug up a Big Sur (11.6.1) machine and it doesn't support `readlink -f`. So it was probably added in Monterey (12).

Thanks for poking upstream!

PR #14266, opened 2022-03-15
exiftool: update to 12.4.0

wyuenho:
exiftool: update to 12.4.0

* change default perl branch to 5.34

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3
Xcode 13.3

###### Verification 
Have you

### DIFF
--- a/perl/exiftool/Portfile
+++ b/perl/exiftool/Portfile
@@ -7,12 +7,12 @@ name                exiftool
 
 perl5.require_variant   yes
 perl5.conflict_variants yes
-perl5.branches          5.26 5.28 5.30
-perl5.default_branch    5.28
+perl5.branches          5.26 5.28 5.30 5.32 5.34
+perl5.default_branch    5.34
 perl5.create_variants   ${perl5.branches}
 
 # keep version in sync with p5-image-exiftool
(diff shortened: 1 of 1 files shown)


ryandesign:
> * change default perl branch to 5.34

Why? Until you see it changed in [the perl-1.0 portgroup](https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/perl5-1.0.tcl#L11-L17), 5.28 is still the preferred perl version in MacPorts.

wyuenho:
I just saw [imageindex](https://github.com/macports/macports-ports/blob/master/graphics/imageindex/Portfile#L29) has done the same and thought it would be okay. I'm not really sure what the policy about default perl version is. It's not documented anywhere.

ryandesign:
The preferred default versions of python and perl and php are documented in their respective portgroups. Users complain when MacPorts installs what they consider too be unnecessary dependencies, especially multiple versions of python or perl. That's why we have a preferred default version that ports should use so that ideally users will only get one version of them. Yes, we should update the preferred perl version to 5.34 but we haven't yet.

PR #14323, opened 2022-03-21
ncurses: fix build on Linux

harens:
ncurses didn't build for me on Raspberry Pi OS. I found an [old patch](https://trac.macports.org/attachment/ticket/48957/ncurses.diff) by @RJVB on trac for this issue that still seems to work.

Closes: https://trac.macports.org/ticket/48957

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

Linux raspberrypi 5.10.52-v7+ #1441 SMP Tue Aug 3 18:10:09 BST 2021 armv7l GNU/Linux

```
> port -v installed ncurses
  ncurses @6.3_0 (active) requested_variants='' platform='linux 5' archs='arm64' date='2022-03-21T21:24:55+0000'
```

###### Verification 
Have you

- [ ] tried a full install with `sudo port -vst install`? (trace mode isn't supported on Linux)

### DIFF
--- a/devel/ncurses/Portfile
+++ b/devel/ncurses/Portfile
@@ -41,8 +41,12 @@ configure.args  --enable-widec \
                 --with-manpage-format=normal \
                 --without-pkg-config \
                 --enable-pc-files \
-                --with-pkg-config-libdir="${prefix}/lib/pkgconfig" \
+                --with-pkg-config-libdir="${prefix}/lib/pkgconfig"
+
+if {${os.platform} eq "darwin"} {
+    configure.args-append \
                 --disable-mixed-case
+}
 
 if {[string match *clang* ${configure.cxx}]} {
     configure.env-append    CXXLIBS=-stdlib=${configure.cxx_stdlib}


devel/ncurses/Portfile
```diff
@@ -44,7 +44,11 @@ configure.args  --enable-widec \
                 --with-pkg-config-libdir="${prefix}/lib/pkgconfig" \
                 --disable-mixed-case
 
-if {[string match *clang* ${configure.cxx}]} {
+platform linux {
+    configure.args-delete --disable-mixed-case
```
ryandesign, line 48: This configure argument's name and description are confusingly misleading:
```
    --enable-mixed-case
        Controls whether the filesystem on which the terminfo database resides
        supports mixed-case filenames (normal for UNIX, but not on other
        systems).  If you do not specify this option, the configure script
        checks the current filesystem.
```
All Mac filesystems support mixed-case filenames. For example, you can have a file named "Portfile", which is composed of an uppercase letter followed by lowercase letters—in other words, mixed case.

What this configure test actually means is: is the filesystem case sensitive? If not specified, it checks at configure time. Our build servers have case-sensitive filesystems, but most of the users installing our binary archives do not, so we specify `--disable-mixed-case` (which should be named `--disable-case-sensitive`) to tell it to assume the filesystem is not case-sensitive even though it is.

Does `--disable-mixed-case` cause some problem on Linux? Since as far as I know it does not cause a problem on case-sensitive Mac systems I would not have expected it to cause a problem on case-sensitive other UNIX systems. If it did cause a problem, depending on what problem it caused, maybe your change here is correct, or maybe it is correct to omit this flag on all non-Darwin systems.

devel/ncurses/Portfile
```diff
@@ -44,7 +44,11 @@ configure.args  --enable-widec \
...
+}
+
+if {${os.platform} eq "darwin" && [string match *clang* ${configure.cxx}]} {
```
ryandesign, line 51: This idiom is used in tons of ports so if it needs to be changed here it needs to be changed in lots of places. However, why would it need to be changed? Why wouldn't we want to specify what C++ standard library to use on non-macOS systems? If the problem is that MacPorts base isn't setting `configure.cxx_stdlib` (or not setting it correctly) on non-macOS systems, let's fix that in base instead.

RJVB:
Well, the argument is misleading on Darwin but not on the traditional systems where this software was developed... but that's not really the point here.

My own version (on which this PR is based, apparently) uses the reverse logic to add the argument on Darwin, instead of adding it first and then removing it on Linux. I think that's more in line with what Ryan proposes: only use it on Darwin (I use "Lamarck" in some places to refer to anything that's not Darwin ;) )

My reasoning is that it's better to disable case-sensitive terminfo databases only if absolutely necessary. I don't know who specifies the name under which entries are saved; if those are specified by the info provider one could run into name clashes with effects that are going to be hard to triage.
This is my general approach for all Linux versions of the ports that I install; the default goal is to get maximum functionality, not some common denominator minimal feature subset.

devel/ncurses/Portfile
harens: Thanks for the detailed explanation @ryandesign.

For some reason, I can no longer reproduce whatever error this fixed. I don't remember changing anything...

If this fix was at some point needed on both @RJVB's and my Linux machine, maybe it is still best to omit this flag on all non-Darwin systems? I'll push a commit updating this.

devel/ncurses/Portfile
harens: > If the problem is that MacPorts base isn't setting configure.cxx_stdlib (or not setting it correctly) on non-macOS systems, let's fix that in base instead.

The issue was the following: ``clang: error: invalid library name in argument '-stdlib='``.

You're absolutely right that MacPorts base just isn't setting `configure.cxx_stdlib` correctly. As a result, the build completes if I set `cxx_stdlib libstdc++` in `macports.conf`. I should be able to send a PR to fix this at some point.

jmroot:
The lowercase terminfo directories work fine everywhere; no functionality is lost. So there's really no reason to change this.

RJVB:
>You're absolutely right that MacPorts base just isn't setting `configure.cxx_stdlib` correctly.

If you look at my macstrop repo you'll find a version of the MacPorts-devel port that has a number of linux-specific patches that are required for proper functioning.
NB: evidently you cannot install this port through `port install`. However, you can go up to and including `port destroot MacPorts-devel`, and take it from there. Formerly I used `porg` to for the actual (manual) install step. Nowadays I maintain a patch the brings back `port dpkg` and use dpkg to do the install. FWIW, that's what I do on Mac too, to make it easier to back out of a botched upgrade (or hope thereof).

RJVB:
>The lowercase terminfo directories work fine everywhere; no functionality is lost. 

Remember that absence of proof of A is no proof of ~A! So what's your proof and why do you think they maintain a flag for this?

PR #14331, opened 2022-03-22
postgresql96-server: Bugfix startupitem

essandess:
* Bugfix startupitem with `sudo`, not `su`
* Use tcl commands in Portfile rather than `system` calls

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.3 21E230 x86_64
Xcode 13.3 13E113

###### Verification 
Have you

### DIFF
--- a/databases/postgresql96-server/Portfile
+++ b/databases/postgresql96-server/Portfile
@@ -4,6 +4,7 @@ PortSystem 1.0
 
 name                postgresql96-server
 version             9.6.24
+revision            1
 categories          databases
 maintainers         {jwa @jyrkiwahlstedt}
 license             Permissive
@@ -30,7 +31,8 @@ add_users           ${dbuser} group=${dbgrp} \
                     home=${prefix}/var/db/${rname} \
                     realname=PostgreSQL-96\ Server
 
-depends_run         port:${rname}
+depends_run         port:bash \
+                    port:${rname}
 
 use_configure       no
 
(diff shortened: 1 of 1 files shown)


ryandesign:
Do these issues only affect postgresql96-server or also the other older and newer postgresqlXX-server ports? If the latter, please make the same changes in those other ports as well, so as to keep the several ports as similar as possible.

essandess:
> Do these issues only affect postgresql96-server or also the other older and newer postgresqlXX-server ports? If the latter, please make the same changes in those other ports as well, so as to keep the several ports as similar as possible.

This issue is a consequence of removing the shell from user `postgres` in #13463. It doesn’t affect the older postgresqlXX-server ports, although those should still be updated in some way to adopt the approach of this port.

This PR should be merged now to fix the now broken `postgresql96-server` port and someone with more knowledge of the older ports can tackle the separate issue of updating those.

mascguy (approved):

mascguy:
> > Do these issues only affect postgresql96-server or also the other older and newer postgresqlXX-server ports? If the latter, please make the same changes in those other ports as well, so as to keep the several ports as similar as possible.
> 
> This issue is a consequence of removing the shell from user `postgres` in #13463. It doesn’t affect the older postgresqlXX-server ports, although those should still be updated in some way to adopt the approach of this port.
> 
> This PR should be merged now to fix the now broken `postgresql96-server` port and someone with more knowledge of the older ports can tackle the separate issue of updating those.

Ryan, any objections to merging this as-is for now, per Steve's comment?

mascguy:
Let's merge, given that we need the fix.

If necessary, I can help accelerate - and/or collaborate with folks - to incorporate similar changes beyond this.

stronk7:
Hi,

I've faced exactly this very same problem with a macports re-installed from scratch (big-sur => mojave upgrade) and `postgresql13-server` (the 12-server Portfile also seems affected). It was working ok previously.

By default ´sudo port load postgresql13-server´does nothing (and nothing is written to log file).

And only digging into the wrapper command and changing the `su postgres -c` by `sudo -u postgres /opt/local/bin/bash -c` caused the load/unload to start working.

So I'd say that both pg12 and pg13 are, by default, affected by the same issue fixed here (haven't tried others).

Ciao :-)

PR #14352, opened 2022-03-23
mpich: Fix gcc builds; add gcc11

eborisch:
Maintainer update.

See also https://github.com/macports/macports-ports/pull/14348

@mascguy

### DIFF
--- a/_resources/port1.0/group/mpi-1.0.tcl
+++ b/_resources/port1.0/group/mpi-1.0.tcl
@@ -327,9 +327,6 @@ proc mpi.setup {args} {
         # So only enable for 10.8+.
         if {${os.major} < 12} {
             lappend ::mpi.disabled_compilers -gcc9 -gcc10 -gcc11
-        } elseif {$is_mpich} {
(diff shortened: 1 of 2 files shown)


mascguy:
Hey Eric, I think (?) gcc11 was removed during last year's cleanup/pruning effort, to reduce the number of variations we have to support. (Have to go back and review the history, but that's my recollection.)

I'm open to resurrecting it though, if you feel it's beneficial.

Your thoughts?

eborisch:
I guess I lean towards pruning old and following current versions myself; we already use gcc11 for the Fortran support for the clang flavors, so it seemed odd to not have a full gcc11 version. I don’t personally use those, so I can go either way. (Perhaps no one does and the gcc versions can be pruned completely?)

I’ve also considered revisiting if the library still truly needs to be built per-compiler, or if we could build the library once and modify the wrappers to call different compilers for for mpicc, etc. for the different versions.

mascguy:
Sounds great Eric, let's merge then!

mascguy (approved):

PR #14372, opened 2022-03-26
helm: add arm64 architecture

wyuenho:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3
Xcode 13.3

###### Verification 
Have you

### DIFF
--- a/sysutils/helm/Portfile
+++ b/sysutils/helm/Portfile
@@ -19,28 +19,49 @@ set latestVersion       helm-3.8
 subport helm-3.8 {
     set baseVersion     3.8
     set patchNumber     1
-    revision            0
-    checksums           rmd160  d295fd8debbc4793b510490419e271275b3dfab8 \
+    revision            1
+
+    if {${build_arch} eq "arm"} {
+        checksums       rmd160  bb47265d1b1275cc895f680be65acd06c394e88d \
+                        sha256  5f0fea586781fb867b92c10133786949ab6a447f297d5c12e1e8f5dd3a9ed712 \
+                        size    14082964
+    } else {
+        checksums       rmd160  d295fd8debbc4793b510490419e271275b3dfab8 \
                         sha256  3b6d87d360a51bf0f2344edd54e3580a8e8de2c4a4fd92eccef3e811f7e81bb3 \
                         size    14318363
+    }
 }
 
 subport helm-3.7 {
     set baseVersion     3.7
     set patchNumber     2
-    revision            0
-    checksums           rmd160  51cfcf1a0bffeba102afffe8333bebd875a11e94 \
+    revision            1
+
+    if {${build_arch} eq "arm"} {
+        checksums       rmd160  903cf44fa466a1901aaa21a1b40c595041d153c6 \
+                        sha256  260d4b8bffcebc6562ea344dfe88efe252cf9511dd6da3cccebf783773d42aec \
+                        size    13978034
+    } else {
+        checksums       rmd160  51cfcf1a0bffeba102afffe8333bebd875a11e94 \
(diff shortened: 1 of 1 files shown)


sysutils/helm/Portfile
```diff
@@ -183,7 +204,6 @@ if {${subport} == ${name}} {
 } else {
     PortGroup           github 1.0
 
-    supported_archs     x86_64
```
herbygillot, line 186: There are other archs besides just `arm64` and `x86_64`, but per the if statement you have below, those are the only two this will work on as a result of this change.  Therefore `supported_archs` should list just those two.

```
supported_archs     arm64 x86_64
```

sysutils/helm/Portfile
```diff
@@ -19,28 +19,49 @@ set latestVersion       helm-3.8
 subport helm-3.8 {
     set baseVersion     3.8
     set patchNumber     1
-    revision            0
-    checksums           rmd160  d295fd8debbc4793b510490419e271275b3dfab8 \
+    revision            1
+
+    if {${os.arch} eq "arm"} {
```
herbygillot, line 24: If I recall correctly, I think the guidance is to prefer `${build_arch}`:

```suggestion
    if {${build_arch} eq "arm64"} {
```

wyuenho: Where's that guidance? Nobody updates the wiki anymore I've lost track on what's the latest.

Also, this port is just downloading a prebuilt binary, it's not building on the system. The Portfile doesn't even support building on the system.

herbygillot: `build_arch` is the explicit configuration option exposed in the MacPorts configuration files. MacPorts implicitly sets it, and it can also be set by the user: https://guide.macports.org/chunked/internals.configuration-files.html

It is set even if the port does not have an explicit build phase.  The port for `terraform` is also simply downloading a pre-built binary, and uses it in exactly this way: https://github.com/macports/macports-ports/blob/df134795909778139be6702f3fa5397084384639/sysutils/terraform/Portfile#L146

lbschenkel (approved):

lbschenkel:
I believe that strictly speaking the port revision didn't need to be bumped, but that is not a big problem. Thank you.

sysutils/helm/Portfile
herbygillot: Sorry to nitpick, but `${build_arch}` has the same values as `supported_archs`.

Meaning you need to match for `arm64`, not `arm`.

herbygillot:
So this got merged before I could comment, but this is not working correctly since `${build_arch}` needs to match against `arm64`, not `arm`.

lbschenkel:
Ooops, sorry, I got misled by the green build.

lbschenkel:
Since it was my fault, I can push a fix to master. Are you OK with that?

wyuenho:
Yes please

wyuenho:
Sorry I should have fixed the value too, I got distracted.

lbschenkel:
Fixed in https://github.com/macports/macports-ports/commit/7e0315ecf247f6d1fd0d29f12118cf5d4411cf08, sorry about jumping the gun.

herbygillot:
@lbschenkel @wyuenho 

Had to further tweak the logic in https://github.com/macports/macports-ports/commit/d00d247b3af7fcad544e2cb5a5901a3f991170a2

In the original logic, the arm64 distname was being indiscriminately set even for Helm versions that _did not_ have an arm64 binary available.

The commit I added changes the logic to only set the arm64 binary if the `build_arch` is arm64 **AND** the given subport has an arm64 binary available.

wyuenho:
Ah thank you!

PR #14386, opened 2022-03-28
privoxy: Bugfix, add privoxy-pki-bundle subport that depends on other PKI bundles

essandess:
* Fixes: https://trac.macports.org/ticket/64892

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3 21E230 x86_64
Xcode 13.3 13E113

###### Verification 
Have you

### DIFF
--- a/www/privoxy/Portfile
+++ b/www/privoxy/Portfile
@@ -1,12 +1,10 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           gpg_verify 1.0
-PortGroup           perl5 1.0
 
 name                privoxy
 version             3.0.33
-revision            2
+revision            3
 categories          www security net
 platforms           darwin
 license             GPL-2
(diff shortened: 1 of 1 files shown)


essandess:
Note: I don't believe that the new port `privoxy-pki-bundle` should not be built by a buildbot and provided by MacPorts so that the CAs are guaranteed to be as recent as the latest ones provided by the dependent ports. How should this be reflected in the Portfile?

mascguy:
Steve, I need to test locally with these changes, once I can take a few minutes away from work.

In the interim, if you're interested in cleaning up the path copy-pasta too, I took a stab at it here. (I only ran a quick test, but seems to work properly.) Feel free to apply my changes to this PR, if you're comfortable with that.

https://github.com/mascguy/macports-ports/tree/mascguy-privoxy-cleanup

essandess:
@mascguy Thank you very much for the suggestions. (I imagined you had a different issue in mind with all the temp file redirections.)

Honestly, I prefer the full references in the commands to be able to follow along directly, rather than cross checking against a list of indirect references in variables, as well not clearly indicating what's in `destroot` and what's in the final install. Therefore, my preference is for the implementation in the current PR.

mascguy:
> Honestly, I prefer the full references in the commands to be able to follow along directly, rather than cross checking against a list of indirect references in variables, as well not clearly indicating what's in `destroot` and what's in the final install. Therefore, my preference is for the implementation in the current PR.

No worries, I'm fine with that.

I'll merge once I've had a chance to test the fix locally. (Hopefully during lunch today.)

mascguy:
> Steve, activation still fails, at least when upgrading.

Ah, just noticed your comment in the ticket, stating that forced activation is still required.

I'm guessing this isn't a new scenario though. Have you checked the Portfile Recipes page on our wiki? (Haven't re-reviewed it yet, but wouldn't be surprised if this is covered...)

mascguy (approved):

PR #14402, opened 2022-03-29
vlang: new port

harens:
#### Description

A port for the [V Programming Language](https://vlang.io/).

The main complexities were the following:

- The compiler for V has to itself be built. This is fetched from [here](https://github.com/vlang/vc) as a separate distfile.
- Vlang won't run [unless its `tools` directory has write permission](https://github.com/vlang/v/issues/10324). This is since it compiles the subcommands locally during runtime. To get around this, rather than running everything with sudo, the user can add themselves to the `_developer` group.
- `/tmp/v` also needs to be writable during runtime. As well as this, if the directory already exists during buildtime (e.g. upgrading the port from source), [the build will fail](https://github.com/vlang/v/issues/7713) if Vlang can't write to it. The same process as above was carried out to fix this.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.2.1 21D62 arm64
Command Line Tools 21D62

###### Verification 
Have you

### DIFF
--- a/lang/vlang/Portfile
+++ b/lang/vlang/Portfile
@@ -1,0 +1,103 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        vlang v 0.2.4
+name                vlang
+github.tarball_from archive
+revision            0
+
+description         Simple, fast, safe, compiled language for developing maintainable software
+long_description    {*}${description}. Compiles itself in <1s with zero library dependencies.
+
+license             MIT
+categories          lang
+maintainers         {harens @harens} openmaintainer
+
+set main_distfile   ${distfiles}
+
+# Install the compiler for vlang as a separate distfile.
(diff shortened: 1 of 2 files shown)


herbygillot:
Is there not like a single constant somewhere in `vlang`'s source code pointing to `/tmp/v` that can be patched? Would ideally be something like `${prefix}/var/run/vlang` or the like.

mascguy (approved):

harens:
> Is there not like a single constant somewhere in vlang's source code pointing to /tmp/v that can be patched? Would ideally be something like ${prefix}/var/run/vlang or the like.

That's a really good idea. After some playing around, it seems like the latest weekly build supports this but the stable build doesn't (at least not trivially).

From @ryandesign https://github.com/macports/macports-ports/commit/0e3cdd414bd67273d9bddc59e7bb7d1475289c84#r69890568:

> Why would MacPorts create /tmp/v?
>
> /tmp is cleaned up after 3 days or upon system restart.

I didn't realise this. This whole `/tmp` thing with stable Vlang doesn't really make much sense then. I'll send a PR to switch to the weekly builds, where this issue can be patched. This also aligns more with the [upstream developers' desires](https://github.com/vlang/v/discussions/12140#discussioncomment-1670716) of installing V as close to bleeding edge as possible.

PR #14438, opened 2022-04-01
7zip: new port (version 21.07)

srirangav:
#### Description

New port for 7zip (version 21.07)

###### Type(s)

- [ ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.4 20G417 arm64
Xcode 12.5.1 12E507

###### Verification 
Have you

### DIFF
--- a/archivers/7zip/Portfile
+++ b/archivers/7zip/Portfile
@@ -1,0 +1,65 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           makefile 1.0
+PortGroup           clang_dependency 1.0
+
+name                7zip
+version             21.07
+revision            0
+
+categories          archivers
+license             LGPL-2.1
+maintainers         {@srirangav} openmaintainer
+
+description         ${name} is a file archiver with a high compression ratio.
+long_description    ${name} is a file archiver that has a high compression \
+                    ratio in 7z format with LZMA and LZMA2 compression, \
+                    supports packing and unpacking XZ, BZIP2, GZIP, TAR, \
+                    ZIP and WIM archives, and unpacking AR, ARJ, CAB, CHM, \
+                    CPIO, CramFS, DMG, EXT, FAT, GPT, HFS, IHEX, ISO, LZH, \
+                    LZMA, MBR, MSI, NSIS, NTFS, QCOW2, RAR, RPM, SquashFS, \
+                    UDF, UEFI, VDI, VHD, VHDX, VMDK, WIM, XAR and Z archives.
+
+homepage            https://www.7-zip.org/
+master_sites        ${homepage}/a/
+
+supported_archs     arm64 x86_64
+
+use_xz              yes
+distname            7z2107-src
+
+checksums           rmd160  f61c0fd91ecd9b9bad37d92bf2a871371886718b \
+                    sha256  213d594407cb8efcba36610b152ca4921eda14163310b43903d13e68313e1e39 \
+                    size    1230388
+
+use_configure       no
+build.cmd           make
+
+patch.dir           ${workpath}
+patchfiles          patch-7zip_gcc_mak.diff
+      
+if {${build_arch} eq "x86_64"} {
+    build {
+        system -W ${workpath}/CPP/7zip/Bundles/Alone2 \
+        "${build.cmd} -j -f ../../cmpl_mac_x64.mak"
+    }
+    destroot {
+        xinstall -m 755 -d ${destroot}${prefix}/bin
+        xinstall -m 755 \
+                 -s ${workpath}/CPP/7zip/Bundles/Alone2/b/m_x64/7zz \
+                 ${destroot}${prefix}/bin
+    }
+} elseif {${build_arch} eq "arm64"} {
+    build {
+        system -W ${workpath}/CPP/7zip/Bundles/Alone2 \
+        "${build.cmd} -j -f ../../cmpl_mac_arm64.mak"
+    }
+    destroot {
+        xinstall -m 755 -d ${destroot}${prefix}/bin
+        xinstall -m 755 \
+                 -s ${workpath}/CPP/7zip/Bundles/Alone2/b/m_arm64/7zz \
+                 ${destroot}${prefix}/bin
+    }
+}
+
--- a/archivers/7zip/files/patch-7zip_gcc_mak.diff
+++ b/archivers/7zip/files/patch-7zip_gcc_mak.diff
@@ -1,0 +1,11 @@
+--- CPP/7zip/7zip_gcc.mak.orig	2022-04-02 21:21:09.000000000 -0700
++++ CPP/7zip/7zip_gcc.mak	2022-04-02 21:21:39.000000000 -0700
+@@ -18,7 +18,7 @@
+ 
+ 
+ ifneq ($(CC), xlc)
+-CFLAGS_WARN_WALL = -Wall -Werror -Wextra
++CFLAGS_WARN_WALL = -Wall -Wextra
+ endif
+ 
+ # for object file


mascguy:
Sriranga, we presently have a port named `p7zip`, which I think (?) is the official port of 7-Zip for Linux/macOS. It's even called out on the main 7-Zip page:

> 7-Zip works in Windows 10 / 8 / 7 / Vista / XP / 2019 / 2016 / 2012 / 2008 / 2003 / 2000.
> 
> [p7zip](https://sourceforge.net/projects/p7zip/) - the port of the command line version of 7-Zip to Linux/Posix.

Is `p7zip` behind 7-Zip, and/or missing functionality reality to the former? (I'm not opposed to your new port, but want to be sure you're aware of what we already have.)

srirangav:
> Is `p7zip` behind 7-Zip, and/or missing functionality reality to the former? (I'm not opposed to your new port, but want to be sure you're aware of what we already have.)

Hi @mascguy,

I followed the link on the 7z website to the source forge site for portable 7zip (p7zip) and it seems to be stuck at a version 16.02 (see https://sourceforge.net/projects/p7zip/files/p7zip/, showing 16.02 as the latest version, which matches the p7zip Portfile).

The main 7zip has advanced to version 21.x, which is why I submitted this new port.  

Also, according to the support page for p7zip, it looks like there are several security issues that remain unresolved in p7zip that have been fixed in upstream 7zip:

https://sourceforge.net/p/p7zip/support-requests/14/
https://sourceforge.net/p/p7zip/support-requests/10/

While some/most of these might be fixed through patches macports includes,  I thought it still might be better to have a port of the official sources for 7z, since they now support macos.

Thanks,

-ranga

mascguy:
Sounds good, thanks for clarifying!

mascguy:
This port fails to build, with errors related to compiler arguments:

```
:info:build error: unknown warning option '-Wno-extra-semi-stmt'; did you mean '-Wno-extra-semi'? [-Werror,-Wunknown-warning-option]
:info:build error: unknown warning option '-Wno-atomic-implicit-seq-cst' [-Werror,-Wunknown-warning-option]
[...etc...]
```

That's why the CI builds jobs are failing, for the two platforms covered (10.15 and 11.x). I've also confirmed locally on macOS 10.13, and I'm seeing the same failures.

mascguy:
So it looks like the project's makefiles are trying to enable some of the newer (newest?) Clang options, with the assumption about the version. So you'll need to patch those, if necessary, to disable that behavior.

Haven't dug too deeply into it, but was able to fix the build via the following patch:

```
--- CPP/7zip/warn_clang_mac.mak.orig	2022-04-02 10:19:27.000000000 -0400
+++ CPP/7zip/warn_clang_mac.mak	2022-04-02 10:19:38.000000000 -0400
@@ -17,23 +17,10 @@
   -Wno-cast-qual \
   -Wno-padded \
   -Wno-exit-time-destructors \
   -Wno-weak-vtables \
 
-CFLAGS_WARN_CLANG_12= $(CFLAGS_WARN_CLANG_3_8) \
-  -Wno-extra-semi-stmt \
-  -Wno-zero-as-null-pointer-constant \
-  -Wno-deprecated-dynamic-exception-spec \
-  -Wno-c++98-compat-pedantic \
-  -Wno-atomic-implicit-seq-cst \
-  -Wconversion \
-  -Wno-sign-conversion \
-  -Wno-suggest-override \
-  -Wno-suggest-destructor-override \
-
-CFLAGS_WARN_MAC = \
-  -Wno-poison-system-directories \
-  -Wno-c++11-long-long \
-  -Wno-atomic-implicit-seq-cst \
+CFLAGS_WARN_CLANG_12=$(CFLAGS_WARN_CLANG_3_8)
 
+CFLAGS_WARN_MAC=
 
 CFLAGS_WARN = $(CFLAGS_WARN_CLANG_12) $(CFLAGS_WARN_MAC)
```

ryandesign:
I would rather say the problem is that `-Werror` has been used, and that the fix should be to remove just that.

srirangav:
Hi @mascguy, @ryandesign,

I have updated my commit with a patch that removes -Werror, as suggested by @ryandesign.  I don't have older version of MacOS to try this out on, but hopefully this fixes the problem.

Thanks,

-ranga

mascguy (approved):

mascguy:
> Hi @mascguy, @ryandesign,
> 
> I have updated my commit with a patch that removes -Werror, as suggested by @ryandesign. I don't have older version of MacOS to try this out on, but hopefully this fixes the problem.

Tested on macOS 10.13, and looks good there as well. Thanks for the contribution!

mascguy:
Hmmm, while it built fine on my local macOS 10.13 installation, it fails on the buildbots for 10.13 and earlier. These are the same errors I saw originally, so more is needed.

https://build.macports.org/builders/ports-10.13_x86_64-builder/builds/146642/steps/install-port/logs/stdio

@ryandesign, your thoughts?

l2dy:
Thanks for submitting this port! I didn't notice they (7-Zip) introduced official macOS support last year. The relationship between 7-Zip for Linux/macOS and p7zip are described in the 7-Zip readme.txt.

> Now there are two different ports of 7-Zip for Linux/macOS:
> 
> 1) p7zip - another port of 7-Zip for Linux, made by an independent developer.
>    The latest version of p7zip now is 16.02, and that p7zip 16.02 is outdated now.
> 
> 2) 7-Zip for Linux/macOS - this package - it's new code with all changes from latest 7-Zip for Windows.
> 
> These two ports are not identical. 
> Note also that some Linux specific things can be implemented better in p7zip than in new 7-Zip for Linux.

The outdated p7zip 16.02 should be considered insecure unless proven otherwise.

ryandesign:
> @ryandesign, your thoughts?

I agree, it doesn't build.

```
../../../../C/AesOpt.c:362:21: error: assigning to '__m256i' (vector of 4 'long long' values) from incompatible type 'int'
      AVX__WOP_KEY (AVX__AES_DEC, 0)
      ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
```

I'm not familiar with this error.

i0ntempest:
> Thanks for submitting this port! I didn't notice they (7-Zip) introduced official macOS support last year. The relationship between 7-Zip for Linux/macOS and p7zip are described in the 7-Zip readme.txt.
> 
> > Now there are two different ports of 7-Zip for Linux/macOS:
> > 
> > 1. p7zip - another port of 7-Zip for Linux, made by an independent developer.
> >    The latest version of p7zip now is 16.02, and that p7zip 16.02 is outdated now.
> > 2. 7-Zip for Linux/macOS - this package - it's new code with all changes from latest 7-Zip for Windows.
> > 
> > These two ports are not identical.
> > Note also that some Linux specific things can be implemented better in p7zip than in new 7-Zip for Linux.
> 
> The outdated p7zip 16.02 should be considered insecure unless proven otherwise.

There is an additional fork of p7zip at version 17.04 with various improvements, which homebrew uses: https://github.com/jinfeihan57/p7zip
But since we now have the official 7zip v21, is updating p7zip still necessary?

srirangav:
> I agree, it doesn't build.
> 
> ```
> ../../../../C/AesOpt.c:362:21: error: assigning to '__m256i' (vector of 4 'long long' values) from incompatible type 'int'
>       AVX__WOP_KEY (AVX__AES_DEC, 0)
>       ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
> ```
> 
> I'm not familiar with this error.

Hi @mascguy, @ryandesign,

Based on another bug marked as closed in trac, it looks like this is a compiler issue:

https://trac.macports.org/ticket/55800

The fix for that bug was to blacklist mp-clang < 3.9.  I would be happy to add the same lines to this Portfile, however, I don't have an older version of MacOS to try this out.  Let me know and I'll revise accordingly.

Thanks,

-ranga

ryandesign:
> fix for that bug was to blacklist mp-clang < 3.9

and Apple clang < 800. Based on the information in that other ticket, it's fine to just make that change without further testing and then see what happens on the buildbot machines.

srirangav:
Hi @ryandesign, @mascguy,

I have submitted a new PR with the current compiler blacklist provided in libvpx, on the assumption that since the same compiler error was encountered for that port the compiler blacklist currently used for that port should work here.  

Thanks,

-ranga

PR #14514, opened 2022-04-07
clang-*: tweak -stdlib; allow low deployment target

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3.1 21E258 x86_64
Xcode 13.3 13E113

###### Verification 
Have you

### DIFF
--- a/lang/llvm-10/Portfile
+++ b/lang/llvm-10/Portfile
@@ -11,8 +11,8 @@ set llvm_version_no_dot 10
 set clang_executable_version 10
 set lldb_executable_version 10
 name                    llvm-${llvm_version}
-revision                2
-subport                 clang-${llvm_version} { revision 5 }
+revision                3
+subport                 clang-${llvm_version} { revision 6 }
 subport                 lldb-${llvm_version} { revision 1 }
 set suffix              mp-${llvm_version}
 set sub_prefix          ${prefix}/libexec/llvm-${llvm_version}
@@ -478,16 +478,14 @@ post-destroot {
     }
 }
 
-if {${os.major} < 11} {
-    variant emulated_tls description { enable c11/c++11 thread_local_storage support on older systems using emulated-tls } {
-        # use emulated-tls to support thread_local on systems prior to 10.7
-        patchfiles-append      9000-patch-llvm-7.0-support-emulated-tls.diff
-        if {${subport} eq "clang-${llvm_version}"} {
-            patchfiles-append  9000-patch-clang-7.0-support-emulated-tls.diff
-        }
+variant emulated_tls description { enable c11/c++11 thread_local_storage support on older systems using emulated-tls } {
+    # use emulated-tls to support thread_local on systems prior to 10.7
+    patchfiles-append      9000-patch-llvm-7.0-support-emulated-tls.diff
+    if {${subport} eq "clang-${llvm_version}"} {
+        patchfiles-append  9000-patch-clang-7.0-support-emulated-tls.diff
     }
-    default_variants-append +emulated_tls
 }
+default_variants-append +emulated_tls
 
 if {${subport} eq "llvm-${llvm_version}"} {
     variant polly description {Provide the polly polyhedral optimizer} {}
@@ -544,65 +542,56 @@ if {${subport} eq "llvm-${llvm_version}"} {
         default_variants-append +defaultlibcxx
     }
 
-    variant libstdcxx description {-stdlib=macports-libstdc++ searches for MacPorts libstdc++} {
+    variant libstdcxx description {-stdlib=libstdc++_macports searches for MacPorts libstdc++} {
 
         patchfiles-append 9001-macports-libstdcxx.diff
 
         post-patch {
+            if { ${configure.build_arch} eq "arm64" } {
+                # see https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483
+                set gcc_arch    aarch64
+            } else {
+                set gcc_arch    ${configure.build_arch}
+            }
+
             reinplace "s|@@MACPORTS_GCC_INCLUDE_DIR@@|${prefix}/include/gcc/c++|g" \
                 ${worksrcpath}/tools/clang/lib/Frontend/InitHeaderSearch.cpp \
                 ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
-            reinplace "s|@@MACPORTS_HOST_NAME@@|${build_arch}-apple-darwin${os.major}|g" \
+            reinplace "s|@@MACPORTS_HOST_NAME@@|${gcc_arch}-apple-darwin${os.major}|g" \
                 ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
             reinplace "s|@@MACPORTS_libstdc++@@|${prefix}/lib/libgcc/libstdc++.6.dylib|g" \
                 ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
 
-            # GCC was built for ${build_arch}-apple-darwin${os.major}
+            # GCC was built for ${configure.build_arch}-apple-darwin${os.major}
             # if GCC was also built universal, it has an extra include directory
-            # if ${build_arch} is 32-bit, there is an extra 64-bit subdirectory
-            # if ${build_arch} is 64-bit, there is an extra 32-bit subdirectory
-            # MacPorts GCC can not compile across platforms
-            switch ${build_arch} {
+            # if ${configure.build_arch} is 32-bit, there is an extra 64-bit subdirectory
+            # if ${configure.build_arch} is 64-bit, there is an extra 32-bit subdirectory
+            # MacPorts GCC cannot compile across platforms
+            switch ${configure.build_arch} {
                 x86_64 {
-                    set gcc_subdirectory_ppc    ""
-                    set gcc_subdirectory_ppc64  "ppc64"
-                    set gcc_subdirectory_i386   "i386"
-                    set gcc_subdirectory_x86_64 ""
+                    reinplace {s|@@MACPORTS_TEST_32_64@@|getTriple().getArch() == llvm::Triple::x86_64 ? "" : "i386"|g} \
+                        ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
                 }
                 i386 {
-                    set gcc_subdirectory_ppc    ""
-                    set gcc_subdirectory_ppc64  "ppc64"
-                    set gcc_subdirectory_i386   ""
-                    set gcc_subdirectory_x86_64 "x86_64"
+                    reinplace {s|@@MACPORTS_TEST_32_64@@|getTriple().getArch() == llvm::Triple::x86_64 ? "x86_64" : ""|g} \
+                        ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
                 }
                 ppc64 {
-                    set gcc_subdirectory_ppc    "ppc"
-                    set gcc_subdirectory_ppc64  ""
-                    set gcc_subdirectory_i386   ""
-                    set gcc_subdirectory_x86_64 "x86_64"
+                    reinplace {s|@@MACPORTS_TEST_32_64@@|getTriple().getArch() == llvm::Triple::ppc64 ? "" : "ppc"|g} \
+                        ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
                 }
                 ppc {
-                    set gcc_subdirectory_ppc    ""
-                    set gcc_subdirectory_ppc64  "ppc64"
-                    set gcc_subdirectory_i386   ""
-                    set gcc_subdirectory_x86_64 "x86_64"
+                    reinplace {s|@@MACPORTS_TEST_32_64@@|getTriple().getArch() == llvm::Triple::ppc64 ? "ppc64" : ""|g} \
+                        ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
+                }
+                arm64 {
+                    reinplace {s|@@MACPORTS_TEST_32_64@@|""|g} \
+                        ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
                 }
             }
-
-            reinplace "s|@@MACPORTS_GCC_SUBDIRECTORY_x86_64@@|${gcc_subdirectory_x86_64}|g" \
-                ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
-            reinplace "s|@@MACPORTS_GCC_SUBDIRECTORY_i386@@|${gcc_subdirectory_i386}|g" \
-                ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
-            reinplace "s|@@MACPORTS_GCC_SUBDIRECTORY_ppc64@@|${gcc_subdirectory_ppc64}|g" \
-                ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
-            reinplace "s|@@MACPORTS_GCC_SUBDIRECTORY_ppc@@|${gcc_subdirectory_ppc}|g" \
-                ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
         }
     }
-    # at present, only the non-arm archs can use this variant
-    if {${build_arch} in "x86_64 i386 ppc64 ppc"} {
-        default_variants-append +libstdcxx
-    }
+    default_variants-append +libstdcxx
 
     post-patch {
         reinplace "s|@@PREFIX@@|${prefix}|" \
--- a/lang/llvm-10/files/9001-macports-libstdcxx.diff
+++ b/lang/llvm-10/files/9001-macports-libstdcxx.diff
@@ -1,31 +1,100 @@
---- orig/tools/clang/include/clang/Lex/HeaderSearchOptions.h
-+++ new/tools/clang/include/clang/Lex/HeaderSearchOptions.h
-@@ -184,6 +184,9 @@ public:
-   /// Use libc++ instead of the default libstdc++.
-   unsigned UseLibcxx : 1;
+From 4e850191efa6659de85a4ff9be7c896cfb0bb1ec Mon Sep 17 00:00:00 2001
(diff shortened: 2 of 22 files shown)


Gcenx:
While doing this could you also fix the c++ headers ditto command for the ports that got the back-ported c++ headers search patch. The c++ headers for those ports should be installed into /include that fixes building using those compilers when an SDK doesn’t ship c++ headers

kencu:
> While doing this could you also fix the c++ headers ditto command for the ports that got the back-ported c++ headers search patch. The c++ headers for those ports should be installed into /include that fixes building using those compilers when an SDK doesn’t ship c++ headers

I am not currently aware of any incorrect location for the c++ headers installed by clang, so careful here ....

Gcenx:
> I am not currently aware of any incorrect location for the c++ headers installed by clang, so careful here ....

After the c++ headers patch was back-ported clang expects to find c++ headers within the MacOSX.SDK/include/c++ then will fallback to it’s own shipped headers that it looks for within /include e.g. (llvm-10/include/c++) else it won’t find c++ headers

I’d experienced this problem while tweaking the clang-11-bootstrap port it couldn’t build any c++ project until I’d copied the headers into /include.

kencu:
I don't know about the clang11-bootstrap port exactly, but we carefully did this in all the clang ports:


https://github.com/macports/macports-ports/blob/038aebdfd61f82321497d1025fee8adfa7dd4fd6/lang/llvm-11/Portfile#L472

I'm sorry, I'm a bit slow this morning. What is wrong exactly with this? It has worked for years and years and years like this....

kencu:
clang is set to look in that location for it's headers...

anyway, this is off topic for @MarcusCalhoun-Lopez 's thing here, but what you're describing is not a trivial thing, and needs it's own full separate discussion I think

MarcusCalhoun-Lopez:
> While doing this could you also fix the c++ headers ditto command for the ports that got the back-ported c++ headers search patch. The c++ headers for those ports should be installed into /include that fixes building using those compilers when an SDK doesn’t ship c++ headers

I am afraid I do not understand the problem well enough to look into it.

Gcenx:
> clang is set to look in that location for it's headers...
> 
> anyway, this is off topic for @MarcusCalhoun-Lopez 's thing here, but what you're describing is not a trivial thing, and needs it's own full separate discussion I think

This is where clang looks for its internal  headers 
`CHECK-LIBCXX-TOOLCHAIN-1: "-internal-isystem" "[[TOOLCHAIN]]/usr/bin/../include/c++/v1"`

I’m assuming that when Macports is setting TOOLCHAIN env that the Xcode shipped c++ headers are getting used over the clang shipped ones.

clang-11-bootstrap wipes all env options lightly why it only seemed to happen with that.

kencu:
> > clang is set to look in that location for it's headers...
> > anyway, this is off topic for @MarcusCalhoun-Lopez 's thing here, but what you're describing is not a trivial thing, and needs it's own full separate discussion I think
> 
> This is where clang looks for its internal headers `CHECK-LIBCXX-TOOLCHAIN-1: "-internal-isystem" "[[TOOLCHAIN]]/usr/bin/../include/c++/v1"`
> 
> I’m assuming that when Macports is setting TOOLCHAIN env that the Xcode shipped c++ headers are getting used over the clang shipped ones.
> 
> clang-11-bootstrap wipes all env options lightly why it only seemed to happen with that.

macports-clang has alwsys found the right headers for me, and I believe for all builds.

I still don’t see a problem to fix here, except maybe in clang11-bootstrap…haven’t seen the issue there as yet to be able to comment.

MarcusCalhoun-Lopez:
> I originally turned the emulated-tls support off on newer systems because the libc++abi.dylib we are using to support TLS on systems < 10.7 has been modified to add the function `cxa_thread_atexit`.
> 
> https://github.com/macports/macports-ports/blob/9e41d7ef209888539b971eb8050ce550556eeb3d/lang/libcxx/Portfile#L141
> 
> This function is an exit function for thread cleanup and does not exist anywhere on 10.7+.
> 
> So there will be cases where trying to build something on 10.7 using emulated-tls with a deployment target of < 10.7 will fail to find the cxa_thread_atexit function.
> 
> The emulated_tls patch for clang modifies clang to look for this function on older systems:
> 
> https://github.com/macports/macports-ports/blob/9e41d7ef209888539b971eb8050ce550556eeb3d/lang/llvm-9.0/files/9000-patch-clang-7.0-support-emulated-tls.diff#L23
> 
> So this will work sometimes, if there happens to be no call for `cxa_thread_atexit`.
> 
> Unless that function is provided some other way, weak linked, or --- etc.

Please forgive me if I am missing something obvious, but this is what I have found by looking through code and symbol tables:
* `__cxa_thread_atexit` and `_tlv_atexit` are "[global destructor(s)](https://github.com/llvm/llvm-project/blob/c1a6fe196d8c23b6f614d6d94f313fcb23da7d85/clang/lib/CodeGen/ItaniumCXXABI.cpp#L2520)" that are required to properly deal with thread-local storage (e.g. `thread_local std::vector<int> vec;`)
* 10.6 and lower: `__cxa_thread_atexit` is provided by MacPorts installed libc++
`/opt/local/bin/nm -mo /usr/lib/libc++abi.dylib | grep ___cxa_thread_atexit && echo "yes"`
* 10.7-10.8: `__cxa_thread_atexit` is **not** provided by the system, and the system provides libc++
* 10.9 and above: `__cxa_thread_atexit` is provided by /usr/lib/libSystem.B.dylib
`nm -mo /usr/lib/system/libsystem_c.dylib | grep ___cxa_thread_atexit && echo "yes"`
* all systems: `__cxa_thread_atexit` exists in MacPorts provided libgcc
`/opt/local/bin/nm -mo /opt/local/lib/libgcc/libstdc++.6.dylib | grep ___cxa_thread_atexit && echo "yes"`
* 10.6 and lower: `_tlv_atexit` is **not** provided anywhere
* 10.7 and above: `_tlv_atexit` is provided by /usr/lib/libSystem.B.dylib
`nm -mo /usr/lib/system/libdyld.dylib | grep __tlv_atexit && echo "yes"`
* By default, Clang uses `_tlv_atexit` on [macOS](https://github.com/llvm/llvm-project/blob/c1a6fe196d8c23b6f614d6d94f313fcb23da7d85/clang/lib/CodeGen/ItaniumCXXABI.cpp#L2531)
* The **only** reason Clang uses `__cxa_thread_atexit` on older systems is because [we tell it to](https://github.com/macports/macports-ports/blob/40a33737f805cb60b299068fa69ffbeb5086e59b/lang/llvm-14/files/0019-10.6-and-less-use-emulated-TLS-before-10.7.patch#L39)

Please forgive my ignorance, but isn't this exactly the situation that [legacy-support](https://github.com/macports/macports-legacy-support) was designed for?
Can we bypass the symbol `__cxa_thread_atexit` entirely?
Would the following be a possibility?
* 10.6 and lower: implment `_tlv_atexit` in legacy-support (/opt/local/lib/libMacportsLegacySystem.B.dylib) using the code from [libcxxabi](https://github.com/llvm/llvm-project/blob/main/libcxxabi/src/cxa_thread_atexit.cpp) or [libstdc++](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/libsupc%2B%2B/atexit_thread.cc)
* Have MacPorts Clang **always** use `_tlv_atexit` (which is is the default anyway)

I am afraid I am out of my comfort zone, so there might be important aspects I am missing.

MarcusCalhoun-Lopez:
Perhaps
* 10.6 and lower: implement `_tlv_atexit` in legacy-support (/opt/local/lib/libMacportsLegacySystem.B.dylib) by **searching** for an appropriate atexit function (`dlsym(RTLD_NEXT, "_tlv_atexit")` then `dlsym(RTLD_NEXT, "__cxa_thread_atexit")`)
At least one of them should be found somewhere.

kencu:
You are getting it!

IF you can implement tlv_atexit that would be fantastic! It is hard!

The very best description I have ever read about how it is implemented (assembly code in dyld) and what needs to be done was written by our very own @saagarjha right here:

https://github.com/iains/gcc-darwin-arm64/issues/1

and it is a masterpiece. It is well beyond my skills.

It might be possible to put cxx_thread_atexit somewhere other than libc++abi.dylib, for sure. I put it there because that is where it is on unix. But I am (as we all know) just a lowly hacker and I did this way back in 2017 when I knew even less than the little I know now. I am certain it can be improved!

The biggest surprise to me was hearing this:
```
10.9 and above: __cxa_thread_atexit is provided by /usr/lib/libSystem.B.dylib
```
 what the heck is that symbol doing in there??? I had no idea it was there. 

Appreciate all your interest and strides here @MarcusCalhoun-Lopez -- it's amazing what hacking in on this teaches us about the inner workings of the system, libc++, libc++abi, and dyld... I've learned so much! Yet know so little!

saagarjha:
Sorry for jumping into this without much context, but if I'm understanding correctly you want thread-local destructors for C++? Two questions I have for you are: 1. what does C++ code for old systems do without `tlv_atexit`? Surely it must be doing *something*…did Mac OS X just not support destructors running at unload back then?

If you want to go with writing your own version, you can steal the one from LLVM or write your own–it's fundamentally not that difficult to write a mediocre implementation, you just make a thread-safe list and run through it when needed, but you'll run into the limitations their implementation mentions. Not entirely sure what level of fidelity you need here.

MarcusCalhoun-Lopez:
Sorry, when I said implement `_tlv_atexit `, I meant it in the [weakest possible sense](https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering):
```
/* MP support header */
#include "MacportsLegacySupport.h"

/* _tlv_atexit wrap */
#if __MP_LEGACY_SUPPORT__TLV_ATEXIT_WRAP__

#include <stdlib.h>
#include <dlfcn.h>

/* signature from https://opensource.apple.com/source/dyld/dyld-852.2/src/threadLocalVariables.c.auto.html */
void _tlv_atexit(void (*func)(void*), void* objAddr) {
    void (*real__tlv_atexit)(void (*func)(void*), void* objAddr);
    int (*real___cxa_thread_atexit)(void (*dtor)(void*), void*, void*);

    real__tlv_atexit = dlsym(RTLD_NEXT, "_tlv_atexit");
    if (real__tlv_atexit != NULL) {
        /* _tlv_atexit exists in libSystem.B.dylib on 10.7 and later */
        real__tlv_atexit(func, objAddr);
        return;
    }

    real___cxa_thread_atexit = dlsym(RTLD_DEFAULT, "__cxa_thread_atexit");
    if (real___cxa_thread_atexit != NULL) {
        /* __cxa_thread_atexit exists in                                                                                                                                                                                                    
               MacPorts provided libc++ on 10.6 and earlier                                                                                                                                                                                 
               libSystem.B.dylib on 10.8 and later                                                                                                                                                                                          
               MacPorts provides libstdc++                                                                                                                                                                                                  
                                                                                                                                                                                                                                            
           in all cases, the third parameter (dso_handle) seems to be ignored
        */
        real___cxa_thread_atexit(func, objAddr, NULL);
        return;
    }

    exit(EXIT_FAILURE);
}

#endif /*__MP_LEGACY_SUPPORT__TLV_ATEXIT_WRAP__*
```

MarcusCalhoun-Lopez:
I added wrapper functions for `__cxa_thread_atexit` and `_tlv_atexit` in the legacy-support [pull request](https://github.com/macports/macports-legacy-support/pull/47/commits/2190af7b70fc7e44c6756ad22f60366e5ddec762).
So a users can use set `MACOSX_DEPLOYMENT_TARGET` and the resulting binaries can run on older systems as long as they link against the legacy-support library, which will find the appropriate atexit function at runtime.

Fortunately, the Rust bootstrap binaries do not seem to use the atexit symbols.

MarcusCalhoun-Lopez:
By linking the legacy-support libraries (https://github.com/macports/macports-legacy-support/pull/47 and https://github.com/macports/macports-ports/commit/60c085270457527ae1fcc09cda4e1b30630ec2f4), `_tlv_atexit` and `__cxa_thread_atexit` should now be available on all systems.
Might this pull request then move forward?

MarcusCalhoun-Lopez:
> We cannot commit changes that require the devel legacy support port, its only there to make it easier to test new features.
> 
> If you can confirm it works the thing to do is to cut a new release of legacy-support, and then update the primary port to use that version. Only then can the new features be used in general.

I take your point.
However, there is nothing in this pull request that **requires** legacy-support at all.
I was just noting the fact that if the lack of `_tlv_atexit` and/or `__cxa_thread_atexit` do cause any issues, we have the tools to figure out a solution.
Since modifying `macosx_deployment_target` is so rare (and generally not a good idea anyway), it is not yet clear what a general solution might be.
For [building Rust on newer systems](https://github.com/macports/macports-ports/pull/14277), the issue does not arise at all.
Should this pull request be put on hold until legacy-support-devel becomes legacy-support?

cooljeanius:
> You are getting it!
> 
> IF you can implement tlv_atexit that would be fantastic! It is hard!
> 
> The very best description I have ever read about how it is implemented (assembly code in dyld) and what needs to be done was written by our very own @saagarjha right here:
> 
> [iains/gcc-darwin-arm64#1](https://github.com/iains/gcc-darwin-arm64/issues/1)
> 
> and it is a masterpiece. It is well beyond my skills.

Hi, I'm coming here from there; if we get this fixed in MacPorts clang, will it make it easier to get @iains's arm64-gcc issue fixed, too? Wondering because it's one of the things blocking gccgo from working on darwin...

MarcusCalhoun-Lopez:
With https://github.com/macports/macports-ports/commit/efab893916bf78dff1337bc95ff2e706303df43a, `_tlv_atexit` and `__cxa_thread_atexit` are available [on all systems](https://github.com/macports/macports-legacy-support/commit/345091d1a4478c0b74b675cd71e5617808dcee6a).
Are there any objections to merging this PR?

kencu:
Marcus, help me walk through the process exactly.

Let's imagine you are on 10.9, and you are building something with this modified compiler with the target set to 10.6.

So you build it through, and the executable is linked against /usr/lib/libc++.dylib (which re-exports all the needed stuff from libc++abi.dylib). And when the tlv_atexit symbols are looked for, then dyld resolves that.

So you are good to go on 10.7 and up I believe.

Now we move that binary to 10.6. The whole idea is that is should run there. First of all, I presume you tried this, and it works. 

But I don't really understand how that binary knows it has to use /opt/local/lib/macportslegacysupport.dylib to find tlv_atexit. Is it linked against that? Do we need to mandate it being available somehow? Or should macportslegacysupport.a it be linked statically?

MarcusCalhoun-Lopez:
> Marcus, help me walk through the process exactly.
> 
> Let's imagine you are on 10.9, and you are building something with this modified compiler with the target set to 10.6.
> 
> So you build it through, and the executable is linked against /usr/lib/libc++.dylib (which re-exports all the needed stuff from libc++abi.dylib). And when the tlv_atexit symbols are looked for, then dyld resolves that.
> 
> So you are good to go on 10.7 and up I believe.
> 
> Now we move that binary to 10.6. The whole idea is that is should run there. First of all, I presume you tried this, and it works.
> 
> But I don't really understand how that binary knows it has to use /opt/local/lib/macportslegacysupport.dylib to find tlv_atexit. Is it linked against that? Do we need to mandate it being available somehow? Or should macportslegacysupport.a it be linked statically?

You points are well taken.
As a matter of fact, as I was thinking of a proper response, I realized an [infinite recursive symbol search](https://github.com/macports/macports-ports/commit/de0004544a41da14dc4d55ec3739ef82e4a886eb) was possible.
I am sorry about the error.

It seems to me that we don't have enough test cases to form a general solution.
Please correct me if I am wrong, but I am aware of only two efforts on this front.
* Improve the [buildbot](https://trac.macports.org/ticket/60878).
* Build a custom [bootstrap Rust compiler](https://github.com/macports/macports-ports/pull/14277).

For the Rust compiler, neither `_tlv_atexit` nor `__cxa_thread_atexit` seem to be built into the bootstrap compiler
(despite [this code](https://github.com/rust-lang/rust/blob/8d68f2f94f9b5cc49958844c935ee84fd63c87f7/library/std/src/sys/unix/thread_local_dtor.rs#L78)), so there is no current problem.

In the case of the buldbot, I am not sure how to test any ideas.

I originally wrote this pull request to build the bootstrap Rust compiler.
So I guess, we have to ask what other scenarios we want to support?

kencu:
I’m OK with the fixes just getting rust working for now, certainly…. without rust, the os versions are doomed.

Just wanted to be sure we understood what child we made here :)

MarcusCalhoun-Lopez:
> I’m OK with the fixes just getting rust working for now, certainly…. without rust, the os versions are doomed.
> 
> Just wanted to be sure we understood what child we made here :)

You are absolutely right.
Do you think we should mark this pull request as a work in progress until these issues are resolved or proceed despite its shortcomings?

kencu:
what we have to do, perhaps, is add tlv_atexit into our modified libcxxabi, right next to cxa_thread_atexit…then it will always be found.

Does this make logical sense?

kencu:
> > I’m OK with the fixes just getting rust working for now, certainly…. without rust, the os versions are doomed.
> > Just wanted to be sure we understood what child we made here :)
> 
> You are absolutely right. Do you think we should mark this pull request as a work in progress until these issues are resolved or proceed despite its shortcomings?

I find myself constantly battling practicality vs purity. So long as we understand the hiccups, I’m fine with practical most of the time! I have learned here in MacPorts that purity has the best long- term outcomes, tho…

MarcusCalhoun-Lopez:
> what we have to do, perhaps, is add tlv_atexit into our modified libcxxabi, right next to cxa_thread_atexit…then it will always be found.
> 
> Does this make logical sense?

I am not 100% sure `_tlv_atexit ` will always be found.
If you compile a program on a newer machine, `nm -mo a.out | grep atexit` yields
` (undefined) external __tlv_atexit (from libSystem)`

With the two-level namespace being what it is, I am not sure `_tlv_atexit` will be found since the program will only look in libSystem instead of libcxxabi.

kencu:
> With the two-level namespace being what it is, I am not sure `_tlv_atexit` will be found since the program will only look in libSystem instead of libcxxabi.

Right.

I knew that somehow sounded too easy.

MarcusCalhoun-Lopez:
> I find myself constantly battling practicality vs purity. So long as we understand the hiccups, I’m fine with practical most of the time! I have learned here in MacPorts that purity has the best long- term outcomes, tho…

Fair point.
Unless I hear different, perhaps we can merge this pull request later today and then use it as a starting point for handling the atexit functions.

PR #14546, opened 2022-04-10
gcc-devel: improve header search locations

MarcusCalhoun-Lopez:
Setting the local prefix to the MacPorts `${prefix}` does little good
since the sysroot is prepended.
The same reasoning applies to the default value of `/usr/local`.
Therefore, do not set the local prefix.

Since the `-stdlib=` uses the system libc++, it should also use the
corresponding system headers.
By including the sysroot in the path, the resulting compiler
respects the environment variable `SDKROOT`.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3.1 21E258 arm64
Xcode 13.3 13E113

###### Verification 
Have you

### DIFF
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -34,8 +34,8 @@ if {${os.arch} eq "arm"} {
 
     # Version must follow same scheme as with GCC snapshots below <version>-<commit-date>
     version         12-20220327
-    revision        0
-    subport         libgcc-devel { revision 0 }
+    revision        1
+    subport         libgcc-devel { revision 1 }
 
     checksums       rmd160  76551e6d93d2a0a15afca6d66ed36f5f23b92b0a \
                     sha256  a21729570ea54460986e0593bef3729356a06d8a9466a7ce5deca7d02948b57b \
@@ -45,8 +45,8 @@ if {${os.arch} eq "arm"} {
     # Use regular GCC releases and snapshsots
 
     version         12-20220403
-    revision        0
-    subport         libgcc-devel { revision 0 }
+    revision        1
+    subport         libgcc-devel { revision 1 }
 
     checksums       rmd160  7ab9d4c3ed8ada8ca619d21d22a524014b1aee01 \
                     sha256  f534aeab7a9809191130361c9f05e1b83bc8014c9fb833ca73549759221098e0 \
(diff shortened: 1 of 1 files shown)


mascguy:
Makes sense, and seems reasonable. Though I'll defer to folks more seasoned in the toolchain arena.

Gcenx:
This won’t work on macOS Mojave as the SDK doesn’t contain c++ headers and unless the headers packages  is installed there is no /usr/include.

macOS Catalina would also be affected as it’s default SDK also doesn’t contain c++ headers and there’s no headers package in Xcode11 command line tools for Catalina.

For these systems the Xcode c++ headers would need to be used as a fallback.

MarcusCalhoun-Lopez:
Thank you @Gcenx for the feedback.
I was a little careless on this one.
I have modified the code to try to take into account all possibilities.

michaelld:
Seems reasonable & the code makes sense. I'll test tomorrow when I'm at work & also looking into upgrading gcc-devel to 20220410 ... hopefully the ARM64 version has been fixed since a commit broke it 1.5 weeks ago.

cjones051073:
TBH I am not that convinced that, using the Xcode headers, particularly on older systems, is a better idea that using a more up to date set of headers from a recent macports clang. Note, gcc only supports an up to date LLVM release.

cjones051073:
b.t.w. if we do go with something like this, because the same logic will be needed in a number of ports (gcc12 now has the same variant), the logic should be moved to a new stub port `gcc-libcxx-headers` or whatever, that just installs a sym-link to  the correct headers in a standard place in ${prefeix}. The gcc ports can then just depend on this port and use that single location, instead of duplicating the same logic.

Gcenx:
> b.t.w. if we do go with something like this, because the same logic will be needed in a number of ports (gcc12 now has the same variant), the logic should be moved to a new stub port `gcc-libcxx-headers` or whatever, that just installs a sym-link to the correct headers in a standard place in ${prefeix}. The gcc ports can then just depend on this port and use that single location, instead of duplicating the same logic.

Or update then use `macports-libcxx` that provides the llvm libcxx along with the headers

kencu:
Iain tells me that at present none of the libc++ header sets work quite properly with gcc when building against libc++.

He is working on a set of libc++ headers that will work properly, and will let us know when he has them available.

cjones051073:
I am going to close this, as I do not believe it should be deployed. I remain unconvinced that using the Xcode/CLT headers is better than what the ports currently do, which is take a recent macports clang version.

indeed, things are not perfect, I have run into occasional issues trying to use libc++ with g++, however, things are much better with the newer clang headers than the older Xcode ones.

PR #14555, opened 2022-04-11
gcc10-bootstrap: fixup builds

Gcenx:
- remove compact code thats part of base
- add upstream xcode12 patch
- fix universal
- fix up usage notes

#### Description

All builds were +universal installing without issue, macOS Mojave was using `MacOSX10.13.sdk` with deployment target set to 10.13 to enable i386 architecture.

<br>

This PR does **not** make `gcc` and other binaries function more like a normal macports-gcc compiler, this still functions more like a cross-compiler. When using `gcc` or other normally named binaries it will only support a single arch so when launched as i386 `gcc` will compile for i386, if launched as x86_64 `gcc` will compile for x86_64 etc.

###### Type(s)

- [x] bugfix

###### Tested on

macOS 11.6 20G165 arm64
Xcode 13.2.1 13C100

macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505

command line tools install
macOS 10.9.5 13F1911 x86_64

###### Verification 

### DIFF
--- a/lang/gcc10-bootstrap/Portfile
+++ b/lang/gcc10-bootstrap/Portfile
@@ -106,6 +106,10 @@ patchfiles-append patch-darwin21.diff
 # https://trac.macports.org/ticket/63161
 patchfiles-append patch-build-i686.diff
 
+# Bootstrap fails with Clang 12.0.5 (XCode 12.5)
+# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100340
+patchfiles-append patch-xcode12-fix.diff
+
 # sterilize MacPorts build environment; we want nothing picked up from MP prefix
 compiler.cpath
 compiler.library_path
@@ -176,20 +180,25 @@ if {${os.major} >= 18 && ${configure.sdkroot} ne ""} {
     configure.args-append --with-sysroot="[regsub {MacOSX1[0-9]\.[0-9]+\.sdk} ${configure.sdkroot} {MacOSX.sdk}]"
 }
 
-# clang (as) from Xcode 12.5 has various problems with gcc build
-if { ${os.platform} eq "darwin" && \
-         ( [ vercmp ${xcodeversion} 12.5 ] >= 0 || [ vercmp ${cltversion} 12.5 ] >= 0 ) } {
-    pre-configure {
-        ui_warn "Applying '--without-build-config' workaround to Xcode ${xcodeversion} / CLT ${cltversion}"
-        ui_warn "If versions > 12.5 please check if it is still required"
-    }
-    # gcc has build issues on macOS 11.3 with the use of Xcode clang as 'as'
-    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100340
-    # https://trac.macports.org/ticket/62775
-    configure.args-append  --without-build-config
+if {${os.major} < 18 || ${os.major} > 19} {
+    default_variants    +universal
 }
 
-default_variants    +universal
+platform powerpc {
+    configure.universal_archs ppc ppc64
+}
+if { ${os.platform} eq "darwin" && ${os.major} >= 20 } {
+    platform i386 {
+        configure.universal_archs x86_64 arm64
+    }
+    platform arm {
+        configure.universal_archs x86_64 arm64
+    }
+} else {
+    platform i386 {
+        configure.universal_archs i386 x86_64
+    }
+}
 
 platform darwin 8 {
     configure.args-append \
@@ -199,10 +208,6 @@ platform darwin 8 {
 merger_arch_flag            yes
 merger_arch_compiler        yes
 merger_must_run_binaries    yes
-
-if {![info exists universal_possible]} {
-    set universal_possible [expr {${os.universal_supported} && [llength ${configure.universal_archs}] >= 2}]
-}
 
 set merger_host(arm64)  aarch64-apple-${os.platform}${os.major}
 set merger_host(i386)   i386-apple-${os.platform}${os.major}
@@ -216,14 +221,6 @@ foreach {arch target} [array get merger_host] {
 }
 
 if {${universal_possible} && [variant_isset universal]} {
-    configure.args-replace \
-                     --disable-multilib \
-                     --enable-multilib
-
-    configure.args-replace \
-                     --disable-multiarch \
-                     --enable-multiarch
-
     set universal_targets ""
     foreach arch ${configure.universal_archs} {
         set universal_targets "${universal_targets},$merger_host(${arch})"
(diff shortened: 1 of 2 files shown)


mascguy (approved):
LGTM. @catap / anyone?

catap:
@mascguy / @Gcenx I'd love to review and test it, but I need a couple of days.

lang/gcc10-bootstrap/Portfile
```diff
@@ -176,20 +180,25 @@ if {${os.major} >= 18 && ${configure.sdkroot} ne ""} {
...
+        configure.universal_archs x86_64 arm64
+    }
+    platform arm {
+        configure.universal_archs x86_64 arm64
+    }
+} else {
+    platform i386 {
+        configure.universal_archs i386 x86_64
+    }
+}
```
catap, line 201: I'm confusing about this block. Any reason when `configure.universal_archs` may have different value from which you have put here.

lang/gcc10-bootstrap/Portfile
```diff
@@ -200,10 +209,6 @@ merger_arch_flag            yes
 merger_arch_compiler        yes
 merger_must_run_binaries    yes
 
-if {![info exists universal_possible]} {
-    set universal_possible [expr {${os.universal_supported} && [llength ${configure.universal_archs}] >= 2}]
```
catap, line 204: You had removed the onle `set universal_possible` which was in this portfile. This means that condition `if {${universal_possible} ...` is always false.

lang/gcc10-bootstrap/Portfile
```diff
@@ -261,21 +258,17 @@ you must use per target compiler. The easy way is using muniversal PG:
 
 PortGroup               muniversal 1.0
 
-if {![info exists universal_possible]} {
```
catap, line 264: why?

catap (commented):
A few fast remarks

lang/gcc10-bootstrap/Portfile
Gcenx: I’d borrow led this for gcc-devel?

Later versions of macOS will still say there an i386 platform when when they only run x86_64. This would account for this.

lang/gcc10-bootstrap/Portfile
Gcenx: This compact code is already part of Macports-base

lang/gcc10-bootstrap/Portfile
Gcenx: Same as above this compact code is already part of Macports-base

lang/gcc10-bootstrap/Portfile
catap: Got it, thanks

lang/gcc10-bootstrap/Portfile
catap: I had it here before, but for some reason had removed it. Let me try to recall why.

Gcenx: I’ll just blame me for that one idk lol

kencu:
@Gcenx -- I hereby nominate you to field all of @barracuda158 's tickets when this port won't build anything universal for him :>

Gcenx:
@kencu _techinically_ that could be fixed using the poor man’s driverdriver script from Iain.

_I should modify the notes to always recommend using a target compiler even for none universal builds._

But on a more serious note as the intended purpose of this is to help bootstrap to Macports-clang/gcc ports I really don’t want us to put in too much effort making it a replacement for the standard macports provided compiler toolchains.

lang/gcc10-bootstrap/Portfile
```diff
@@ -176,20 +180,25 @@ if {${os.major} >= 18 && ${configure.sdkroot} ne ""} {
...
-    pre-configure {
-        ui_warn "Applying '--without-build-config' workaround to Xcode ${xcodeversion} / CLT ${cltversion}"
-        ui_warn "If versions > 12.5 please check if it is still required"
-    }
-    # gcc has build issues on macOS 11.3 with the use of Xcode clang as 'as'
-    # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100340
-    # https://trac.macports.org/ticket/62775
-    configure.args-append  --without-build-config
+if {${os.major} < 18 || ${os.major} > 19} {
+    default_variants    +universal
```
catap, line 184: Why? does enable `+universal` on `os.major` with value `18` brokes anything?

catap, line 184: Why? does enable `+universal` on `os.major` with value `18` brokes anything?

kencu:
> @kencu _techinically_ that could be fixed using the poor man’s driverdriver script from Iain.
> 
> _I should modify the notes to always recommend using a target compiler even for none universal builds._
> 
> But on a more serious note as the intended purpose of this is to help bootstrap to Macports-clang/gcc ports I really don’t want us to put in too much effort making it a replacement for the standard macports provided compiler toolchains.

If we built this like this, it is no more effort, and probably much easier:

architecture a = i386 (or ppc)
architecture b = x86_64 (or ppc64)

1. build architecture a as multilib
2. build architecture b as multilib
3. delete all the libgcc files from architecture a
4. let muniversal merge the trees (no conflicts now)

then gcc10-bootstrap would work as expected, and the compiler could build either arch a or b transparently without any changes to any portfiles.

The way you have it here, nothing will work without special treatment to everything. Now if "everything" is just clang11-bootstrap, nobody cares.  But for @barracuda153 there, everything is everything, and so he is DOA.

So my vote is to do it like I describe here, obviously, as I've been saying the same thing since the beginning. But you're right, I"m not actually **doing** it, only opining :> If you do it the way you have it here, it is highly likely somebody (ryan or josh or chris or michael or maybe me) will come along later and fix it as I mentioned.

catap:
I agree with @kencu special that builds with current portfile were failed on conflict with merges. It may came from another typo in rebasing, or just something simpler.

I need a bit more time to sort my life and as soon as it ready => will work on it.

kencu:
I feel I should not comment, to be honest. I'm clearly not willing to do the actual work, so I should (and will) just be quiet and let you folks who are willing to the work do it.

I'll stay out of this and let you folks work it out. This method here has advantages for cross-processor building and perhaps it  is time to just let that all work through!

Have at it, and thanks for doing it!

Gcenx:
As far as I’m concerned `gcc10-bootstrap` was only ever going to be used for `clang-11-bootstrap` and doing what you’ve mentioned with libgcc may have other consequences we’re not yet aware (see below)

I believe @barracuda153 had been attempting to use this as a replacement for macports-gcc as those hasn’t been working correctly for them, but I don’t see this being anything other than a bootstrap getting a modern clang on 10.6+.

@catap depending on the arches libgcc will conflict but for some it won’t still the only reasonable ways to make this functional is replacing the standard gcc binaries with wrappers that automatically use the correct cross-compiler for the job. But again _why_ make this function like a macports-gcc when it’s intended purpose was to quickly get to a modern clang?

barracuda156:
> As far as I’m concerned `gcc10-bootstrap` was only ever going to be used for `clang-11-bootstrap` and doing what you’ve mentioned with libgcc may have other consequences we’re not yet aware (see below)
> 
> I believe @barracuda153 had been attempting to use this as a replacement for macports-gcc as those hasn’t been working correctly for them, but I don’t see this being anything other than a bootstrap getting a modern clang on 10.6+.
> 
> @catap depending on the arches libgcc will conflict but for some it won’t still the only reasonable ways to make this functional is replacing the standard gcc binaries with wrappers that automatically use the correct cross-compiler for the job. But again _why_ make this function like a macports-gcc when it’s intended purpose was to quickly get to a modern clang?

@Gcenx I do not think `clang` is the sole use case. `gcc11` cannot be built with default Apple compilers on older systems, and it is very natural to have `gcc10-bootstrap` account for that case. I am not sure there is another way in fact.

catap:
as promised I've just pushed the fix for destroot: https://github.com/catap/macports-ports/commits/gcc10-bootstrap

@Gcenx I've incorporated all your changes and before I open a PR to macports I'd love to ask you to help me test it :) It is very time consuming and before I did it I'd love to test it properly on all available macOS versions which requires a lot of time.

Anyway it should be build and should work as close as possible to "universal compiler".

lang/gcc10-bootstrap/Portfile
catap: after a lot of attempts to remember why I feel that the only reason is attemplt to make port file simpler.

catap: after a lot of attempts to remember why I feel that the only reason is attemplt to make port file simpler.

barracuda156:
@evanmiller @catap So what is the status on Tiger, has `gcc10-bootstrap` been fixed for it or not yet?

catap:
@barracuda156 may I ask you to test https://github.com/catap/macports-ports/commits/gcc10-bootstrap on tiger to which you have an access :)

I plan open a PR later today if all tests are hold

barracuda156:
> @barracuda156 may I ask you to test https://github.com/catap/macports-ports/commits/gcc10-bootstrap on tiger to which you have an access :)
> 
> I plan open a PR later today if all tests are hold

@catap I will, provided the earlier reported error on Tiger has been addressed, at least in principle.

PR #14568, opened 2022-04-12
gdb-apple: fix build on 10.5/10.6

evanmiller:
#### Description

* Split `patch-gdb-availability.diff` into 2 pieces to be more manageable
* Promote the larger piece to be applied unconditionally, testing function-relevant macros instead of MAC_OS_X_VERSION_MIN_REQUIRED (this fixes a build problem on 10.5)
* Fix a duplicate symbol linking error witnessed on 10.6 with clang-11
* Promote `patch-macosx-tdep.c.diff` to be applied unconditionally
* Add `size` to the checksums

CC @kencu and @barracuda156

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode x.y

###### Verification 
Have you

### DIFF
--- a/devel/gdb-apple/Portfile
+++ b/devel/gdb-apple/Portfile
@@ -37,19 +37,21 @@ distname        gdb-${version}
 master_sites    http://opensource.apple.com/tarballs/gdb
 
 checksums       rmd160  90290c3950dd06c1691a477b9ee302c00b681ed1 \
-                sha256  d25b056a826015e1e0ea338b9b8f03e40dce3ff4696faf95df259ae1a42698da
+                sha256  d25b056a826015e1e0ea338b9b8f03e40dce3ff4696faf95df259ae1a42698da \
+                size    18298209
 
 depends_build   port:gettext port:zlib port:flex port:texinfo
 
 depends_lib     port:libiconv port:ncurses port:sqlite3
 
-patchfiles      patch-macosx-nat.c.diff
+patchfiles      patch-gdb-headers.diff \
+                patch-macosx-nat.c.diff \
+                patch-macosx-tdep.c.diff
 
 if {${os.platform} eq "darwin" && ${os.major} < 10} {
     patchfiles-append \
-                patch-macosx-tdep.c.diff \
-                patch-macosx-nat-threads.c.diff \
-                patch-gdb-availability.diff
+                patch-gdb-backtrace.diff \
+                patch-macosx-nat-threads.c.diff
 
     if {${os.major} == 8} {
         depends_build-append port:libmacho-headers
--- a/devel/gdb-apple/files/patch-gdb-backtrace.diff
+++ b/devel/gdb-apple/files/patch-gdb-backtrace.diff
@@ -1,0 +1,59 @@
+Check the deployment target before calling backtrace.
+
+This patch requires patch-gdb-headers.diff.
+
+--- gdb/remote.c
++++ gdb/remote.c
+@@ -64,7 +64,10 @@
+ #include "macosx-nat-dyld.h"
+ #include "macosx-nat-dyld-process.h"
+ #endif
++#ifdef HAVE_EXECINFO_H
+ #include <execinfo.h>
++#endif
++#include <AvailabilityMacros.h>
+ 
+ /* Prototypes for local functions.  */
+ static void cleanup_sigint_signal_handler (void *dummy);
+@@ -478,11 +481,13 @@ struct memory_packet_config
+ static void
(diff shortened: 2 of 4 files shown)


mascguy (approved):
Approving on behalf of Jeremy.

But we can leave this open for three days, to provide others the opportunity to review.

kencu:
looks ok to me too.

barracuda156:
@evanmiller I fully understand that my exotic case is no one's concern, haha, but for the record, new patch fails on 10.6 PPC on this:
```
./macosx/macosx-nat-dyld.c: In function ‘macosx_locate_dyld_via_taskinfo’:
./macosx/macosx-nat-dyld.c:809: error: ‘struct task_dyld_info’ has no member named ‘all_image_info_addr’
./macosx/macosx-nat-dyld.c:810: error: ‘struct task_dyld_info’ has no member named ‘all_image_info_size’
./macosx/macosx-nat-dyld.c:812: error: ‘struct task_dyld_info’ has no member named ‘all_image_info_addr’
./macosx/macosx-nat-dyld.c:869: warning: unused variable ‘dyld_here’
./macosx/macosx-nat-dyld.c: In function ‘macosx_get_osabi_from_dyld_entry’:
./macosx/macosx-nat-dyld.c:1926: warning: passing argument 1 of ‘xfree’ discards qualifiers from pointer target type
./macosx/macosx-nat-dyld.c:1929: warning: passing argument 1 of ‘xfree’ discards qualifiers from pointer target type
./macosx/macosx-nat-dyld.c: In function ‘dyld_fix_path’:
./macosx/macosx-nat-dyld.c:3631: warning: unused variable ‘prefix’
make[1]: *** [macosx-nat-dyld.o] Error 1
```
While the old one was transparent enough to change, the new one is not.

mascguy:
> @evanmiller I fully understand that my exotic case is no one's concern, haha, but for the record, new patch fails on 10.6 PPC on this:
>
> [...]
>
> While the old one was transparent enough to change, the new one is not.

Evan, let me know if there's anything I can to help. (Even if it's just merging the ultimate PR/change.)

kencu:
10.6 PPC is about 1/2 way between 10.6 and 10.5 in SDK features. so things that might be expected to work on 10.6 don’t work. Sometimes you can force a 10.5 path for it, but it’s a frustrating lot of extra work and testing for a system that is exceptionally niche in use and interest. So you really need to be using it, and care…

I was initially intrigued, but then uninterested once all the many warts were made visible. I can’t see anything 10.6 for PPC can do that 10.5 doesn’t do better, so far at least. 

The kludgy workarounds often needed to pathway 10.6 for PPC are IMHO generally too hack for MacPorts.

But sometimes trying build on that system turns up more generic logic errors in base or portfiles, and those are reasonable to fix.

barracuda156:
@kencu The earlier version of @evanmiller patch here had a condition on the OS (>=10500) – which in fact should have been >=10600 – so it was transparent to modify it for the case of 10.6 PPC. New patch made it unnecessarily non-trivial. Anyone else who would want to make `gdb-apple` work for 10.6 PPC now will need to reinvent the bicycle or dig out the old version of the patch.
Does the new conditioning give any benefits? I do not ask anyone to make others' life difficult for the sake of a niche case, but I think it is reasonable not to make niche life more difficult without a good reason.

>I can’t see anything 10.6 for PPC can do that 10.5 doesn’t do better, so far at least.

I have demonstrated a couple of such cases in fact, though, admittedly, nothing tremendously significant.

evanmiller:
> @kencu The earlier version of @evanmiller patch here had a condition on the OS (>=10500) – which in fact should have been >=10600 – so it was transparent to modify it for the case of 10.6 PPC. New patch made it unnecessarily non-trivial. Anyone else who would want to make `gdb-apple` work for 10.6 PPC now will need to reinvent the bicycle or dig out the old version of the patch. Does the new conditioning give any benefits? I do not ask anyone to make others' life difficult for the sake of a niche case, but I think it is reasonable not to make niche life more difficult without a good reason.
> 
> > I can’t see anything 10.6 for PPC can do that 10.5 doesn’t do better, so far at least.
> 
> I have demonstrated a couple of such cases in fact, though, admittedly, nothing tremendously significant.

If you need a pre-processor line to evaluate to false, you can always change it to `#if 0` locally. The new logic tries to condition on the features that the SDK self-reports as available. This is generally less error-prone than conditioning on OS versions and releases, but as you discovered there are exceptions to the rule.

PR #14569, opened 2022-04-12
texlive: update to 2022

drkp:
This updates the texlive ports to TeX Live 2022, which was released last week.

This is not ready for merge yet, but as usual I am posting it here first in case anyone else wants to try it. It is pretty close to complete, however; the main thing left is testing.

Unless any unexpected problems crop up, I will probably merge in the next week.


evanmiller:
Looking through the linter warnings, it looks like many of the new checksums are missing the file `size`.

drkp:
> Looking through the linter warnings, it looks like many of the new checksums are missing the file `size`.

Good call, fixed the generator to add the size checksum too.

I think this should be complete so once I finish a couple more tests I will merge.

PR #14586, opened 2022-04-15
py-filelock: update to 3.6.0

dgilman:
#### Description

Note the epoch bump: this will downgrade any python 3.6 users from filelock 3.4 to 3.2.1. There's little change in filelock so I don't expect this to break any 3.6 users. But at any rate python 3.6 is no longer supported upstream and there's little point in giving it special support in MacPorts.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H1824 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/python/py-filelock/Portfile
+++ b/python/py-filelock/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           python 1.0
 
 name                py-filelock
-version             3.4.0
+version             3.6.0
 revision            0
 
 platforms           darwin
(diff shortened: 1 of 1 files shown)


reneeotten (changes requested):
version 3.4.0 (in fact even 3.4.1 is still supported for Python 3.6). So yes, please do give the PY36 a special treatment here in the sense that you update that one to 3.4.1 and no `epoch` increase please

PR #14631, opened 2022-04-19
clang-11-bootstrap: new port

catap:
#### Description

Should be merged after https://github.com/macports/macports-ports/pull/14624

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

macOS 12.3.1 21E258 x86_64
Xcode 13.3.1 13E500a

macOS 12.0.1 21A559 arm64
Xcode 13.1 13A1030d

###### Verification 
Have you

### DIFF
--- a/lang/clang-11-bootstrap/Portfile
+++ b/lang/clang-11-bootstrap/Portfile
@@ -1,0 +1,282 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           muniversal 1.0
+
+# clang-11 is the last clang which can be build with the last version of cmake to not require c++11.
+name                clang-11-bootstrap
+
+version             11.1.0
+revision            0
+epoch               0
+
+platforms           darwin
+categories          lang
+maintainers         {@catap korins.ky:kirill} openmaintainer
+
+license             NCSA
+
+description         clang-11 with minimal MacPorts patches pegged at version ${version}.
+long_description    ${description} This port depends only on bootstrap dependencies.
+
+homepage            https://llvm.org/
+
+
+master_sites        https://github.com/llvm/llvm-project/releases/download/llvmorg-${version}
+distname            llvm-project-${version}.src
+
+checksums           rmd160  f566b4b75c8f30418f19069a9a84864ead766401 \
+                    sha256  74d2529159fd118c3eac6f90107b5611bccc6f647fdea104024183e8d5e25831 \
+                    size    84065492
+
+# no Clang / LLVM for PowerPC yet
+supported_archs     arm64 i386 x86_64
+
+depends_extract     port:xz-bootstrap
+depends_skip_archcheck-append \
+                    xz-bootstrap
+
+extract.suffix      .tar.xz
+extract.cmd         ${prefix}/libexec/xz-bootstrap/bin/xz
+
+depends_build-append \
+                    port:python27-bootstrap
+depends_skip_archcheck-append \
+                    python27-bootstrap
+
+# clang is linked against gcc's libstdc++.6.dylib and libgcc_s.1.1.dylib
+depends_lib-append  port:gcc10-bootstrap
+
+# Use cmake-bootstrap
+depends_build-replace \
+                    path:bin/cmake:cmake port:cmake-bootstrap
+depends_skip_archcheck-append \
+                    cmake-bootstrap
+configure.cmd       ${prefix}/libexec/cmake-bootstrap/bin/cmake
+
+
+worksrcdir          llvm-project/llvm
+patch.dir           ${workpath}/llvm-project/llvm
+
+post-extract {
+    ln -s ${workpath}/${distname}                       ${workpath}/llvm-project
+    ln -s ${workpath}/llvm-project/clang                ${worksrcpath}/tools/clang
+    ln -s ${workpath}/llvm-project/compiler-rt          ${worksrcpath}/projects/compiler-rt
+    ln -s ${workpath}/llvm-project/libcxx               ${worksrcpath}/projects/libcxx
+    ln -s ${workpath}/llvm-project/libcxxabi            ${worksrcpath}/projects/libcxxabi
+    ln -s ${workpath}/llvm-project/lld                  ${worksrcpath}/projects/lld
+    ln -s ${workpath}/llvm-project/clang-tools-extra    ${worksrcpath}/tools/clang/tools/extra
+}
+
+patch.pre_args      -p1
+patchfiles          0002-Define-EXC_MASK_CRASH-and-MACH_EXCEPTION_CODES-if-th.patch \
+                    0003-MacPorts-Only-Don-t-embed-the-deployment-target-in-t.patch \
+                    0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch \
+                    0006-Only-call-setpriority-PRIO_DARWIN_THREAD-0-PRIO_DARW.patch \
+                    0007-patch-llvm8-tools-dsymutil-symbolmap-use-older-cfname-and-fix-uuid-on-leopard.diff \
+                    0008-patch-lib-support-unix-path-copyfileclone-on-older-systems.diff \
+                    1002-MacPorts-Only-Fix-name-of-scan-view-executable-insid.patch \
+                    1004-Default-to-fragile-ObjC-runtime-when-targeting-darwi.patch \
+                    1005-Fixup-libstdc-header-search-paths-for-older-versions.patch \
+                    1007-Fix-float.h-to-work-on-Snow-Leopard-and-earlier.patch \
+                    1008-compilerrt-fuzzer-missingdefs.diff \
+                    1009-compilerrt-sanitizer-missingdefs.diff \
+                    2002-Work-around-no-libdispatch-on-10.6.patch \
+                    3001-Fix-missing-long-long-math-prototypes-when-using-the.patch \
+                    3002-implement-atomic-using-mutex-lock_guard-for-64b-ops-.patch \
+                    3003-patch-libcxx-chrono-restore-steadyclock-fallback.diff \
+                    leopard-no-asan.patch \
+                    5000-patch-compilerrtdarwinutils-find-macosxsdkversion.diff \
+                    5002-patch-toolchains-darwin-add-back-pre-10.6-link-libs.diff \
+                    patch-clang-fix-include-next-sysroot-cpp-headers.diff \
+                    Fix-build-issues-pre-Lion-due-to-missing-a-strnlen.patch \
+                    fix-build-clang-by-gcc10.diff
+
+# sterilize MacPorts build environment; we want nothing picked up from MP prefix
+compiler.cpath
+compiler.library_path
+
+configure.cxx_stdlib
+
+configure.cflags
+configure.cxxflags
+configure.cppflags
+configure.optflags
+configure.ldflags
+
+configure.universal_cflags
+configure.universal_cxxflags
+configure.universal_cppflags
+configure.universal_ldflags
+configure.universal_args
+
+configure.ccache    no
+configure.distcc    no
+
+# sterilize PATH
+configure.env-append \
+                    PATH=${workpath}/bins:/usr/bin:/bin:/usr/sbin:/sbin
+build.env-append    PATH=${workpath}/bins:/usr/bin:/bin:/usr/sbin:/sbin
+
+cmake.build_type    Release
+cmake.generator     {Unix Makefiles}
+
+# Install everything into libexec
+cmake.install_prefix \
+                    ${prefix}/libexec/${name}
+cmake.install_rpath
+
+configure.pre_args-replace \
+                    {*}[cmake::system_prefix_path] \
+                    -DCMAKE_SYSTEM_PREFIX_PATH="${cmake.install_prefix}\;/usr"
+
+configure.pre_args-replace \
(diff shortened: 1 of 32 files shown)


lang/clang-11-bootstrap/Portfile
```diff
@@ -0,0 +1,275 @@
...
+    ln -s ${workpath}/${distname}                       ${workpath}/llvm-project
+    ln -s ${workpath}/llvm-project/clang                ${worksrcpath}/tools/clang
+    ln -s ${workpath}/llvm-project/compiler-rt          ${worksrcpath}/projects/compiler-rt
+    ln -s ${workpath}/llvm-project/libcxx               ${worksrcpath}/projects/libcxx
+    ln -s ${workpath}/llvm-project/libcxxabi            ${worksrcpath}/projects/libcxxabi
+    ln -s ${workpath}/llvm-project/lld                  ${worksrcpath}/projects/lld
+    ln -s ${workpath}/llvm-project/clang-tools-extra    ${worksrcpath}/tools/clang/tools/extra
+
+    file mkdir      ${cmake.build_dir}/lib
+    ln -s ${prefix}/libexec/gcc10-bootstrap/lib/libgcc_s.1.1.dylib ${cmake.build_dir}/lib
```
Gcenx, line 77: These two lines aren’t needed, clang will have the path to libgcc hardcoded

lang/clang-11-bootstrap/Portfile
```diff
@@ -0,0 +1,275 @@
...
+                    port:python27-bootstrap
+depends_skip_archcheck-append \
+                    python27-bootstrap
+
+# gcc10-bootstrap is used as compiler and as libc++
+# to avoid such dependency this port should also build libc++
+# and use 2-stage bootstrap
+depends_build-append \
+                    port:gcc10-bootstrap
+depends_skip_archcheck-append \
```
Gcenx, line 53: Remove this as well want gcc10-bootstrap arch to match now that universal seems to be correctly fixed.

lang/clang-11-bootstrap/Portfile
catap: @Gcenx without this it is failed to bootstrap :(

catap: But it worth to try without it

Gcenx: That's weird as I'd removed this and didn't have any issues with it finding the libgcc, the c++ headers on the other hand needed to be installed into include.

catap: @Gcenx I'll re-run tests on something old without this line to confirm.

To be honest I've added it a while ago, when `gcc10-bootstrap` port was not so good. Probably it isn't needed after all.

lang/clang-11-bootstrap/Portfile
```diff
@@ -0,0 +1,275 @@
...
+        lappend merger_build_env(${arch}) BUILD_ARCH=${arch}
+    }
+}
+
+merger_arch_flag            yes
+merger_arch_compiler        yes
+merger_must_run_binaries    yes
+
+post-destroot {
+    system "ditto ${worksrcpath}/projects/libcxx/include ${destroot}${cmake.install_prefix}/lib/c++/v1"
```
Gcenx, line 263: I still don't understand _why_ this was even in the normal clang ports as they should be installed into /include not /lib, but those should be installed correctly by some install target anyway

catap: @Gcenx to be honest I've just grab it from original port :)

lang/clang-11-bootstrap/Portfile
```diff
@@ -0,0 +1,275 @@
...
+
+depends_build-append \
+                    port:python27-bootstrap
+depends_skip_archcheck-append \
+                    python27-bootstrap
+
+# gcc10-bootstrap is used as compiler and as libc++
+# to avoid such dependency this port should also build libc++
+# and use 2-stage bootstrap
+depends_build-append \
```
Gcenx, line 51: I really don't like the idea of always using

```
# OS X Yosemite default Xcode is new enough to compile directly
if {${os.major} < 14} {
    # gcc10-bootstrap is used as compiler and as libc++
    # to avoid such dependency this port should also build libc++
    # and use 2-stage bootstrap
    depends_build-append \
                    port:gcc10-bootstrap
}
```

This would speed things up on more recent versions of OS X/macOS

Gcenx: Then also put the other libgcc block behind a similar block

catap: @Gcenx it makes things simpler and I really don't see any reason why someone is needed `clang-11-bootstrap` on something after Yosemite.

lang/clang-11-bootstrap/Portfile
Gcenx: The original port has the headers in both /include and /lib due to the used install target I'm assuming, unless someone can explain _why_ there's a second copy in /lib it doesn't seem to be needed after all the testing I've done with this already.

lang/clang-11-bootstrap/Portfile
Gcenx: Getting directly to modern cctools/ld64 this would be required to avoid the usual toolchain dance that currently happens within macports

catap: Thus, it also depends on python27-bootstrap on all systems, but it really needed such dependency on a few.

catap: > Getting directly to modern cctools/ld64 this would be required to avoid the usual toolchain dance that currently happens within macports

yes, but I prefer to keep things simpler here.

And if I may have one simpler way to build it everywhere, let use it.

Gcenx: For an initial run I guess that makes sense, getting a stage2 build that’s not locked go libgcc would be preferable.

catap: @Gcenx I see this compiler as a tool to build `libcxx` where it is needed which allows to build `cmake` and `clang` and everything via usual toolset.

But it is the next step.

Gcenx: `gcc10-bootstrap` > `clang-11-bootstrap` > `libcxx` > `clang-11-bootstrap` (stage2) > `cctools`/`lds64` > `clang`

I believe `clang-11-bootstrap` could be used to break the current clang dependency hell when attempting to get to a more modern macports-clang

Gcenx: > You may build cctools and lds64 by clang-11-bootstrap.

Already been doing this since you'd first proposed this

A stage2 version of clang-11-bootstrap would ensure it no longer required gcc10-bootstrap, we'd first build lbcxx (on 10.6 and below) then rebuild clang while forcing it to use libcxx over libgcc. I think any c++ projects build using this will end up getting permanently linked to libgcc.

lang/clang-11-bootstrap/Portfile
catap: @Gcenx without this line I can't compile anything by this complier:
```
catap@Kirills-mini-m1 ~ % port contents clang-11-bootstrap | grep 'iostream'
catap@Kirills-mini-m1 ~ % cd /tmp 
catap@Kirills-mini-m1 /tmp % cat > test.cc
#include <iostream>
using namespace std;
int main() {
   cout << "Hello, World!" << endl; // This prints Hello, World!
   return 0;
}
catap@Kirills-mini-m1 /tmp % /opt/local/libexec/clang-11-bootstrap/bin/clang++ test.cc 
test.cc:1:10: fatal error: 'iostream' file not found
#include <iostream>
         ^~~~~~~~~~
1 error generated.
catap@Kirills-mini-m1 /tmp % 
```

anyway, I agree that path looks quite wrong, special that `clang` is looking into `/opt/local/libexec/clang-11-bootstrap/bin/../include/c++/v1`

catap:
Marked it as a draft because it seems that I was to early to open it. Need more time to polish.

lang/clang-11-bootstrap/Portfile
kencu: it will…but we had decided before that having clang11-bootstrap linked to gcc10-bootstrap was **OK**. 

If you do a 2 stage build of clang11-bootstrap you need to be very careful which libcxx you are building against and exactly which libcxx you intend to install.

up to now, we install on 10.6 a specially modified libcxx from llvm 5.0 that works. There are reasons tonthink that installing anything newer that the libcc from llvm 7 might caise trouble (ABI change), but it might be OK.

needs testing.

catap: @Gcenx I feel that `stage2 clang` is usual `clang-X` port in macports. Am I wrong?

Gcenx: @catap no Macports clang ports are not stage2 compilers.

Currently the only compilers that actually go through the bootstrapping are crossgcc (manually stages) and gcc10-bootstrap (using the bootstrap target)

kencu: > @kencu the stage2 clang would be built against the system install libcxx or in 10.6 and belows cases this would be the custom libcxx 5.0
> 
> Being permanently linked to gcc10-bootstrap will be irritating considering things like cctools/ld64 end up being permanently linked to the LTO library from the compiler.
> 
> The idea is to use this compiler to break the current compiler toolchain dance that’s happening in macports.

Once again I feel I should not comment but -- 

clang11-bootstrap **cannot** depend on anything that it installs. Otherwise you are right back in the mess.

So IMHO leave clang11-bootstrap linked against gcc10-bootstrap. It will be much simpler in the end, I believe.

If it drives you crazy to have that space wasted, then you can copy the needed libgcc parts into clang11-bootstrap at the end of the build and change it to look for them there.

I read that clang is designed to look first in it's own lib trees for libgcc if it built against libgcc -- I haven't tried this, but if that is true, then this will be really easy.

Gcenx: > clang11-bootstrap **cannot** depend on anything that it installs. Otherwise you are right back in the mess.

Exactly what I was talking about on macports-wine _if_ cctools/ld64 gets built with LTO it will be locked to clang-11-bootstrap. I guess that can be figured out later however as **libcxx** will be fine.

> So IMHO leave clang11-bootstrap linked against gcc10-bootstrap. It will be much simpler in the end, I believe.
> 
> If it drives you crazy to have that space wasted, then you can copy the needed libgcc parts into clang11-bootstrap at the end of the build and change it to look for them there.
> 
> I read that clang is designed to look first in it's own lib trees for libgcc if it built against libgcc -- I haven't tried this, but if that is true, then this will be really easy.

I'd also considered that but never gotten around to trying it, having libgcc copied into clang-11-bootstrap does sounds like the best possible compromise, at least then clang-11-bootstrap won't require gcc10-bootstrap just to function.

kencu: > > clang11-bootstrap **cannot** depend on anything that it installs. Otherwise you are right back in the mess.
> 
> Exactly what I was talking about on macports-wine _if_ cctools/ld64 gets built with LTO it will be locked to clang-11-bootstrap. I guess that can be figured out later however as **libcxx** will be fine.

That can't happen like that then. 

cctools/ld64 can't be built against the llvm in clang11-bootstrap. 

clang11-bootstrap will have to build the llvm that is used by cctools/ld64 some other way. The system must function with clang11-bootstrap uninstalled. clang11-bootstrap is a build-time-only compiler, and can then be unistalled.

If a disposable cctools/ld64 is needed to bootstrap, then that will have to be used too. This is needed anyway, because otherwise you're hamstrung with bootstrapping ld64.

So still -- gcc10-bootstrap -> clang11-bootstrap -> builds everything else, starting with whatever is needed and going from there.

kencu: running a double-build of clang11-bootstrap doesn't really solve any of this, AFAICT. You still wind up in loops depending on libc++. 

It is also plenty hard to do, passing all the fancy config stuff we need for building clang11 on 10.6 (and earlier) into the second bootstrap build is a big PITA (sure, I looked at this many times before)...

Gcenx: Let’s ignore cctools/ld64 for the moment.

> running a double-build of clang11-bootstrap doesn't really solve any of this, AFAICT. You still wind up in loops depending on libc++

That’s only the case _if_ your expecting to do this in a single pass, separating the stages into its own subports that require the prior stage works around these kind of annoyances. Aka what’s already being done in the crossgcc group each compiler stage it’s it’s own port.

lang/clang-11-bootstrap/Portfile
Gcenx: Make this section

```
post-destroot {
    system "ditto ${worksrcpath}/projects/libcxx/include ${destroot}${cmake.install_prefix}/lib/c++/v1"
    file copy ${prefix}/libexec/gcc10-bootstrap/lib/libstdc++.6.dylib ${destroot}${cmake.install_prefix}/lib
}
```

That _should_ make it so `gcc10-bootstrap` isn't a runtime dep anymore, at least it worked during my testing.

catap:
Finally, I had time to finish this one.

I've tested it by building a simple C++ program:
```c++
#include <iostream>
using namespace std;
int main() {
   cout << "Hello, World!" << endl;
   return 0;
}
```
as
```
/opt/local/libexec/clang-11-bootstrap/bin/clang++ test.cc
```
on two systems:

macOS 12.4 21F79 arm64
Xcode 13.1 13A1030d

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

=> it produces a working binary

kencu:
ok, let's get this show on the road!

catap:
@kencu thanks! You were very fast!

PR #14635, opened 2022-04-20
[volk-]gnss-sdr: update to 0.0.17 and devel to 20220420-0449308c

carlesfernandez:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.5
Xcode 12.2 (12B45b)

###### Verification 
Have you

### DIFF
--- a/science/gnss-sdr/Portfile
+++ b/science/gnss-sdr/Portfile
@@ -25,11 +25,11 @@ if {${subport} eq "gnss-sdr"} {
     long_description    ${description}: \
         This port is kept up with the GNSS-SDR release, which is typically updated every few months.
 
-    github.setup    gnss-sdr gnss-sdr 0.0.16 v
-    revision        1
-    checksums       rmd160  0dd06ff13e825054180f2b809e6256e921e816fd \
-                    sha256  4100c6c53bec2dbd66b4aba8202f56c0664b3704be1e13309cbc6d8e5d71c8c5 \
-                    size    4455559
+    github.setup    gnss-sdr gnss-sdr 0.0.17 v
+    revision        0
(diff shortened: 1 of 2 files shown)


michaelld:
@carlesfernandez OK so the issue is 2-fold & interrelated between them: (1) MacPorts' `lapack` is _not_ guaranteed to produce `libblas.dylib` ... for example if the `+openblas` variant is used, then `lapack` uses `openblas` & does not use, compile, or install its internal `blas` code. (2) The [code for finding blas/lapack on Darwin](https://github.com/gnss-sdr/gnss-sdr/blob/main/cmake/Modules/AvoidAccelerate.cmake) seems confused and a bit outdated. For example the `find_library(BLAS_LIBRARIES` is looking for `libblas.dylib` which as noted in (1) may or not exist; further, this call doesn't even use the `BLAS_ROOT_USER_DEFINED` path(s) (as does the subsequent `find_library(LAPACK_LIBRARIES`). I agree that using Accelerate isn't wise, since it has bugs & we generally try to avoid using it. That said, there has got to be a more robust way to determine BLAS and LAPACK. Please advise!

michaelld (approved):
Builds cleanly for me locally when I install straight up `lapack` (no `+openblas`; this is the default install). We still need to clean up the `lapack +openblas` case, but I feel comfortable merging this as is.

carlesfernandez:
@michaelld Ok, I was not seeing those errors. I'll work on the CMake part in order to cover all variant cases. From what I'm seeing, adding `libopenblas.dylib` to the list in `find_library(BLAS_LIBRARIES ...)` does the trick, but let me check all possible combinations.

Thanks for your help!

PR #14674, opened 2022-04-23
ghc-ppc-bootstrap: new port; latest release of GHC for PowerPC

barracuda156:
#### Description

Macports currently has no `ghc` for PowerPC systems. This port addresses this problem. It is based on the latest available build for PowerPC on Haskell.org.
Once installed, it can be used to build `ghc` up to @7.6.3. Which means that we can build existing `ghc-bootstrap` for three archs (obviously, at the moment it cannot be built for PPC): https://ports.macports.org/port/ghc-bootstrap/details/

I can update `ghc-bootstrap` port to include PPC, however it would mean that while for i386 and x86_64 the port is downloaded and installed, for PPC it will need to be built using `ghc-ppc-bootstrap`. This sort of fails the name of -bootstrap, however it appears to be the only feasible solution for now.

@kencu @catap @evanmiller @mascguy @neverpanic What do you think?

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6 PPC
Xcode 3.2

The compiler itself is reasonably tested, I used it to build `ghc` @7.6.2 and @7.6.3. (Optional framework variant is added just because the original pkg installs a framework. I did not test framework variant beyond verifying it destroots correctly.)

P. S. There is some bug in the base of `ghc` around 7.4.x–7.6.x versions which does not allow using them to build 7.7.x–7.10.x. So the pathway that I currently propose can lead to functional `ghc` @7.6.3 but not later. Long thread here: https://trac.macports.org/ticket/64698

Once/if I succeed in cross-building `ghc` 9.x for PowerPC, this port can be updated accordingly.

### DIFF
--- a/lang/ghc-ppc-bootstrap/Portfile
+++ b/lang/ghc-ppc-bootstrap/Portfile
@@ -1,0 +1,76 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                ghc-ppc-bootstrap
+set canonicalname   ghc
+version             7.0.4
+revision            0
+categories          lang haskell
+maintainers         @barracuda156
+license             BSD
+platforms           darwin
+supported_archs     ppc
+universal_variant   no
+installs_libs       no
+
+description         The Glorious Glasgow Haskell Compilation System
+long_description    This is a package that installs a binary bootstrap ghc compiler.
+
+homepage            http://haskell.org/${canonicalname}
+master_sites        https://downloads.haskell.org/ghc/7.0.4/krabby/
+distname            GHC-${version}-powerpc
+distfiles			${distname}.pkg
+
+checksums           rmd160  38cc042793d2b4f4a6b901c6bb7f6a1f6959cce8 \
+                    sha256  81cc84e18d30d35d4103b21d146b264abd7f00879ae64ef492bc83d67231fb7c \
+                    size    171783399
+
+depends_extract     port:xar
+
+worksrcdir          ${canonicalname}-${version}
+
+extract {
+    system -W ${workpath} "mkdir -p ${worksrcpath}/pkg"
+    system -W ${workpath} "mkdir -p ${worksrcpath}/files"
+    system -W ${distpath} "cp ${distpath}/${distfiles} ${worksrcpath}/pkg"
+    system -W ${worksrcpath}/pkg "${prefix}/bin/xar -xf ${worksrcpath}/pkg/GHC-${version}-powerpc.pkg -C ${worksrcpath}/files"
+    system -W ${worksrcpath}/files/ghc.pkg "cat Payload | gunzip -dc | cpio -i"
+    
+if {![variant_isset framework]} {
+    system -W ${worksrcpath}/files/ghc.pkg "mv GHC.framework/Versions/7.0.4-powerpc/usr ${worksrcpath}/${name}"
+    system -W ${worksrcpath}/files/ghc.pkg "cp -R GHC.framework/Versions/7.0.4-powerpc/Tools ${worksrcpath}/${name}/share"
+    }
+}
+
+use_configure     no
+
+build {}
+
+if {![variant_isset framework]} {
+	destroot.destdir prefix=${destroot}/share
+	set path ${prefix}/share
+	destroot {
+		copy ${worksrcpath}/${name} ${destroot}${path}
+	}
+	
+	post-destroot {
+		# Delete dylibs; they arenʼt used by the bootstrap ghc and are incorrectly linked against /usr/local, causing rev-upgrade to complain.
+		fs-traverse f ${destroot}${path}/${name}/lib {
+			if {[file isfile ${f}]} {
+				if {[file extension ${f}] == ".dylib"} {
+					delete ${f}
+				}
+			}
+		}
+	}
+}
+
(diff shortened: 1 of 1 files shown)


kencu:
I think you're making good progress now as a MacPorts contributor. There are a few things that people will discussion about the details here, but the main point is that you're on a path!

barracuda156:
@evanmiller Could you please test this when you got time to?

evanmiller:
> @evanmiller Could you please test this when you got time to?

It builds OK, but after installation I get an error about the port being broken (e.g. linking to a non-existent file).

barracuda156:
@evanmiller Never mind, I tested it on Tiger now, it complains about `libiconv` being too old. Apparently it needs to depend on Macports `libiconv` instead. I will try fixing it now.

barracuda156:
@kencu Have you got any idea how to make `ghc` to invoke Macports `libiconv` at destroot stage (
I tried adding:
```
platform darwin 8 {
    depends_build port:libiconv
    depends_lib   port:libiconv
    destroot.env-append DYLD_FALLBACK_LIBRARY_PATH=${prefix}/lib
}
```
No effect.

Specifically this part needs a newer `libiconv`, apparently:
```
	post-activate {
		catch {system "${prefix}/share/ghc-ppc-bootstrap/bin/ghc-pkg -v recache"}
	}
```
As @evanmiller notes, if fails:
```
dyld: Library not loaded: /usr/lib/libiconv.2.dylib
  Referenced from: /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/ghc-pkg
  Reason: Incompatible library version: ghc-pkg requires version 7.0.0 or later, but libiconv.2.dylib provides version 5.0.0
Command failed: /opt/local/share/ghc-ppc-bootstrap/bin/ghc-pkg -v recache
Killed by signal: 5
--->  Cleaning ghc-ppc-bootstrap
--->  Removing work directory for ghc-ppc-bootstrap
Incompatible library version: /opt/local/share/ghc-ppc-bootstrap/bin/hpc requires version 7.0.0 or later, but /usr/lib/libiconv.2.dylib provides version 5.0.0
Incompatible library version: /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/ghc requires version 7.0.0 or later, but /usr/lib/libiconv.2.dylib provides version 5.0.0
Incompatible library version: /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/ghc-pkg requires version 7.0.0 or later, but /usr/lib/libiconv.2.dylib provides version 5.0.0
Incompatible library version: /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/haddock requires version 7.0.0 or later, but /usr/lib/libiconv.2.dylib provides version 5.0.0
Incompatible library version: /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/hsc2hs requires version 7.0.0 or later, but /usr/lib/libiconv.2.dylib provides version 5.0.0
Incompatible library version: /opt/local/share/ghc-ppc-bootstrap/lib/ghc-7.0.4/runghc requires version 7.0.0 or later, but /usr/lib/libiconv.2.dylib provides version 5.0.0
--->  Found 6 broken files, matching files to ports
--->  Found 1 broken port, determining rebuild order
```

kencu:
well, if I was going to try to do this, I would probably try to first of all build the right version of libiconv (likely the one that is on Leopard) and that had the right symbol names in it (there is a whole project to be understood there, but in short, libiconv builds with different names for system installs vs user installs and this ghc is built agains a version with the system install names).

And then I would use install_name_tool to change the libiconv used in ghc everywhere it is used.

It might work. I'd give it 70:30 odds.

barracuda156:
> well, if I was going to try to do this, I would probably try to first of all build the right version of libiconv (likely the one that is on Leopard) and that had the right symbol names in it (there is a whole project to be understood there, but in short, libiconv builds with different names for system installs vs user installs and this ghc is built agains a version with the system install names).
> 
> And then I would use install_name_tool to change the libiconv used in ghc everywhere it is used.
> 
> It might work. I'd give it 70:30 odds.

Thanks! Do you recall any example ports that deal with a similar issue? (Please don’t waste time to search, of course, I ask just in case you call pull a name from memory. Otherwise I will look for it myself.)

barracuda156:
@kencu @evanmiller Well, it turned out I already had this `libiconv` failure on Tiger with my first version of `ghc-ppc-bootstrap` locally (based on `ghc` 7.0.1): https://trac.macports.org/ticket/64698#comment:19
Not that I put a huge effort to fix it back then, but portfile magic did not work.

barracuda156:
@mascguy This PR can be merged IMO. Compiler builds and works on 10.5.8 (Leopard) and 10.6 PPC.
Tiger fails, but we have many ports that do not build for Tiger. At the moment we have no version of `ghc` for PPC at all. With this PR we get one to bootstrap.

Once this is merged, I can update `ghc-bootstrap` port to 7.6.3 and incorporate PPC version there.

barracuda156:
@evanmiller I thought that we could just do this: 
```
system "install_name_tool -change /usr/lib/libiconv.2.dylib ${prefix}/lib/libiconv.dylib ${destroot}${path}/${name}/lib/ghc-7.0.4/ghc"
```
However, apparently it leaves GHC broken:
```
38-216:~ svacchanda$ /opt/local/share/ghc-ppc-bootstrap/bin/ghc --version
Bus error
```
There is one more issue: recache does not work:
```
Command failed: /opt/local/share/ghc-ppc-bootstrap/bin/ghc-pkg -v recache
Killed by signal: 10
```
This is on Tiger, I need to verify elsewhere.

PR #14686, opened 2022-04-25
jdk17: new submission

breun:
#### Description

New port for the Oracle Java SE 17 Development Kit.

Note that this is not the same as what's provided by the existing `openjdk17-oracle` port. This port is not for the GPL-licensed Oracle OpenJDK, but for the [NFTC](https://java.com/freeuselicense)-licensed Oracle Java SE JDK. The benefit of this new port is that it still receives updates for Java 17, while updates for Oracle OpenJDK 17 have already ended.

###### Tested on

macOS 12.3.1 21E258 x86_64
Xcode 13.3.1 13E500a

###### Verification 
Have you

### DIFF
--- a/java/jdk17/Portfile
+++ b/java/jdk17/Portfile
@@ -1,0 +1,90 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem       1.0
+
+name             jdk17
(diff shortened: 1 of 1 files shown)


usersxx:
How about renaming it to `jdk17` as jdk is the non open source distribution provided by oracle. (see: https://en.m.wikipedia.org/wiki/Java_Development_Kit)

breun:
I guess there will be only one JDK which is not explicitly a build of OpenJDK (although Oracle JDK is almost identical to Oracle OpenJDK these days), so we could indeed omit the `-oracle` postfix. I have renamed the port.

PR #14691, opened 2022-04-26
ocaml: update to @4.14.0, fix PPC build

barracuda156:
#### Description

UPD. While build for ppc succeeds, test target fails at the moment. I will see if that can be fixed. Please ignore this PR until I can offer an update.

This PR updates `ocaml` to @4.14.0 and hopefully fixes PPC build. The idea is to keep PPC patches completely separate, because they include replacing the whole set of Power-arch files. That is, they are to be used only on Darwin and when the build arch is ppc/ppc64.

This addresses the following ticket: https://trac.macports.org/ticket/55385
@kencu Please take a look if this looks okay for you.

###### Type(s)

- [ x] bugfix
- [x ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode 3.2.6

macOS 10.6 PPC (10A190)
Xcode 3.2

macOS 10.5.8 for ppc32
Xcode 3.1.4

I will check on a PPC machine soon. @evanmiller If you got time, you are very welcome to try it too.

###### Verification 
Have you

- [ x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x ] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [ x] referenced existing tickets on [Trac](https://trac.macports.org/wiki/Tickets) with full URL? 

P. S. Initially I thought to bring back Darwin PPC support into existing Power-specific code and submit PR with upstream. However, upstream told me they cannot care less and closed my ticket: https://github.com/ocaml/ocaml/issues/11216
So now I see no point in improving the source code; it will be rather time-consuming, but carries zero benefit on our level, since there is no use-case of building for PowerPC and POWER at the same time. Darwin ASM files are borrowed from `ocaml` 4.0.2, the last version which had them before upstream decided that “connecting such a system to the Internet should be prohibited by law” (real quote).
I made a few minor changes, one to fix a duplicate symbol error and a couple of stylistic.
After that plus patching `configure` to identify arch and OS correctly, the build went quite far, but failed on this:
```
/opt/local/bin/gcc-mp-11 -arch ppc -shared                    -flat_namespace -undefined suppress -Wl,-no_compact_unwind -read_only_relocs suppress                    -L/opt/local/lib -Wl,-headerpad_max_install_names -arch ppc -o libasmrun_shared.so startup_aux.npic.o startup_nat.npic.o main.npic.o fail_nat.npic.o roots_nat.npic.o signals.npic.o signals_nat.npic.o misc.npic.o freelist.npic.o major_gc.npic.o minor_gc.npic.o memory.npic.o alloc.npic.o compare.npic.o ints.npic.o floats.npic.o str.npic.o array.npic.o io.npic.o extern.npic.o intern.npic.o hash.npic.o sys.npic.o parsing.npic.o gc_ctrl.npic.o eventlog.npic.o md5.npic.o obj.npic.o lexing.npic.o unix.npic.o printexc.npic.o callback.npic.o weak.npic.o compact.npic.o finalise.npic.o custom.npic.o globroots.npic.o backtrace_nat.npic.o backtrace.npic.o dynlink_nat.npic.o debugger.npic.o meta.npic.o dynlink.npic.o clambda_checks.npic.o afl.npic.o bigarray.npic.o memprof.npic.o domain.npic.o skiplist.npic.o codefrag.npic.o power_libasmrunpic.o -lm 
ld: relocation used in _caml_system__code_begin from power_libasmrunpic.o not allowed in slidable image
make[3]: *** [libasmrun_shared.so] Error 1
```
In order to get past that, I added some settings for PPC in Portfile. In result:
```
macmini:macports-ports svacchanda$ port -v installed ocaml
  ocaml @4.14.0_0 (active) requested_variants='' platform='darwin 10' archs='ppc' date='2022-04-26T08:47:31+0800'
```


evanmiller:
Did you try running any tests?

barracuda156:
> Did you try running any tests?

@evanmiller Not yet, need to get some sleep finally first LOL.

I will build it on the Quad tonight or tomorrow and run tests there. Rosetta may have problems of its own which do not apply to native PPC.

pmetzger:
What you're doing here is essentially maintaining a separate, out of tree PowerPC patch set for OCaml. This would sort of work for 4.14, but will fail utterly fail when 5.0 comes out because of the multicore support. I don't think this is a good idea, as much as I appreciate the sentiment. If you really want to restore this, we're going to have to convince the OCaml maintainers to buy it back, which I suspect they won't want to do.

pmetzger:
BTW, even now, PowerPC wouldn't work quite right. There have been lots of changes in OCaml in recent years (like guarantees about the number of parameters available to a tail recursive function without forcing it not to be tail recursive) that have required changes to the code generators. Even if this sorta/kinda works it is going to bite us in the rear at some point even just for 4.14, let alone for 5.0 where the system has loads of hooks for multicore in the code generator.

barracuda156:
@pmetzger By the way, if you prefer, I could actually make PPC fix into a patch-only form without any extra files. It will be a long patch, but it can be made into a single diff file. I can also maintain it as long as `ocaml` 4.x is around.

barracuda156:
@pmetzger @evanmiller Okay, while it does build on 10.5.8 for ppc, test target fails, so please allow me some time to see what's going on.

barracuda156:
@pmetzger @evanmiller  UPD. It actually may be not that `ocaml` is broken but rather something to do with `world` target vs default `world.opt` target. The test suite complains about `ocamlopt` missing.
Helloworld works normally:
```
36-206:~ svacchanda$ cd /Users/svacchanda/Dev 
36-206:Dev svacchanda$ bbedit hello.ml
36-206:Dev svacchanda$ ocamlc -o hello hello.ml
36-206:Dev svacchanda$ /Users/svacchanda/Dev/hello 
Hello world!

36-206:Dev svacchanda$ /opt/local/bin/ocamlc -version
4.14.0
```

barracuda156:
@pmetzger With some warnings, but I was able to build `opam` with my `ocaml` on 10.5.8.
```
36-206:Dev svacchanda$ port -v installed opam
  opam @2.1.1_0 (active) requested_variants='-universal'
platform='darwin 9' archs='ppc' date='2022-04-27T14:22:19+0800'
```
[opam_log.txt](https://github.com/macports/macports-ports/files/8569387/opam_log.txt)

barracuda156:
@pmetzger Could you please help me a bit and suggest an idea how to fix the following? This is what prevents `world.opt` target to get built:
```
/opt/local/bin/gcc-mp-11 -arch ppc -shared                    -flat_namespace -undefined suppress -Wl,-no_compact_unwind -read_only_relocs suppress                    -L/opt/local/lib -Wl,-headerpad_max_install_names -arch ppc -o libasmrun_shared.so startup_aux.npic.o startup_nat.npic.o main.npic.o fail_nat.npic.o roots_nat.npic.o signals.npic.o signals_nat.npic.o misc.npic.o freelist.npic.o major_gc.npic.o minor_gc.npic.o memory.npic.o alloc.npic.o compare.npic.o ints.npic.o floats.npic.o str.npic.o array.npic.o io.npic.o extern.npic.o intern.npic.o hash.npic.o sys.npic.o parsing.npic.o gc_ctrl.npic.o eventlog.npic.o md5.npic.o obj.npic.o lexing.npic.o unix.npic.o printexc.npic.o callback.npic.o weak.npic.o compact.npic.o finalise.npic.o custom.npic.o globroots.npic.o backtrace_nat.npic.o backtrace.npic.o dynlink_nat.npic.o debugger.npic.o meta.npic.o dynlink.npic.o clambda_checks.npic.o afl.npic.o bigarray.npic.o memprof.npic.o domain.npic.o skiplist.npic.o codefrag.npic.o power_libasmrunpic.o -lm 
ld: relocation used in _caml_system__code_begin from power_libasmrunpic.o not allowed in slidable image
make[3]: *** [libasmrun_shared.so] Error 1
```

barracuda156:
@iains Just in case, is this a known `ld` error to you? This is a failure of specific target of `ocaml` on Leopard when building shared libs.
```
ld: relocation used in _caml_system__code_begin from power_libasmrunpic.o not allowed in slidable image
```
Apple Open Source `ld` changelog has the following which sounds somewhat similar:
```
	<rdar://problem/5645446> SWB: failures due to ld: pointer in read-only segment not allowed in slidable image, used in ...
	* src/MachOReaderRelocatable.hpp: in Reader<x86>::addRelocReference() handle weak pc-rel 32-bit vanilla relocs properly
```
https://opensource.apple.com/source/ld64/ld64-85.2.1/ChangeLog.auto.html

barracuda156:
The following suggests that support for `-read_only_relocs suppress` flag has been dropped in Apple linker: https://sympa.inria.fr/sympa/arc/caml-list/2008-02/msg00115.html
And rather `ld_classic` has to be used.

pmetzger:
So let me repeat: even now, the OCaml port is going to break subtly on PowerPC. There are a bunch of subtle bugs that have been fixed in the code generators for other architectures but which haven't been fixed for PowerPC.

Here's a good example: let's say that you have ocaml code that depends on tail recursion, uses the tail recursion annotation to say "don't compile if this can't be made tail recursive",  and uses more than a trivial number of passed parameters to a function; on x86_64 etc. the guarantee you request will be respected, but in the older PowerPC code generator it will fail because it hasn't been maintained. This is not a small matter; the reason I know about this particular bug so much is I was involved in a security critical problem that it triggered.

This was clearly significant work for you and I'm reluctant to tell you to give up when you obviously did so much to get this to work, but it's probably not a good idea unless you can get upstream to re-adopt it and fix the things that haven't been maintained in the interim.

kencu:
or @barracuda156 -- spin off an "ocaml-ppc" port and let it run at that. Then you can control the narrative, and put in the notes any issues with the possible hiccups.

or perhaps you can see where LeopardPorts and TigerPorts came into being -- you could just do that too.

iains:
> `-read_only_relocs suppress` is supported by the apple linker. You might have to do some tricks to GET it to the linker, like this for example ` -Wl,-read_only_relocs,suppress` but it is supported.
> 
> I've used it hundreds of times -- it's often needed for 32 bit code, i386 and ppc.

It would be good to examine some of those times, and make sure that we do not have some code-gen issue.  User space code should not require too much shenanigans of this sort.

barracuda156:
> To the best of my knowledge, there is no need to use ld_classic on Darwin9 outside linking stuff for the kernel. What needs to be fixed is whatever code-gen is giving rise to the wrong relocations.

@iains Got it, thank you very much!

barracuda156:
> So let me repeat: even now, the OCaml port is going to break subtly on PowerPC. There are a bunch of subtle bugs that have been fixed in the code generators for other architectures but which haven't been fixed for PowerPC.
> 
> Here's a good example: let's say that you have ocaml code that depends on tail recursion, uses the tail recursion annotation to say "don't compile if this can't be made tail recursive", and uses more than a trivial number of passed parameters to a function; on x86_64 etc. the guarantee you request will be respected, but in the older PowerPC code generator it will fail because it hasn't been maintained. This is not a small matter; the reason I know about this particular bug so much is I was involved in a security critical problem that it triggered.
> 
> This was clearly significant work for you and I'm reluctant to tell you to give up when you obviously did so much to get this to work, but it's probably not a good idea unless you can get upstream to re-adopt it and fix the things that haven't been maintained in the interim.

@pmetzger Well, if you will happen to have a mood and time, and can look into the code for ppc arch, that could greatly help. (I understand that you may have a better use for your time, but just in case.)

There are just two major components which are arch-dependent: `power.S` in `runtime` (former `asmrun`) and few files in `asmcomp/power`, of which the difficult ones are `emit.mlp` and `proc.ml`. It does not look inconceivably complex, though _for me_ it is complicated, of course.

The question is ultimately whether it is preferable to have nothing or something but imperfect. I definitely cannot fix `ocaml` for PPC on my own to the level it exists for officially supported archs. But if someone more knowledgeable – you or someone else – joins in, it can be doable, I guess.

@kencu I don’t mind making a dedicated PPC port, I just thought that it is normally preferred not to “multiply entities”. In some case, like that of `ghc`, it is simply unavoidable; here PPC may have been integrated in a way that guarantees that nothing else is adversely affected.
But yeah, I am okay with a separate port.

barracuda156:
@kencu @iains We actually had a similar `icu` reloc error recently: https://trac.macports.org/attachment/ticket/64398/icu-ppc-universal-fail-after-adding-read-only-relocs-suppress.log
```
:info:build ld: relocation used in icu_67::FixedDecimal::getFractionalDigits(double, int) from plurrule.o not allowed in slidable image
```
And that time it turned out to be a bug in `gcc` for powerpc64 target.

pmetzger:
To start, it would be nice to do the update to 4.14.0. Let's split that off.

kencu:
> @kencu @iains We actually had a similar `icu` reloc error recently: https://trac.macports.org/attachment/ticket/64398/icu-ppc-universal-fail-after-adding-read-only-relocs-suppress.log
> 
> ```
> :info:build ld: relocation used in icu_67::FixedDecimal::getFractionalDigits(double, int) from plurrule.o not allowed in slidable image
> ```
> 
> And that time it turned out to be a bug in `gcc` for powerpc64 target.

that one, indeed, was a very unique situation. Quite unlike the othe 1,000s of read_only_reloc errors.

I have this error all sorted out in the archives someplace. If I get to the dungeon this week I’ll dig it up from it’s 32-bit grave for you…

kencu:
> I just thought that it is normally preferred not to “multiply entities”.

Oh, yeah, sure, beyond a doubt, absolutely, 100% agree, yes ... but here the "ask" for you get your PR committed is to have upstream reinstate support for a long-dead architecture that nearly nobody uses. Everyone knows that ain't never gonna happen :>

So -- your choices are -- wait six months and close this PR. Or -- set up your own separate ocaml-ppc port. Personally, I would (and did) just set up my own LeopardPorts/TigerPorts overlay and move on -- you can have anything in there you want to have -- but in case you really really really want ocaml in MacPorts for some reason, it could happen...

I just wanted you to know there is an option for you.

barracuda156:
> > I just thought that it is normally preferred not to “multiply entities”.
> 
> Oh, yeah, sure, beyond a doubt, absolutely, 100% agree, yes ... but here the "ask" for you get your PR committed is to have upstream reinstate support for a long-dead architecture that nearly nobody uses. Everyone knows that ain't never gonna happen :>
> 
> So -- your choices are -- wait six months and close this PR. Or -- set up your own separate ocaml-ppc port. Personally, I would (and did) just set up my own LeopardPorts/TigerPorts overlay and move on -- you can have anything in there you want to have -- but in case you really really really want ocaml ppc support in MacPorts for some reason, it could happen...
> 
> I just wanted you to know there is an option for you.

@kencu Yes, I totally get your reasoning. I do think that getting it into Macports may carry some positive utility, since someone might find it there, while having it in my local repo is a sure way to exactly zero users ever :)
Upstream is indeed out of question here: unlike the case of `ghc`, where they at least said PPC can be considered if I manage to implement it back, you saw the reply of `ocaml` developer.

So yeah, I fill follow your suggestion, I guess, to make a separate `ocaml` port and keep it at v. 4.x, while the main one gonna move to v. 5. But let me see if it can be improved first.

P. S. I was trying to build `world.opt` target with `disable-shared`, since `ld` failure is unclear in terms of how to fix it, and that kept failing on `asmcomp` portion. Well, after a night of torturing the code, I made through `asmcomp` – only to see it fail on atomics of `stdlib` :)
_Gonna update the patches just in case someone bumps in to save it._

barracuda156:
I will try to improve on the last patch today. Basically, try-debug-fix is not the best approach here. I was able to solve an immediate issue at hand (building `samcomp` stage), but it did no good to the compiler and might have broken something else. I have no other way but read through on Mac OSX ABI and PPC assembler and go through `emit` and `proc` files to bring their current versions in compliance with Darwin ABI instead of making ad hoc random fixes based on compilation errors.

barracuda156:
UPD. Bad news, it takes time to go through and fix. Good news, once I built [a broken] asmcomp, I got an output of stdlib errors, which show where the newly build native compiler fails. Like:
```
L102:
	addi	r4, 31, 4
	lwz	r11, 44(r1)
	mtlr	r11
	li	r5, 1024
	stw	r5, -4(r4)
	stw	r3, 0(r4)
	mr	r3, r4
	addi r1, r1, 48
	blr
L101:
	b	_caml_call_gc
	.size	_camlCamlinternalAtomic__make_9, . - _camlCamlinternalAtomic__make_9
	.globl	_camlCamlinternalAtomic__get_12
	.align	2
_camlCamlinternalAtomic__get_12:
L103:
	lwz	r3, 0(r3)
	addi r1, r1, 32
	blr
	.size	_camlCamlinternalAtomic__get_12, . - _camlCamlinternalAtomic__get_12
	.globl	_camlCamlinternalAtomic__set_15
	.align	2
_camlCamlinternalAtomic__set_15:
	addi	1, 1, -48
	mflr	0
	stw	0, 44(1)
L104:
	addis	r11, 0, ha16(L_caml_modify$non_lazy_ptr)
	lwz	r11, lo16(L_caml_modify$non_lazy_ptr)(r11)
	mtctr	r11
	bctrl
	lwz	r11, 44(r1)
	mtlr	r11
	li	r3, 1
	addi r1, r1, 48
	blr
	.size	_camlCamlinternalAtomic__set_15, . - _camlCamlinternalAtomic__set_15
	.globl	_camlCamlinternalAtomic__exchange_19
	.align	2
_camlCamlinternalAtomic__exchange_19:
	addi	1, 1, -48
	mflr	0
	stw	0, 44(1)
```
So here in several instances registers got passed to assembler without proper names, which resulted in junk. This should be traceable and fixable, hopefully.

barracuda156:
@pmetzger @gasche Could someone do me a favor? I have _almost_ fixed `world.opt` target for PPC, however there remains one problem: since @4.10, a number of symbols were moved into `Caml_state`, and it does not work if this form is not used (as it was in @4.0.2) – everything builds, but linking fails with undefined symbols (I omit similar refs to same symbols):
```
Undefined symbols:
  "_Is_in_code_area", referenced from:
      _handle_signal in libasmrun.a(signals_nat.n.o)
  "_caml_backtrace_pos", referenced from:
      _caml_raise_exn in libasmrun.a(power.o)
      _caml_raise_exn in libasmrun.a(power.o)
  "_caml_last_return_address", referenced from:
      _caml_system__code_begin in libasmrun.a(power.o)
      _caml_system__code_begin in libasmrun.a(power.o)
  "_caml_try_leave_blocking_section_hook", referenced from:
      _handle_signal in libasmrun.a(signals_nat.n.o)
  "_caml_young_ptr", referenced from:
      _caml_system__code_begin in libasmrun.a(power.o)
      _caml_system__code_begin in libasmrun.a(power.o)
  "_caml_bottom_of_stack", referenced from:
      _caml_system__code_begin in libasmrun.a(power.o)
      _caml_system__code_begin in libasmrun.a(power.o)
  "_caml_gc_regs", referenced from:
      _caml_system__code_begin in libasmrun.a(power.o)
      _caml_system__code_begin in libasmrun.a(power.o)
  "_caml_young_limit", referenced from:
      _caml_system__code_begin in libasmrun.a(power.o)
  "_caml_exception_pointer", referenced from:
      _caml_system__code_begin in libasmrun.a(power.o)
      _caml_raise_exception in libasmrun.a(power.o)
  "_caml_backtrace_active", referenced from:
      _caml_raise_exn in libasmrun.a(power.o)
      _caml_raise_exn in libasmrun.a(power.o)
  "_caml_execute_signal", referenced from:
      _handle_signal in libasmrun.a(signals_nat.n.o)
     (maybe you meant: _caml_execute_signal_exn)
  "_caml_call_gc", referenced from:
```

At the same trying to retain `Caml_state` syntax leads to another bunch of errors:
```
/opt/local/bin/gcc-mp-11 -arch ppc -c -O2 -fno-strict-aliasing -fwrapv -pthread -Wall -Wdeclaration-after-statement -fno-common -fexcess-precision=standard -g -pipe -Os -arch ppc -D_FILE_OFFSET_BITS=64 -DCAML_NAME_SPACE  -DCAMLDLLIMPORT= -DNATIVE_CODE -DTARGET_power -DMODEL_ppc -DSYS_rhapsody -DDEBUG -I/opt/local/include -o startup_aux.nd.o startup_aux.c
../runtime/caml/domain_state.tbl:17:Unknown pseudo-op: .equ
../runtime/caml/domain_state.tbl:17:Rest of line ignored. 1st junk character valued 100 (d).
power.S:69:Expected comma after segment-name
power.S:75:Expected comma after segment-name
power.S:81:Can't relocate expression. Absolute 0 assumed.
power.S:81:Parameter syntax error (parameter 3)
power.S:84:Can't relocate expression. Absolute 0 assumed.
power.S:84:Parameter syntax error (parameter 3)
power.S:216:Parameter syntax error (parameter 3)
power.S:221:Unknown pseudo-op: .size
power.S:221:Rest of line ignored. 1st junk character valued 99 (c).
power.S:225:Expected comma after segment-name
power.S:225:Rest of line ignored. 1st junk character valued 32 ( ).
power.S:228:Unknown pseudo-op: .cfi_register
power.S:228:Rest of line ignored. 1st junk character valued 54 (6).
power.S:232:Can't relocate expression. Absolute 0 assumed.
power.S:232:Parameter syntax error (parameter 3)
power.S:233:Can't relocate expression. Absolute 0 assumed.
power.S:233:Parameter syntax error (parameter 3)
power.S:235:Can't relocate expression. Absolute 0 assumed.
power.S:235:Parameter syntax error (parameter 3)
power.S:236:Can't relocate expression. Absolute 0 assumed.
power.S:246:Parameter syntax error (parameter 3)
power.S:249:Unknown pseudo-op: .cfi_endproc
power.S:250:Unknown pseudo-op: .size
power.S:250:Rest of line ignored. 1st junk character valued 99 (c).
power.S:254:Expected comma after segment-name
power.S:254:Rest of line ignored. 1st junk character valued 32 ( ).
power.S:276:Unknown pseudo-op: .size
power.S:276:Rest of line ignored. 1st junk character valued 99 (c).
power.S:280:Expected comma after segment-name
power.S:283:Invalid mnemonic 'lg32'
power.S:288:Can't relocate expression. Absolute 0 assumed.
power.S:300:Parameter syntax error (parameter 3)
power.S:301:Can't relocate expression. Absolute 0 assumed.
power.S:302:Parameter syntax error (parameter 3)
power.S:307:Unknown pseudo-op: .size
power.S:307:Rest of line ignored. 1st junk character valued 99 (c).
power.S:311:Expected comma after segment-name
power.S:311:Rest of line ignored. 1st junk character valued 32 ( ).
power.S:316:Parameter error: expression must be absolute (parameter 3)
power.S:316:Invalid mnemonic 'ha'
power.S:321:Unknown pseudo-op: .cfi_adjust_cfa_offset
power.S:321:Rest of line ignored. 1st junk character valued 40 (().
power.S:325:Unknown pseudo-op: .cfi_offset
power.S:325:Rest of line ignored. 1st junk character valued 54 (6).
power.S:367:Can't relocate expression. Absolute 0 assumed.
power.S:371:Parameter syntax error (parameter 3)
power.S:378:Unknown pseudo-op: .cfi_adjust_cfa_offset
power.S:378:Rest of line ignored. 1st junk character valued 49 (1).
power.S:381:Can't relocate expression. Absolute 0 assumed.
power.S:381:Parameter syntax error (parameter 3)
power.S:393:Unknown pseudo-op: .cfi_adjust_cfa_offset
power.S:393:Rest of line ignored. 1st junk character valued 45 (-).
power.S:397:Can't relocate expression. Absolute 0 assumed.
power.S:397:Parameter syntax error (parameter 3)
power.S:403:Can't relocate expression. Absolute 0 assumed.
power.S:403:Parameter syntax error (parameter 3)
power.S:457:Unknown pseudo-op: .cfi_endproc
power.S:458:Unknown pseudo-op: .size
power.S:458:Rest of line ignored. 1st junk character valued 99 (c).
power.S:462:Expected comma after segment-name
power.S:469:Unknown pseudo-op: .size
power.S:469:Rest of line ignored. 1st junk character valued 99 (c).
power.S:471:Expected comma after segment-name
power.S:478:Parameter error: expression must be absolute (parameter 3)
power.S:478:Invalid mnemonic 'ha'
power.S:480:Unknown pseudo-op: .size
power.S:480:Rest of line ignored. 1st junk character valued 99 (c).
power.S:482:Expected comma after segment-name
power.S:489:Parameter error: expression must be absolute (parameter 3)
power.S:489:Invalid mnemonic 'ha'
power.S:491:Unknown pseudo-op: .size
power.S:491:Rest of line ignored. 1st junk character valued 99 (c).
power.S:493:Expected comma after segment-name
power.S:499:Rest of line ignored. 1st junk character valued 32 ( ).
power.S:500:Unknown pseudo-op: .type
power.S:500:Rest of line ignored. 1st junk character valued 99 (c).
power.S:500:Invalid mnemonic 'object'
power.S:509:unknown section type: %progbits
If your assembler produced syntax errors, it is probably
unhappy with the preprocessor. Check your assembler, or
try producing power.o by hand.
make[3]: *** [power.o] Error 2
```
It is likely a simple fix, but I just dunno how to make it. This portion is problematic in `runtime/power.S`:
```
.set    domain_curr_field, 0
#define DOMAIN_STATE(c_type, name) \
.equ    domain_field_caml_##name, domain_curr_field ; \
.set    domain_curr_field, domain_curr_field + 1
#include "../runtime/caml/domain_state.tbl"
#undef DOMAIN_STATE
```
Could you please suggest me an appropriate way to include this, so that `Caml_state` works as supposed to?

barracuda156:
After fixing symbols, I got to here:
```
:info:build ../../ocamlc.opt -depend -slash -I dynlink_compilerlibs \
:info:build 	dynlink_compilerlibs/binutils.ml dynlink_compilerlibs/config.ml dynlink_compilerlibs/build_path_prefix_map.ml dynlink_compilerlibs/misc.ml dynlink_compilerlibs/identifiable.ml dynlink_compilerlibs/numbers.ml dynlink_compilerlibs/arg_helper.ml dynlink_compilerlibs/clflags.ml dynlink_compilerlibs/profile.ml dynlink_compilerlibs/consistbl.ml dynlink_compilerlibs/terminfo.ml dynlink_compilerlibs/warnings.ml dynlink_compilerlibs/local_store.ml dynlink_compilerlibs/load_path.ml dynlink_compilerlibs/int_replace_polymorphic_compare.ml dynlink_compilerlibs/lazy_backtrack.ml dynlink_compilerlibs/location.ml dynlink_compilerlibs/longident.ml dynlink_compilerlibs/docstrings.ml dynlink_compilerlibs/syntaxerr.ml dynlink_compilerlibs/ast_helper.ml dynlink_compilerlibs/ast_mapper.ml dynlink_compilerlibs/attr_helper.ml dynlink_compilerlibs/builtin_attributes.ml dynlink_compilerlibs/ident.ml dynlink_compilerlibs/path.ml dynlink_compilerlibs/primitive.ml dynlink_compilerlibs/type_immediacy.ml dynlink_compilerlibs/shape.ml dynlink_compilerlibs/types.ml dynlink_compilerlibs/btype.ml dynlink_compilerlibs/subst.ml dynlink_compilerlibs/predef.ml dynlink_compilerlibs/datarepr.ml dynlink_compilerlibs/cmi_format.ml dynlink_compilerlibs/persistent_env.ml dynlink_compilerlibs/env.ml dynlink_compilerlibs/debuginfo.ml dynlink_compilerlibs/lambda.ml dynlink_compilerlibs/runtimedef.ml dynlink_compilerlibs/instruct.ml dynlink_compilerlibs/opcodes.ml dynlink_compilerlibs/bytesections.ml dynlink_compilerlibs/dll.ml dynlink_compilerlibs/meta.ml dynlink_compilerlibs/symtable.ml dynlink_compilerlibs/asttypes.ml dynlink_compilerlibs/parsetree.ml dynlink_compilerlibs/outcometree.ml dynlink_compilerlibs/cmo_format.ml dynlink_compilerlibs/cmxs_format.ml \
:info:build 	dynlink_compilerlibs/binutils.mli dynlink_compilerlibs/config.mli dynlink_compilerlibs/build_path_prefix_map.mli dynlink_compilerlibs/misc.mli dynlink_compilerlibs/identifiable.mli dynlink_compilerlibs/numbers.mli dynlink_compilerlibs/arg_helper.mli dynlink_compilerlibs/clflags.mli dynlink_compilerlibs/profile.mli dynlink_compilerlibs/consistbl.mli dynlink_compilerlibs/terminfo.mli dynlink_compilerlibs/warnings.mli dynlink_compilerlibs/local_store.mli dynlink_compilerlibs/load_path.mli dynlink_compilerlibs/int_replace_polymorphic_compare.mli dynlink_compilerlibs/lazy_backtrack.mli dynlink_compilerlibs/location.mli dynlink_compilerlibs/longident.mli dynlink_compilerlibs/docstrings.mli dynlink_compilerlibs/syntaxerr.mli dynlink_compilerlibs/ast_helper.mli dynlink_compilerlibs/ast_mapper.mli dynlink_compilerlibs/attr_helper.mli dynlink_compilerlibs/builtin_attributes.mli dynlink_compilerlibs/ident.mli dynlink_compilerlibs/path.mli dynlink_compilerlibs/primitive.mli dynlink_compilerlibs/type_immediacy.mli dynlink_compilerlibs/shape.mli dynlink_compilerlibs/types.mli dynlink_compilerlibs/btype.mli dynlink_compilerlibs/subst.mli dynlink_compilerlibs/predef.mli dynlink_compilerlibs/datarepr.mli dynlink_compilerlibs/cmi_format.mli dynlink_compilerlibs/persistent_env.mli dynlink_compilerlibs/env.mli dynlink_compilerlibs/debuginfo.mli dynlink_compilerlibs/lambda.mli dynlink_compilerlibs/runtimedef.mli dynlink_compilerlibs/instruct.mli dynlink_compilerlibs/opcodes.mli dynlink_compilerlibs/bytesections.mli dynlink_compilerlibs/dll.mli dynlink_compilerlibs/meta.mli dynlink_compilerlibs/symtable.mli dynlink_compilerlibs/asttypes.mli dynlink_compilerlibs/parsetree.mli dynlink_compilerlibs/outcometree.mli dynlink_compilerlibs/cmo_format.mli dynlink_compilerlibs/cmxs_format.mli \
:info:build         > dynlink_compilerlibs/.depend
:info:build /bin/sh: line 1: 35407 Bus error               ../../ocamlc.opt -depend -slash -I dynlink_compilerlibs dynlink_compilerlibs/binutils.ml dynlink_compilerlibs/config.ml dynlink_compilerlibs/build_path_prefix_map.ml dynlink_compilerlibs/misc.ml dynlink_compilerlibs/identifiable.ml dynlink_compilerlibs/numbers.ml dynlink_compilerlibs/arg_helper.ml dynlink_compilerlibs/clflags.ml dynlink_compilerlibs/profile.ml dynlink_compilerlibs/consistbl.ml dynlink_compilerlibs/terminfo.ml dynlink_compilerlibs/warnings.ml dynlink_compilerlibs/local_store.ml dynlink_compilerlibs/load_path.ml dynlink_compilerlibs/int_replace_polymorphic_compare.ml dynlink_compilerlibs/lazy_backtrack.ml dynlink_compilerlibs/location.ml dynlink_compilerlibs/longident.ml dynlink_compilerlibs/docstrings.ml dynlink_compilerlibs/syntaxerr.ml dynlink_compilerlibs/ast_helper.ml dynlink_compilerlibs/ast_mapper.ml dynlink_compilerlibs/attr_helper.ml dynlink_compilerlibs/builtin_attributes.ml dynlink_compilerlibs/ident.ml dynlink_compilerlibs/path.ml dynlink_compilerlibs/primitive.ml dynlink_compilerlibs/type_immediacy.ml dynlink_compilerlibs/shape.ml dynlink_compilerlibs/types.ml dynlink_compilerlibs/btype.ml dynlink_compilerlibs/subst.ml dynlink_compilerlibs/predef.ml dynlink_compilerlibs/datarepr.ml dynlink_compilerlibs/cmi_format.ml dynlink_compilerlibs/persistent_env.ml dynlink_compilerlibs/env.ml dynlink_compilerlibs/debuginfo.ml dynlink_compilerlibs/lambda.ml dynlink_compilerlibs/runtimedef.ml dynlink_compilerlibs/instruct.ml dynlink_compilerlibs/opcodes.ml dynlink_compilerlibs/bytesections.ml dynlink_compilerlibs/dll.ml dynlink_compilerlibs/meta.ml dynlink_compilerlibs/symtable.ml dynlink_compilerlibs/asttypes.ml dynlink_compilerlibs/parsetree.ml dynlink_compilerlibs/outcometree.ml dynlink_compilerlibs/cmo_format.ml dynlink_compilerlibs/cmxs_format.ml dynlink_compilerlibs/binutils.mli dynlink_compilerlibs/config.mli dynlink_compilerlibs/build_path_prefix_map.mli dynlink_compilerlibs/misc.mli dynlink_compilerlibs/identifiable.mli dynlink_compilerlibs/numbers.mli dynlink_compilerlibs/arg_helper.mli dynlink_compilerlibs/clflags.mli dynlink_compilerlibs/profile.mli dynlink_compilerlibs/consistbl.mli dynlink_compilerlibs/terminfo.mli dynlink_compilerlibs/warnings.mli dynlink_compilerlibs/local_store.mli dynlink_compilerlibs/load_path.mli dynlink_compilerlibs/int_replace_polymorphic_compare.mli dynlink_compilerlibs/lazy_backtrack.mli dynlink_compilerlibs/location.mli dynlink_compilerlibs/longident.mli dynlink_compilerlibs/docstrings.mli dynlink_compilerlibs/syntaxerr.mli dynlink_compilerlibs/ast_helper.mli dynlink_compilerlibs/ast_mapper.mli dynlink_compilerlibs/attr_helper.mli dynlink_compilerlibs/builtin_attributes.mli dynlink_compilerlibs/ident.mli dynlink_compilerlibs/path.mli dynlink_compilerlibs/primitive.mli dynlink_compilerlibs/type_immediacy.mli dynlink_compilerlibs/shape.mli dynlink_compilerlibs/types.mli dynlink_compilerlibs/btype.mli dynlink_compilerlibs/subst.mli dynlink_compilerlibs/predef.mli dynlink_compilerlibs/datarepr.mli dynlink_compilerlibs/cmi_format.mli dynlink_compilerlibs/persistent_env.mli dynlink_compilerlibs/env.mli dynlink_compilerlibs/debuginfo.mli dynlink_compilerlibs/lambda.mli dynlink_compilerlibs/runtimedef.mli dynlink_compilerlibs/instruct.mli dynlink_compilerlibs/opcodes.mli dynlink_compilerlibs/bytesections.mli dynlink_compilerlibs/dll.mli dynlink_compilerlibs/meta.mli dynlink_compilerlibs/symtable.mli dynlink_compilerlibs/asttypes.mli dynlink_compilerlibs/parsetree.mli dynlink_compilerlibs/outcometree.mli dynlink_compilerlibs/cmo_format.mli dynlink_compilerlibs/cmxs_format.mli > dynlink_compilerlibs/.depend
:info:build ../../ocamlc.opt -g -nostdlib -I ../../stdlib -c -for-pack Dynlink_compilerlibs -strict-sequence -principal -absname -w +a-4-9-40-41-42-44-45-48-66-70 -bin-annot -safe-string -strict-formats -I byte \
:info:build           -I dynlink_compilerlibs -o dynlink_compilerlibs/binutils.cmo dynlink_compilerlibs/binutils.ml
:info:build ../../ocamlc.opt -g -nostdlib -I ../../stdlib -c -for-pack Dynlink_compilerlibs -strict-sequence -principal -absname -w +a-4-9-40-41-42-44-45-48-66-70 -bin-annot -safe-string -strict-formats -I byte \
:info:build           -I dynlink_compilerlibs -o dynlink_compilerlibs/config.cmo dynlink_compilerlibs/config.ml
:info:build ../../ocamlc.opt -g -nostdlib -I ../../stdlib -c -for-pack Dynlink_compilerlibs -strict-sequence -principal -absname -w +a-4-9-40-41-42-44-45-48-66-70 -bin-annot -safe-string -strict-formats -I byte \
:info:build           -I dynlink_compilerlibs -o dynlink_compilerlibs/build_path_prefix_map.cmo dynlink_compilerlibs/build_path_prefix_map.ml
:info:build ../../ocamlc.opt -g -nostdlib -I ../../stdlib -c -for-pack Dynlink_compilerlibs -strict-sequence -principal -absname -w +a-4-9-40-41-42-44-45-48-66-70 -bin-annot -safe-string -strict-formats -I byte \
:info:build           -I dynlink_compilerlibs -o dynlink_compilerlibs/misc.cmo dynlink_compilerlibs/misc.ml
:info:build make[4]: *** [dynlink_compilerlibs/build_path_prefix_map.cmo] Bus error
:info:build make[4]: *** Waiting for unfinished jobs....
:info:build make[4]: *** [dynlink_compilerlibs/misc.cmo] Bus error
:info:build make[4]: *** [dynlink_compilerlibs/binutils.cmo] Bus error
:info:build make[4]: *** [dynlink_compilerlibs/config.cmo] Bus error
:info:build make[4]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_ocaml/ocaml/work/ocaml-4.14.0/otherlibs/dynlink'
:info:build make[3]: *** [all] Error 2
:info:build make[3]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_ocaml/ocaml/work/ocaml-4.14.0/otherlibs'
:info:build make[2]: *** [otherlibraries] Error 2
:info:build make[2]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_ocaml/ocaml/work/ocaml-4.14.0'
:info:build make[1]: *** [opt.opt] Error 2
:info:build make[1]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_ocaml/ocaml/work/ocaml-4.14.0'
:info:build make: *** [world.opt] Error 2
:info:build make: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_ocaml/ocaml/work/ocaml-4.14.0'
:info:build Command failed:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_ocaml/ocaml/work/ocaml-4.14.0" && /usr/bin/make -j4 -w world.opt 
:info:build Exit code: 2
:error:build Failed to build ocaml: command execution failed
:debug:build Error code: CHILDSTATUS 31182 2
:debug:build Backtrace: command execution failed
:debug:build     while executing
:debug:build "system {*}$notty {*}$callback {*}$nice $fullcmdstring"
:debug:build     invoked from within
:debug:build "command_exec -callback portprogress::target_progress_callback build"
:debug:build     (procedure "portbuild::build_main" line 8)
:debug:build     invoked from within
:debug:build "$procedure $targetname"
```

gasche:
`dynlink_compilerlibs` is probably the first point in the compiler built where the just-built `ocamlc.opt` is used to in turn build files. I believe that the errors suggest that your `ocamlc.opt` (this is the bytecode compiler, compiled itself with the native compiler) does not work properly -- it may be that running `./ocamlc.opt --help` also fails with a "Bus Error"? -- but for me it is difficult to say more from those reports.

gasche:
Note: (not related to the errors you are asking about, just a general point.) On some occasions the Debian people have stopped supported the native OCaml compiler on some of their architectures, but kept supporting the bytecode compiler which is much easier to port and maintain -- but generates slower programs. Would that be an option for you? Basically, if you configure with `--disable-native-compiler`, do you get a working system that passes the testsuite?

I think that it would be a sensible option if your objective is not to use darwin-PPC machines as development environments for OCaml, but rather simply to use end-user programs/packages implemented in OCaml, and there you are willing to accept a performance hit. (Bytecode-compiled programs are typically 5x slower than native-compiled programs.)

barracuda156:
> Note: (not related to the errors you are asking about, just a general point.) On some occasions the Debian people have stopped supported the native OCaml compiler on some of their architectures, but kept supporting the bytecode compiler which is much easier to port and maintain -- but generates slower programs. Would that be an option for you? Basically, if you configure with `--disable-native-compiler`, do you get a working system that passes the testsuite?
> 
> I think that it would be a sensible option if your objective is not to use darwin-PPC machines as development environments for OCaml, but rather simply to use end-user programs/packages implemented in OCaml, and there you are willing to accept a performance hit. (Bytecode-compiled programs are typically 5x slower than native-compiled programs.)



> Note: (not related to the errors you are asking about, just a general point.) On some occasions the Debian people have stopped supported the native OCaml compiler on some of their architectures, but kept supporting the bytecode compiler which is much easier to port and maintain -- but generates slower programs. Would that be an option for you? Basically, if you configure with `--disable-native-compiler`, do you get a working system that passes the testsuite?
> 
> I think that it would be a sensible option if your objective is not to use darwin-PPC machines as development environments for OCaml, but rather simply to use end-user programs/packages implemented in OCaml, and there you are willing to accept a performance hit. (Bytecode-compiled programs are typically 5x slower than native-compiled programs.)

@gasche Thank you for responding! Yes, absolutely, this will do. Speed is not an issue of concern in this case.
`--disable-native-compiler` implies building `world` target, not `world.opt`, right?

`world` is building already, though I am not sure about tests at the moment. It worked to build `opam` though, so not completely broken at least.

I will try your suggestion.

barracuda156:
@gasche @pmetzger So here are test results:
```
List of failed tests:
    tests/lib-unix/kill/'unix_kill.ml' with 1.1 (bytecode) 
    tests/lib-threads/'signal.ml' with 1.1.1.1.1.1.1.1 (check-program-output) 
    tests/lib-threads/'delayintr.ml' with 1.1.1.1.1.1.1.1 (check-program-output) 

List of unexpected errors:
    tests/lib-systhreads/'eintr.ml' with 1.1.1 (bytecode) 

Summary:
  2083 tests passed
  513 tests skipped
    3 tests failed
  531 tests not started (parent test skipped or failed)
    1 unexpected errors
  3131 tests considered
#### Something failed. Exiting with error status.
```

gasche:
This sounds much better to me indeed. The remaining failures look related to signals or pthread, they may be the kind of things that differs on a different kernel, but I'm surprised that they would work on darwin/{amd,arm}64 and fail on darwin/ppc. Who knows; it may just be that darwin/ppc machines have an older userland whose pthread implementation is less compatible.

I think you could *reasonably*:
- ignore the testsuite failure
- or patch the tests to disable them on darwin/ppc (but I'm not 100% sure how to do this)
- or try to fix the issues, but this requires expertise I don't have

It might be possible to disable those tests by tweaking the following in configure.ac:

```
AS_CASE([$host],
  [*-*-mingw32|*-pc-windows],
    [unix_or_win32="win32"
    unixlib="win32unix"
    ],
  [unix_or_win32="unix"
  unixlib="unix"])
AS_CASE([$host],
  [*-*-cygwin*|*-*-mingw32|*-pc-windows],
    [exeext=".exe"],
  [exeext=''])
```

If you make sure that darwin/ppc ends up with `unix_or_win32="darwin"` (or something else different from `unix`) but `unixlib` remains `unix`), then things might magically work. I have not tried. I'm not sure it is worth patching just to disable some tests in the testsuite.

barracuda156:
> This sounds much better to me indeed. The remaining failures look related to signals or pthread, they may be the kind of things that differs on a different kernel, but I'm surprised that they would work on darwin/{amd,arm}64 and fail on darwin/ppc. Who knows; it may just be that darwin/ppc machines have an older userland whose pthread implementation is less compatible.

@gasche I recall I have encountered few threads-related errors with other ports on PPC. AFAIK threads were properly implemented around 10.7, and for PPC the best we got is either 10.5.8 or half-way-through developer release of 10.6. So something may work and something may not.

> I think you could _reasonably_:
> * ignore the testsuite failure
> * or patch the tests to disable them on darwin/ppc (but I'm not 100% sure how to do this)
> * or try to fix the issues, but this requires expertise I don't have
> If you make sure that darwin/ppc ends up with `unix_or_win32="darwin"` (or something else different from `unix`) but `unixlib` remains `unix`), then things might magically work. I have not tried. I'm not sure it is worth patching just to disable some tests in the testsuite.

Thank you! Probably fixing tests isn’t really crucial indeed.

Do I get it correctly, that unless native compiler is properly fixed (and then `world.opt` enabled), arch-specific files are not involved at all (i.e. `asmcomp/"arch"` and `runtime/"arch".S`)? In which case it is enough for now to patch `configure` and add config args in a portfile.

gasche:
Yes, those files are only used for the native compilers. Passing `--no-native-compiler` should suffice to configure OCaml in bytecode-only mode, which only installs `ocamlc`. Note that some user packages may assume that `ocamlopt` (the native compiler) is provided, and thus break to build if they haven't been tweaked to support the bytecode-only compilation. The Debian people have experience packaging bytecode-only on some architectures (mips, sparc I suppose?), so presumably all the Debian packages work in this mode. (It's also possible that their downstream patches would be of interest to you.)

barracuda156:
> Yes, those files are only used for the native compilers. Passing `--no-native-compiler` should suffice to configure OCaml in bytecode-only mode, which only installs `ocamlc`. Note that some user packages may assume that `ocamlopt` (the native compiler) is provided, and thus break to build if they haven't been tweaked to support the bytecode-only compilation. The Debian people have experience packaging bytecode-only on some architectures (mips, sparc I suppose?), so presumably all the Debian packages work in this mode. (It's also possible that their downstream patches would be of interest to you.)

@gasche Great, thanks! Yes, I have bumped into complaints of some ports that depend on `ocaml` re missing `ocamlopt` when I wanted to test my compiler. So at least in principle `ocamlc` should be able to build everything, though substantially slower? (I will check Debian patches, thank you for the tip.)

@pmetzger What do you think if we just include a patch to `configure` for Darwin PPC to be recognized and dedicated flags in the portfile? That will be a minimal change, and it will allow to build a usable compiler on PPC systems (which has been broken for ages in Macports).

barracuda156:
I am also not giving up on fixing the native compiler – don’t want to have spent time in vain, but that may take some time. Need to figure out some PPC assembler specifics and see to what extent @4.14.0 code for POWER is compatible with Darwin. _Honestly, I am not even sure it was correctly implemented in `ocaml` earlier._

P. S. Due to the linker error which does not seem to be caused by broken native compiler, if the latter succeeds, likely it has to be built with `--disable-shared`.

gasche:
If you have a given piece of OCaml code, you can run it by either compiling it with `ocamlopt`, which gives a native executable for the users' platform (if supported), for example an ELF executable on Linux, or with `ocamlc`, which produces a bytecode program (not a native executable), which must be interpreted by the `ocamlrun` bytecode interpreter (written in C, very portable, works on most systems).

For end-user programs implemented in OCaml, users don't really see the different, they can basically click on the resulting binary and it will launch and do stuff (but the ocamlc-compiled version will be slower). The difference is:
- for people willing to develop in OCaml, maybe they want to work with `ocamlopt` (for example to benchmark their programs in the real-world production case)
- for OCaml packages, maybe the build system is robust and can work with either compilers, or maybe it's a hard-coded script that calls `ocamlopt` directly and that fails

kencu:
if ocamlopt is frequently called by scripts but unavailable on PPC, it wouldn’t be horrible to make it a symlink to ocamlc to make things work more smoothly.

Assuming those accept similar flags, etc.

gasche:
No, they don't accept the same flags. But again, Debian people have experience packaging OCaml with architecture without a native backend, and it seems to work okay for them, so I think that could be fine.

barracuda156:
> No, they don't accept the same flags. But again, Debian people have experience packaging OCaml with architecture without a native backend, and it seems to work okay for them, so I think that could be fine.

@gasche Somewhat surprisingly, looks like NetBSD `ocaml` port has Darwin PPC specifics: https://github.com/NetBSD/pkgsrc/blob/trunk/lang/ocaml/Makefile

pmetzger:
@barracuda156 I think using the bytecode compiler / runtime for PPC would be just fine and I support that. We shouldn't have any real problem with this anyway.

barracuda156:
@pmetzger Great!

Should I make a separate PR without assembler patches? Or just remove them here?

(The current state of this PR will actually only build bytecode compiler, which tests fine (results above). That is, assembler patch does not hurt, but it is incomplete and requires further work. So perhaps better not include it until it is properly fixed. I will try that, but it may not be fast.)

pmetzger:
Let's do this in several parts:

1. Add the patches to use the bytecode compiler on PPC
2. Upgrade to 4.14
3. Possibly do something about the assembler patch.

Upgrading to 4.14.0 will also require bumping and upgrading a bunch of stuff in the ocaml/ category.

barracuda156:
> Let's do this in several parts:
> 
> 1. Add the patches to use the bytecode compiler on PPC

@pmetzger Done here: https://github.com/macports/macports-ports/pull/14845
I have built it now on 10.6.8 Rosetta and used it to build `opam`. All good.

I added condition for `clang` dependency, otherwise Macports tried to force it, which gonna fail on PPC.

pmetzger:
Okay, perhaps this should now be closed given that the PPC bytecode interpreter fix is in, and a 4.14.0 patch should be separately submitted?

barracuda156:
@pmetzger Yes, closing this for now (if assembler is fixed later, that should be a new PR anyway).

pmetzger:
That's not a problem, but I might update everything to 4.14.0 some time soon.

barracuda156:
> That's not a problem, but I might update everything to 4.14.0 some time soon.

Great!

barracuda156:
@pmetzger BTW apparently that was some issue with an earlier `gcc`, I have built now `ocaml` 4.12.1 with `gcc12` on 10.6.8 Rosetta from existing portfile, it worked fine (that is, the problematic flag `--enable-imprecise-c99-float-ops` is unnecessary).

PR #14729, opened 2022-04-30
Remove py-isce2 port due to lack of users

piyushrpt:
#### Description

- isce2 has been updated numerous times since this port was last updated.
- Primary maintainers also have moved away from osx due to changes in architecture and other work related limitations
- No issues have been raised on trac or github related to this port
- We believe, this port can be safely deprecated
- Thank you to the ports community for supporting our work.


mazhe:
I agree with @piyushrpt assesment and this proposal, my own usecases also do not involve macos installations.

Thanks

mascguy (approved):
Searched existing portfiles, and no references exist for this port. So we're good-to-go.

Also rebased against master, so that we can merge.

PR #14755, opened 2022-05-04
FAudio: restore port and update to v22.05

richardg867:
#### Description

The FAudio port was previously removed by #13875 due to redundancy; however, we at [86Box/86Box](https://github.com/86Box/86Box) actually use both MacPorts and FAudio. This PR restores the port and updates it to the latest 22.05 version.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.3.1 21E258 arm64
Xcode 13.3.1 13E500a

###### Verification
Have you


mojca:
Thank you very much for the contribution. This looks good enough to be brought back.

A question to fellow MacPorts developers (@richardg867: please don't blindly follow my suggestion until other agree). Does it sound reasonable to do this in two commits:
* git revert 13e3e185af1939706d8f044ca33961d0b30cbc0e (optionally/potentially amend the message with the reference to the PR)
* update the port and do `git commit -m "FAudio: update to v22.05"`

That way we would keep a clear reference to the commit that deleted the port and immediately see the changes.

@Gcenx

mojca:
@richardg867: I implemented the suggested changes and pushed them to your branch.

I just wanted to point out something: it's generally a super bad idea to create pull requests from your master branch. Until this gets merged, maybe avoid pulling (without rebasing) from the upstream master branch.

audio/FAudio/Portfile
```diff
@@ -18,28 +18,18 @@ long_description        an XAudio reimplementation that focuses solely on develo
...
-                        sha256  5599d4351b716e1dc8716b3ddd5855fddbf7a48f47558e04648a8a731cb39637 \
-                        size    1134780
+checksums               rmd160  dcdea2f9212da074dbae6485c250d78b3fa7d0b3 \
+                        sha256  d80a9598ec9e94652e8b7f5cf503b7c9792c8f905e63c9d93aa5d03a8efa2cf5 \
+                        size    1109448
 
 # remove set deployment target and hard-coded RPATH setting
 patchfiles              patch-faudio-remove-deployment-target.diff
 
-configure.args          -DGSTREAMER=OFF \
```
mojca: Is this option gone from FAudio now or did we just remove it from the port?
If it's not gone, maybe we can explicitly set it to ON or OFF (it's not exactly clear to me from the logic below what it should be).

mojca:
> After this gets restored it would be nice if you’d also submit a formal port for 86Box. As then at least something will be using it in macports itself along with giving you more feedback on multiple versions of macOS via Macports buildbots.

I fully agree.

audio/FAudio/Portfile
Gcenx: This was removed from FAudio after v22.01, the only real consumer of FAudio + gstreamer was wine.

Wine now imported the FAudio sources and builds as PE format where access to other codecs is handled by other wine components.

mojca (approved):

richardg867:
Thanks for applying the requested changes.

> I just wanted to point out something: it's generally a super bad idea to create pull requests from your master branch. Until this gets merged, maybe avoid pulling (without rebasing) from the upstream master branch.

I'm aware; I've somehow failed to create a new branch, and only noticed the change was on master when I went to make the pull request.

> After this gets restored it would be nice if you’d also submit a formal port for 86Box. As then at least something will be using it in macports itself along with giving you more feedback on multiple versions of macOS via Macports buildbots.

I will discuss this with the project's other maintainers. Keeping the port up to date with our monthly release schedule will be a challenge, because none of us have Macs (I did all the testing for this port on a remote build node).

PR #14807, opened 2022-05-10
qemu: ppc32 is not a supported host

evanmiller:
#### Description

Building on ppc32 fails with an error that requires assembly programming to fix. Disable it.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11

###### Verification 
Have you

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -16,6 +16,9 @@ platforms               darwin
 maintainers             {raimue @raimue} \
                         openmaintainer
 
+# ppc32 is not a supported host on Darwin
+supported_archs         i386 x86_64 ppc64 arm64
+
 description             Emulator for various architectures
 long_description        QEMU is a generic and open source machine emulator. \
                         It can run OSes and programs made for one machine on a \


raimue:
Is there a report that shows the build failure on pcc? I think it used to work in the past...

evanmiller:
qemu 7.0 fails to configure on PPC32:

```
../qemu-7.0.0/common-user/meson.build:1:0 ERROR: Include dir host/ppc does not exist.
```

With some patching I got qemu 6.2 to build, but running x86 code always terminated with `SIGILL`.

barracuda156:
> qemu 7.0 fails to configure on PPC32:
> 
> ```
> ../qemu-7.0.0/common-user/meson.build:1:0 ERROR: Include dir host/ppc does not exist.
> ```

@evanmiller Does it have `ppc64` Big endian? If yes, then `ppc32` may be feasible/easy to fix. If not, then no luck, perhaps.

mascguy:
Just wanted to quickly check in, and see where things stand with this PR?

barracuda156:
> Just wanted to quickly check in, and see where things stand with this PR?

@mascguy It is not clear to me why we just don’t fix `ppc32` :)
Let me see what’s there in the source.

barracuda156:
@mascguy If the whole problem is with https://github.com/macports/macports-ports/pull/14807?notification_referrer_id=NT_kwDOBXwLlrMzNTkzMzIzOTc1OjkyMDE1NTEw#issuecomment-1126991563 then it's a tiny file, we can certainly deal with that.

barracuda156:
@evanmiller I don’t have time right now to try building the whole thing, but could you check if this results in a sensible output?
https://github.com/macports/macports-ports/compare/master...barracuda156:qemu

barracuda156:
@evanmiller Have you had a chance to test that?

barracuda156:
> Is there a report that shows the build failure on pcc? I think it used to work in the past...

@raimue Will this work for `ppc32`? https://github.com/macports/macports-ports/compare/master...barracuda156:qemu

mascguy:
So... where do things stand?

barracuda156:
> So... where do things stand?

@mascguy I hoped @evanmiller gonna reply, but nothing so far. Myself I can do testing only by second half of August. (Still away from TW.)

I would suggest to postpone removing PPC support, since it is quite likely restorable, if not with my patch here then with a similar one.

evanmiller:
> @evanmiller I don’t have time right now to try building the whole thing, but could you check if this results in a sensible output?
> [master...barracuda156:qemu](https://github.com/macports/macports-ports/compare/master...barracuda156:qemu)

Building now...

evanmiller:
The build is successful with @barracuda156's patches (combined with some of my local patches disabling CoreAudio on Tiger). However, as with previous versions of qemu, I get a `SIGILL` (Illegal instruction) when attempting to run `qemu-system-i386-unsigned` and `qemu-system-x86_64-unsigned` on a PPC32 machine. So I stand by this PR.

barracuda156:
@evanmiller Thanks for checking. Well, I’m away from PPC hardware until second half of August, so cannot run a test on Leopard now, but I think it makes sense to confirm a failure on 10.5.8 prior to removing PPC support altogether. (I have no specific interest in Qemu, so if fixing existing code fails, I probably won’t pursue it any further.)

barracuda156:
@evanmiller Maybe report the issue to the upstream? Apparently their code is broken.

raimue:
It looks like we cannot resolve this quickly with a patch... Shall we move the restoration of ppc32 support into a Trac ticket? And apply the one proposed here in the PR with an additional reference to that ticket?

evanmiller:
It's worth noting that I am testing on a G3 – it's possible that it works on G4, so it might be worth holding off if someone can test that processor.

mascguy:
> It's worth noting that I am testing on a G3 – it's possible that it works on G4, so it might be worth holding off if someone can test that processor.

I have a PowerBook G4 running 10.4 and 10.5, so I might be able to test there.

Assuming that the various dependencies, as well as `qemu` itself, don't require days to build... I can help with testing.

mascguy:
> I have a PowerBook G4 running 10.4 and 10.5, so I might be able to test there.

With the patch from @barracuda156 , I'm seeing a build failure. Will post details shortly, once I've transferred the logfile.

Just to confirm, I'm building with just one variant enabled: `+target_ppc`. (With the Intel variants - and `usb` - all explicitly disabled.) Is that correct for a PPC machine?

barracuda156:
@mascguy Just to be clear, does it build without my patch?

(I did not verify myself if the build succeeds, and will be able to do that only by end of August.)

barracuda156:
> Great point, I'll try to retry this today, if possible.

@mascguy Thanks!
Could you share a log of a failure btw? If it’s smth with my patch code, I code look at it.

mascguy:
> @mascguy Thanks! Could you share a log of a failure btw? If it’s with with my patch code, I can look at it.

Sure:

[qemu-build-10.5-ppc-error.log.gz](https://github.com/macports/macports-ports/files/9113152/qemu-build-10.5-ppc-error.log.gz)

barracuda156:
@evanmiller By the way, do we even need these safe calls? The file `safe-syscall.inc.S` has been moved from `linux-user`: https://github.com/qemu/qemu/commit/bbf15aaf7c7506c88062288b3ae122b882f65e69
Now if we check earlier versions of `host`, `safe-syscall.inc.S` was there only for `ppc64`, not for `ppc32`:
https://github.com/qemu/qemu/tree/stable-5.0/linux-user/host
https://github.com/qemu/qemu/tree/stable-6.1/linux-user/host
(Of course I have no idea if it ever worked at all.)
If this file is not required, it can be just removed.

P. S. Instead of breaking PPC in Macports I actually suggest restoring `ppc32` in the upstream. Even if `qemu` does not work on PPC as is, that can be a good start to fix it. Restoring seems trivial and perhaps no one would object. Adding Darwin version of `ppc64` may be a bit more tricky, but still doable, as long as no one objects.

mascguy:
Haven't forgotten about this, I'll try to revisit soon...

barracuda156:
> Haven't forgotten about this, I'll try to revisit soon...

@mascguy Yeah, let’s keep the PR in draft but not close. I will revisit it in a while again. Rosetta is of little use for this case, I need PPC hardware.

hamishmb:
I have an iBook G4 if you need someone with hardware to test this.

I haven't done anything even remotely low-level, so I'd probably need some handholding, but I'd be happy to test when I get some time.

reneeotten:
the `qemu` port is currently at version 8.0.2 - is anything here still relevant? If so, please update the PR for this newer version, otherwise please close

barracuda156:
@reneeotten I will rebase this. Perhaps also worth opening an issue with the upstream.

barracuda156:
@reneeotten Trying to fix the build now. If it works, I will rebase, adding needed fix-ups. What we got now will not build as is on ppc.

barracuda156:
@mascguy @evanmiller Okay, I have fixed the build itself, but – unsurprisingly so – it fails to work. Possibly default choices are wrong, if so, it might work. Or possibly everything is fundamentally broken.
signal-2023-08-19-015223_002

mascguy:
> @mascguy @evanmiller Okay, I have fixed the build itself, but – unsurprisingly so – it fails to work. Possibly default choices are wrong, if so, it might work. Or possibly everything is fundamentally broken.

You're definitely making progress! And on the flip side, it's possible that there's only one - or a few - small wrinkles to work out.

Perhaps Evan or @kencu might have some thoughts/suggestions...

barracuda156:
> it's possible that there's only one - or a few - small wrinkles to work out.

Let’s hope!

> Perhaps Evan or @kencu might have some thoughts/suggestions...

For the reference, updated patches are here: https://github.com/macports/macports-ports/pull/19940

P. S. Make `ucontext` should be used instead for coroutine? But by default it does not configure with it, complaining about missing `makecontext`.

pmetzger:
I think we should close this. Objections?

PR #14816, opened 2022-05-11
python39-310: fix +lto PPC, python38-310: add SLPPC case

barracuda156:
#### Description

Fixes `+lto` variant for PPC, adds a case of building for `ppc` on 10.6.x.
My trac ticket : https://trac.macports.org/ticket/64904

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

macOS 10.6 PPC (10A190)
Xcode 3.2

macOS 10.5.8
Xcode 3.1.4

###### Verification 
Have you

### DIFF
--- a/lang/python310/Portfile
+++ b/lang/python310/Portfile
@@ -42,6 +42,11 @@ if {${os.platform} eq "darwin" && ${os.major} < 10} {
                        patch-threadid-older-systems.diff
 }
 
+if {${os.platform} eq "darwin" && ${os.major} == 10 && ${build_arch} eq "ppc"} {
+    # this is needed for a PPC build on SL
+    patchfiles-append  patch-threadid-older-systems.diff
+}
+
 depends_build       port:pkgconfig
 depends_lib         port:bzip2 \
                     port:expat \
@@ -242,6 +247,10 @@ variant optimizations description {enable expensive, stable optimizations (inclu
 
 variant lto description {enable Link-Time Optimization} {
     configure.args-append   --with-lto
+    if {${build_arch} eq "ppc" || ${build_arch} eq "ppc64"} {
+        compiler.blacklist-append *gcc-4.*
+        patchfiles-append   patch-enable_dynamic.diff
+    }
 }
 
 platform darwin {
--- a/lang/python310/files/patch-enable_dynamic.diff
+++ b/lang/python310/files/patch-enable_dynamic.diff
@@ -1,0 +1,20 @@
+--- configure.orig	2022-03-24 04:12:04.000000000 +0800
++++ configure	2022-03-29 15:03:21.000000000 +0800
+@@ -6755,7 +6755,7 @@
+       case $ac_sys_system in
+         Darwin*)
+           # Any changes made here should be reflected in the GCC+Darwin case below
+-          LTOFLAGS="-flto -Wl,-export_dynamic"
++          LTOFLAGS="-flto"
+           LTOCFLAGS="-flto"
+           ;;
+         *)
+@@ -6766,7 +6766,7 @@
+     *gcc*)
+       case $ac_sys_system in
+         Darwin*)
+-          LTOFLAGS="-flto -Wl,-export_dynamic"
++          LTOFLAGS="-flto"
+           LTOCFLAGS="-flto"
+           ;;
+         *)
--- a/lang/python38/Portfile
+++ b/lang/python38/Portfile
@@ -47,6 +47,11 @@ if {${os.platform} eq "darwin" && ${os.major} < 10} {
                        patch-threadid-older-systems.diff
 }
 
+if {${os.platform} eq "darwin" && ${os.major} == 10 && ${build_arch} eq "ppc"} {
+     # this is needed for a PPC build on SL
+     patchfiles-append  patch-threadid-older-systems.diff
+}
+
 depends_build       port:pkgconfig
 depends_lib         port:bzip2 \
                     port:expat \
--- a/lang/python39/Portfile
+++ b/lang/python39/Portfile
@@ -42,6 +42,11 @@ if {${os.platform} eq "darwin" && ${os.major} < 10} {
                        patch-threadid-older-systems.diff
 }
 
+if {${os.platform} eq "darwin" && ${os.major} == 10 && ${build_arch} eq "ppc"} {
+    # this is needed for a PPC build on SL
+    patchfiles-append  patch-threadid-older-systems.diff
+}
+
 depends_build       port:pkgconfig
 depends_lib         port:bzip2 \
                     port:expat \
@@ -241,6 +246,10 @@ variant optimizations description {enable expensive, stable optimizations (inclu
 
 variant lto description {enable Link-Time Optimization} {
     configure.args-append   --with-lto
+    if {${build_arch} eq "ppc" || ${build_arch} eq "ppc64"} {
+        compiler.blacklist-append *gcc-4.*
+        patchfiles-append   patch-enable_dynamic.diff
+    }
 }
 
 platform darwin {
--- a/lang/python39/files/patch-enable_dynamic.diff
+++ b/lang/python39/files/patch-enable_dynamic.diff
@@ -1,0 +1,20 @@
+--- configure.orig	2022-03-24 05:12:08.000000000 +0800
++++ configure	2022-03-29 14:21:47.000000000 +0800
+@@ -6696,7 +6696,7 @@
+       case $ac_sys_system in
+         Darwin*)
+           # Any changes made here should be reflected in the GCC+Darwin case below
+-          LTOFLAGS="-flto -Wl,-export_dynamic"
++          LTOFLAGS="-flto"
+           LTOCFLAGS="-flto"
+           ;;
+         *)
+@@ -6707,7 +6707,7 @@
+     *gcc*)
+       case $ac_sys_system in
+         Darwin*)
+-          LTOFLAGS="-flto -Wl,-export_dynamic"
++          LTOFLAGS="-flto"
+           LTOCFLAGS="-flto"
+           ;;
+         *)


barracuda156:
@jmroot @ryandesign @kencu Do you think we should rather implement `+lto` fix for `{${os.major} < 14` instead of restricting it to PowerPC? I presume it is failing due to `clang` being broken. With a new `gcc` it should build for Intel too, not just for PPC.

kencu:
rather than add a clang blocker when building for PPC to 12,000 Portfiles, it would sure be nice if base just never brought it forward as a possibility in the first place…

kencu:
at least until Iain or someone fixes clang for PPC … 10 years and counting now…

barracuda156:
> rather than add a clang blocker when building for PPC to 12,000 Portfiles, it would sure be nice if base just never brought it forward as a possibility in the first place…

There is something odd going on specifically with `python3`: if I just write `compiler.blacklist-append` there, Macports just ignores it and uses `gcc-4.2`, which fails. It might not be the case on 10.6 PPC (I am away from that machine atm), but on 10.6.8 I get this problem. What works instead is `compiler.blacklist` with `-append`. Then Macports enforces the ban, as it should, and uses a newer `gcc`. I added `clang` just to be on the safe side: it won’t hurt, as long as `ppc` is concerned. This is not relevant to other ports, it is a contextual thing here.

However if you think we rather fix `+lto` for all systems prior to 10.9, then blacklisting `clang` will be presumably required.

barracuda156:
@kencu So I changed to:
```
variant lto description {enable Link-Time Optimization} {
     configure.args-append   --with-lto
     if {${os.major} < 15} {
         compiler.blacklist  *clang* *gcc-4.*
         patchfiles-append   patch-enable_dynamic.diff
     }
}
```
And built `python310+lto` on 10.6.8 for `x86_64`:
```
macmini:bin svacchanda$ ./port -v installed python310
  python310 @3.10.4_0 requested_variants='' platform='darwin 10' archs='x86_64' date='2022-05-10T04:22:21+0800'
  python310 @3.10.4_1 requested_variants='' platform='darwin 10' archs='x86_64' date='2022-05-10T21:31:47+0800'
  python310 @3.10.4_1+lto (active) requested_variants='+lto' platform='darwin 10' archs='x86_64' date='2022-05-14T17:55:27+0800'
```

kencu:
but we’re not going to blacklist *clang* on all these systems!

barracuda156:
> but we’re not going to blacklist all clangs on all these systems!

@kencu We don’t need to do that in a general case (and I do not do it). It is only for `+lto` variant, which is _not default_ on those systems.

barracuda156:
@kencu But hey, I do not insist on fixing anything for Intel! If you think it is preferable to restrict the fix to PPC, I am okay with that, and will revert the relevant condition to its previous version.

barracuda156:
@kencu @mascguy @jmroot @ryandesign Is the current fix for LTO encompassing old Intel systems controversial? I have tested it on 10.6.8 for x86_64, and it has built successfully. However if there are any objections, I will revert it to PPC-only.

ryandesign:
I haven't been paying attention to this PR. I see you're proposing blacklisting all clangs in the lto variant on 10.7 and earlier, while the ticket this PR claims to close only discusses an issue that occurs building for PowerPC. Blacklisting all clangs on an Intel system is suspicious and would not normally be the kind of thing we'd want to do since clang is our preferred compiler on Intel systems. I have not looked into this issue, don't know what the problem is that you're fixing, don't know if it can be solved another way that still allows the use of clang.

barracuda156:
> I haven't been paying attention to this PR. I see you're proposing blacklisting all clangs in the lto variant on 10.7 and earlier, while the ticket this PR claims to close only discusses an issue that occurs building for PowerPC. Blacklisting all clangs on an Intel system is suspicious and would not normally be the kind of thing we'd want to do since clang is our preferred compiler on Intel systems. I have not looked into this issue, don't know what the problem is that you're fixing, don't know if it can be solved another way that still allows the use of clang.

@ryandesign OK, let me indeed revert to PPC-only fix. I cannot test Intel stuff on every system.

barracuda156:
@ryandesign @mascguy @kencu I squashed commits and returned to PPC-only fix. No Intel systems are affected in any way.

kencu:
I would **not** support adding a bunch of 10.6 on PPC fixes into macports.

That is a holy mess of a system. 

Opening the door to a bunch of fixes like that is simply a recipe for disaster.

If you want to support 10.6 on PPC for the 10 users who want that, use an overlay.

IMHO. Whoever wants to push this can do as they please, of course.

barracuda156:
> I would **not** support adding a bunch of 10.6 on PPC fixes into macports.
> 
> That is a holy mess of a system.
> 
> Opening the door to a bunch of fixes like that is simply a recipe for disaster.
> 
> If you want to support 10.6 on PPC for the 10 users who want that, use an overlay.
> 
> IMHO. Whoever wants to push this can do as they please, of course.

I will do as you say, of course.

But allow me to express my opinion:
a) Generally speaking, SL PPC equally means a completely standard case of 10.6.x Rosetta (Rosetta may need additional fixes, but for the most part these overlap).
b) Nothing is changed for any system or configuration outside one specific case. It cannot possibly have any adverse effects.
c) It also does not seem an extra burden for anyone to maintain, since the conditional merely applies existing patch in case the build is for `ppc`. Literally no work is required any further from anyone. It will automatically apply the same patch which is already used for Leopard.
d) If this benefits 10 users and hurts no one, it is Pareto-optimal to do :)

kencu:
I am not the boss of anything, so we'll see what everyone says.

We all appreciate your participation!

barracuda156:
> I am not the boss of anything, so we'll see what everyone says.
> 
> We all appreciate your participation!

Thank you.

Be sure, I am not planning to push tons of specific fixes for 10.6 PPC for dozens of ports, and definitely nothing that could potentially break smth for others. Also, I will maintain whatever fixes are accepted, so no extra effort is needed from other developers here.

jmroot:
Please work with upstream to add support for LTO with gcc on Darwin (and threadid on older systems) in a form they find acceptable. Once that happens, we can tweak the compiler selection on our end if necessary.

barracuda156:
> Please work with upstream to add support for LTO with gcc on Darwin (and threadid on older systems) in a form they find acceptable. Once that happens, we can tweak the compiler selection on our end if necessary.

We both know upstream won’t implement any fixes for old systems, and _that’s why Macports uses patches for those instead of “working with upstream”._
This is the level of support one can expect from the upstream – 2016 ticket with a ready patch for `python27`: https://bugs.python.org/issue26487

barracuda156:
P. S. It seems that `python38` has no maintainer atm, unless that is simply an error in the portfile. If anyone thinks `python38` is okay to implement PPC fixes, update me.

There is no LTO variant for `python38`, so that probably won’t work. But it can be made to allow building as `ppc+ppc64` on Leopard, as well as on 10.6 PPC.

Otherwise I give up.

kencu:
I think you might consider making an overlay repo, perhaps called "PPCSnowLeopardPorts" for example, and then you could tell people interested in using MacPorts on a system like 10.6 on PPC to use that overlay.

Then you can change anything you want, whenever you want, without any kind of permission from anyone on MacPorts. You can have issues in the GitHub repo relating to 10.6 on PPC, and interested users can contribute fixes there. 

You can control port updates so that routine version bumps won't cause the whole ecosystem to crash down on you.

And you won't clog up MacPorts, which is already difficult to maintain with it's primary target mix, with a lot of fragile spaghetti code that will break in a few weeks or months with the next upgrade, and which absolutely nobody can hope to test or debug.

I beseech you to be reasonable, see things from the point of view of the 99.99999999% of MacPorts users who don't use 10.6 on PPC and never will, and be aware of the limitations on people's time and abilities.

kencu:
And - 10.6 of PPC is IMHO a huge waste of time, by the way. 

I think you're much better off adding needed support to 10.5 on PPC, like the libblocksruntime port, the newer ld64 with patches from Iain, and someday maybe clang/llvm than you will ever be in trying to mess around with 10.6 on PPC with all it's myriad of shortcomings.

barracuda156:
@kencu The only real advantage of 10.5.8 is `ppc64`, but it is more broken than 10.6 PPC, incomparably so. You know this better than me.

It is impossible to set up and use Macports as `ppc64` on Leopard (because tons of ports are broken for `ppc64` and some never existed for it), and it is painful to use it as `+universal` (again, some stuff is hopelessly broken, some takes weeks or months to fix, and then once you fix that, a PR is just discarded or ignored LOL).
So while in theory Leopard sounds great for G5, in practice – we can't even make `gcc10-bootstrap` work properly there (5th revision failing, apparently no one gonna fix – and I perfectly understand it is non-trivial, it took months for me to fix Pythons, and `gcc` is a far more complex thing).

P. S. We have few users and even fewer testers exactly because of policies that discourage people to make any ongoing endeavors. Sure enough, I can make my own repo and do my stuff there. Macports gets –1 person who could do something for systems which already suffer from scarcity of people able and willing to support them. Further down the road, some 10–20 users who could become engaged in Macports because it offers support for old systems which are not supported elsewhere, won't use it, because setting it up on 10.6 PPC requires manual tweaks here & there and skills which regular users may not possess of. I can see what we lose in this process, even if such loss is marginal – I don't really see what we gain in any sense.

barracuda156:
@kencu
> I beseech you to be reasonable, see things from the point of view of the 99.99999999% of MacPorts users who don't use 10.6 on PPC and never will, and be aware of the limitations on people's time and abilities.

You have seen my code here. LTO patch removed flags which caused the build to fail and banned old `gcc` which does not compile `-flto`. There are hundreds of similar patches across Macports. I can't think of any logical reason this could possibly be seen as controversial.
SLPPC case was using already existing Leopard patch and applying it for a specific configuration. There are exactly zero maintenance costs involved at the margin, as long as Leopard is supported.

kencu:
> @kencu The only real advantage of 10.5.8 is `ppc64`, but it is more broken than 10.6 PPC, incomparably so. You know this better than me.

This is completely wrong.

10.5.8 has all the required system parts as PPC code, and was completely debugged by Apple as best as could be.

10.6 of PPC is a horrible mishmash of mixed up bits, system parts that are Intel only, badly debugged components, and and inconsistent mess of available functions that are neither consistent with the existing 10.5 nor 10.6 systems that are out there in common use.

Just throw it away -- it was a terrible idea for people to get you going down that road in the first place.

kencu:
anyway, you know my opinion for many months now.

kencu:
> You have seen my code here. LTO patch removed flags which caused the build to fail and banned old `gcc` which does not compile `-flto`. There are hundreds of similar patches across Macports. I can't think of any logical reason this could possibly be seen as controversial. SLPPC case was using already existing Leopard patch and applying it for a specific configuration. There are exactly zero maintenance costs involved at the margin, as long as Leopard is supported.

I think you see that youu might well have been able to get in a fix that was of general use to users of 10.4 and 10.5 PPC, and 10.6 Intel, if it was clean and solid and made sense.

However, at one point your PR required clang be outlawed as a build compiler on 10.6, which is not something that could be supported. 

And then all this 10.6 on PPC stuff got thrown in, and I don't think anyone wants to open that door even a crack.

barracuda156:
> > @kencu The only real advantage of 10.5.8 is `ppc64`, but it is more broken than 10.6 PPC, incomparably so. You know this better than me.
> 
> This is completely wrong.
> 
> 10.5.8 has all the required system parts as PPC code, and was completely debugged by Apple as best as could be.

I guess my statement above was unclear. I did not mean 10.5.8 is broken as such, it’s not, of course. I meant that building ports for `ppc64` is broken in Macports. And this fact makes the only – IMHO – real advantage of 10.5.8 irrelevant.
If you or anyone doubts that `ppc64` is broken in Macports – well, I can list specific cases that fail. And there are many.

> 10.6 of PPC is a horrible mishmash of mixed up bits, system parts that are Intel only, badly debugged components, and an inconsistent mess of available functions that are neither consistent with the existing 10.5 nor 10.6 systems that are out there in common use.

I acknowledge that you are far more knowledgeable here, so cannot argue. But I can say as a matter of fact that everything I can do for `ppc32` on Leopard, I can do on 10.6 PPC. There are literally a couple of exceptions (I think `g95` does not build on 10.6 PPC as-is and `qt4-mac`), but there are ports that build on 10.6 PPC and fail on 10.5.8 too.

> Just throw it away -- it was a terrible idea for people to get you going down that road in the first place.

_Literally the only reason_ I got involved with Macports (and software development in general) is 10.6 PPC and your help in that thread on Macrumors. I leave you to judge whether my contribution here is of any value.
I not a developer in my _dayjob_ or education. I am doing this as a hobby, and 10.6 PPC has been the reason I got an interest in it.

barracuda156:
> I think you see that youu might well have been able to get in a fix that was of general use to users of 10.4 and 10.5 PPC, and 10.6 Intel, if it was clean and solid and made sense.
> However, at one point your PR required clang be outlawed as a build compiler on 10.6, which is not something that could be supported.

That was needed for Intel in order to make LTO fix work. `+lto` was not a default option for those systems. `clang` was not banned in a general case, but only in a variant, which otherwise was unavailable anyway. Honestly I can’t see a problem here: you either have no LTO on Intel before 10.9 or get LTO with using `gcc` (`clang` was only blacklisted for a single port variant).
But I am not interested to push any Intel fixes, so consider this matter dropped.

> And then all this 10.6 on PPC stuff got thrown in, and I don't think anyone wants to open that door even a crack.

@kencu @jmroot @mascguy @ryandesign @catap Tagging everyone to bring attention to the matter.

I am okay with throwing away 10.6 PPC line. I can re-do this PR in order to include LTO for PPC fix, which was approved earlier by @mascguy in: https://trac.macports.org/ticket/64904
Everyone knows this won’t be fixed upstream. And if anyone cares about Leopard still, you should be okay with my LTO fix for PPC. To re-iterate, LTO part has nothing to do with 10.6 PPC case.

P. S. @jmroot I apologize personally if my inclusion of 10.6 PPC appeared disturbing. My intention was to fix a problem for few users, given that the fix had zero effect for everyone else. I did not intend to make anyone upset with that.

jmroot:
> There are exactly zero maintenance costs involved at the margin, as long as Leopard is supported.

Patches stop applying cleanly with new versions all the time.

> Everyone knows this won’t be fixed upstream.

You never know if you never try. They have code supporting deployment targets back to 10.3, at least in theory.

> I did not intend to make anyone upset with that.

I'm not upset, FWIW.

kencu:
Fixes in Portfiles that are of a "generally doing it right" nature, which in the process also fix 10.6 on PPC, are certainly 100% acceptable and are much appreciated!

You might very well be in an excellent position to discover areas that need such improvement when trying to build things on nonstandard systems.

barracuda156:
> > There are exactly zero maintenance costs involved at the margin, as long as Leopard is supported.
> 
> Patches stop applying cleanly with new versions all the time.
> 
> > Everyone knows this won’t be fixed upstream.
> 
> You never know if you never try. They have code supporting deployment targets back to 10.3, at least in theory.

@jmroot Thank you for responding.

Will you be okay with LTO fix for PPC being included here in a case upstream does nothing about it? I will file a case with them today.

barracuda156:
> You might very well be in an excellent position to discover areas that need such improvement when trying to build things on nonstandard systems.

At least we have discovered two bugs in `gcc` for `powerpc*` targets (one on 10.5.8 for `ppc64`, another on 10.6 PPC)!

barracuda156:
Ticket with upstream: https://github.com/python/cpython/issues/93045

jmroot:
To be clear, I don't think Python upstream will work on a fix if you just file an issue, you will need to give them a PR. I would advise taking the approach of testing whether the linker supports `-export_dynamic` in the configure script, and then using that option conditionally based on the result.

barracuda156:
@jmroot Thank you. I will try that.

mojca:
PS: I would still submit a pull request if I was you.
Even if the request / report itself gets closed, coming up with a well-tested patch is more likely to succeed, and it would likely be checked by other people in the team as well.

PR #14825, opened 2022-05-12
nss: fix for ppc64 build on Leopard

barracuda156:
#### Description

This finally fixes `nss` for `ppc64` and `+universal` on Leopard.

Closes my own ticket: https://trac.macports.org/ticket/64914
I have reported the problem to upstream, so hopefully we can retire the patch after awhile: https://bugzilla.mozilla.org/show_bug.cgi?id=1769063

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8
Xcode 3.1.4

###### Verification 
Have you

### DIFF
--- a/net/nss/Portfile
+++ b/net/nss/Portfile
@@ -6,7 +6,7 @@ PortGroup           compiler_blacklist_versions 1.0
 
 name                nss
 version             3.78
-revision            0
+revision            1
 set NSS_VMAJOR      [lindex [split ${version} .] 0]
 set NSS_VMINOR      [lindex [split ${version} .] 1]
 set NSS_VPATCH      [lindex [split ${version} .] 2]
@@ -46,11 +46,11 @@ depends_lib     port:nspr \
                 port:zlib \
                 port:sqlite3
 
-destroot.dir ${destroot.dir}/dist
-build.dir    ${build.dir}/nss
+destroot.dir    ${destroot.dir}/dist
+build.dir       ${build.dir}/nss
 
 # external tests require C++11
-build.args   NSS_DISABLE_GTESTS=1
+build.args      NSS_DISABLE_GTESTS=1
 
 # disable use of -Werror
 build.args-append NSS_ENABLE_WERROR=0
@@ -176,6 +176,11 @@ if {${universal_possible} && [variant_isset universal]} {
             system "env DYLD_LIBRARY_PATH=${destroot}${prefix}/lib/nss ${destroot}${prefix}/bin/shlibsign -i ${base}.dylib -o ${chk}"
         }
     }
+}
+
+platform darwin 9 {
+    # PPC64 code in mpcpucache.c is invalid in Mac OSX ABI and causes a build failure
+    patchfiles-append patch-ppc64.diff
 }
 
 platform darwin 8 {
--- a/net/nss/Portfile
+++ b/net/nss/Portfile
@@ -5,8 +5,8 @@ PortGroup           muniversal 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
 name                nss
-version             3.78
-revision            1
+version             3.79
+revision            0
 set NSS_VMAJOR      [lindex [split ${version} .] 0]
 set NSS_VMINOR      [lindex [split ${version} .] 1]
 set NSS_VPATCH      [lindex [split ${version} .] 2]
@@ -28,9 +28,9 @@ set my_release      NSS_[strsed ${version} {g/\./_/}]_RTM
 master_sites        https://ftp.mozilla.org/pub/mozilla.org/security/nss/releases/${my_release}/src/ \
                     ftp://ftp.mozilla.org/pub/mozilla.org/security/nss/releases/${my_release}/src/
 
-checksums           rmd160  5147da7f09cfa3ae2f1a9937a0b6235c98b52692 \
-                    sha256  f455f341e787c1167328e80a84f77b9a557d595066dda6486a1874d72da68800 \
-                    size    84815720
+checksums           rmd160  5e278dd0eb923f4d19bafa2439e2ce97ca4c7e12 \
+                    sha256  ebdf2d6a96613b6fe70ad579e9f983e0e94e0110171cfb2999db633d3394a514 \
+                    size    84830113
 
 # build fails with gcc-4.2 on Intel, but succeeds with gcc-4.2 on PPC
 # cc1: error: unrecognized command line option "-mpclmul"
@@ -46,11 +46,11 @@ depends_lib     port:nspr \
                 port:zlib \
                 port:sqlite3
 
-destroot.dir    ${destroot.dir}/dist
-build.dir       ${build.dir}/nss
+destroot.dir ${destroot.dir}/dist
+build.dir    ${build.dir}/nss
 
 # external tests require C++11
-build.args      NSS_DISABLE_GTESTS=1
+build.args   NSS_DISABLE_GTESTS=1
 
 # disable use of -Werror
 build.args-append NSS_ENABLE_WERROR=0
@@ -179,7 +179,7 @@ if {${universal_possible} && [variant_isset universal]} {
 }
 
 platform darwin 9 {
-    # PPC64 code in mpcpucache.c is invalid in Mac OSX ABI and causes a build failure
+    # See: https://bugzilla.mozilla.org/show_bug.cgi?id=1769063
     patchfiles-append patch-ppc64.diff
 }
 
--- a/net/nss/files/patch-ppc64.diff
+++ b/net/nss/files/patch-ppc64.diff
@@ -1,0 +1,11 @@
+--- nss/lib/freebl/mpi/mpcpucache.c.orig	1970-01-01 08:00:00.000000000 +0800
++++ nss/lib/freebl/mpi/mpcpucache.c	2022-05-12 21:28:37.000000000 +0800
+@@ -705,7 +705,7 @@
+ #define MPI_GET_PROCESSOR_LINE_SIZE_DEFINED 1
+ #endif
+ 
+-#if defined(__ppc64__)
++#if !defined(__APPLE__) && defined(__ppc64__)
+ /*
+  *  Sigh, The PPC has some really nice features to help us determine cache
+  *  size, since it had lots of direct control functions to do so. The POWER
--- a/net/nss/files/patch-ppc64.diff
+++ b/net/nss/files/patch-ppc64.diff
@@ -1,11 +1,15 @@
---- nss/lib/freebl/mpi/mpcpucache.c.orig	1970-01-01 08:00:00.000000000 +0800
-+++ nss/lib/freebl/mpi/mpcpucache.c	2022-05-12 21:28:37.000000000 +0800
-@@ -705,7 +705,7 @@
- #define MPI_GET_PROCESSOR_LINE_SIZE_DEFINED 1
- #endif
+# See: https://bugzilla.mozilla.org/show_bug.cgi?id=1769063
+--- nss/lib/freebl/mpi/mpcpucache.c.orig	2022-05-26 17:54:33.000000000 +0800
++++ nss/lib/freebl/mpi/mpcpucache.c	2022-06-04 07:52:38.000000000 +0800
+@@ -727,10 +727,7 @@
+ static inline void
+ dcbzl(char *array)
+ {
+-    register char *a asm("r2") = array;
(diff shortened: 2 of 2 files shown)


net/nss/Portfile
```diff
@@ -6,7 +6,7 @@ PortGroup           compiler_blacklist_versions 1.0
 
 name                nss
 version             3.78
-revision            0
+revision            1
```
kencu, line 9: no need to bump revision if ppc64 never built

net/nss/Portfile
```diff
@@ -46,11 +46,11 @@ depends_lib     port:nspr \
                 port:zlib \
                 port:sqlite3
 
-destroot.dir ${destroot.dir}/dist
-build.dir    ${build.dir}/nss
+destroot.dir    ${destroot.dir}/dist
+build.dir       ${build.dir}/nss
```
kencu, line 50: don’t alter whitespace here, that’s for a separate commit if needed

net/nss/Portfile
```diff
@@ -46,11 +46,11 @@ depends_lib     port:nspr \
...
 
 # external tests require C++11
-build.args   NSS_DISABLE_GTESTS=1
+build.args      NSS_DISABLE_GTESTS=1
```
kencu, line 53: ditto abt whitespace changed

net/nss/Portfile
```diff
@@ -178,6 +178,11 @@ if {${universal_possible} && [variant_isset universal]} {
     }
 }
 
+platform darwin 9 {
```
kencu, line 181: no need to restrict the fix by platform — if ppc64 never built, and only ppc64 code is being altered, the patch can be always applied without fear of altering other builds

net/nss/files/patch-ppc64.diff
```diff
@@ -0,0 +1,11 @@
+--- nss/lib/freebl/mpi/mpcpucache.c.orig	1970-01-01 08:00:00.000000000 +0800
++++ nss/lib/freebl/mpi/mpcpucache.c	2022-05-12 21:28:37.000000000 +0800
+@@ -705,7 +705,7 @@
+ #define MPI_GET_PROCESSOR_LINE_SIZE_DEFINED 1
+ #endif
+ 
+-#if defined(__ppc64__)
++#if !defined(__APPLE__) && defined(__ppc64__)
```
kencu, line 8: just disabling this code might he OK….as you say, ppc gets by without it. But it looks like this code was supposed to work, reading through the code upstream…

kencu (changes requested):

net/nss/files/patch-ppc64.diff
barracuda156: > just disabling this code might he OK….as you say, ppc gets by without it. But it looks like this code was supposed to work, reading through the code upstream…

Apparently the upstream did not have in mind that Darwin supported `ppc64`. Otherwise they would have had a case for `ppc` there too, which they do not. IMO it is quite obviously intended for ELF, and that’s it fails on Leopard for `ppc64`.

net/nss/Portfile
barracuda156: @kencu I think it is appropriate to restrict it to Darwin, though removing 9 won’t hurt. On Linux and BSD `ppc64` will presumably work as-is, though I cannot personally confirm that. Could you check it on Debian?

kencu: you already restricted it with the Apple test in the patch, right? 

So, that covers it.

net/nss/files/patch-ppc64.diff
kencu: I don’t see anything ELF in there… or relating to any ABI at all, at least in this one I found browsing:

https://chromium.googlesource.com/chromium/deps/nss/+/ccb083050bac653bd6b98c38237fdf93c5d64a7a/nss/lib/freebl/mpi/mpcpucache.c#695

and Apple is at least mentioned at one point…

kencu: so it looks to me like it was supposed to work. Exactly what happens when you try to compile this? It was not clear from the ticket what went wrong.

net/nss/Portfile
barracuda156: > you already restricted it with the Apple test in the patch, right?
> 
> So, that covers it.

Fair enough.

net/nss/files/patch-ppc64.diff
barracuda156: > so it looks to me like it was supposed to work. Exactly what happens when you try to compile this? It was not clear from the ticket what went wrong.

This:
```
:info:build mpi/mpcpucache.c: In function ‘dcbzl’:
:info:build mpi/mpcpucache.c:730: error: nested functions are disabled, use -fnested-functions to re-enable
:info:build mpi/mpcpucache.c:730: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘asm’
:info:build mpi/mpcpucache.c:730: warning: implicit declaration of function ‘asm’
:info:build mpi/mpcpucache.c:730: error: lvalue required as left operand of assignment
:info:build mpi/mpcpucache.c:732: error: ‘a’ undeclared (first use in this function)
:info:build mpi/mpcpucache.c:732: error: (Each undeclared identifier is reported only once
:info:build mpi/mpcpucache.c:732: error: for each function it appears in.)
:info:build mpi/mpcpucache.c:731: error: invalid lvalue in asm output 0
:info:build make[4]: *** [Output.OBJD/Darwin_SINGLE_SHLIB/mpcpucache.o] Error 1
:info:build make[4]: Leaving directory `/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_net_nss/nss/work/nss-3.78-ppc64/nss/lib/freebl'
:info:build make[3]: *** [freebl_FREEBL_BUILD_SINGLE_SHLIB] Error 2
```

kencu: anyway, if no processor-specific method works the code falls back to a generic default value, so no big deal if Apple ppc64 doesn’t use an optimal method I suppose, even if it did compile and work.

barracuda156: > I don’t see anything ELF in there… or relating to any ABI at all, at least in this one I found browsing:
> 
> https://chromium.googlesource.com/chromium/deps/nss/+/ccb083050bac653bd6b98c38237fdf93c5d64a7a/nss/lib/freebl/mpi/mpcpucache.c#695
> 
> and mac is at least mentioned at one point…

Yes, I have read the relevant source code, of course. In fact, there is an instruction in Darwin ABI `dcbzl128` which may be what they would want, but their code fails, as a matter of fact, on Leopard. My assumption re ELF is based on this: I mean, Leopard did not change, so if the code was working back then, it will work now too. But it does not.

barracuda156: > anyway, if no processor-specific method works the code falls back to a generic default value, so no big deal if Apple ppc64 doesn’t use an optimal method I suppose, even if it did compile and work.

Indeed. So then I just remove a platform restriction for the patch now, and we are good to go with this?

kencu: so this little bit is what fails to compile? I should compile on Darwin:



/* clear the cache line that contains 'array' */
--
  | static inline void dcbzl(char *array)
  | {
  | register char *a asm("r2") = array;
  | __asm__ __volatile__( "dcbzl %0,r0" : "=r" (a): "0"(a) );
  | }

barracuda156: > so this little bit is what fails to compile? It should compile on Darwin:
> 
> /* clear the cache line that contains 'array' */   static inline void dcbzl(char *array)   {   register char *a asm("r2") = array;   **asm** **volatile**( "dcbzl %0,r0" : "=r" (a): "0"(a) );   }

Thank you! I can check this once I am back to the Quad. No G5 machine at home.

barracuda156: @kencu By the way, maybe you could suggest how to change this line in `libuv` code for Darwin `ppc64`? https://github.com/libuv/libuv/issues/3623#issuecomment-1124859031

kencu: upstream has rewritten this function like this now:

https://github.com/nss-dev/nss/blob/671d89b6c4a6f41707bb044534751098e2e3f211/lib/freebl/mpi/mpcpucache.c#L725


```
/* clear the cache line that contains 'array' */
static inline void
dcbzl(char *array)
{
    register char *a asm("r2") = array;
    __asm__ __volatile__("dcbzl %0,0"
                         : "=r"(a)
                         : "0"(a));
}
```

which is pretty similar.

anyway if it won't compile, it won't I guess. No big loss to fall back to the (slower) default fallback.

kencu: I had trouble getting it to compile on godbolt until I removed the "static"

https://godbolt.org/z/esG4Gj7TP

barracuda156:
@kencu Sorry for a delay, I tried using suggested code, but it still fails:
```
mpi/mpcpucache.c: In function 'dcbzl':
mpi/mpcpucache.c:729:20: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'asm'
  729 |   register char *a asm("r2") = array;
      |                    ^~~
mpi/mpcpucache.c:729:20: warning: implicit declaration of function 'asm' [-Wimplicit-function-declaration]
mpi/mpcpucache.c:729:30: error: lvalue required as left operand of assignment
  729 |   register char *a asm("r2") = array;
      |                              ^
mpi/mpcpucache.c:730:47: error: 'a' undeclared (first use in this function)
  730 |   __asm__ __volatile__( "dcbzl %0,r0" : "=r" (a): "0"(a) );
      |                                               ^
mpi/mpcpucache.c:730:47: note: each undeclared identifier is reported only once for each function it appears in
mpi/mpcpucache.c:730:3: error: invalid lvalue in 'asm' output 0
  730 |   __asm__ __volatile__( "dcbzl %0,r0" : "=r" (a): "0"(a) );
      |   ^~~~~~~
make[4]: *** [Output.OBJD/Darwin_SINGLE_SHLIB/mpcpucache.o] Error 1
make[4]: Leaving directory `/opt/local/var/macports/build/_opt_PPCLeopardPorts_net_nss/nss/work/nss-3.78-ppc64/nss/lib/freebl'
make[3]: *** [freebl_FREEBL_BUILD_SINGLE_SHLIB] Error 2
make[3]: Leaving directory `/opt/local/var/macports/build/_opt_PPCLeopardPorts_net_nss/nss/work/nss-3.78-ppc64/nss/lib/freebl'
make[2]: *** [freebl] Error 2
make[2]: Leaving directory `/opt/local/var/macports/build/_opt_PPCLeopardPorts_net_nss/nss/work/nss-3.78-ppc64/nss/lib'
make[1]: *** [lib] Error 2
make[1]: Leaving directory `/opt/local/var/macports/build/_opt_PPCLeopardPorts_net_nss/nss/work/nss-3.78-ppc64/nss'
make: *** [all] Error 2
```
Here is the patch I applied instead of mine:
```
--- nss/lib/freebl/mpi/mpcpucache.c.orig	1970-01-01 08:00:00.000000000 +0800
+++ nss/lib/freebl/mpi/mpcpucache.c	2022-05-18 18:14:28.000000000 +0800
@@ -724,14 +724,11 @@
 #include "memory.h"
 
 /* clear the cache line that contains 'array' */
-static inline void
-dcbzl(char *array)
-{
-    register char *a asm("r2") = array;
-    __asm__ __volatile__("dcbzl %0,0"
-                         : "=r"(a)
-                         : "0"(a));
-}
+  static inline void dcbzl(char *array)
+  {
+  register char *a asm("r2") = array;
+  __asm__ __volatile__( "dcbzl %0,r0" : "=r" (a): "0"(a) );
+  }
 
 #define PPC_DO_ALIGN(x, y) ((char *)((((long long)(x)) + ((y)-1)) & ~((y)-1)))
 
```

barracuda156:
@kencu Actually wait a bit with this PR, or if you got a minute, check this: https://bugzilla.mozilla.org/show_bug.cgi?id=1769063

There is a version of `__dcbzl` from Apple (and even its 32-bit version), however it needs 2 arguments: https://opensource.apple.com/source/llvmgcc42/llvmgcc42-2118/more-hdrs/ppc_intrinsics.h.auto.html

barracuda156:
Closing, replaced with: https://github.com/macports/macports-ports/pull/15054/

PR #14845, opened 2022-05-14
ocaml: fix build for PPC, bytecode compiler

barracuda156:
#### Description

This fixed build of `ocaml` @4.1.2 for PPC (bytecode compiler, world target).
Ref from: https://github.com/macports/macports-ports/pull/14691

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/lang/ocaml/Portfile
+++ b/lang/ocaml/Portfile
@@ -14,12 +14,12 @@ PortGroup           compiler_blacklist_versions 1.0
 name                ocaml
 epoch               1
 version             4.12.1
-revision            0
+revision            1
 set major_vers      [join [lrange [split ${version} .] 0 1] .]
 platforms           darwin
-# Note that i386 on Mac is truly no longer supported by upstream
+# Note that ppc, ppc64 and i386 on Mac are no longer supported by upstream
 # We may have to remove support at some point.
-supported_archs     i386 x86_64 arm64
+supported_archs     i386 x86_64 arm64 ppc ppc64
 maintainers         {pmetzger @pmetzger} openmaintainer
 categories          lang ocaml
 license             LGPL
@@ -47,18 +47,21 @@ compiler.blacklist  gcc-4.0 *gcc-4.2 {clang < 400}
 # debug problems with these patches.
 if {${os.major} < 11} {
     # need to add strnlen patch as not in library
-    patchfiles-append  patch-strnlen-socketaddr.diff
+    patchfiles-append  patch-strnlen-socketaddr.diff \
+                       patch-configure-darwin-ppc.diff
 
+    if {${build_arch} eq "i386" || ${build_arch} eq "x86_64"} {
     # compiler selection pulls in clang-3.4 on 10.6 and less
     # and this compiler is needed for this port to run once installed
     # so force this compiler and require it at runtime
     compiler.whitelist     macports-clang-3.4
     depends_run-append     port:clang-3.4
+    }
 }
 
 # Note: the port maintainer has no access to an i386 host, and cannot
 # debug problems on one.
-if {${build_arch} eq "i386"} {
+if {${build_arch} eq "i386" || ${build_arch} eq "ppc"} {
     patchfiles-append  patch-configure-darwin32.diff
 }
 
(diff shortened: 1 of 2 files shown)


pmetzger:
This seems really complicated just to default to using the bytecode compiler on PPC. Some of lines in the patch do things like reindenting existing lines, and a lot of what is here doesn't seem to make much sense to me.

Could we get a truly minimal version of this? At the very least it would make it easier to review.

barracuda156:
> This seems really complicated just to default to using the bytecode compiler on PPC. Some of lines in the patch do things like reindenting existing lines, and a lot of what is here doesn't seem to make much sense to me.
> 
> Could we get a truly minimal version of this? At the very least it would make it easier to review.

@pmetzger Sure, will be done.

lang/ocaml/Portfile
```diff
@@ -49,35 +49,49 @@ if {${os.major} < 11} {
...
-configure.env-append \
-    "AS=${configure.cc} -Wno-trigraphs -c" \
-    "ASPP=${configure.cc} -Wno-trigraphs -c" \
-    "PARTIALLD=${configure.cc} -Wl,-r"
+if {${build_arch} ne "ppc" && ${build_arch} ne "ppc64"} {
+    configure.env-append \
+       "AS=${configure.cc} -Wno-trigraphs -c" \
+       "ASPP=${configure.cc} -Wno-trigraphs -c" \
+       "PARTIALLD=${configure.cc} -Wl,-r"
+}
```
pmetzger, line 84: Why are you doing this only on ppc?

lang/ocaml/Portfile
```diff
@@ -49,35 +49,49 @@ if {${os.major} < 11} {
...
+if {${build_arch} eq "i386" || ${build_arch} eq "ppc"} {
     patchfiles-append  patch-configure-darwin32.diff
 }
 
+if {${build_arch} eq "ppc" || ${build_arch} eq "ppc64"} {
+    patchfiles-append  patch-configure-darwin-ppc.diff
+    compiler.blacklist-append *clang*
+    configure.args-append --enable-imprecise-c99-float-ops
+}
+
```
pmetzger, line 72: Is the imprecise-c99-float-ops actually necessary?

barracuda156: For Rosetta, necessary, otherwise the build fails. For other configurations doesn’t hurt, but not required.

barracuda156:
@pmetzger `Wno-trigraphs` is a flag used for `clang` and `x86_64`, see `configure`, lines 14238–14239, 14290–14291. It is also applied there in a case of FreeBSD (_which does not include `ppc` or `ppc64` there_).

lang/ocaml/Portfile
pmetzger: Why are we building under Rosetta? I'm also not convinced that it can't hurt if people have concerns about their numerics. The OCaml documentation specifically warns against this flag. It should only be turned on in rare circumstances if absolutely needed because of flaws in c99 support.

pmetzger:
I believe I had to add the no-trigraphs flag because otherwise sometimes the x86 support breaks.

You are not making the least intrusive possible set of changes here to allow ppc to build with the bytecode runtime support. I don't feel comfortable altering several of the things that have been changed here. I suspect they'll cause breakage in conditions you may not be familiar with. I would recommend that we make the *minimal* set of changes needed to make PPC build with the bytecode compiler and leave it there.

barracuda156:
> I believe I had to add the no-trigraphs flag because otherwise sometimes the x86 support breaks.
> 
> You are not making the least intrusive possible set of changes here to allow ppc to build with the bytecode runtime support. I don't feel comfortable altering several of the things that have been changed here. I suspect they'll cause breakage in conditions you may not be familiar with. I would recommend that we make the _minimal_ set of changes needed to make PPC build with the bytecode compiler and leave it there.

@pmetzger
1. I will remove a flag that fixes Rosetta build. Not essential.
2. `Wno-trigraphs` is still being applied for Intel, as before. No changes to i386, x86_64 and ARM has been made.

barracuda156:
UPD. Give me a few minutes, I will fix everything. Git pushed wrong commit.

barracuda156:
@pmetzger All changes done, as you have told me. Hopefully now it is okay.

Thank you, and sorry that it took a long time to sort out.

pmetzger:
Merged. Thanks for your work, @barracuda156 !

barracuda156:
@pmetzger Thank you very much! Great!

P. S. I will try slowly to fix assembler part, and if that succeeds (and passes tests), will update you.

pmetzger:
Let's do the update to 4.14.0 instead first.

barracuda156:
@pmetzger Yes, fair enough.

PR #14854, opened 2022-05-15
ld64-274: allow build on Mac OS X Leopard

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L31a i386
Xcode 3.1.4 9M2809

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/rust-1.0.tcl
+++ b/_resources/port1.0/group/rust-1.0.tcl
@@ -729,10 +729,9 @@ proc rust::rust_pg_callback {} {
         depends_lib-append              port:openssl${openssl_ver}
     }
 
-    if { [string match "macports-clang*" [option configure.compiler]] && [option os.major] == 10 } {
+    if { [string match "macports-clang*" [option configure.compiler]] && [option os.major] < 11 } {
         # by default, ld64 uses ld64-127 when 9 <= ${os.major} < 11
-        # Rust fails to build when ${os.major} >= 10 and ld64 uses ld64-127
-        # ld64-274 does not seem to build when ${os.major} < 10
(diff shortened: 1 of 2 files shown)


kencu:
you didn't like this one?

https://ports.macports.org/port/libblocksruntime/

I actually have this ld64 running on my Intel Leopard system too, using the above. Just never pushed it through to MacPorts for the usual reasons. I used the same mods as you.

kencu:
forcing SnowLeopard to use ld64-274 would be a step back for most of us.... I use ld64-latest.

There is also this bootstrapping issue where you can't build ld64-274 unless you have already installed ld64-127, and once you have installed ld64-127 we never sorted out a method whereby macports can somehow automatically upgrade it to ld64-274 for you.

neither of those are deal-breakers. Users can do stuff manually once they get into this rarefied atmosphere.

The real problem is trying to make a buildbot happy, as it always has to build from bare system roots and there is just too much code to write to make it automatic.  We may be reaching the point where that expectation is no longer realistic, although our newly revitalized efforts to simplify bootstrapping by getting rid of the clang-3.4 pathway and with this path instead:

```
gcc10bootstrap -> clang11bootstrap -> build a current toolchain
```

might make it possible.

kencu (approved):

kencu:
LGTM!

Thanks as always, Marcus!

MarcusCalhoun-Lopez:
> you didn't like this one?
> 
> https://ports.macports.org/port/libblocksruntime/

Thanks. I should have looked more carefully for existing ports.

kencu:
who would have thought we would have so quietly snuck that in on you, though? :>

PR #14858, opened 2022-05-16
MoltenVK: Update to 1.1.10 & add workaround for 10.11

Gcenx:
#### Description

- Update to v1.1.10

- Repackage an already built archive this avoids needing to worry about mounting the DMG (v1.1.9 bump after v1.1.10 packages are done)

###### Type(s)

- [x] enhancement

###### Tested on

macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505

(workaround set to darwin 18 while testing)

###### Verification 

### DIFF
--- a/graphics/MoltenVK/Portfile
+++ b/graphics/MoltenVK/Portfile
@@ -3,10 +3,10 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        KhronosGroup MoltenVK 1.1.9 v
+github.setup        KhronosGroup MoltenVK 1.1.10 v
 epoch               1
 
-set sdkversion      1.3.211.0
+set sdkversion      1.3.216.0
 categories          graphics
 maintainers         {ryandesign @ryandesign} openmaintainer
 platforms           macosx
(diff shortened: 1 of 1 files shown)


chrstphrchvz:
I was not personally interested in 10.11. For some unknown reason I had encountered an hdiutil error on 10.15 resembling the one seen on 10.11.

Gcenx:
> I was not personally interested in 10.11. For some unknown reason I had encountered an hdiutil error on 10.15 resembling the one seen on 10.11.

Ah yes you had that issue, I should go check that closed PR as I remember someone wanted MoltenVK on 10.11

Gcenx:
@pmetzger a solution to support macOS 10.11 again, not exactly ideal but this being the only solution I can come up with without LunarG changing there packing or MoltenVK directly providing artifacts.

ryandesign (commented):
Remind me again what is the reason why the portfile was changed to go to such efforts to use a binary rather than build from source like it used to?

graphics/MoltenVK/Portfile
```diff
@@ -21,53 +21,73 @@ long_description    ${name} is an implementation of the high-performance, \
...
+    if {${os.major} < 17} {
         archive_sites
         distfiles
         depends_build
         known_fail  yes
         pre-fetch {
-            ui_error "${subport} @${version} requires macOS Sierra or later"
-            return -code error "incompatible OS X version"
+            ui_error "${subport} @${version} requires macOS El Capitan or later"
+            return -code error "incompatible macOS version"
```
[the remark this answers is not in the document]
Gcenx: It's common to revert to any version as macOS now but I'll revert to the old naming.

graphics/MoltenVK/Portfile
```diff
@@ -21,53 +21,73 @@ long_description    ${name} is an implementation of the high-performance, \
...
-    # vulkan and vk_video are provided via vulkan-headers
-    file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
+        # vulkan and vk_video are provided via vulkan-headers
+        file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
+    }
 }
 
 platform darwin {
-    if {${os.major} < 16} {
+    if {${os.major} < 17} {
```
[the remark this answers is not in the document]
Gcenx: Yes that's a mistake it indeed should have been lowered to 15

Gcenx:
> Remind me again what is the reason why the portfile was changed to go to such efforts to use a binary rather than build from source like it used to?

Only the very latest versions of macOS/Xcode can built MoltenVK from source, the options were repackage the VulkanSDK or build using a buildbot and repackage on older versions of macOS. I do this from my MoltenVK mirror but I doubt macports would want to use that over say LunarG provided archive.

Repacking the archive has always been a viable option in the case of MoltenVK as the projects min target is 10.11, however you've never accepted a bump to MoltenVK due to `wine`

mascguy:
Ryan, are you fine with merging this?

Gcenx:
> Ryan, are you fine with merging this?

_If_ this gets merged after packages are built it then needs a bump to the workaround so it install 1.10.0.

I’ll just hope that LunarG hasn’t bumped the SDK since this was last updated as they love you so this and remove the older version from download. This makes these slow updates even worse.

Gcenx:
Anything?

mascguy:
> _If_ this gets merged after packages are built it then needs a bump to the workaround so it install 1.10.0.

Dean, can you clarify what needs a bump?

graphics/MoltenVK/Portfile
```diff
@@ -21,52 +21,72 @@ long_description    ${name} is an implementation of the high-performance, \
                     industry-standard Vulkan graphics and compute API, that \
                     runs on Apple's Metal graphics framework.
 
-distname            vulkansdk-macos-${sdkversion}
-use_dmg             yes
+if {${os.major} == 16} {
```
Gcenx, line 24: > Dean, can you clarify what needs a bump?

This section here would need to be bumped _after_ `MoltenVK-1.10.1_0.darwin_16.x86_64` package is avaliable from packages.macports

mascguy: > This section here would need to be bumped _after_ `MoltenVK-1.10.1_0.darwin_16.x86_64` package is avaliable from packages.macports

What would it need to be changed to? (Sorry for being dense, but I'm not in-tune with all of the details...)

Gcenx: > What would it need to be changed to? (Sorry for being dense, but I'm not in-tune with all of the details...)

After the packages are build this section would need to be bumped to use the 1.10.0 so update `version` here and the checksums to match the prebuild 1.10.0 darwin16 package once it's available for download from packages.macports.

```
if {${os.major} == 16} {
    version             1.1.9
    master_sites        https://packages.macports.org/MoltenVK/
    distname            MoltenVK-${version}_0.darwin_16.x86_64
    checksums           sha256  84c49e85712d2bb19f5d00bd91df4dae9518bc31ca3265c56c1fdabae25fcf08 \
                        rmd160  bf7e89fc3f67c0aeb392149eae4ce51ee06d4f06 \
                        size    7026887
    use_bzip2           yes
    extract.suffix      .tbz2
```

mascguy: > After the packages are build this section would need to be bumped to use the 1.10.0 so update `version` here and the checksums to match the prebuild 1.10.0 darwin16 package once it's available for download from packages.macports.

Could this be accomplished via a bootstrap-ish subport, which the parent depends on?

Gcenx: > Could this be accomplished via a bootstrap-ish subport, which the parent depends on?

All this port does is unpack the LunarG VulkanSDK and grab the files, the _workaround_ does exactly what a bootstrap port would accomplish.

chrstphrchvz: Because the workaround is for 10.11 it should instead say:
```suggestion
if {${os.major} == 15} {
```

A bump (of what exactly—the revision?) doesn’t seem necessary, the 10.11 builds will just fail until the 10.12 build has been published. I would think it would be preferable to instead hold `version` back on 10.11 until it can be updated. But whatever the approach entails I think others might still find it confusing.

Gcenx: > Because the workaround is for 10.11 it should instead say

Done

> A bump (of what exactly—the revision?) doesn’t seem necessary, the 10.11 builds will just fail until the 10.12 build has been published. I would think it would be preferable to instead hold version back on 10.11 until it can be updated withot anything failing. But whatever the approach entails I think others might still find it confusing.

After the MoltenVK-1.10.0 packages are built the workaround section could then be updated to 1.10.0 to match the rest.

Gcenx:
MoltenVK-v1.1.11 has been marked not checked if for an updated VulkanSDK url but it seems this isn’t going anywhere so I’m not rebasing this _again_

PR #14859, opened 2022-05-16
yabai: update to version 4.0.0

Ci7rix:
* update to version 4.0.0

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.3.1 21E258 x86_64
Xcode 13.3.1 13E500a

###### Verification 
Have you


TheKevJames:
@Ci7rix looks like some checksum errors here:
```
  Error: Checksum (size) mismatch for yabai-4.0.0.tar.gz
  checksums           rmd160  951e2b20428fae7ada2ecbe2b75476e5b8405100 \
                      sha256  efadb820b4485208b9267739e4205e2273f1f4a30f270ff411c7c00934fb96ff \
                      size    1525695
  Error: Failed to checksum yabai: Unable to verify file checksums
```

Ci7rix:
@TheKevJames oops, sorry for that.
I just fixed it. I'm a noob with Git, is there something else I should do ?

Ci7rix:
Hi @TheKevJames,

Checks have failed. On my side the build is working fine.

Could you maybe help me with that, if time permits ?
Is there a way I could reproduce the build issue on my computer ? Like installing the same build tools ?  

Thank you.

Ci7rix:
Hi

According to what I could find, the build on macOS 10.15 version does not work because it requires a version identical or higher than Xcode 12.2 in order to build a universal binary (see here https://github.com/koekeishiya/yabai/issues/1191#issuecomment-1069711506).

For macOS 11, if I understand correctly the 'safeAreaInsets' property does not exist yet in this version. I found a similar problem here: https://trac.macports.org/ticket/63824#comment:1

If I understand correctly, it is necessary to set the minimal version of Xcode to 12.2 and the minimal version of macOS to 12.0.

If so, how do I do it?

Thanks

TheKevJames:
Hi @Ci7rix ! My understanding is that you've done the right thing in specifying the minimum xcode version. I think the OS version follows from the xcode version, though: so, for example, since you specified 12.2 as the minimum xcode version, that means the port is meant to build on the earliest version of OSX that supports Xcode v12.2 (OSX 10.15.4, according to [this doc](https://developer.apple.com/support/xcode/)).

I would suspect the right answer here is to mark the minimum Xcode version as 13.3, since that's the oldest Xcode that supports only OSX v12 or newer.

That said, I'm pretty new to working with portfiles, so the above is mostly guesswork!

TheKevJames (approved):
LGTM assuming this got the build to pass.

I unfortunately don't have permissions to approve test workflows or to merge PRs -- @ryandesign and @reneeotten helped me when I first added this port, maybe one of you would be able to help out again here?

Ci7rix:
Hi @TheKevJames,

Thank you for your help !

l2dy:
Approved by maintainer, merging now. Thanks for your contribution.

PR #14899, opened 2022-05-21
qmake5 PortGroup: reduce spurious noisy error

kencu:
this xcrun command is supposed to fail if the specific
sdk macosx${configure.sdk_version} is not found. However,
this error shows up on user's screens causing considerable
confusion about what the true build error is.

### DIFF
--- a/_resources/port1.0/group/qmake5-1.0.tcl
+++ b/_resources/port1.0/group/qmake5-1.0.tcl
@@ -57,8 +57,16 @@ pre-configure {
     #
     # avoid --show-sdk-path since it is not available on all platforms
     # see https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261
-    if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld} result]} {
-        configure.sdk_version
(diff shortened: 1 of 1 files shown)


kencu:
also, I think the reason that
```
use_xcode yes
```
is often needed is that this command calls through to xcodebuild to do it's work, and I believe xcodebuild only exists if a full xcode installation is present and is being used.

I think.

kencu:
I suppose you could also say that the presence of the ```-sdk macosx``` option should be actually verified as well, rather than just assumed to work.

_resources/port1.0/group/qmake5-1.0.tcl
```diff
@@ -57,7 +57,8 @@ pre-configure {
     #
     # avoid --show-sdk-path since it is not available on all platforms
     # see https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261
-    if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld} result]} {
+    if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld 2>@1 > /dev/null} result]} {
```
jmroot, line 60: The `result` could also be removed since it is unused.

kencu: done, thx

kencu:
hmm.. this doesn't seem to be working right now. I'll have to take a closer look at it later.

PR #14903, opened 2022-05-21
algol68g: update to 3.0.6

barracuda156:
#### Description

Update to 3.0.6. Add patches for macOS.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6 PPC
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/lang/algol68g/Portfile
+++ b/lang/algol68g/Portfile
@@ -21,19 +21,16 @@ checksums           rmd160  7849acdadd91ca6a9cac1537bfaac102e1969de7 \
 
 compiler.c_standard 1999
 
-depends_build       port:autoconf \
-                    port:automake \
-                    port:libtool
-
 depends_lib         port:gmp \
                     port:gsl \
                     port:mpfr \
                     port:ncurses \
-                    port:r \
+                    port:R \
                     port:readline
 
 patchfiles          patch-a68glib.diff \
                     patch-alloc.diff \
+                    patch-darwin-ppc.diff \
                     patch-includedir.diff
 
 post-patch {
@@ -45,11 +42,12 @@ use_autoreconf      yes
 configure.args      --disable-postgresql \
                     --disable-plotutils
 
-platform darwin powerpc {
-    patchfiles-append \
-                    patch-darwin-ppc.diff
-    configure.args-append \
-                    --enable-quadmath=no
+platform darwin 10 {
+    # Rosetta needs this, otherwise it uses a wrong arch and clang
+    if {${build_arch} eq "ppc"} {
+        configure.args-append --build=powerpc-apple-darwin10
+        compiler.blacklist *clang*
+    }
 }
 
 variant compiler description {Enable compiler, experimental} {
--- a/lang/algol68g/Portfile
+++ b/lang/algol68g/Portfile
@@ -1,3 +1,5 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
 PortSystem          1.0
 
 name                algol68g
@@ -6,10 +8,9 @@ revision            0
 categories          lang devel
 license             GPL-3
 maintainers         nomaintainer
-description         Algol68 implementation as defined by the Revised Report
-long_description \
-    Algol68G is an implementation of Algol 68 as defined by the Revised \
-    Report. It ranks among the most complete implementations of the language.
+description         Algol 68 implementation as defined by the Revised Report
+long_description    Algol68G is an implementation of Algol 68 as defined by the Revised Report. \
+                    It ranks among the most complete implementations of the language.
 
 homepage            http://jmvdveer.home.xs4all.nl/algol.html
 master_sites        http://jmvdveer.home.xs4all.nl/
--- a/lang/algol68g/Portfile
+++ b/lang/algol68g/Portfile
@@ -1,10 +1,9 @@
 PortSystem          1.0
 
 name                algol68g
-version             2.8.1
-revision            6
+version             3.0.6
+revision            0
 categories          lang devel
-platforms           darwin
 license             GPL-3
 maintainers         nomaintainer
 description         Algol68 implementation as defined by the Revised Report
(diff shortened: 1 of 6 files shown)


lang/algol68g/Portfile
```diff
@@ -15,14 +14,55 @@ long_description \
...
+                    size    621615
 
-depends_lib         port:gsl port:readline port:ncurses
+compiler.c_standard 1999
 
-configure.args      --disable-postgresql
+depends_build       port:autoconf \
+                    port:automake \
+                    port:libtool
+
```
ryandesign, line 27: `use_autoreconf yes` adds these dependencies for you.
```suggestion
```

lang/algol68g/files/patch-darwin-ppc.diff
```diff
@@ -0,0 +1,24 @@
...
+ #
+-*86-*-*darwin* | *86_64-*-*darwin*) 
++*86-*-*darwin* | *86_64-*-*darwin* | powerpc*-*-*darwin*) 
+   AC_DEFINE(BUILD_BSD, 1, [Define this if DARWIN was detected]) 
+   AC_DEFINE(HAVE_IEEE_754, 1, [Define this if IEEE_754 compliant]) 
+   AC_MSG_WARN([configuring interpreter-only])
+
+
+--- configure.orig	2022-03-13 04:55:26.000000000 +0800
++++ configure	2022-05-21 18:53:02.000000000 +0800
```
ryandesign, line 15: Patch either configure.ac if you are regenerating the configure script with `use_autoreconf yes` (which you are) or patch configure if you're not, not both.

lang/algol68g/files/patch-includedir.diff
```diff
@@ -0,0 +1,24 @@
+--- configure.orig	2022-03-13 04:55:26.000000000 +0800
++++ configure	2022-05-21 21:25:34.000000000 +0800
```
ryandesign, line 2: Since you're using `use_autoreconf yes`, patch configure.ac, not configure, for this change to have any effect. Or possibly the newly generated configure script already contains this change.

lang/algol68g/files/patch-includedir.diff
```diff
@@ -0,0 +1,24 @@
...
+
+
+--- Makefile.am.orig	2022-03-13 04:55:25.000000000 +0800
++++ Makefile.am	2022-05-21 21:55:19.000000000 +0800
+@@ -105,7 +105,7 @@
+ ./src/include/a68g-stddef.h \
+ ./src/include/a68g-transput.h \
+ ./src/include/a68g-types.h
+-includedir = /usr/local/include/algol68g
++includedir = /@@PREFIX@@/include/algol68g
```
ryandesign, line 21: `@@PREFIX@@` already begins with a slash.
```suggestion
+includedir = @@PREFIX@@/include/algol68g
```

lang/algol68g/Portfile
```diff
@@ -15,14 +14,55 @@ long_description \
...
+depends_lib         port:gmp \
+                    port:gsl \
+                    port:mpfr \
+                    port:ncurses \
+                    port:r \
```
ryandesign, line 32: This port name is capitalized.
```suggestion
                    port:R \
```

lang/algol68g/Portfile
```diff
@@ -15,14 +14,55 @@ long_description \
...
+}
+
+use_autoreconf      yes
+
+configure.args      --disable-postgresql \
+                    --disable-plotutils
+
+platform darwin powerpc {
+    patchfiles-append \
+                    patch-darwin-ppc.diff
```
ryandesign, line 50: Unless this patchfile will cause problems for non-PowerPC systems, please include it unconditionally. It is likely that the next person to update this port will not be using a PowerPC computer, and by including the patch unconditionally, it ensures that if the patch needs to be updated, that person will not inadvertently forget to do so.

ryandesign (commented):
We don't want to maintain these patches forever. Are they from upstream? (Provide URL.) Or have you reported them to upstream? (Provide URL or state how they were sent to upstream—e.g. by email, bug report form, etc.)

lang/algol68g/Portfile
barracuda156: @ryandesign Thank you. This makes perfect sense.

lang/algol68g/files/patch-includedir.diff
barracuda156: I will fix these soon. Thank you for checking.

barracuda156:
@ryandesign Not yet sent to upstream, but I will do it today.

barracuda156:
@ryandesign Apparently `clang` on 10.6.8 with Xcode 3.2.6 does not support `__attribute__((aligned()))`, and `configure` fails on that. (In fact, it tries to pull broken `clang` even when building for `ppc` in Rosetta.)

Should I blacklist `clang` for OS <11 (this is not arch-specific, it seems) or just leave this? 

P. S. All your suggestions worked, I have finally built `algol68g` on 10.6.8, took awhile for dependencies.

kencu:
> @ryandesign Apparently `clang` on 10.6.8 with Xcode 3.2.6 does not support `__attribute__((aligned()))`, and `configure` fails on that. (In fact, it tries to pull broken `clang` even when building for `ppc` in Rosetta.)
> 
> Should I blacklist `clang` for OS <11 (this is not arch-specific, it seems) or just leave this?
> 
> P. S. All your suggestions worked, I have finally built `algol68g` on 10.6.8, took awhile for dependencies.
> 
> UPD. Never mind, perhaps later `clang` gonna work fine on 10.6, only native Xcode one fails. No need to blacklist, I guess users will figure out that they need to choose another compiler.

macports will never choose the xcode clang as a compiler on 10.6.8, so this issue will never arise.

barracuda156:
> macports will never choose the xcode clang as a compiler on 10.6.8, so this issue will never arise.

Ok, great.

barracuda156:
> We don't want to maintain these patches forever. Are they from upstream? (Provide URL.) Or have you reported them to upstream? (Provide URL or state how they were sent to upstream—e.g. by email, bug report form, etc.)

@ryandesign So I could not locate any upstream repo (Sourceforge one does not appear to be what is needed), but I sent an e-mail to the address from algol68g page with a brief description and the link to this PR.

barracuda156:
@ryandesign I got a reply from the developer that patches will be included in the next release. So it is a temporary measure now.

>  I appreciate your feedback and patches, as I have no Apple hardware at my disposition.
> Your patches will be incorporated in the next release.

PR #14908, opened 2022-05-21
libiconv: minor tweaks

MarcusCalhoun-Lopez:
###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L31a i386
Xcode 3.1.4 9M2809

###### Verification 
Have you

### DIFF
--- a/textproc/libiconv/Portfile
+++ b/textproc/libiconv/Portfile
@@ -29,6 +29,7 @@ long_description \
     conversion.
 
 depends_build           port:gperf
+depends_skip_archcheck  perf
 
 patchfiles              patch-src-Makefile.in-darwin.diff
 
--- a/textproc/libiconv/Portfile
+++ b/textproc/libiconv/Portfile
@@ -76,9 +76,6 @@ post-destroot {
     }
 }
 
-if {![info exists universal_possible]} {
-    set universal_possible [expr {${os.universal_supported} && [llength ${configure.universal_archs}] >= 2}]
-}
 if {${universal_possible} && [variant_isset universal]} {
     # When cross-compiling, can guess wrong endian value.
     lappend merger_configure_env(ppc)     cl_cv_sys_endian='big endian'
--- a/textproc/libiconv/Portfile
+++ b/textproc/libiconv/Portfile
@@ -1,7 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem              1.0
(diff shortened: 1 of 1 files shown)


textproc/libiconv/files/patch-c99.diff
```diff
@@ -6,7 +6,7 @@
  };
 -#ifdef __GNUC__
 +#if (defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || defined(__GNUC_STDC_INLINE__)
-+static inline
++inline
```
ryandesign, line 9: This patch isn't needed anymore at all, and I already had the deletion staged, so I pushed it, which will create a conflict in this PR. You should just be able to drop the commit that modified this patch and then force-push to this branch.

MarcusCalhoun-Lopez: As suggested, I have dropped the fix part of the pull request since it has already been fixed.
All that remains is the minor tweaks that should not change the installed files.

barracuda156:
@MarcusCalhoun-Lopez I suggest to remove restriction for Xcode <3.2.6 on 10.6.x. As a matter of fact, `libiconv` builds fine with Xcode 3.2 (so presumably gonna build with any Xcode 3.2.x).

ryandesign:
> restriction for Xcode <3.2.6 on 10.6.x.

`git blame` shows it was added to fix https://trac.macports.org/ticket/59150, which was a build failure of libiconv-bootstrap, which has since been deleted. If this fix isn't needed for the main libiconv port then it can of course be removed.

barracuda156:
@MarcusCalhoun-Lopez If there are no objections, please remove the ban on Xcodes <3.2.6 on 10.6.

barracuda156:
@ryandesign Thank you!

kencu:
specifically, libiconv (bootstrap or not) did not build on Intel systems with an old clang that was included with a less-than-current 3.2.6 version of xcode.

I’m quite certain it still doesn’t, but I’m not going to find that old clang to prove it.

barracuda builds with gcc on 10.6 ppc, so would never see the clang build error.

barracuda156:
@kencu It should not use any `clang`, apparently, at least by default:
```
$ port deps libiconv
Full Name: libiconv @1.17_0
Build Dependencies:   gperf
```
(Though I am not sure how Xcode decides whether to pick its gcc, llvm-gcc or clang.)

I cannot really test with every version of Xcode 3.2.x, but I would assume that if Xcode from 10a190 can build it, any released Xcode 3.2.x should either.
If it is of interest for someone, I can try on 10a190 `i386`, which has the same old Xcode. (I actually have a VM with 10.6.8, `libc++` and clangs, i.e. default Macports set-up for Intel, but with Xcode 4.2.)

PR #14909, opened 2022-05-21
gobject-introspection: use MacPorts arch settings

MarcusCalhoun-Lopez:
There are parts of the build system that do not respect the `-arch`
flag unless CFLAGS and LDFLAGS are set in the build phase.

Also ensure Python is run as the same architecture as `build_arch`.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L31a i386
Xcode 3.1.4 9M2809

###### Verification 
Have you

- [] checked your Portfile with `port lint --nitpick`?

### DIFF
--- a/gnome/gobject-introspection/Portfile
+++ b/gnome/gobject-introspection/Portfile
@@ -3,6 +3,7 @@
 PortSystem          1.0
 PortGroup           meson 1.0
 PortGroup           active_variants 1.1
+PortGroup           muniversal 1.1
 
 name                gobject-introspection
 conflicts           gobject-introspection-devel
@@ -53,8 +54,8 @@ patchfiles          patch-fix-rpath-gir-typelib.diff \
 post-patch {
     reinplace "s|libcairo-gobject.2.dylib|${prefix}/lib/libcairo-gobject.2.dylib|g" ${worksrcpath}/gir/meson.build
     reinplace "s|@MP_LIB@|${prefix}/lib/|g" ${worksrcpath}/giscanner/shlibs.py
-    reinplace "s|@MP_PYTHON_CMD@|${configure.python}|g" ${worksrcpath}/tools/meson.build
-    reinplace "s|#!/usr/bin/env python3|#!${configure.python}|g" \
+    reinplace "s|@MP_PYTHON_CMD@|/usr/bin/arch ${configure.python}|g" ${worksrcpath}/tools/meson.build
+    reinplace "s|#!/usr/bin/env python3|#!/usr/bin/arch ${configure.python}|g" \
         ${worksrcpath}/misc/update-glib-annotations.py \
         misc/update-gtkdoc-tests.py \
         misc/update-vulkan-gir.py \
@@ -64,7 +65,21 @@ post-patch {
 configure.python    ${prefix}/bin/python${py_ver}
 
 configure.args      -Ddoctool=enabled \
-                    -Dpython=${configure.python}
+                    -Dpython=${workpath}/python
+
+post-extract {
+    set python_file [open "${workpath}/python" w 0755]
+    puts ${python_file} "#!/bin/sh"
+    puts ${python_file} "exec /usr/bin/arch ${configure.python} \"$@\""
+    close ${python_file}
+}
+
+post-destroot {
+    reinplace "s|#!/usr/bin/arch ${configure.python}|#!${configure.python}|g" \
+        ${destroot}${prefix}/bin/g-ir-annotation-tool \
+        ${destroot}${prefix}/bin/g-ir-doc-tool \
+        ${destroot}${prefix}/bin/g-ir-scanner
+}
 
 # By default, gir will attempt to link programs using
 # the same compiler used to build Python itself. However,
@@ -78,6 +93,12 @@ configure.args      -Ddoctool=enabled \
 # in the gobject_introspection PortGroup, so dependent projects
 # should not require modification.
 build.env-append    CC=${configure.cc}
+
+foreach arch ${muniversal.architectures} {
+    # There seem to be parts of the build system that do not have the `-arch` flag.
+    build.env.${arch}-append    "CFLAGS=${configure.cflags}   [option configure.cc_archflags.${arch}]"
+    build.env.${arch}-append    "LDFLAGS=${configure.ldflags} [option configure.ld_archflags.${arch}]"
+}
 
 platform darwin 8 {
     # Tiger does not support RPATHs at this time


MarcusCalhoun-Lopez:
This pull request may be too inelegant to be of use, but it is the only way I could get gobject-introspection to build with the universal variant.

mascguy:
I'll defer to others who are building for PPC, and/or Universal.

@evanmiller , @barracuda156 , anyone...?

gnome/gobject-introspection/Portfile
```diff
@@ -79,6 +94,12 @@ configure.args      -Ddoctool=enabled \
 # should not require modification.
 build.env-append    CC=${configure.cc}
 
+foreach arch ${muniversal.architectures} {
+    # There seem to be parts of the build system that do not have the `-arch` flag.
+    build.env.${arch}-append    "CFLAGS=${configure.cflags}   [option configure.ld_archflags.${arch}]"
+    build.env.${arch}-append    "LDFLAGS=${configure.ldflags} [option configure.cc_archflags.${arch}]"
```
kencu, line 100: these look backwards? (LDFLAGS should have configure configure.ld_archflags)

kencu: not that it makes any difference, one hopes...

barracuda156:
> I'll defer to others who are building for PPC, and/or Universal.
> 
> 
> 
> @evanmiller , @barracuda156 , anyone...?

@mascguy Let me check.
By the way, `gobject-introspection` is broken for +universal on Leopard. I was only able to build pre-meson version.

gnome/gobject-introspection/Portfile
MarcusCalhoun-Lopez: Thank you for noticing.
The order should now be fixed.

barracuda156:
@mascguy @MarcusCalhoun-Lopez And on 10.6.8 Rosetta I had to do this in order to build 1.72.0: https://github.com/mesonbuild/meson/issues/10351#issuecomment-1131530681

barracuda156:
@MarcusCalhoun-Lopez This is awesome! Your PR fixes `+universal` on Leopard. Thank you very much. This was one issue which I could not solve myself.

P. S. @mascguy There are two problems however, unrelated to `gobject-introspection`, that have to be solved in Macports, in order to build this for `ppc+ppc64`.

1. `meson` failure:
```
gobject-introspection 1.72.0

  User defined options
    Cross files: ppc64-darwin
    prefix     : /opt/local
    doctool    : enabled
    python     : /opt/local/var/macports/build/_opt_PPCLeopardPorts_gnome_gobject-introspection/gobject-introspection/work/python

Found ninja-1.11.0 at /opt/local/bin/ninja

ERROR: An exe_wrapper is needed but was not found. Please define one in cross file and check the command and/or add it to PATH.
```
This is fixed by adding the following to cross-files:
```
[properties]
needs_exe_wrapper = false
```
I can make a PR for this, should be okay.

2. Well, as I have mentioned a number of times, Pythons are broken for ppc+ppc64, and here, I am afraid, @jmroot won't allow me to fix them.
I have them fixed locally, but it requires changes to how universal is built, not just ppc-specific fix.

barracuda156:
If anyone interested to test though, here is what allows to build `python39` and `python310` as `+lto +universal` on Leopard:

1. Add `PortGroup muniversal 1.0`
2. Replace the following chunk of code with the caption below:
```
platform darwin {
    set framewpath  ${frameworks_dir}/Python.framework
    set framewdir   ${framewpath}/Versions/${branch}
    set confdir     config-${branch}-darwin

    post-configure {
        # poll() misbehaves on 10.8 and older
        # See https://trac.macports.org/ticket/18376
        if {${os.major} <= 12} {
            set oldmtime [file mtime ${worksrcpath}-${build_arch}/pyconfig.h]
            system -W ${worksrcpath}-${build_arch} "ed - pyconfig.h < ${filespath}/pyconfig.ed"
            file mtime ${worksrcpath}-${build_arch}/pyconfig.h $oldmtime
        }
    }

    post-build {
        set buildlibdir [lindex [glob -directory ${worksrcpath}-${build_arch}/build lib.*-*-*-${branch}] 0]
        # preserve mtime of sysconfig data file to avoid rebuilding things after changing it
        set oldmtime [file mtime ${buildlibdir}/_sysconfigdata__darwin_darwin.py]

        # Without this, LINKFORSHARED is set to
        # ... $(PYTHONFRAMEWORKDIR)/Versions/$(VERSION)/$(PYTHONFRAMEWORK)
        # (this becomes Python.framework/Versions/3.10/Python) which doesn't
        # work for dependents that incorrectly use this variable to find out
        # how to link against python (see ticket #15099); instead we mirror
        # the behavior of `python-config --ldflags` here.
        set lfs_pattern {^([[:space:]]*'LINKFORSHARED':).*}
        set lfs_replacement "\\1 '-L${framewdir}/lib/python${branch}/${confdir} -lpython${branch} -ldl -framework CoreFoundation',"
        reinplace -E s|${lfs_pattern}|${lfs_replacement}| \
            ${buildlibdir}/_sysconfigdata__darwin_darwin.py

        # remove -arch flags from the config
        reinplace -E {s|-arch [a-z0-9_]+||g} \
            ${buildlibdir}/_sysconfigdata__darwin_darwin.py

        # also remove gettext overlinking
        reinplace "s|-lintl||" \
            ${buildlibdir}/_sysconfigdata__darwin_darwin.py

        file mtime ${buildlibdir}/_sysconfigdata__darwin_darwin.py $oldmtime
    }

    post-destroot {
        foreach dir { Headers Resources Python Versions/Current } {
            file delete ${destroot}${framewpath}/${dir}
        }

        ln -s ${framewdir}/share/man/man1/python${branch}.1 ${destroot}${prefix}/share/man/man1/
        ln -s ${framewdir}/lib/pkgconfig/python-${branch}.pc ${destroot}${prefix}/lib/pkgconfig/
        ln -s ${framewdir}/lib/pkgconfig/python-${branch}-embed.pc ${destroot}${prefix}/lib/pkgconfig/

        # Also make the sysconfig changes in the Makefile
        reinplace {s|^\(LINKFORSHARED=\).*$|\1 -L$(LIBPL) -lpython$(VERSION)$(ABIFLAGS) $(LIBS) $(SYSLIBS)|} \
            ${destroot}${framewdir}/lib/python${branch}/${confdir}/Makefile

        reinplace -E {s|-arch [a-z0-9_]+||g} \
           ${destroot}${framewdir}/lib/python${branch}/${confdir}/Makefile

        reinplace "s|-lintl||" \
           ${destroot}${framewdir}/lib/python${branch}/${confdir}/Makefile
    }
}

post-destroot {
    foreach unversioned {2to3 idle3 pydoc3 python3 python3-config} {
        delete ${destroot}${prefix}/bin/${unversioned}
    }
}

notes "
To make this the default Python or Python 3 (i.e., the version run \
by the 'python' or 'python3' commands), run one or both of:

    sudo port select --set python $pythonNoDot
    sudo port select --set python3 $pythonNoDot
"

variant optimizations description {enable expensive, stable optimizations (including PGO)} {
    configure.args-append   --enable-optimizations
}

variant lto description {enable Link-Time Optimization} {
    configure.args-append   --with-lto
    if {${os.major} < 11} {
    patchfiles-append	    patch-enable_dynamic.diff
    compiler.blacklist	    *gcc-4.0 *gcc-4.2
    }
}
```
3. Add my patch for LTO: https://github.com/macports/macports-ports/pull/14816/files#diff-77fb5d41c051fffdedeba53d495bf1a00818b75e81ecd2c28c59dfb87018f8a8

P. S. This is not an elegant solution for universal, but so far this is the only way I was able to build `+universal` and get working binaries. Alternative approaches failed: https://trac.macports.org/ticket/64916
If someone suggests a better way, it will be greatly appreciated.

Gcenx:
> needs_exe_wrapper = true

That’s very weird and more reserved for mingw builds so shouldn’t have caused any issue

kencu:
> If anyone interested to test though, here is what allows to build `python39` and `python310` as `+lto +universal` on Leopard:

It would appear that with gcc at least, "lto" offers about a 1% increase in speed on Intel processors... not sure about PowerPC. 

https://www.researchgate.net/publication/47374866_Optimizing_real_world_applications_with_GCC_Link_Time_Optimization

barracuda156:
> > needs_exe_wrapper = true
> 
> 
> 
> That’s very weird and more reserved for mingw builds so shouldn’t have caused any issue

Otherwise the build fails complaining about exe_wrapper. I wouldn’t even find such setting without a reason.

barracuda156:
But LTO is not the point here (and it is easy to fix anyway). What is broken is +universal for all Pythons on PPC, and only `python27` is fixed with a simple patch. Other require portfile changes I have pointed to above (or show me a better solution if you know).

kencu:
oh, you mentioned lto specifically, so I therefore assumed it was important to your need for a patch in some fashion

barracuda156:
@kencu Got it. Just mentioned it along, in case someone will be interested to reproduce what I have locally. LTO should not have impact on gobject-introspection build.

Gcenx:
Seems like meson incorrectly assumed if your system can natively run said executable so it looked for a “wrapper” and failed.

barracuda156:
> Seems like meson incorrectly assumed if your system can natively run said executable so it looked for a “wrapper” and failed.

@Gcenx Well, that’s on a stock Leopard. If someone has an idea how to fix it in `meson`, suggestions are welcomed. Otherwise, adding this to cross-file works:
```
[properties]
needs_exe_wrapper = false
```

Gcenx:
@barracuda156 I guess it might make sense to add that for the ppc/ppc64 cross-files.

I don’t see other solutions being accepted into Macports

kencu:
Do we know what is really wrong?

Somewhere in it's python code meson (I presume) asks itself if ppc Leopard can run ppc64 binaries and thinks it cannot, so therefore wants an exe wrapper? Is that it?

kencu:
This all has nothing to do with MacPorts, but if nobody is ever going to try to fix it, and if adding a line to say "skip the wrapper" works around it in a general fashion, I guess nobody will complain too much about that.

In general, however, you will find over and over that fixes to work around problems without fixing the actual problems come back to bite you, and the upper ups in MacPorts will not like them much. Too many times, these kind of "hacks" come back to bite you again, and then you have what often turns out to be more work to do to fix the fixes as well as the problem.

barracuda156:
@kencu I do not know the inner mechanism of that, but it is presumably `meson` screwing up. At the same time, I think this is the only such issue I have encountered. Only Leopard + `gobject-introspection`.

Given that we need our cross-files anyway, I tend to think cross-file solution is easy and painless. (It is also very easy to revert, if ever needed.)

barracuda156:
P. S. Cross-files usage is actually an officially recommended way to deal with specific build settings.

kencu:
> P. S. Cross-files usage is actually an officially recommended way to deal with specific build settings.

Oh, I know -- I was the person who came up with the MacPorts implementation, as you may remember.

Adding "wrong" settings to cross files to fix busted code in meson is what I'm taking about here.

Gcenx:
Going off `meson` own documentation is mentions that sometimes it will make incorrect assumptions about the need for an exe wrapper. simply setting `needs_exe_wrapper = false` lets `meson` know this won't be needed.

kencu:
simply indeed.

So meson just says it doesn't do this right.

Heh.

kencu:
Back to what I said above:

> This all has nothing to do with MacPorts, but if nobody is ever going to try to fix it, and if adding a line to say "skip the wrapper" works around it in a general fashion, I guess nobody will complain too much about that.

barracuda156:
@kencu Yes, I remember.

Do we have a conceivable case when `exe_wrapper` *is* needed for PPC though? If this setting simply tells `meson` not to do something wrong, I recon it is fine.
I agree that fixing upstream is preferable, but we both know that no one gonna bother to fix this for us. I can look into `meson` code just in case, but don't want to kill many hours to fix something that is otherwise fixed for all practical purposes trivially.

Gcenx:
> Do we have a conceivable case when `exe_wrapper` _is_ needed for PPC though? If this setting simply tells `meson` not to do something wrong, I recon it is fine. I agree that fixing upstream is preferable, but we both know that no one gonna bother to fix this for us. I can look into `meson` code just in case, but don't want to kill many hours to fix something that is otherwise fixed for all practical purposes trivially.

The problem here is `meson` doesn't believe your mac can run the pcc/pcc64 binaries so defaults to requiring an exe wrapper. By setting `needs_exe_wrapper = false` you've skipping this check so `meson` knows the current system _can_ indeed run this code.

barracuda156:
This might be wrong:
```
    elif trial.startswith(('powerpc64', 'ppc64')):
        trial = 'ppc64'
    elif trial.startswith(('powerpc', 'ppc')) or trial in {'macppc', 'power macintosh'}:
        trial = 'ppc'
```
Let me look into this a bit and maybe try a PR to `meson`.

barracuda156:
And then it has:
```
    if env.machines.matches_build_machine(for_machine):
        exe_wrap: T.Optional[ExternalProgram] = None
    else:
        exe_wrap = env.get_exe_wrapper()
```
So if it gets something like `macppc` from somewhere and then `ppc64`, it decides we are cross-compiling and consequently `exe_wrapper` needed. Crazy :)

barracuda156:
From `mesonbuild` log:
```
Running test binary command: /opt/local/var/macports/build/_opt_PPCLeopardPorts_gnome_gobject-introspection/gobject-introspection/work/build-ppc64/meson-private/sanitycheckc.exe
C compiler for the build machine: /usr/bin/gcc-4.2 (gcc 4.2.1 "powerpc-apple-darwin9-gcc-4.2.1 (GCC) 4.2.1 (Apple Inc. build 5577)")
C linker for the build machine: /usr/bin/gcc-4.2 ld64 85.2.1
Build machine cpu family: ppc
Build machine cpu: power macintosh
Host machine cpu family: ppc64
Host machine cpu: ppc64
Target machine cpu family: ppc64
Target machine cpu: ppc64
```

barracuda156:
Okay, I fixed it in `meson`.

barracuda156:
Not exactly sure if `gobject-introspection` is at fault here, but `pango` fails to build with `+universal` on 10.5.8 (`muniversal 1.0` added to portfile):
```
  File "/opt/local/bin/g-ir-scanner", line 99, in <module>
    sys.exit(scanner_main(sys.argv))
  File "/opt/local/lib/gobject-introspection/giscanner/scannermain.py", line 590, in scanner_main
    transformer = create_transformer(namespace, options)
  File "/opt/local/lib/gobject-introspection/giscanner/scannermain.py", line 409, in create_transformer
    transformer.register_include(include_obj)
  File "/opt/local/lib/gobject-introspection/giscanner/transformer.py", line 140, in register_include
    self._parse_include(filename)
  File "/opt/local/lib/gobject-introspection/giscanner/transformer.py", line 230, in _parse_include
    self._parse_include(dep_filename)
  File "/opt/local/lib/gobject-introspection/giscanner/transformer.py", line 223, in _parse_include
    parser.parse(filename)
  File "/opt/local/lib/gobject-introspection/giscanner/girparser.py", line 59, in parse
    tree = parse(filename)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/xml/etree/ElementTree.py", line 1229, in parse
    tree.parse(source, parser)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/xml/etree/ElementTree.py", line 580, in parse
    self._root = parser._parse_whole(source)
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 231, column 0
```

mascguy:
Adding @michaelld to the mix as well.

barracuda156:
@MarcusCalhoun-Lopez Any idea what could be the issue?

pmetzger:
I'm going to propose that we close this soon. If anyone wants to reopen it later in order to finish the work, they are of course free to do so.

pmetzger:
I'm closing this for now. It can be reopened when someone wants to work on it.

PR #14914, opened 2022-05-21
qmake5 PortGroup: improve sdk detection

kencu:
- hide output from sdk-finding command

the xcrun command is supposed to fail if the specific
sdk macosx${configure.sdk_version} is not found. However,
this error shows up on user's screens causing considerable
confusion about what the true build error is.

- ensure sdk macosx is found if we are going to use it


kencu:
I think it's right this time. I had to go learn some more tcl, specifically the ordering of the ```2>@1 and > /dev/null``` commands had to be the other way around.

PR #14931, opened 2022-05-23
net-snmp: add perl variants

i0ntempest:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 x86_64
Xcode 13.4 13F17a

###### Verification 
Have you

### DIFF
--- a/net/net-snmp/Portfile
+++ b/net/net-snmp/Portfile
@@ -27,7 +27,9 @@ long_description        This is net-snmp, a derivative of CMU's SNMP \
                         handle SNMP traps, a version of the unix 'netstat' \
                         command using SNMP, a Tk/perl MIB browser.
 
-perl5.branches          5.28
+perl5.branches          5.28 5.30 5.32 5.34
+perl5.default_branch    5.28
+perl5.create_variants   ${perl5.branches}
 
 depends_build           port:automake
 


net/net-snmp/Portfile
```diff
@@ -27,7 +27,9 @@ long_description        This is net-snmp, a derivative of CMU's SNMP \
                         handle SNMP traps, a version of the unix 'netstat' \
                         command using SNMP, a Tk/perl MIB browser.
 
-perl5.branches          5.28
+perl5.branches          5.28 5.30 5.32 5.34
+perl5.default_branch    5.32
```
ryandesign, line 31: Why default to 5.32? 5.28 is the preferred default per https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/perl5-1.0.tcl#L17.

ryandesign (commented):
Why are perl variants desired?

i0ntempest:
> Why are perl variants desired?

To reduce clutters. My system has 4 Perl versions installed due to ports using different versions of them. Also, mismatching Perl versions cause build failure with po4a (say you have perl5 +5_34 installed and po4a by default uses 5.30 iirc).
As for the default version, I’ll put them back to 5.28.

ryandesign:
Reducing the number of installed perl versions is exactly why we specify a preferred default perl version. The only ports that use a perl other than 5.28 should be those that are incompatible with 5.28. Of course we should use a newer perl by default in the near future but that hasn't been done yet.

If po4a experiences a build failure when perl5 is installed with a different variant then that is a bug in po4a; file a ticket.

i0ntempest:
Just put the default back to 5.28. Users should still have the choice though, I tend to make everything to use the newest Perl.

PR #14958, opened 2022-05-27
privoxy-pki-bundle: Set the version and revision to match curl-ca-bundle

essandess:
* Fixes: https://lists.macports.org/pipermail/macports-dev/2022-May/044278.html

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/www/privoxy/Portfile
+++ b/www/privoxy/Portfile
@@ -547,6 +547,11 @@ TLS_PRIVOXY_ROOT_CA
 }
 
 subport ${name}-pki-bundle {
+    revision        1
+    # revision 1 corresponds to curl-ca-bundle cert_date 20220325
+    # Please update the revision and comment above whenever subport
+    # curl-ca-bundle contents change
+
     license         MIT
 
     description     PKI Bundle for ${name}


JDLH:
To my outsider's eyes (not familiar with either port involved) this change is part of, but not quite, what the [macports-dev thread](https://lists.macports.org/pipermail/macports-dev/2022-May/044278.html) suggests. 

Adding a revision to the `privoxy-pki-bundle` subport looks like an important and helpful addition. This is what should get incremented each time the `curl-ca-bundle` port gets changed.

I don't know whether adding a version string to the `privoxy-pki-bundle` subport helps. My naive impression is that  trying to make the version number match port `curl-ca-bundle`'s history rather than its own history might be confusing. 

The comment, `# set the version and revision to match port curl-ca-bundle`, seems curious to me. It reads like a developer explaining the rationale for the most recent change they made. What I expect to see here is an imperative instruction addressed to future port updaters. Something like: "Increment the revision number here each time the  port curl-ca-bundle changes. This will force this port to rebuild, and incorporate the new curl-ca-bundle contents."

The macports-dev thread suggests putting a similar comment in the Portfile for port `curl-ca-bundle`. It should read something like, "Each time you update this port, please also increment the revision numbers of the following ports: subport privoxy-pki-bundle of port privoxy […and any others…]. Those ports depend on the contents of this port. Incrementing their revision numbers ensures that they are rebuilt to use the new contents of this port."

essandess:
> I don't know whether adding a version string to the `privoxy-pki-bundle` subport helps. My naive impression is that trying to make the version number match port `curl-ca-bundle`'s history rather than its own history might be confusing.

Rather, a direct equivalence to `curl-ca-bundle`'s version is what's required, as this port relies (mostly) on an up-to-date `/opt/local/share/curl/curl-ca-bundle.crt`. If an indirect `revision` number is used instead, then there's a confusing correspondence problem of which revision of `privoxy-pki-bundle` goes with which version of `curl-ca-bundle`, whether it's been updated, and then an annoying query into git history timestamps. Tying to the version makes this whole problem go away.

Additionally, if the port `privoxy-pki-bundle`  must ever be revised, that can be accomplished by making the appropriate edits and simply doing a revbump to keep the version correspondence with `curl-ca-bundle`.

JDLH:
> Rather, a direct equivalence to `curl-ca-bundle`'s version is what's required, as this port relies (mostly) on an up-to-date `/opt/local/share/curl/curl-ca-bundle.crt`. If an indirect `revision` number is used instead, then there's a confusing correspondence problem of which revision of `privoxy-pki-bundle` goes with which version of `curl-ca-bundle`, whether it's been updated, and then an annoying query into git history timestamps. Tying to the version makes this whole problem go away.

OK, so it sounds like you want a version numbering scheme which gives a "correspondence" between a version of port `curl-ca-bundle` and a version of port `privoxy-pki-bundle`. Some comments.

First, does including a **version** keyword get you that? The **version** keyword you are matching tracks the contents of port `curl`. I do not see a **version** keyword in subport `curl-ca-bundle`. Instead, it uses **certdata_updated**, **certdata_commit**, and **certdata_date**. 

Second, take a look at the diff for the [most recent update to `curl-ca-bundle`](https://github.com/macports/macports-ports/commit/c7b23eee221f62b365cf63ea3de247680948e370). What I see there is that no **version** value changes. The **certdata_date** value appears to be the primary identifier, and it changes. If you want to use a **version** keyword in `privoxy-pki-bundle`, maybe it should track the **certdata_date** value.

Third, as far as I understand, it does not matter to users of these ports whether version numbers correspond. MacPorts will always build `privoxy-pki-bundle` so that it uses whatever contents are in `curl-ca-bundle` at build time. The ony thing you need from the versioning system is that every time the contents of `curl-ca-bundle` changes, some part of the `privoxy-pki-bundle` version/revision identifiers becomes greater. Incrementing the revision number will accomplish this. Having a `privoxy-pki-bundle` increment the revision number just to be on the safe side won't hurt anything (except to cause users to rebuild when they did not need to). You could just maintain correspondence via a definition and comment like: 

```
revision 3
# Revision 3 corresponds to curl-ca-bundle cert_date 20220325 .
# Whenever subport curl-ca-bundle contents change, please update the revision definition above, 
# update the comment above. 
```

But I suppose what really matters is not what I think, but what will be most effective at getting maintainers of subport `curl-ca-bundle` to make the appropriate changes in subport `privoxy-pki-bundle` — as well as what will actually function in the desired way.

ryandesign:
I agree with Jim that there's no reason to modify the version or revision of privoxy-pki-bundle to match that of curl-ca-bundle. Instead, all that's needed is for the revision of privoxy-pki-bundle to be increased whenever curl-ca-bundle is updated, and a comment should be added to the curl-ca-bundle portfile to remind the maintainer of that port (me) to do that.

essandess:
Thanks done.

mascguy (approved):

PR #14998, opened 2022-05-31
scons: update to 4.3.0

barracuda156:
#### Description

Update to 4.3.0.
Github binary is broken, but PyPi version builds and works fine.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6 PPC (10A190)

###### Verification 
Have you


barracuda156:
Bot errors on:
```
/opt/local/Library/Frameworks/Python.framework/Versions/3.10/bin/python3.10 setup.py --no-user-cfg build -j3 
[287](https://github.com/macports/macports-ports/runs/6670955899?check_suite_focus=true#step:8:290)
[288](https://github.com/macports/macports-ports/runs/6670955899?check_suite_focus=true#step:8:291)
    File "/opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_devel_scons/scons/work/SCons-4.3.0/setup.py", line 4, in <module>
[289](https://github.com/macports/macports-ports/runs/6670955899?check_suite_focus=true#step:8:292)
      from setuptools import setup
[290](https://github.com/macports/macports-ports/runs/6670955899?check_suite_focus=true#step:8:293)
  ModuleNotFoundError: No module named 'setuptools'
```

Isn’t dependency on setuptools defined automatically by the `python 1.0` PG?

barracuda156:
> @barracuda156 no, as you can see in the PG file

Will fix this soon, thanks.

barracuda156:
@ryandesign Thank you for fixing https for the homepage!

Why do we have checksums in such a weird arrangement in this port though?

ryandesign:
I think it was Michael who started placing the checksums near the version (since they get updated together) a few years ago, and I started doing that in all my ports some time after that, and others have picked it up as well.

When this PR is squashed and merged, we'll want to add to the commit message the URLs of the two tickets that this resolves: https://trac.macports.org/ticket/59303 and https://trac.macports.org/ticket/60957. I had been holding off on updating scons to version 4 because of the assumption that it would break something. But I'm ok with updating now, since it's been a long enough time that scons 4 fixes should either be available from the developers of those ports already, or else the ports are abandoned and no fix will likely ever appear from the developers so we'll have to develop our own sometime anyway.

I see that with this update you have removed the post-destroot block that installed the documentation. I'd rather retain the documentation if possible. If you have any comments about why you had to remove it let me know, otherwise I'll see if I can figure out a way to keep it.

ryandesign:
I see that you switched the download from sourceforge to pypi. Any particular reason? It looks like sourceforge is still where they release their files.

barracuda156:
> I think it was Michael who started placing the checksums near the version (since they get updated together) a few years ago, and I started doing that in all my ports some time after that, and others have picked it up as well.

Got it. I was just surprised at that, no objections whatsoever. Whichever convention is fine IMO.

> I see that with this update you have removed the post-destroot block that installed the documentation. I'd rather retain the documentation if possible. If you have any comments about why you had to remove it let me know, otherwise I'll see if I can figure out a way to keep it.

Due to this: https://github.com/SCons/scons/issues/3950
I was not able to make it build with docs from GitHub, and PyPi does not have those docs, from what I understood. (There was some reason why I initially switched from SF to GH, and then after failed attempts with GH, to PyPi, but tbh I cannot say now which in particular.)

Please feel free to modify the PR in any way you see appropriate. I am not an expert in Python ports, I just updated this in order to build some other port.

PR #15004, opened 2022-06-01
lldb-xx: migrate codesign info to notes

mascguy:
#### Description

Proposed change to the various `lldb-xx` subports, moving code signing instructions to port notes. This ensures that the user can easily revisit those via `port notes lldb-xx`. Notes are also less likely to be missed/lost, vs. outputting them via `ui_msg`.

p.s. The only downside is that notes might be line-wrapped, depending on the width of a user's terminal. But the advantages of notes arguably outweigh that.


kencu (approved):

mascguy:
Thanks Ken!

PR #15008, opened 2022-06-01
gcc11/libgcc11: skip bootstrap comparison for 10.6 and earlier

mascguy:
#### Description

Fix build failure for `libgcc11` on macOS 10.6, by disabling bootstrap comparison.

See: https://trac.macports.org/ticket/65149

NOTE: Change hasn't been tested locally yet: Goal is to first ensure this is the right approach, before waiting an eternity for a build test. (Likewise, CI is being skipped at the moment, to avoid wasting resources.)


kencu:
seems fine… all these years making all these systems do bootstrap comparison has been mostly (completely?) a waste of time anyway.

PR #15011, opened 2022-06-01
ffmpeg: prefer libsmbclient from samba4 instead of samba3

markmentovai:
ffmpeg+gpl3’s libsmbclient dependency may be satisfied by either samba3
or samba4. This updates its preference to samba4.

The port’s revision is unchanged because no port rebuild is required. If
samba3 is installed, it will continue to satisfy this dependency. As
samba3 and samba4 conflict, there is no ambiguity within a MacPorts
installation, and ffmpeg will continue to depend on whichever samba
version is installed absent user intervention.

This also updates notes to refer to libsmbclient by the correct name.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4 13F17a

###### Verification 
Have you

### DIFF
--- a/multimedia/ffmpeg/Portfile
+++ b/multimedia/ffmpeg/Portfile
@@ -376,7 +376,7 @@ variant gpl3 requires gpl2 description {Enable GPL code, license will be GPL-3+}
                             --enable-libsmbclient \
                             --enable-libaribb24
 
-    depends_lib-append      path:lib/pkgconfig/smbclient.pc:samba3 \
+    depends_lib-append      path:lib/pkgconfig/smbclient.pc:samba4 \
                             port:libaribb24
 
     license                 GPL-3+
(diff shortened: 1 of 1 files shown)


mascguy:
I'm with you in-principle, relative to this change.

But practically speaking, this will be irritating for users: That's because virtually every other Samba dependent relies on `samba3`, and most of those require that version.

So if we default to `samba4`, and users attempt to install anything else that requires `samba3`, that will fail with a conflict error. So while folks are welcome to use either with `ffmpeg`, we want to follow a path that results in the least potential user pain. And at the moment, that's `samba3`.

Does that make sense?

markmentovai:
I see. Thanks for the explanation.

OK, withdrawn.

PR #15053, opened 2022-06-03
libLASi: fix universal build

barracuda156:
#### Description

This PR enables `+universal` build.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8
Xcode 3.1.4

###### Verification 
Have you

P. S. A typography commit is not just a matter of æsthetics in this case, it fixes code coloring in BBEdit.

### DIFF
--- a/print/libLASi/Portfile
+++ b/print/libLASi/Portfile
@@ -3,6 +3,10 @@
 PortSystem              1.0
 PortGroup               cmake 1.1
 
+if {[string match *gcc* ${configure.compiler}]} {
+    PortGroup           muniversal 1.0
+}
+
 name                    libLASi
 version                 1.1.3
 revision                1


mascguy (approved):
LGTM. @tenomoto ?

kencu:
I believe all the reinplacing is no longer needed and most likely just adding muniversal (probably only if the compiler is gcc) would do it.

There is not much enthusiasm in MacPorts in general for adding muniversal just to fix universal builds with gcc, by the way… as that causes too many problems to support 0.00000001% of users, and there is no end to it.

barracuda156:
> I believe all the reinplacing is no longer needed and most likely just adding muniversal (probably only if the compiler is gcc) would do it.

I can check that once back to my PPC machines.

mascguy:
Rebased against master, fixing merging conflicts. Though more Universal changes still pending.

kencu:
I suspect just remove all this, and we'll be just fine:

```
if {[variant_isset universal]  && [info exists universal_archs_supported]} {
    post-configure {
        foreach arch ${universal_archs_supported} {
            if {[file isfile ${worksrcpath}-${arch}/build/src/CMakeFiles/LASi.dir/link.txt]} {
                reinplace "s|@rpath|${prefix}/lib|" ${worksrcpath}-${arch}/build/src/CMakeFiles/LASi.dir/link.txt
            }
        }
    }
} else {
    post-configure {
        reinplace "s|@rpath|${prefix}/lib|" ${workpath}/build/src/CMakeFiles/LASi.dir/link.txt
    }
}

cmake.out_of_source     yes
```

mascguy:
> I suspect just remove all this, and we'll be just fine:
> [...]

Done!

kencu:
this port requires a few dozen universal deps to be built, so I'm not up for that.

I suspect (but don't know for sure) that the universal build will work fine without muniversal if the compiler is not gcc.

I'm personally not in favour of adding muniversal to 25,000 ports just to support universal building with gcc that is relevant to 0.00000001% of users. If it must be added, perhaps it would be added only if the compiler is actually a version of gcc that needs it.

The better fix for this is to fix newer versions of gcc to be a multi-compiler like gcc-4.2 is and like clang is. That might be a reasonable project for someone to take on (not me).

mascguy:
> this port requires a few dozen universal deps to be built, so I'm not up for that.
> 
> I suspect (but don't know for sure) that the universal build will work fine without muniversal if the compiler is not gcc.
> 
> I'm personally not in favour of adding muniversal to 25,000 ports just to support universal building with gcc that is relevant to 0.00000001% of users. If it must be added, perhaps it would be added only if the compiler is actually a version of gcc that needs it.
> 
> The better fix for this is to fix newer versions of gcc to be a multi-compiler like gcc-4.2 is and like clang is. That might be a reasonable project for someone to take on (not me).

So are you in favor of simply closing this PR then, or...?

barracuda156:
@kencu @mascguy Well, here we have a fix that does solve existing failure for this specific port. If it does not break anything for anyone else, why not have it fixed?

Making gcc universal is great, but it won’t be made soon.

mascguy:
> @kencu @mascguy Well, here we have a fix that does solve existing failure for this specific port. If it does not break anything for anyone else, why not have it fixed?
> 
> Making gcc universal is great, but it won’t be made soon.

Agreed.

Ken, originally it sounded like you weren't opposed to the change. Have you simply thought more about all of this from a big-picture perspective, and changed your mind?

kencu:
I'm not in charge of anything, but others, including Ryan, have expressed concerns about adding muniversal to Ports solely to support building universal with gcc.

I don't think we should do that, in general, but again, not up to me.

If you were going to do it, I would do something like this pseudocode:

```
if (gcc but not gcc-4.2) then
  #workaround gcc that can't build universal at present
  PortGroup muniversal 1.0
endif
```

and then it would affect nobody, and be easy to find to strip out if / when somebody fixes gcc.

But that is me. I'm a neat freak.

kencu:
Oh, I haven't changed my mind, by the way. It's the same comment I made three weeks ago (above) and elsewhere.

kencu:
> @kencu @mascguy Well, here we have a fix that does solve existing failure for this specific port. If it does not break anything for anyone else, why not have it fixed?

Nobody knows if it breaks things for others. It might. I believe this is completely untested, except on 10.5 PPC.

> Making gcc universal is great, but it won’t be made soon.

Be great to keep the pressure on that fix, though, instead of 25,000 muniveral additions to MacPorts that all need to be stripped out.

IMHO -- almost nobody on 10.4 and 10.5 cares a whit about universal anyway. They build for the native compiler they have, and are happy to have that. But that's my take.

mascguy:
> I believe all the reinplacing is no longer needed and most likely just adding muniversal (probably only if the compiler is gcc) would do it.

Ah gotcha, missed your original comment. Seems like a good idea, to help prevent potential impact to others.

barracuda156:
> If you were going to do it, I would do something like this pseudocode:
> 
> ```
> if (gcc but not gcc-4.2) then
>   #workaround gcc that can't build universal at present
>   PortGroup muniversal 1.0
> endif
> ```
> 
> and then it would affect nobody, and be easy to find to strip out if / when somebody fixes gcc.
> 
> But that is me. I'm a neat freak.

@kencu @mascguy This looks perfectly good to me, since we don’t use `clang` on PPC systems anyway.
Could someone add this condition in a proper way to the PR? (And then in all future cases I will include that, so we don’t face identical issues anymore.)

P. S. Off topic, but what has been the purpose of `muniversal` if not to fix building `+universal` when it fails otherwise?

mascguy:
> > If you were going to do it, I would do something like this pseudocode:
> > ```
> > if (gcc but not gcc-4.2) then
> >   #workaround gcc that can't build universal at present
> >   PortGroup muniversal 1.0
> > endif
> > ```
> @kencu @mascguy This looks perfectly good to me, since we don’t use `clang` on PPC systems anyway. Could someone add this condition in a proper way to the PR? (And then in all future cases I will include that, so we don’t face identical issues anymore.)

I'm wondering whether we should create a new portgroup for such a case - something like `muniversal-conditional`, or whatever the best name for it is - to encapsulate this. That would also make it easy to tweak the logic, without having to go back and change the ports that need it. (I wouldn't dare muck around in `muniversal`, for fear of breaking something.)

mascguy:
> @mascguy I think I suggested something like that elsewhere. The problem we get on PPC is typical: whenever C or C++ standard is higher than supported by `gcc42`, `+universal` fails without `muniversal` PG (plus in those instances where `gcc42` fails for other reasons).
> 
> P. S. I have to mention that, unfortunately, simply adding `muniversal` does not often solve the problem, and additional fixes are needed. Usually they are rather trivial. Few disastrous ports like Pythons 3.x and `gobject-introspection` are a pain though.

Great point: If there other gymnastics that you routinely have to include in ports - and which could be selectively enabled, via options defined by the new pg - those could also be included.

Granted, we wouldn't want to rush this, and we'd do it via a collaborative PR. But definitely something that might be worthwhile, particularly if we can eliminate common copy-pasta across ports.

kencu:
> P. S. Off topic, but what has been the purpose of `muniversal` if not to fix building `+universal` when it fails otherwise?

There are a number of ports that can't build with the default "multiarch" clang style of building universal. Most of these configure themselves differently depending on the different arch they are building for.

There are a number of ports that build perfectly fine with clang's "multiarch" building style, but FAIL to build when muniversal is added, for various reasons. We don't want to screw around with these, or discover 10,000 more with this issue that we don't know about now.

Look at all the fun I had trying to get icu to build universal, which requires muniversal sometimes but fails with it at other times:

https://github.com/macports/macports-ports/blob/e4dc1e08a821d31b02994d4c2052951ed18f55cf/devel/icu/Portfile#L47

There are complexities to this you only begin to learn after you have been doing it a long while. You learn, for one thing, not to screw around too much with things that are currently working.

kencu:
I'm not 100% sure why a third muniversal portgroup would be needed, myself, but I will leave this with you to sort out if you are motivated to do so.

mascguy:
> I'm not 100% sure why a third muniversal portgroup would be needed, myself, but I will leave this with you to sort out if you are motivated to do so.

You're right, and perhaps I'm simply being overly-cautious.

Indeed, we could (carefully) add opt-in functionally to `muniversal`, for legacy scenarios. And so long as it's well-scrutinized before rollout, hopefully it shouldn't be a problem.

PR #15068, opened 2022-06-05
ddrescue: fix compilation error related to std::exit

mascguy:
### Description

File `loggers.cc` references `std::exit`, without including `cstdlib`. This works with some compilers, but not others.

Fixes: https://trac.macports.org/ticket/65300

### Tested on

* macOS 10.7
* macOS 10.9
* macOS 10.11


jmroot (changes requested):
Changing compilers is not the right fix for a missing include.

mascguy:
> Changing compilers is not the right fix for a missing include.

PR updated

mascguy:
Retested on 10.7, 10.9, and 10.11, with latest approach. And looks good on all three.

raimue (approved):
Looks good to me. :+1:

PR #15076, opened 2022-06-06
util-linux: include getopt

markmentovai:
There is a separate getopt port, which conflicts. The conflict is marked appropriately.

The util-linux version of getopt was synced with the last-released version of the standalone getopt (1.1.6, 2014-11-24) at b5f3078e1474
(2014-12-08).

Since then, there have been no changes to the standalone getopt, thus it appears unmaintained. There have been changes to the copy in util-linux, which is still actively maintained.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

@kurthindenburg


mascguy:
> Given that the two `getopt`s are indeed the same thing, with one (`util-linux`) being maintained and the other (frodo) not, the best path forward may be to just treat the frodo source as dead and stop using it. Instead, we could make the `getopt` port depend on `util-linux`, which would assume the responsibility for providing the `getopt` utility and man page, with no conflicts declared.

That sounds like a better approach, particularly if our current port for `getopt` is based on unmaintained/outdated code.

Anyone else?

ryandesign:
Options:

1.
   * make the util-linux port also install the getopt program, manpage, etc.
   * mark the getopt port `replaced_by util-linux`
   * change all ports that depend on port getopt so that they depend on port util-linux instead
2.
   * update the getopt port to install the getopt program, manpage, etc. from util-linux
   * potentially make getopt a subport of util-linux to make updates easier
   * potentially make port util-linux depend on port getopt
   * this could be an opportunity to make the other util-linux utilities available as subports as well instead of in one monolithic port, though if the utilities are small and build quickly this may be of negligible benefit

markmentovai:
@ryandesign Due to the small size of the `util-linux` build (or the portion used by MacPorts, anyway), I went with option 1 from [your comment](https://github.com/macports/macports-ports/pull/15076#issuecomment-1151714708), implemented here now.

mascguy:
Hey Mark, not trying to rush you. But just wanted to quickly check in, and see where things stand with this PR?

markmentovai:
@mascguy I thought we were ready to go here as of [this comment](https://github.com/macports/macports-ports/pull/15076#issuecomment-1155251393).

mascguy (approved):

PR #15082, opened 2022-06-07
submission: speedtest

linuxgemini:
This commit adds the Portfile for Ookla's own Speedtest.net CLI.

Signed-off-by: İlteriş Yağıztegin Eroğlu <ilteris@asenkron.com.tr>

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 x86_64
Xcode 13.3.1 13E500a

###### Verification 
Have you

### DIFF
--- a/net/speedtest/Portfile
+++ b/net/speedtest/Portfile
@@ -1,0 +1,51 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+version             1.1.1.84
+revision            0
+
+name                speedtest
+categories          net
+maintainers         {@linuxgemini linuxgemini.space:macportspackaging} openmaintainer
+
+homepage            https://speedtest.net/apps/cli
+license             restrictive nomirror noncommercial
+
+description         official cli for speedtest.net
+long_description    command line interface for testing internet bandwidth using speedtest.net
+
+master_sites        https://install.speedtest.net/app/cli/
+distname            ookla-speedtest-${version}-macosx-x86_64
+extract.suffix      .tgz
+extract.mkdir       yes
+
+universal_variant   yes
+supported_archs     x86_64 arm64
+
+checksums           rmd160  c16334261d6858b21ef08ebb4886ced3c303420e \
+                    sha256  153f76e4e502a2bb470c7f7c3b4b8afaa50eb98bb11e86216fcb1a8b92676e94 \
+                    size    1019149
+
+use_configure       no
+build {}
+
+destroot {
+    set installdir ${prefix}/bin
+
+    xinstall -m 755 -W ${worksrcpath} \
+        speedtest \
+        ${destroot}${installdir}
+}
+
+post-destroot {
+    set manpagedir ${prefix}/share/man/man5
+
+    xinstall -m 444 -W ${worksrcpath} \
+        speedtest.5 \
+        ${destroot}${manpagedir}
+}
+
+# ookla does not provide a public filetree other than their homebrew pointer file
+# at https://github.com/teamookla/homebrew-speedtest/blob/master/speedtest.rb
+livecheck.type      none


linuxgemini:
> @linuxgemini we're typically not really in favor of just downloading binaries... and the licensing of this isn't ideal either. What does this that cannot be done with, for example, the `speedtest-cli` port?

Yeah, I have my issues about the licensing as well (may as well just close it to be honest). AFAICT speedtest-cli is feature-complete compared to the official binary. Other than a rather minor issue about server selection (which there is a fix but the PR hasn't been merged since last December).

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
...
+
+PortSystem          1.0
+
+version             1.1.1.84
+revision            0
+
+name                speedtest
+categories          net
+platforms           darwin
+supported_archs     noarch
```
ryandesign, line 11: It is not noarch.
```suggestion
```

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+version             1.1.1.84
+revision            0
+
+name                speedtest
+categories          net
+platforms           darwin
```
ryandesign, line 10: This line can be omitted. (It is the default.)
```suggestion
```

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
...
+
+description         official cli for speedtest.net
+long_description    command line interface for testing internet bandwidth using speedtest.net
+
+master_sites        https://install.speedtest.net/app/cli/
+distname            ookla-speedtest-${version}-macosx-x86_64
+extract.suffix      .tgz
+
+# ookla already ships this as an universal binary
+universal_variant   no
```
ryandesign, line 25: Since the binary is universal, the port should have a universal variant. Either the port can force the universal variant on, to convey that the software that is installed is universal, or ideally the universal variant can be optional, and if the user does not select it, then the port uses `lipo` to remove the architectures that were not requested.

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
...
+                    sha256  153f76e4e502a2bb470c7f7c3b4b8afaa50eb98bb11e86216fcb1a8b92676e94 \
+                    size    1019149
+
+use_configure       no
+build {}
+
+destroot {
+    set installdir ${prefix}/bin
+
+    xinstall -d ${destroot}${installdir}
```
ryandesign, line 38: MacPorts creates standard directories for you.
```suggestion
```

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
...
+    xinstall -m 755 -W ${workpath} \
+        speedtest \
+        ${destroot}${installdir}
+}
+
+post-destroot {
+    set docdir ${prefix}/share/man/man5
+
+    xinstall -d ${destroot}${docdir}
```
ryandesign, line 47: MacPorts creates standard directories for you.
```suggestion
```
We usually use the name "docdir" to refer to a location within ${prefix}/share/doc where additional documentation files are installed, so it is confusing to use such variable name for a different purpose, such as installing manpages here. Since you're only using the variable once, you don't need a variable at all.

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
...
+                    size    1019149
+
+use_configure       no
+build {}
+
+destroot {
+    set installdir ${prefix}/bin
+
+    xinstall -d ${destroot}${installdir}
+    xinstall -m 755 -W ${workpath} \
```
ryandesign, line 39: Since evidently this distfile does not extract with an enclosing directory, we prefer for the port to ask MacPorts to create that enclosing directory by using `extract.mkdir yes`. Then all references to `${workpath}` in the Portfile will need to be changed to `${worksrcpath}` as per usual.

net/speedtest/Portfile
```diff
@@ -0,0 +1,54 @@
...
+    set docdir ${prefix}/share/man/man5
+
+    xinstall -d ${destroot}${docdir}
+    xinstall -m 644 -W ${workpath} \
+        speedtest.5 \
+        ${destroot}${docdir}
+}
+
+# ookla does not provide a public filetree
+livecheck.type      none
```
ryandesign, line 54: How did you discover that the version number was 1.1.1.84? Is there no web page where that number is listed that could be used for livecheck?

linuxgemini: Hi Ryan, Ookla themselves have a tap for homebrew at https://github.com/teamookla/homebrew-speedtest/blob/master/speedtest.rb

ryandesign: Ok, then how did they discover that this was the latest version?

net/speedtest/Portfile
linuxgemini: Resolved in b64b2932c58a1adfdb20129e1234c07b65dcb6cc

net/speedtest/Portfile
linuxgemini: TIL about `lipo`, thank you.

Resolved in b64b2932c58a1adfdb20129e1234c07b65dcb6cc.

net/speedtest/Portfile
linuxgemini: Resolved in b64b2932c58a1adfdb20129e1234c07b65dcb6cc.

net/speedtest/Portfile
linuxgemini: Resolved in b5c45453f88a930d37fc00cb64eaf0283cf5e57c.

net/speedtest/Portfile
linuxgemini: Resolved in b5c45453f88a930d37fc00cb64eaf0283cf5e57c.

net/speedtest/Portfile
linuxgemini: Resolved in b5c45453f88a930d37fc00cb64eaf0283cf5e57c.

linuxgemini:
Closing the PR due to the nature of the official distribution being binary-only with no direct way to check for updates _and_ having a quite restrictive license.

PR #15103, opened 2022-06-09
VLC3: upgrade, +universal support and various fixes

RJVB:
#### Description
This upgrades VLC to 3.0.17.3 (the latest version for which binary builds seem to exist). It also reintroduces arm64 support, a +universal variant and various small fixes.

###### Type(s)
- [X ] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on
OS X 10.9.5

###### Verification 
Have you

- [X ] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [X ] tried a full install with `sudo port -vst install`?


barracuda156:
@RJVB Is it possible to add at least a pre-built (DMG) variant for PPC? There are versions for Leopard and Tiger here: https://www.videolan.org/vlc/download-macosx.html
(Leopard version will also apply to 10.6 PPC, so if it does not hurt, please include that case too.)

P. S. In theory VLC2 should build for PPC, at least its earlier versions, but I understand that you cannot fix it without PPC hardware to test it. (I will try fixing it somewhat later.)

barracuda156:
@RJVB Thank you for responding!

Of course I did not mean I expect you to fix the current – or any earlier – version to build for PPC. (I will see later on if I can fix that, and if I get any success, I gonna make a PR.)
What I meant is merely having pre-built DMG downloaded and installed via the port in a case of 10.4–10.6 coupled with `ppc` arch. This is only a suggestion. I could do that myself, but I am away from PPC hardware this month and just cannot test anything.

RJVB:
>What I meant is merely having pre-built DMG downloaded and installed via the port in a case of 10.4–10.6 coupled with `ppc` arch. This is only a suggestion. I could do that myself, but I am away from PPC hardware this month and just cannot test anything.

I would still argue that this will have to be done in the VLC2 portfile, because port:VLC2 is the one that provides VLC 2.x . With this current PR, PPC users should again be getting port:VLC2 when they try to install port:VLC, because they'll be running an OS that is not supported by VLC v3 .

barracuda156:
> What I meant is merely having pre-built DMG downloaded and installed via the port in a case of 10.4–10.6 coupled with `ppc` arch. This is only a suggestion. I could do that myself, but I am away from PPC hardware this month and just cannot test anything.
> I would still argue that this will have to be done in the VLC2 portfile, because port:VLC2 is the one that provides VLC 2.x . With this current PR, PPC users should again be getting port:VLC2 when they try to install port:VLC, because they'll be running an OS that is not supported by VLC v3 .

Okay, fair enough.

RJVB:
Older OS versions are supposed to use port:VLC2 and as far as I know there must be a binary image of the default variant configuration.

I don't know if VLC 2.2.8 still builds on PPC; apparently 2.1.0 and newer only support 64 bits (and Snow Leopard or up) at least as far as the GUI plugin is concerned. You'd have to try building it on PPC.

If you look at the Portfile you will maybe understand why I dropped my efforts to keep trying to build the code after 3.0.11 or so. There are too many patchfiles already. While it would not be impossible to manage 2 versions of those where needed to support building an older VLC 2.x on PPC I am frankly not interested in trying to implement that.

Someone motivated enough could go through the git history of port:VLC to find the latest version that built on PPC. But be aware that the update to 2.1.0 was #4b132a0b03a3d6ec545a084fd3be1bb63d4418a9 in Sept. 2013, long before I was named maintainer, and that building such old code against the current versions of its dependencies (FFMpeg in particular) can be a daunting task.

Or alternatively, somebody could give port:VLC2 a conditional section that works like port:VLC now does. Adding such a section isn't very hard (take a look at my personal copy of port:VLC in my macstrop repo for how to add such a section); what you do in it ... I'd rather not know :). From what I can tell Videolan still provide the x86_64, i386 and PPC DMGs for VLC 2.0.9 which is better than I feared it might be.

barracuda156:
@RJVB I will try to deal with this in coming days. If we find which version builds for PPC (perhaps for VLC2), there is little to no maintenance burden: just peg the version, no updates.

multimedia/VLC/Portfile
```diff
@@ -76,15 +96,15 @@ build               {}
 
 destroot {
     xinstall -m 755 -d ${destroot}${applications_dir}
-    system "ditto --rsrc ${workpath}/VLC-${version}/VLC.app ${destroot}${applications_dir}/VLC.app"
+    system "ditto --rsrc ${workpath}/VLC-${version}/VLC.app ${destroot}${applications_dir}/VLC3.app"
```
sideeffect42, line 99: I personally find it a bit strange to name the app VLC3.app.

Is there a case where someone would like to have VLC2 and VLC3 installed at the same time?

multimedia/VLC/Portfile
```diff
@@ -97,18 +117,43 @@ post-destroot {
     foreach l {libvlc.pc vlc-plugin.pc} {
         xinstall -m 644 ${filespath}/${l} ${destroot}${vlcprefix}/lib/pkgconfig
         reinplace "s|@PREFIX@|${prefix}|g" ${destroot}${vlcprefix}/lib/pkgconfig/${l}
-        reinplace "s|@VERSION@|${version}|g" ${destroot}${vlcprefix}/lib/pkgconfig/${l}
     }
-    system "echo \"#!/bin/sh\nexec \\\"${applications_dir}/VLC.app/Contents/MacOS/VLC\\\" \\\"\\\$\@\\\"\" > ${destroot}${prefix}/bin/vlc3"
+    reinplace "s|\$\{VERSION\}|${version}|g" ${destroot}${vlcprefix}/lib/pkgconfig/vlc-plugin.pc
+    reinplace "s|@VERSION@|${version}|g" ${destroot}${vlcprefix}/lib/pkgconfig/libvlc.pc
+    system "echo \"#!/bin/sh\nexec \\\"${applications_dir}/VLC3.app/Contents/MacOS/VLC\\\" \\\"\\\$\@\\\"\" > ${destroot}${prefix}/bin/vlc3"
```
sideeffect42, line 123: Couldn't the launcher script be replaced with a symlink in this case?

multimedia/VLC/Portfile
barracuda156: > Is there a case where someone would like to have VLC2 and VLC3 installed at the same time?

IMO highly unlikely, but possibly for testing purposes (?).

RJVB:
On Monday March 20 2023 12:38:13 sideeffect42 wrote:

>Is there a case where someone would like to have VLC2 and VLC3 installed at the same time?

I have but I also seem to recall that there were certain features that were dropped from VLC3. Either way, some day VLC 4 will be released, there is no good reason NOT to make the versions co-installable so we can just as well start versioning the application name now so there's no ambiguity.

> Couldn't the launcher script be replaced with a symlink in this case?

It could, but in the past it has been judged cleaner to use a script and I agree with that. Note the use of `exec`  which makes the script overlay itself with VLC (and that this is not a new feature).

multimedia/VLC/Portfile
sideeffect42: Couldn't testing be done by switching VLC version using `port deactivate`/`port activate`?

barracuda156: It can, just conflict should be set in the portfile.

sideeffect42:
> > Couldn't the launcher script be replaced with a symlink in this case?
>
> It could, but in the past it has been judged cleaner to use a script and I agree with that. Note the use of `exec` which makes the script overlay itself with VLC (and that this is not a new feature).

I understand.
After checking the ports tree I've seen that the same launcher script is also used by other ports (e.g. audacity, ipe, unison).

pmetzger:
This has now been stuck for a year and a half. Should we simply close it?

RJVB:
>This has now been stuck for a year and a half. Should we simply close it?

Probably...

pmetzger:
Closing for now.

barracuda156:
@RJVB @sideeffect42 What was the status of this on arm64 btw? It would be nice to have it building from source in the latest version, of course.

RJVB:
>@RJVB @sideeffect42 What was the status of this on arm64 btw? It would be nice to have it building from source in the latest version, of course.

AFAICT it should work, but you should still be able to get the proposed version of the port directory from this PR, and try for yourself. I spent quite a bit of time on figuring out the various available build versions and even implementing a +universal variant.

I ditched the build-from-source approach as it was always a very complicated business and IIRC I couldn't build the latest source on my OS version anymore. The only reason I still made a port rather than deciding to let people install from the upstream images themselves is that it allows me to provide a (working!) libVLC. I don't think there's anything in stock MacPorts that depends on that library though.

PR #15105, opened 2022-06-10
cmake-bootstrap: Backport a fix for find_library

markmentovai:
CMake versions older than 3.18 (including `cmake-bootstrap`, pegged at 3.9) have a bug in `find_library` that causes them to find libraries when filesystem nodes are present at a candidate path, even if they are not usable. This can happen “innocently” since macOS 11, where expanded use of the dyld shared cache has caused the removal of most OS dylibs from the filesystem.

For example, when instructed to look for shared library `iodbc.2`, a CMake with this bug running on macOS 11 or later might notice a symbolic link at `/usr/lib/libiodbc.2.dylib` and return it. That dylib is not usable, as it’s “dangling”: it refers to `libiodbc.2.1.18.dylib`, which isn’t present on disk, it’s only in the dyld shared cache. This CMake bug might cause a build to attempt to link against `/usr/lib/libiodbc.2.dylib`, which is, of course, impossible, as the symbolic link points nowhere. The expected behavior is for CMake to discard that unusable path and continue its search, eventually locating `$SDKROOT/usr/lib/libiodbc.2.tbd`.

This backports [the fix present in CMake 3.18](https://gitlab.kitware.com/cmake/cmake/-/commit/f2c903fb9a02) which causes `find_library` to check that a file is readable before returning it. A symbolic link that points nowhere will not pass this check. This improves `cmake-bootstrap`’s reliability on newer OS versions where OS dylibs are only in the dyld shared cache and not in the filesystem.

This also updates `cmake-bootstrap` to 3.9.6. The `cmake-bootstrap` port is intended to be pegged to the last version of CMake to not require C++11. In reality, [the diff between the existing 3.9.4 and 3.9.6](https://gitlab.kitware.com/cmake/cmake/-/compare/v3.9.4...v3.9.6) is minimal and does not introduce a C++11 requirement. There’s no reason for the port not to be at the terminal version on the CMake 3.9 branch, 3.9.6.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

@ryandesign (recently touched this port in 5b37f3a51044), @kencu (originated this port in 6d03d8692113).

### DIFF
--- a/devel/cmake-bootstrap/Portfile
+++ b/devel/cmake-bootstrap/Portfile
@@ -12,12 +12,12 @@ maintainers         nomaintainer
 dist_subdir         cmake
 
 set branch          3.9
-version             ${branch}.4
+version             ${branch}.6
 distname            cmake-${version}
 
-checksums           rmd160  bb9c30f2726413f0d7d7b92e0294b61438151d62 \
-                    sha256  b5d86f12ae0072db520fdbdad67405f799eb728b610ed66043c20a92b4906ca1 \
-                    size    7705052
+checksums           rmd160  858257d95fde810195e4cf7254072dd8db608cc4 \
+                    sha256  7410851a783a41b521214ad987bb534a7e4a65e059651a2514e6ebfc8f46b218 \
+                    size    7705400
 
 homepage            https://cmake.org
 master_sites        ${homepage}/files/v${branch}/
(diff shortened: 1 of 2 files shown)


kencu:
LGTM

PR #15114, opened 2022-06-11
google-cloud-sdk: python310 dependency

jrabinow:
#### Description

Updates google-cloud-sdk port to depend on python3.10 instead of python3.9. This had already been attempted in the past, but failed.

[The bug](https://issuetracker.google.com/issues/205005959) that prevented [this PR](https://github.com/macports/macports-ports/pull/13936) from being merged has been marked duplicate of a fixed bug [here](https://issuetracker.google.com/issues/202172882)
Also, as I pointed out [here](https://github.com/macports/macports-ports/pull/13936#issuecomment-1034485449), the dependency version in the portfile is actually ignored: the shellscript shipped in this package just invokes whatever python3 it can find on the system.
I believe it's time to reconsider this change.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.4 21F79 x86_64
Xcode 13.3 13E113

###### Verification 
Have you

### DIFF
--- a/devel/google-cloud-sdk/Portfile
+++ b/devel/google-cloud-sdk/Portfile
@@ -5,7 +5,7 @@ PortGroup           python 1.0
 
 name                google-cloud-sdk
 version             389.0.0
-revision            0
+revision            1
 categories          devel python
 license             Apache-2
 maintainers         {breun.nl:nils @breun} openmaintainer
@@ -41,7 +41,7 @@ master_sites        https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/
 
 worksrcdir          ${name}
 
-python.default_version 39
+python.default_version 310
 
 post-patch {
     reinplace "s|\"disable_updater\": false|\"disable_updater\": true|" ${worksrcpath}/lib/googlecloudsdk/core/config.json


breun:
Python was upgraded from 3.8 to 3.9 for this port via https://github.com/macports/macports-ports/commit/30b2f28eeea5769e5765b6e54d4245114b337cd7, but I was not involved with that change.

I personally would prefer to stick to the latest officially supported version of Python mentioned at https://cloud.google.com/sdk/docs/install#mac.

There might not be any known issues with 3.9 (or 3.10) right now, but looking at https://issuetracker.google.com/issues/170125513#comment19 and further there were some issues on 3.9 and 3.10 in the past.

jrabinow:
Yes, in the past there were issues - it seems they were fixed though, so this seems to me  more a discussion on tradeoff between stability and being up-to-date.

Looking at the other platforms, I see:
- windows says that python3 is supported while python3.5 to 3.8 are "preferred", whatever that means
- generic linux and mac specify python 3 supported, (3.5 to 3.8, with 3.7 preferred)
- Debian/Ubuntu and RH/Centos/Fedora just have a note about upgrading from python2 to python3

My suspicion is that google's infra/language team has blessed python3.7 as being the "official" version for use everywhere internally, and they haven't gotten around to updating beyond. Another possible explanation is that no one updated the preferences in those docs since python 3.8 came out.
I'll also mention I've been [running gcloud with python3.10 on my personal laptop for a few months now](https://github.com/jrabinow/homedir-scripts/blob/d9260f69bf63f8ccf1994bf1983beb312594a79a/platform_specific/mac/macports_update#L39) without issues.

In the end it's your call.

> Would you be able to add that change

I have my reserves about patching this one. As I mentioned here https://github.com/macports/macports-ports/pull/13936#issuecomment-1034485449, this could easily turn into an uphill battle. If it was guaranteed to be a one-time thing, maybe, but I'm too short on bandwidth these days to create additional friction for myself in the future (and AFAIK I have to say things seem to work well enough today)

breun:
I also don't know what Google means with 'preferred' versions, but I think it's smart to stay within the version range defined by Google as 'supported'. I know that this is already not true anyway since the update to 3.9, but I also didn't want to downgrade the Python dependency to 3.8 again as long as there are no issues reported.

I guess the Google Cloud SDK shell scripts could be patched to replace the `CLOUDSDK_PYTHON=$(order_python python3 python2 python2.7 python)` line with `CLOUDSDK_PYTHON=${macports-prefix}/bin/python${python-version}`, because that would be guaranteed to exist with a MacPorts dependency on a specific `python*` port.

jrabinow:
Sorry for only getting back to you now, I'm afraid I don't have much bandwidth.

Ok then, I guess that patch could make sense, the reserves I shared earlier haven't changed but I can see that having that patch could be an improvement. I'll close this PR since it sounds unlikely you'll want to approve it soon, but I'd just like to ask, if google-cloud-sdk is never officially supported with python3.10, do we just leave things as they are forever? Would it make sense to wait and if it's still not supported in 2023-2024 for example, we could do this then?

Typically in situations like these, I tend to do as much testing as possible, release, and if a problem crops up, revert. I'm a bit on the "move fast and break things" side though, so if that's not an option here, I understand.

breun:
I've created https://github.com/macports/macports-ports/pull/15182 to make Google Cloud SDK use the Python binary from MacPorts. Just the `gcloud` binary already has a ton of things it can do, and I can't come close to testing all the things it can do, so I've decided to stick to the officially supported and recommended Python versions for now.

PR #15117, opened 2022-06-12
libiptcdata: add intl-related libs to link path; update deps for gettext; add missing deps

mascguy:
### Description

* Port currently fails to build for Big Sur and Monterey, due to Intl-related libs not present on link path.
* Also update deps for `gettext`, and add other missing deps
* Presently a blocker for `rawtherapee`, on Big Sur and Monterey

Fixes: https://trac.macports.org/ticket/65330

CC: @jasonliu--

### Tested on
macOS 11.6.7
Xcode 13.0


mascguy:
Reviewers: Additional deps determined based on both review of configure output, combined with trace mode.

mascguy:
One other note: A rev-bump is included in the changes, to validate and rebuild across-the-board.

graphics/libiptcdata/Portfile
```diff
@@ -24,6 +26,18 @@ master_sites        sourceforge:project/libiptcdata/libiptcdata/${version}
 checksums           rmd160  64e1b7cd1d5166091227eb651f9945bc3862e39b \
                     sha256  79f63b8ce71ee45cefd34efbb66e39a22101443f4060809b8fc29c5eebdcee0e
 
-depends_lib         port:gettext
+# Add missing intl-related libs to link path
+patchfiles-append   patch-Makefile.in-intl.diff
+
+depends_build-append \
+                    port:cctools \
```
jmroot, line 33: Why is cctools needed?

graphics/libiptcdata/Portfile
```diff
@@ -24,6 +26,18 @@ master_sites        sourceforge:project/libiptcdata/libiptcdata/${version}
...
+                    port:gettext \
+                    port:gtk-doc \
```
jmroot, line 35: Shouldn't the prebuilt docs in the distribution be sufficient, making a gtk-doc dep unnecessary?

mascguy: Good point, doc generation is disabled by default. I'll remove.

graphics/libiptcdata/Portfile
mascguy: The configure script looks for basic tools like `nm`, `ranlib`, and `strip`, though I suppose most everything does as well.

Generally speaking, when should we depend on `cctools`, vs. allowing a port to use the versions provided by macOS? I'm thinking back to `ffmpeg`, for example, where it's critical to depend on it. (Okay, that's a slightly different scenario, since it also involved `libffi`. But figured I'd mention it.)

[63443 - ffmpeg-4.4_3+gpl2: build failures: libffi dylib version bump breaks core ports like LLVM](https://trac.macports.org/ticket/63443)

jmroot: Just like the compiler, we use the other toolchain components provided by Xcode unless they don't work for some reason.

mascguy:
Maintainer timeout reached, merging.

PR #15120, opened 2022-06-12
sslscan: update to 2.0.13

breun:
#### Description

Update to sslscan 2.0.13.

###### Tested on

macOS 12.4 21F79 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/net/sslscan/Portfile
+++ b/net/sslscan/Portfile
@@ -3,8 +3,8 @@
 PortSystem 1.0
 PortGroup github 1.0
 
-github.setup    rbsec sslscan 2.0.10
-revision        1
+github.setup    rbsec sslscan 2.0.13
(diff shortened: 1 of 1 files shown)


net/sslscan/Portfile
```diff
@@ -19,9 +19,9 @@ license         {GPL-3+ OpenSSLException}
 
 depends_lib     port:openssl
```
herbygillot, line 20: Maybe use the `openssl` portgroup?

breun: I’m not familiar with that PortGroup and I don’t see it documented on https://guide.macports.org/chunked/reference.portgroup.html. What benefit would using that PortGroup bring for this port?

PR #15136, opened 2022-06-14
qt*-qtbase: improve SDK detection

markmentovai:
The `qt*-qtbase` ports attempted to build a valid SDK name as `macosx${configure.sdk_version}`. This candidate SDK was validated by checking whether `MacOSX${configure.sdk_version}.sdk` exists in the SDKs directory within the developer directory.

In 9887e90d69f4265f9056cddc45e41551d7400235, identical logic in the `qmake5` port group was revised to instead check for a valid SDK based on `${configure.sdk_version}` by running `xcrun --sdk macosx${configure.sdk_version} --show-sdk-path`. This was later revised by be588d4c1b8bf39b090c7b44cb1e5ef9ecda0dfd to use `--find ld` instead of `--show-sdk-path`, which is a newer option not honored by all versions of `xcrun`.

This change brings the improved logic from the `qmake5` port group (and later updates) to the various `qt*-qtbase` subports, which build via the Qt `configure` script and not `qmake`, and do not otherwise benefit from the improvements already made to the `qmake5` port group.

This is necessary because unlike other recent versions of Xcode, Xcode 14.0b1 14A5228q contains a `MacOSX13.sdk` symbolic link in addition to the expected `MacOSX13.0.sdk` symbolic link. This means that the path existence check for `MacOSX${configure.sdk_version}.sdk` succeeded in this configuration when `configure.sdk_version` was 13, as it will be by default on macOS 13. This resulted in Qt being configured with an SDK identifier of `macosx13`. Although the `MacOSX13.sdk` symbolic link was present, `xcrun` and other Xcode tools do not recognize `macosx13` as a valid SDK identifier, causing a build configured in this way to fail.

The expectation based on other recent versions of Xcode (since Xcode 12 and the macOS 11 SDK) is for the existence check to fail, causing the build to fall back to the generic `macosx` SDK. By instead using `xcrun` with a candidate SDK name (`xcrun --sdk macosx13`), these ports can now check to see whether the SDK identifier to be given to the Qt build is considered valid by the Developer Tools installation. This is much more accurate than the path existence check.

The `qt*-qtbase` build winds up encoding the SDK identifier into its installed files for its own use during `qmake`-based builds as `QMAKE_MAC_SDK`, so it’s most desirable for MacPorts’ purposes to track the SDK major version when possible, but it’s absolutely undesirable to track the SDK minor version, as it will cause `qmake` to fail if a Developer Tools update changes the SDK’s minor version. These minor version bumps have occurred in recent Developer Tools updates that identify SDKs by both major and minor version numbers. Because `QMAKE_MAC_SDK` may only be set to a short identifier (`macosx10.15` or `macosx`) and not a pathname, there is no way to currently identify such SDKs to the tools by just a major version number. The only remaining option is to specify the SDK as the most generic `macosx`. However, if the developer tools do eventually regain the capability to properly specify an SDK by just major version (`macosx13`), the strategy implemented here will allow these ports to take advantage of it.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.0 22A5266r arm64
Xcode 14.0 14A5228q

###### Verification 
Have you

@landonf (9887e90d69f4265f9056cddc45e41551d7400235)
@MarcusCalhoun-Lopez (be588d4c1b8bf39b090c7b44cb1e5ef9ecda0dfd)
@kencu (fbdae6e6c2cc9c2a8b8a625e322aeee6bd00e610)

### DIFF
--- a/aqua/phantomjs-qt/Portfile
+++ b/aqua/phantomjs-qt/Portfile
@@ -368,17 +368,19 @@ foreach {module module_info} [array get modules] {
 
             # starting with Xcode 7.0, the SDK for build OS version might not be available
             # see https://trac.macports.org/ticket/53597
-            if { ${use_xcode} } {
-                if {[vercmp $xcodeversion 4.3] < 0} {
-                    set sdks_dir ${configure.developer_dir}/SDKs
+            #
+            # avoid --show-sdk-path since it is not available on all platforms
+            # see https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261
+            if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld  > /dev/null 2>@1}]} {
+
+                # if no specific sdk can be found, check for a generic macosx sdk
+                if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx --find ld > /dev/null 2>@1}]} {
+                    ui_error "${subport}: no usable SDK can be found"
+                    return -code error "no usable SDK can be found"
                 } else {
-                    set sdks_dir ${configure.developer_dir}/Platforms/MacOSX.platform/Developer/SDKs
+                    ui_debug "${subport}: using generic macosx SDK as macosx${configure.sdk_version} does not exist"
+                    configure.sdk_version
                 }
-            } else {
-                set sdks_dir ${configure.developer_dir}/SDKs
-            }
-            if { ![file exists ${sdks_dir}/MacOSX${configure.sdk_version}.sdk] } {
-                configure.sdk_version
             }
 
             # respect configure.sdk_version
--- a/aqua/qt5/Portfile
+++ b/aqua/qt5/Portfile
@@ -1178,17 +1178,19 @@ foreach {module module_info} [array get modules] {
 
             # starting with Xcode 7.0, the SDK for build OS version might not be available
             # see https://trac.macports.org/ticket/53597
-            if { ${use_xcode} } {
-                if {[vercmp $xcodeversion 4.3] < 0} {
-                    set sdks_dir ${configure.developer_dir}/SDKs
+            #
+            # avoid --show-sdk-path since it is not available on all platforms
+            # see https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261
+            if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld  > /dev/null 2>@1}]} {
+
+                # if no specific sdk can be found, check for a generic macosx sdk
+                if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx --find ld > /dev/null 2>@1}]} {
+                    ui_error "${subport}: no usable SDK can be found"
+                    return -code error "no usable SDK can be found"
                 } else {
-                    set sdks_dir ${configure.developer_dir}/Platforms/MacOSX.platform/Developer/SDKs
+                    ui_debug "${subport}: using generic macosx SDK as macosx${configure.sdk_version} does not exist"
+                    configure.sdk_version
                 }
-            } else {
-                set sdks_dir ${configure.developer_dir}/SDKs
-            }
-            if { ![file exists ${sdks_dir}/MacOSX${configure.sdk_version}.sdk] } {
-                configure.sdk_version
             }
 
             # respect configure.sdk_version
--- a/aqua/qt511/Portfile
+++ b/aqua/qt511/Portfile
@@ -999,17 +999,19 @@ foreach {module module_info} [array get modules] {
 
             # starting with Xcode 7.0, the SDK for build OS version might not be available
             # see https://trac.macports.org/ticket/53597
-            if { ${use_xcode} } {
-                if {[vercmp $xcodeversion 4.3] < 0} {
-                    set sdks_dir ${configure.developer_dir}/SDKs
+            #
+            # avoid --show-sdk-path since it is not available on all platforms
+            # see https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261
+            if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld  > /dev/null 2>@1}]} {
+
+                # if no specific sdk can be found, check for a generic macosx sdk
+                if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx --find ld > /dev/null 2>@1}]} {
+                    ui_error "${subport}: no usable SDK can be found"
+                    return -code error "no usable SDK can be found"
                 } else {
-                    set sdks_dir ${configure.developer_dir}/Platforms/MacOSX.platform/Developer/SDKs
+                    ui_debug "${subport}: using generic macosx SDK as macosx${configure.sdk_version} does not exist"
+                    configure.sdk_version
                 }
-            } else {
-                set sdks_dir ${configure.developer_dir}/SDKs
-            }
-            if { ![file exists ${sdks_dir}/MacOSX${configure.sdk_version}.sdk] } {
-                configure.sdk_version
             }
 
             # respect configure.sdk_version
--- a/aqua/qt513/Portfile
+++ b/aqua/qt513/Portfile
@@ -1002,17 +1002,19 @@ foreach {module module_info} [array get modules] {
 
             # starting with Xcode 7.0, the SDK for build OS version might not be available
             # see https://trac.macports.org/ticket/53597
-            if { ${use_xcode} } {
-                if {[vercmp $xcodeversion 4.3] < 0} {
-                    set sdks_dir ${configure.developer_dir}/SDKs
+            #
+            # avoid --show-sdk-path since it is not available on all platforms
+            # see https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261
+            if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx${configure.sdk_version} --find ld  > /dev/null 2>@1}]} {
+
+                # if no specific sdk can be found, check for a generic macosx sdk
+                if {[catch {exec -ignorestderr /usr/bin/xcrun --sdk macosx --find ld > /dev/null 2>@1}]} {
+                    ui_error "${subport}: no usable SDK can be found"
+                    return -code error "no usable SDK can be found"
                 } else {
-                    set sdks_dir ${configure.developer_dir}/Platforms/MacOSX.platform/Developer/SDKs
+                    ui_debug "${subport}: using generic macosx SDK as macosx${configure.sdk_version} does not exist"
+                    configure.sdk_version
                 }
-            } else {
-                set sdks_dir ${configure.developer_dir}/SDKs
-            }
-            if { ![file exists ${sdks_dir}/MacOSX${configure.sdk_version}.sdk] } {
-                configure.sdk_version
             }
 
             # respect configure.sdk_version
--- a/aqua/qt53/Portfile
+++ b/aqua/qt53/Portfile
@@ -851,17 +851,19 @@ foreach {module module_info} [array get modules] {
 
             # starting with Xcode 7.0, the SDK for build OS version might not be available
             # see https://trac.macports.org/ticket/53597
-            if { ${use_xcode} } {
-                if {[vercmp $xcodeversion 4.3] < 0} {
-                    set sdks_dir ${configure.developer_dir}/SDKs
+            #
+            # avoid --show-sdk-path since it is not available on all platforms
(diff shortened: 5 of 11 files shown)


kencu:
a remaining issue to fix was the fact that xcrun seems to call through to xcodebuild in at least some circumstances with our current coding. This then requires us to have 

use_xcode yes

in portfiles or they still error out. I didn’t fully sort out when it is needed to have xcodebuild available, or if it can be prevented by tweaking the code. There is a ticket where I demonstrated this… I’ll see if I can find it. 

This issue may not be relevant here if these builds already specify use_xcode yes.

kencu:
here’s the ticket where I put down my experiments

https://trac.macports.org/ticket/65293

markmentovai:
@kencu:
> here’s the ticket where I put down my experiments
> 
> https://trac.macports.org/ticket/65293

I don’t follow. There, you show that `xcrun --sdk macosx --find ld` and even `xcrun --sdk macosx --show-sdk-path` both work, even if Xcode is not available.

In any event, the message that [ticket 65293](https://trac.macports.org/ticket/65293) reports [comes from Qt itself](https://github.com/qt/qtbase/blob/v6.3.1/mkspecs/features/mac/sdk.prf#L8) where it runs `xcrun` all on its own. None of that’s changing here. Either it’s already broken for these ports, or it’s not—if it’s already broken, this change won’t make it any worse, and if it’s currently working, Qt’s existing use of `xcrun` indicates that it’ll also work when the Portfile does the same.

I’m sure that there are other bugs here (and some of them are likely on the Qt side), but I don’t think that this change will make them any worse.

kencu:
> @kencu:
> 
> > here’s the ticket where I put down my experiments
> > https://trac.macports.org/ticket/65293
> 
> I don’t follow. 

1. the detection code I fixed up in the qmake5 PG is still not quite 100% right.
2. "use_xcode yes" is still needed, sometimes, and it ideally should not be
3. it may in fact be needed only if the the user has Xcode installed. (In which case xcrun calls through to xcodbuild). 

So ideally we would get it 100% right before copying it into all these qt ports. My suggestion was to perhaps just call the basic "macosx" SDK on darwin 20+, as this would be unlikely to generate an error xcode or not, but there will be other opinions on that.

markmentovai:
@kencu:
> 2. "use_xcode yes" is still needed, sometimes, and it ideally should not be

Can we figure out when, though? The part that I’m having trouble with is that you showed that both `xcrun --sdk macosx --find ld` and `xcrun --sdk macosx --show-sdk-path` were doing the right thing on macOS 12.4, with or without Xcode installed.

> 3. it may in fact be needed only if the the user has Xcode installed. (In which case xcrun calls through to xcodebuild, apparently only on newer systems).

The ports in question may already handle that. [qt5, for example](https://github.com/macports/macports-ports/blob/02697547c3a68d82fbda996b373bbaa254b82ba5/aqua/qt5/Portfile#L32).

kencu:
> @kencu:
> 
> > 2. "use_xcode yes" is still needed, sometimes, and it ideally should not be
> 
> Can we figure out when, though? 

Hey, that would be great!

> The part that I’m having trouble with is that you showed that both `xcrun --sdk macosx --find ld` and `xcrun --sdk macosx --show-sdk-path` were doing the right thing on macOS 12.4, with or without Xcode installed.

Well, not quite. 

The experiment suggests that xcrun calls through to xcodebuild if Xcode is installed, but not if Xcode is not installed. My current thinking is that call through to xcodebuild fails if "use_xcode yes" is not in the portfile. Dunno exactly why just now, if that is right.

markmentovai:
@kencu:
> Well, not quite.
> 
> The experiment suggests that xcrun calls through to xcodebuild if Xcode is installed, but not if Xcode is not installed. My current thinking is that call through to xcodebuild fails if "use_xcode yes" is not in the portfile. Dunno exactly why just now, if that is right.

Ah, I see what you were getting at, then.

Excerpting from a deep-dive training I gave at work on this facility a few months ago:

> `/usr/bin/xcrun` is dispatched via `/usr/lib/libxcselect.dylib`, and will follow whatever the active Developer Tools installation is. `libxcselect.dylib` honors the `DEVELOPER_DIR` environment variable, if set. Otherwise, it follows the symbolic link at `/var/db/xcode_select_link` (which is what `xcode-select --switch` and `xcode-select --reset` maintain), and if that’s not present, it tries a few hard-coded paths including common installation locations for both Xcode.app and Command Line Tools.

If no preferred tools installation has been chosen via `xcode-select --switch` and /Applications/Xcode.app is present, it will be used in preference to a Command Line Tools installation. An easy way to circumvent this, if both are installed but you really want Command Line Tools, is to set `DEVELOPER_DIR`. And, in fact `use_xcode yes` does control this in MacPorts.

Anyway, that’s all kind of background material. I took a closer look at [your findings](https://trac.macports.org/ticket/65293#comment:10) and did spot something suspicious.

> ```
> % /usr/bin/xcrun --sdk macosx12  --find ld
> xcrun: error: SDK "macosx12" cannot be located
> /Library/Developer/CommandLineTools/usr/bin/ld
> ```

That’s bad. We’re expecting `--sdk macosx12` to be an impossible configuration, because the SDK can’t actually be identified that way. It’s either `macosx12.3` (since Xcode 13.3 and CLT equivalent) or `macosx`, but never `macosx12`. But in the Command Line Tools case, while `xcrun --sdk macosx12 --show-sdk-path` fails, `xcrun --sdk macosx12 --find ld` warns about the missing SDK but ultimately succeeds. MacPorts since https://github.com/macports/macports-ports/commit/be588d4c1b8bf39b090c7b44cb1e5ef9ecda0dfd takes that as a success indication (and why wouldn’t it?) and assumes that it means that the SDK is valid. It isn’t.

The comment history at https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235 shows why `--show-sdk-path` didn’t work and why you eventually landed on `--find ld` instead, but unfortunately, it seems that while `--find ld` can be used to validate an SDK identifier with Xcode, it’s totally wrong for Command Line Tools, because it gives false positives.

markmentovai:
https://github.com/macports/macports-ports/commit/9887e90d69f4265f9056cddc45e41551d7400235#commitcomment-49824261 showed that `xcrun --show-sdk-path` doesn’t work on some older OS versions (presumably OS X 10.7, but it wasn’t nailed down explicitly), hence the `xcrun --find ld` technique.

As I mentioned [above](https://github.com/macports/macports-ports/pull/15136#issuecomment-1159774591), [since 5.9](https://github.com/qt/qtbase/blob/5.9/mkspecs/features/mac/sdk.prf), Qt uses `xcrun --show-sdk-path` on its own, so avoiding that construct in MacPorts doesn’t make any sense—Qt’s own build is going to use it (and fail on older OS versions that don’t support it) anyway. And [your experiments](https://trac.macports.org/ticket/65293#comment:10) show that `xcrun --show-sdk-path` works equally well for Xcode.app and Command Line Tools.

There are a handful of `qt5*` ports older than `qt59`, though: `qt53`, `qt55`, `qt56`, `qt57`, and `qt58`. For this discussion, I’m assuming that there’s nothing about them inherently incompatible with older OS versions.

One option would be to sniff out the string `cannot be located` from `xcrun`’s `stderr`, and treat it as failing to locate the SDK by the provided name. The check should still also consider a nonzero exit code from `xcrun` to be a failure, this would just provide another option to detect failure. Unfortunately, it’s not possible to treat _any_ `stderr` output as failure, because recent `xcrun`/`xcselect` tools tend to produce noise on `stderr`.

mascguy:
Mark/Ken, where do things stand? There's no rush, but just wanted to quickly checkin relative to overall status.

kencu:
I'm not sure using xcrun to locate the SDK makes great sense in the end, given all the problems it is causing. MacPorts base already sniffs out the SDK, so why aren't we somehow using that instead?

Otherwise, if that can't work, having a different method for different OS versions rather than one method for all versions might be needed, I guess. 

Our previous method only became a problem once we got into the macosx12.1, 12.2, 12.3, etc business it seems?

markmentovai:
> Our previous method only became a problem once we got into the macosx12.1, 12.2, 12.3, etc business it seems?

Well, here, yes, mostly. Apple screwed this up by switching from naming SDKs for major (annual) releases (10.14, 10.15) to naming them for minor releases (12.0, 12.1), all by accident, as a result of continuing to carry two version components when they changed version number schemes. There’s an indication from changes in Xcode 14.0 betas that they might have some intention of fixing this someday, although as of 14.0b3, it’s still broken (actually, slightly _more_ broken than before).

> Otherwise, if that can't work, having a different method for different OS versions rather than one method for all versions might be needed, I guess.

I don’t recall exactly where `xcrun --show-sdk-path` originated, but it almost certainly works with all of the Xcode versions that could feasibly run on macOS 11 and later, which is where you’d start to see these problems. But given the finding in [bug 65293](https://trac.macports.org/ticket/65293) that `xcrun --find ld` might return a path to `ld` under Command Line Tools (even if it doesn’t do that under non-CLT Xcode.app), I think you probably wouldn’t want to use “does `xcrun --sdk … --find ld` seem to work?” as an indicator _at all_, at least not without checking its `stderr`.

The point here is, although the problem really only surfaced in a big way when the significance of SDK version numbers changed, the results of the investigation also show that we shouldn’t really trust the `xcrun --find ld` strategy (without checking `stderr`), since it’s broken in the CLT case.

> I'm not sure using xcrun to locate the SDK makes great sense in the end, given all the problems it is causing. MacPorts base already sniffs out the SDK, so why aren't we somehow using that instead?

What’s tricky here is that MacPorts knows the SDK name, but correctly doesn’t want to give the SDK name with a full version to Qt, if the full version is more specific than an annual major OS release. The reason is that the SDK name gets baked into Qt’s own build tools, which will forever attempt to use that SDK in the future. That means that if you started with `qt5-qtbase` built while Xcode 13.1 was installed, you’d wind up with a Qt that looks for SDK `macosx12.1`. That’s fine until you upgrade to Xcode 13.3, which no longer contains an SDK by that name, but instead carries `macosx12.3`. Now the `qt5-qtbase` port is broken and it won’t be able to build anything, because the SDK it’s looking for isn’t there (anymore). Since MacPorts packages are versioned by major OS version and not by Xcode or CLT version, it’s not possible to have a single `qt5-qtbase` that relies on `macos12.anything` and work correctly under all circumstances. This is a problem for `qt*-qtbase` builds done locally across Xcode upgrades, and it’s an even bigger problem for binary packages that are done with who-knows-what toolchain versions available, without any knowledge of what’s going to be available on the systems it runs on.

But I guess there’s still hope? If you wanted to rewrite the logic from the ground up without touching `xcrun` or `xcodebuild` at all, it might look like:

```
sdk_version = whatever follows macosx in, say, 12.3
if number of components in sdk_version == 1
  sdk_major_version = sdk_version
else if first component of sdk_version == 10
  sdk_major_version = first two components of sdk_version  # will require special handling for 10.4u if still relevant
else
  panic!
if sdk_major_version == os_version && sdk_version == sdk_major_version
  tell Qt to use SDK "macosx" + sdk_version
else
  tell Qt to use SDK "macosx" without any version
```

Maybe what I’m calling `sdk_major_version` here is interesting enough across the board that it could be available globally.

kencu:
@chrstphrchvz fyi

chrstphrchvz:
I’m aware of this issue but have no thoughts on what to do about it. Xcode 14 will be out any week now.

kencu:
Although the improvements I talked about still make sense to me in general, in the real world this port is still broken and nothing more has been done to fix it in months.

So -- if this PR fixes it, I'd say go with it.

kencu:
This brings the logic from the qmake5 portgroup into the qt5 ports. 

It has been working in the qmake5 PG for months now, and it fixes the build on Ventura with Xcode 14.

markmentovai:
Thanks, @kencu. Was going to poke this pull request this morning for obvious reasons, but you beat me to it.

My [comment](https://github.com/macports/macports-ports/pull/15136#issuecomment-1182630383) might have the foundation of a better general solution, but it’s hairy for sure.

i0ntempest:
Does this affect ports that use qt5 in any way? Some of my ports (say `qbittorrent-qt5`) still does not configure without `configure.sdk_version` workaround.

kencu:
qbittorrent-qt5 should be getting this fix via the qmake5 portgroup…which it is using…

you can build with “-d” to get extra messages from the fix.

If that sheds no light, perhaps some puts sprinkled here and there will lead to what’s happening…

i0ntempest:
I can see `qt5-qtbase: using generic macosx SDK as macosx13 does not exist` in the build log, but I have `MacOSX.sdk     MacOSX12.3.sdk MacOSX12.sdk   MacOSX13.0.sdk MacOSX13.sdk` in `/Library/Developer/CommandLineTools/SDKs/`.
`xcrun --show-sdk-path` doesn't work with `--sdk macosx13`, but works with `--sdk macosx13.0` or just `--sdk macosx`, just like in the qscintilla-qt5 ticket, but that port builds fine on my system.

kencu:
Something else, then. Is there a ticket?

i0ntempest:
Not yet, I'll figure out how many of my ports are affected by this and create a ticket for all of them.

i0ntempest:
https://trac.macports.org/ticket/66113

PR #15148, opened 2022-06-17
docker-machine-driver-qemu: new port

sainnhe:
#### Description

I want to install docker from macports instead of official website, but when I run docker via: `docker ps`, an error message is printed as follows:

```
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
```

This is because the `docker` binary is only a client and you cannot use it to run the docker daemon, we have to install `docker-machine` in order to create VM and attach to it.

The default driver is VirtualBox, but it's not available on Apple Silicon Macs. There is also a port `docker-machine-parallels` that uses Parallels Desktop as driver, but it's not FOSS.

I believe we should have a fully open source solution, so that's what this port is going to do.

[machine-drivers/docker-machine-driver-qemu](https://github.com/machine-drivers/docker-machine-driver-qemu) is a docker machine driver that uses qemu as backend, which is fully open source and supports both x86_64 and aarch64 architectures.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix
- [x] submission

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/docker-machine-driver-qemu/Portfile
+++ b/devel/docker-machine-driver-qemu/Portfile
@@ -1,0 +1,43 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        machine-drivers docker-machine-driver-qemu 0324171328f7bdf2f1e7803dbdc3b18db457c786
+version             20220617
+revision            0
+categories          devel
+license             Apache-2
+maintainers         {@sainnhe gmail.com:sainnhe} openmaintainer
+description         Docker machine driver for pure qemu/kvm - no libvirt
+long_description    A plugin for Docker Machine allowing to \
+                    create Docker hosts locally on QEMU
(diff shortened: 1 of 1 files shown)


devel/docker-machine-driver-qemu/Portfile
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        machine-drivers docker-machine-driver-qemu 0324171328f7bdf2f1e7803dbdc3b18db457c786
+version             20220617
+revision            0
+categories          devel
+platforms           darwin
```
mascguy, line 10: The default platform is now `darwin`, so this declaration can be removed.

sainnhe: > The default platform is now `darwin`, so this declaration can be removed.

Thanks! I've removed that.

mascguy (approved):
LGTM. Anyone else?

PR #15166, opened 2022-06-20
muniversal: strip host information when merging

kencu:
for consideration? (by the way, needed the leading space in the search pattern otherwise cmp thinks the files differ based on space runs I believe).

host information can vary from build arch to build arch
depending on the native compiler and the cross build

but this host information is not relevant to the config files and
scripts, and requires stripping it out individually in ports to
facilitate merging

so just always try it when merging suitable files


MarcusCalhoun-Lopez (commented):
This looks like a good idea!
I do have a question about the space before `--host` and `host_alias`.
Other than jemalloc, I do not have any examples to look at, but can we assume there will always be a space?

kencu:
I guess it could happen there might not be a space :> . I'm not sure how to manage that -- strip space runs out of the files too, at the end of processing? (I assume there is some tricky regex that does that too... ).

PR #15177, opened 2022-06-21
gcc-devel, libgcc-devel: update arm64 to 12-20220615

markmentovai:
#### Description

https://github.com/iains/gcc-12-branch/compare/gcc-12.1-darwin-r0...gcc-12.1-darwin-r1

This brings macOS 13 and Xcode 14 support to gcc-devel, modulo one SDK fix required in Xcode 14.0b1 (already reported to Apple).

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

macOS 13.0 22A5266r arm64
Xcode 14.0 14A5228q

###### Verification 
Have you

@michaelld (most recent updates to gcc-devel), @cjones051073 (some recent updates to gcc-devel)


kencu:
we also need to do this, to un-break all the rpath wreckage.

```
 Adds a new configure option ```--with-darwin-extra-rpath=``` which allows
   a distribution to prepend a run path to the list automatically added by
   the compiler.  This is to allow for cases where the distribution refers to
   installed libraries using symlinks from outside the installation paths.
```

ie in the arm version that is building with iain's fork, add this to the configure args:

```
configure.args-append --with-darwin-extra-rpath=${prefix}/lib/libgcc
```

We should also do this same thing to the gcc11 port (use iain's fork instead of FX's fork, and add this option there too).

And then we should just use that same fork for all our builds, Intel and arm64, so we stop messing ourselves up with different gcc behaviour.)

kencu:
I have been using this rpath-fixed gcc11 from Iain's fork myself for some time now on Monterey Intel to kick the tires.

It fixes all the rpath issues, while still allowing rpath to be used, which Iain requires if we want him to keep maintaining gcc on Darwin.

kencu:
By the way -- Iain added this rpath option _specifically for MacPorts_ to use for our unique gcc library layout.

markmentovai:
@kencu:
> we also need to do this, to un-break all the rpath wreckage.

I’ve updated with this addition.

> We should also do this same thing to the gcc11 port (use iain's gcc11 fork instead of FX's fork, and add this option there too). Then gcc11 will no longer be broken for arm64, and things can work properly again with the buildbot builds.

I think that should be handled in a separate change.

> (And then we should just use that same fork for all our builds, Intel and arm64, so we stop messing ourselves up with different gcc behaviour.)

For `gcc11`, sure.

For `gcc-devel` also, in theory. But right now, the x86_64 `gcc-devel` port has moved on to the GCC trunk (to become GCC 13), while arm64 is being held to https://github.com/iains/gcc-12-branch. Seems like https://github.com/iains/gcc-darwin-arm64 is a bit more out of date, so there isn’t a good arm64 GCC trunk option right now. Since x86_64 `gcc-devel` has already moved past 12, I wouldn’t want to downgrade it just because arm64 can’t move forward yet.

> It fixes all the rpath issues, while still allowing rpath to be used, which Iain requires if we want him to keep maintaining gcc on Darwin.

I really don’t love this entire rpath-based technique, but that’s an entirely different conversation. No need to clutter up this update with that.

PR #15205, opened 2022-06-25
volk: try to tweak compilers to enforce actual c++17 std::filesystem …

michaelld:
#### Description

volk: try to tweak compilers to enforce actual c++17 std::filesystem support

This is currently a test PR to see how QA/CI responds to the tweaks. Once we determine the correct settings, maybe try to move them into the compilers PG?

Closes: https://trac.macports.org/ticket/65377

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.8 20G715
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/science/volk/Portfile
+++ b/science/volk/Portfile
@@ -9,15 +9,16 @@ PortGroup           boost 1.0
 
 compiler.c_standard   2011
 compiler.cxx_standard 2017
-
-# some older Clang say they support SIMD when they don't
-compiler.blacklist-append {clang < 900}
+# requires working std::filesystem, above is not enough :(
+compiler.blacklist-append {clang < 1100} {macports-clang-[4-9].0} {macports-gcc-[3-7]}
 
 name                volk
 categories          science comms
 maintainers         {michaelld @michaelld}
 description         Vector-Optimized Library of Kernels
 long_description    VOLK is the Vector-Optimized Library of Kernels, a library that contains kernels of hand-written single-instruction multiple data (SIMD) code for different mathematical operations, providing portable SIMD code that is optimized for a variety of platforms.
+long_description {*}${long_description} ${subport} \
+    provides the release version, which is typically updated every month or so.
 license             GPL-3
 platforms           darwin
 homepage            https://libvolk.org/
@@ -27,8 +28,6 @@ revision            0
 
 # bump the epoch because I moved the version from 20150707 to 1.0.1
 epoch 1
-long_description {*}${long_description} ${subport} \
-    provides the release version, which is typically updated every month or so.
 
 # Volk requires the submodule "cpu_features" as found here: <
 # https://github.com/google/cpu_features >.  Because we're not


michaelld:
Anybody know the correct `blacklist` entries to enforce viable `std::filesystem` support in C++17? These settings seem reasonable based on internet searching, but I haven't tried / tested these to make sure they work and/or are correct (enough).

ra1nb0w:
don't build. See attachment. Maybe something wrong on my installation?

[main.log](https://github.com/macports/macports-ports/files/8984884/main.log)

kencu:
filesystem support also depends on the libc++ that the system has installed to provide the workings.

Apple controls this here:

https://github.com/llvm/llvm-project/blob/e7bc73739a00e471cb96527415e6819cebfc7e80/libcxx/include/__availability#L231

there is no support before 10.15

kencu:
although you can get it via gcc, and also the libc++ we install on 10.6.8 hasvit :)

ra1nb0w:
yes, you can use gcc but can be issues with linking with other SDR software.
In the past I used https://github.com/gulrak/filesystem but you need to patch all calls with `ghc::` prefix

kencu:
so 10.14 and less are SOL for now

ra1nb0w:
The other solution is to use 2.5.0 for =< 10.14 and newer with others

kencu:
always can do that, sure.

filesystem used to be a separate thing in experimental until it fully rolled out. Some enterprising soul could find a way perhaps to add it to legacysupport

kencu:
that gulrak implementstion of filesystem looks worth exploring for older systems to use. I suspect the namespace thing could be worked through to make it transparently usable

kencu:
here is an important commit for someone who might like to try to bring libc++ filesystem support to older systems.

https://github.com/llvm/llvm-project/commit/998a5c88312066fcc2b2de1358edc76587611354

I am not certain yet if some old clang like clang-7.0 might provide this still…have to check

kencu:
and I’ll mention the macports-libcxx port that can often fix this but sometimes has odr issues

see https://trac.macports.org/ticket/62426

ra1nb0w:
Anyway, we should consider a global (macports wide) solution to overcome std::filesystems for compilers that implements c++17 rather than fix this only. For the moment we can choose the volk version based on macOS version. what do you think @michaelld ?

kencu:
Anyway, you don’t even care to try macports-libcxx? 

that is what I made it for, but up to you. I don’t use volk. That is the only system-wide fix I plan on doing.

And it may not work here, who knows.

kencu (commented):
turns out to be a bit more complicated

ra1nb0w:
> Anyway, you don’t even care to try macports-libcxx?

I read now the message. Can be an idea...

michaelld:
Catching up here ... thanks for the great discussion! My preference is always to do a global solution so that all port that use the feature benefit! That said, I think we've been here before (e.g., [recent Doxygen update](https://trac.macports.org/ticket/65314) sounds like a similar issue even if not `std::filesystem`; please correct me if I'm wrong here).... so I think the better short-term solution is to provide Volk 2.5.0 for "legacy" macOS/OSX & hope for the best moving forward.

michaelld:
@kencu can you describe the complexity with `macports-libcxx`? Here, or maybe it's on some MP issue that you can link in?

kencu:
It's easy to try out macports-libcxx (add the legacysupport PG and enable the option for it) but -- it introduces two different libc++.dylibs into the mixture, one for many things that are being used (system, most of macports stuff) and another for the things that use the new libc++.dylib (eg volk, should it be enabled). 

And so the objects between the two versions of libc++.dylib can be subtly different. Upstream is not interested in making them compatible (too hard to maintain, not going to try). This therefore can sometimes cause crashes or weird errors. Only way to know is to build it and then run the test suite. Still no guarantee.

Safer would be link in the new libc++.a statically when it is needed, eg statically link it into volk. This is so likely to be desirable it should probably be an option in legacysupport if it is possible and squeeks past the various build systems.

If we don't want to use the new libc++ because it is too difficult/unpredictable, and instead want to supply filesystem through an alternate method while still using the system libc++ on < 10.15, then either using the implementation mentioned above (gulrak) or cribbing it out of the libcxx llvm tree before it was folded into libc++.dylib would be starting points.

michaelld:
@kencu thanks for that explanation

michaelld:
you've probably done it before for me & I just don't recall ... ;)

michaelld:
I'm going to revert Volk to 2.5.0 & bump the epoch so-as to get back to a sane install. We (all) can then work out what to do, but at least folks can use Volk again. Closing this PR, but will note it in the Portfile change.

PR #15207, opened 2022-06-25
poppler: update to 22.06.0

michaelld:
#### Description

poppler: update to 22.06.0

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.8 20G715
Xcode 12.3 12C33

###### Verification 
Have you


mascguy:
Hey Michael, I wanted to wait a bit before updating the non-Devel port, as includes a breaking ABI change.

And that will necessitate rev-bumping some dependents. Inkscape is one that I know of for sure, based on local testing. But there may be others as well, though I haven't had an opportunity to test.

mascguy:
Haven't dug into the specific change(s) yet, but it looks like it doesn't break everything.

So in short, I'd like to better understand the specifics - and determine what does/doesn't require a rev-bump - before moving forward. (And I'd prefer to not rely on rev-upgrade to fix anything broken, as that's not fun when various ports have to be rebuilt locally from source.)

Does that all make sense? Your thoughts?

michaelld:
@mascguy that's fine with me. I've been using this Poppler since it was released, without issues. That only means it is probably API-enough compatible with current ports; I can't test for all ports, of course! Any specific ports that depend on poppler that you want to be vetted?

mascguy (approved):
Added necessary rev-bump for Inkscape ports, so we're good to merge.

michaelld:
YW for getting this going & thanks y'all for finishing it off!

PR #15264, opened 2022-07-03
perl5: update 5.34.1; add 5.36.0

mojca:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G9016 x86_64
Xcode 10.1 10B61

###### Verification 
Have you

- [x] tried existing tests with `sudo port test`? (many tests are failing though, esp. with `port -vt test`)

@jmroot @nerdling


jmroot (approved):
Looks fine to me. This is more aggressive than we've been with removing the variants for old versions in the past, but I'm not particularly invested in keeping them. 5.32 is technically still eligible for security fixes, having had its .0 release in the last 3 years: https://perldoc.perl.org/perlpolicy#MAINTENANCE-AND-SUPPORT

mojca:
> Looks fine to me. This is more aggressive than we've been with removing the variants for old versions in the past, but I'm not particularly invested in keeping them. 5.32 is technically still eligible for security fixes, having had its .0 release in the last 3 years: https://perldoc.perl.org/perlpolicy#MAINTENANCE-AND-SUPPORT

I would say that the only reason to keep old variants is because we are usually unable to fix the hundreds of ports in a timely manner. Until we port everything to "the latest" perl, we cannot really remove some older variants.

That said, the only really relevant variant to remove would be 5.28. As long as we keep that one, users running `sudo port upgrade outdated` will be stuck with perl5.28 as default.

By removing the variants we are not really removing the ability to run an older perl version, just removing the option to have that one selected as default. Perl is super backward compatible in general. Unless there are bugs (which happen), I don't really see any serious benefit in supporting gazillion versions, other than the grace period that we need to do the transition. If that transitions takes many years (which is often the case), we are stuck with the intermediate variants.

mojca:
Main question: do we need to wait with removal of old variants for perl before doing the total clean-up of all ports that depend on perl modules (like in the linked ticket)? If yes, then I can remove/adapt one commit, so that we get the latest 5.34 and 5.36 versions without having to worry about dependent ports.

(Judging from the feedback it might take a tiny bit longer to do a proper clean-up.)

mojca:
I avoided removal of variants for 5.28 up to 5.32 for now. Let's clean the small number of remaining ports first and only then put the last nail to old Perls coffin.

PR #15239, opened 2022-06-29
Change default perl5 to 5.34

jmroot:
Older perl versions are EOL: https://www.cpan.org/src/README.html

Also removed perl5.26 variants that can't work because the dependencies no longer exist.

PR #15265, opened 2022-07-03
Switch perl dependencies to p5.34 subports

mojca:
#### Description

Completely untested so far.

Special notes:
* I probably forgot to increase the revisions in some ports and likely made tons of other mistakes.
* ~~`lilypond` may need double-checking~~ (DONE)
* `pulseaudio` could probably switch to default version
* the following ports could use some extra love: `mythtv-core.27`, `mythtv-plugins.27`, `mythweb.27`, `mythweb.28`
* a long list of ports uses default perl version and will probabably stick with 5.28 unless the revision gets increased (I didn't specifically check any of those ports): `PsyncX`, `conf2struct`, `ddclient`, `flamegraph`, `frozenbubble2`, `gscan2pdf`, `macos-fortress-dshield`, `macos-fortress-hosts`, `myrepos`, `stow`
* #14790 could be closed
* I would like to see variants for anything older than 5.34 removed from regular ports before we add subports for p5.36.
* Some subports like `subversion-perlbindings-5.28`, `xapian-bindings-perl5.28` should also be removed once we remove all other variants and subports.

See also:
* https://trac.macports.org/ticket/58361
* #15239 (first batch of changes)
* #15264 (perl version 5.36)

@jmroot 

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/audio/pulseaudio/Portfile
+++ b/audio/pulseaudio/Portfile
@@ -5,6 +5,7 @@
 
 PortSystem          1.0
 PortGroup           active_variants 1.1
+PortGroup           perl5 1.0
 
 name                pulseaudio
 version             14.2
@@ -31,15 +32,13 @@ checksums           rmd160  5e52b8549a0ece2b52df854b5edaa2998c438ccd \
                     sha256  75d3f7742c1ae449049a4c88900e454b8b350ecaa8c544f3488a2562a9ff66f1 \
                     size    1951300
 
-set perl_branch     5.28
-
 depends_build       port:pkgconfig \
                     port:intltool \
                     port:autoconf \
                     port:automake \
                     port:libtool \
                     port:gettext \
-                    port:p${perl_branch}-xml-parser
+                    port:p${perl5.major}-xml-parser
 
 depends_lib         port:libiconv \
                     port:libsndfile \
@@ -84,7 +83,7 @@ platform darwin 8 {
 # reconfigure using upstream autogen.sh for intltool 0.51 compatibility
 post-patch {
     xinstall -m 755 ${filespath}/autogen.sh ${worksrcpath}
-    reinplace "s|@@MP_PERL@@|${prefix}/bin/perl${perl_branch}|" ${worksrcpath}/man/Makefile.am
+    reinplace "s|@@MP_PERL@@|${perl5.bin}|" ${worksrcpath}/man/Makefile.am
 }
 
 # Requires C11.
--- a/audio/pulseaudio/Portfile
+++ b/audio/pulseaudio/Portfile
@@ -131,9 +131,9 @@ if {"${configure.sdkroot}" eq ""} {
 # needs cctools with newer clangs on older systems. Error is:
 # libtool:   error: not configured to extract global symbols from dlpreopened files
 if {${os.platform} eq "darwin" && ${os.major} < 10 && ${build_arch} in "i386 x86_64" && [string match *clang* ${configure.compiler}] } {
-	depends_build-append  port:cctools
-	configure.env-append  NM=${prefix}/bin/nm
-	configure.args-append lt_cv_path_NM=${prefix}/bin/nm
+    depends_build-append  port:cctools
+    configure.env-append  NM=${prefix}/bin/nm
+    configure.args-append lt_cv_path_NM=${prefix}/bin/nm
 }
 
 test.run            no
--- a/mail/signing-party/Portfile
+++ b/mail/signing-party/Portfile
@@ -5,7 +5,7 @@ PortGroup           perl5 1.0
 
 name                signing-party
 version             2.11
-revision            0
+revision            1
 checksums           rmd160  58289ade13f85ec20d65d75569cdda5e774f6bee \
                     sha256  52264331726025f413db8747795d33ef36fef318b3191a1baedfb5f59a6b534f \
                     size    224706
@@ -33,7 +33,7 @@ use_configure       no
 depends_build       port:automake \
                     port:autoconf
 
-perl5.major         5.30
+perl5.major         5.34
 depends_lib         port:perl${perl5.major} \
                     port:p${perl5.major}-class-methodmaker \
                     port:p${perl5.major}-mailtools \
--- a/net/pidgin/Portfile
+++ b/net/pidgin/Portfile
@@ -1,10 +1,11 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           perl5 1.0
 
 name                pidgin
 version             2.14.8
-revision            0
+revision            1
 categories          net
 platforms           darwin
 license             GPL-2+
@@ -30,10 +31,7 @@ use_bzip2           yes
 # Python is only required if --enable-dbus is used, and only for
 # the build process to generate the dbus code, not for running pidgin.
 
-# set 'pbranch' to desired perl version for perl plugin support
-# currently 5.26 (#55208)
-
-set pbranch         5.30
+perl5.major         5.34
 
 depends_build       port:pkgconfig \
                     port:intltool \
@@ -41,14 +39,14 @@ depends_build       port:pkgconfig \
                     port:automake \
                     port:libtool \
                     port:gsed \
-                    port:python39
+                    port:python310
 
 depends_lib         path:lib/pkgconfig/glib-2.0.pc:glib2 \
                     port:libxml2 \
                     port:gconf \
                     port:libidn \
                     port:gnutls \
-                    port:perl${pbranch}
+                    port:perl${perl5.major}
 
 patchfiles          patch-configure.ac.diff \
                     patch-libpurple-internal.h.diff \
@@ -85,10 +83,10 @@ configure.args      --disable-avahi \
                     --disable-tk \
                     --disable-vv \
                     --mandir=${prefix}/share/man \
-                    --with-python3=${prefix}/bin/python3.9 \
+                    --with-python3=${prefix}/bin/python3.10 \
                     --without-x
 
-configure.args-append ac_cv_path_perlpath=${prefix}/bin/perl${pbranch}
+configure.args-append ac_cv_path_perlpath=${perl5.bin}
 
 configure.ldflags-append    -bind_at_load
 
--- a/net/wget/Portfile
+++ b/net/wget/Portfile
@@ -2,6 +2,7 @@
 
 PortSystem              1.0
 PortGroup               muniversal 1.0
+PortGroup               perl5 1.0
 
 name                    wget
 version                 1.21.3
@@ -61,14 +62,12 @@ configure.args          --enable-option-checking \
 # older system perl versions (10.6 and earlier) do not support pod2man --utf8
 # https://trac.macports.org/ticket/50164
 
-set perl_version        5.30
-
 depends_build           bin:makeinfo:texinfo \
-                        port:perl${perl_version} \
+                        port:perl${perl5.major} \
                         port:pkgconfig
 
-depends_test            port:p${perl_version}-libwww-perl \
-                        port:p${perl_version}-io-socket-ssl
+depends_test            port:p${perl5.major}-libwww-perl \
+                        port:p${perl5.major}-io-socket-ssl
 
 depends_lib \
     port:libiconv \
@@ -84,8 +83,8 @@ patchfiles              prefix.patch
 
 configure.checks.implicit_function_declaration.whitelist-append strchr
 
-configure.perl          ${prefix}/bin/perl${perl_version}
-configure.env-append    POD2MAN=${prefix}/bin/pod2man-${perl_version}
+configure.perl          ${perl5.bin}
+configure.env-append    POD2MAN=${prefix}/bin/pod2man-${perl5.major}
 
 # Ensure libuuid doesn't get used even if the ossp-uuid port is installed.
 configure.args-append       ac_cv_header_uuid_uuid_h=no
--- a/sysutils/amttools/Portfile
+++ b/sysutils/amttools/Portfile
@@ -1,11 +1,12 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           perl5 1.0
 
 name                amttools
 set my_distname     amtterm
 version             1.6
-revision            0
+revision            1
 categories          sysutils
 license             GPL-2+
 maintainers         {ionic @Ionic} openmaintainer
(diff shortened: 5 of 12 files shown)


audio/pulseaudio/Portfile
```diff
@@ -31,7 +31,7 @@ checksums           rmd160  5e52b8549a0ece2b52df854b5edaa2998c438ccd \
                     sha256  75d3f7742c1ae449049a4c88900e454b8b350ecaa8c544f3488a2562a9ff66f1 \
                     size    1951300
 
-set perl_branch     5.28
+set perl_branch     5.34
```
mojca, line 34: This is only used by `depends_build`. Maybe it could switch to the `perl5` portgroup and simply use whatever is defined as the default perl there.

gnome/gnome-voice/Portfile
```diff
@@ -23,6 +25,8 @@ checksums           rmd160  7e45780cb9aa8105639f5a01a1d7fc368d7d43ee \
                     sha256  2c92a4acf3474c76758cead7469ed73b42f9352b9c949dd9febea36b05f39014 \
                     size    105728
 
+perl5.branches      5.34
```
mojca, line 28: Only used by `depends_build`. Maybe this line could go?

multimedia/mythtv.27/Portfile
```diff
@@ -10,10 +10,10 @@ set shorthash       3c7582e9
 set majorversion    .27
 set minorversion    .7
 set lastcommit      20180201
-set corerev         4
-set pluginsrev      3
+set corerev         5
+set pluginsrev      4
 github.setup        MythTV mythtv ${shorthash}
-revision            0
+revision            1
```
mojca, line 16: I'm not sure about all these revisions.

multimedia/mythtv.27/Portfile
```diff
@@ -65,7 +65,7 @@ if { ${subport} eq ${name} } {
 subport mythtv-core${majorversion} {
 	name                mythtv-core${majorversion}
 	version             0${majorversion}${minorversion}-Fixes-${lastcommit}
-    revision            1
+    revision            2
```
mojca, line 68: This one seems ignored anyway. And spacing is messed up in the whole file.

ctreleaven: The myth*.27 ports may be dead by now.  I used to have one user that was stuck on that version due to hardware constraints so I left them in place.  I haven't heard from him for a couple of years now so it is probably safe to just remove them.  I'll try to find some time in the near future to do that.

gnome/evolution-data-server/Portfile
```diff
@@ -52,7 +52,7 @@ depends_lib         path:lib/pkgconfig/gtk+-3.0.pc:gtk3 \
                     port:db53
 
 # our choice of perl for installed scripts
-depends_lib-append  port:perl5.28
+depends_lib-append  port:perl5.34
```
jmroot, line 55: More work is needed, see https://trac.macports.org/ticket/65409

gnome/gnome-voice/Portfile
jmroot: Doesn't appear to be telling the build system which perl to use.

mail/signing-party/Portfile
```diff
@@ -33,7 +33,7 @@ use_configure       no
 depends_build       port:automake \
                     port:autoconf
 
-perl5.major         5.30
+perl5.major         5.34
```
jmroot, line 36: Most ports using the perl5 portgroup and a single version of perl are setting `perl5.branches`, which is why I missed this one. Setting `perl5.major` for a single value makes more sense to me though. We should probably standardise on one or the other?

multimedia/mythtv.27/Portfile
jmroot: Yeah, I don't think the metaport revision needs to be bumped. See what I did when updating mythtv.28.

multimedia/mythtv.27/Portfile
jmroot: I agree this line should just be removed in any case, as it's redundant and potentially confusing.

net/pidgin/Portfile
```diff
@@ -33,15 +33,15 @@ use_bzip2           yes
 # set 'pbranch' to desired perl version for perl plugin support
 # currently 5.26 (#55208)
```
jmroot, line 34: Comment is no longer accurate.

net/wget/Portfile
```diff
@@ -5,7 +5,7 @@ PortGroup               muniversal 1.0
 
 name                    wget
 version                 1.21.3
-revision                0
+revision                1
```
jmroot, line 8: Does this need a rev bump for just a build dep change?

sysutils/littleutils/Portfile
```diff
@@ -27,6 +29,8 @@ checksums           rmd160  f53a0cbd12f40004a0dbeecea08d4d9ae82e4ba6 \
                     sha256  dfbedc748992d69b65f101ee655b1834cba918e22e39b94e6e1a16a1ab12812d \
                     size    212952
 
+perl5.branches      5.34
```
jmroot, line 32: Doesn't seem to be telling the build which perl to use.

tex/texlive-bin-extra/Portfile
```diff
@@ -35,6 +36,8 @@ texlive.formats      \
...
 
+perl5.branches      5.34
```
jmroot, line 39: Doesn't seem to be telling the build which perl to use.

textproc/lilypond/Portfile
```diff
@@ -84,11 +85,11 @@ depends_build-append \
...
-                    port:p5.28-encode \
-                    port:p5.28-pod-escapes \
-                    port:p5.28-pod-simple \
-                    port:p5.28-podlators \
-                    port:p5.28-scalar-list-utils \
+                    port:p${perl5.major}-encode \
+                    port:p${perl5.major}-pod-escapes \
+                    port:p${perl5.major}-pod-simple \
+                    port:p${perl5.major}-podlators \
+                    port:p${perl5.major}-scalar-list-utils \
```
jmroot, line 92: Doesn't seem to be telling the build which perl to use.

textproc/vislcg3/Portfile
```diff
@@ -39,7 +40,10 @@ checksums               rmd160 393f7cb49d40c265a1fcd29c13b999278712ee8d \
 
 boost.depends_type      build
 
-depends_lib-append      port:icu port:p5.28-getopt-long
+perl5.branches          5.34
```
jmroot, line 43: Doesn't seem to be telling the build which perl to use.

jmroot:
> * a long list of ports uses default perl version and will probabably stick with 5.28 unless the revision gets increased (I didn't specifically check any of those ports): `PsyncX`, `conf2struct`, `ddclient`, `flamegraph`, `frozenbubble2`, `gscan2pdf`, `macos-fortress-dshield`, `macos-fortress-hosts`, `myrepos`, `stow`

It's worse than that. When installing one of these, the user's system will see them as depending on perl5.34 (or whatever arbitrary perl5 happens to be installed), and the deps will be installed for that version. But the binaries will be built for 5.28 and will not work after installation due to missing dependencies.

It's only safe to depend on a non-deterministic perl version when it's a build-time-only dep.

sysutils/littleutils/Portfile
Schamschula: It uses`/opt/local/bin/perl`

Schamschula: But I have `perl5 @5.28.3_0+perl5_34 (active)`

jmroot: Needs to use the perl version that cryptx is installed for, I would assume.

Schamschula: Yes. Somehow my comment didn't come over, but this should do it:

`configure.args      PROGPERL=${prefix}/bin/perl${perl5.major}`

gnome/evolution-data-server/Portfile
mascguy: > More work is needed, see https://trac.macports.org/ticket/65409

Fixed via separate commit: https://github.com/macports/macports-ports/commit/55d6436349def9ce35944a3e389f149329390f05

So changes to this port can be removed from the PR.

mail/signing-party/Portfile
mojca: I'm indifferent. I happily leave it to others to decide what looks better and then we can unify it in other ports.

I would also prefer to standardise the syntax, for example in other ports using
```
set perl_branch     5.34
```
and alike. If we do that, we could migrate to a newer version much faster and more easily in the future.

I tried to do that to some ports in the past, but then some maintainers complained that including the whole portgroup was an overkill when all they needed was a single simple variable.

jmroot: Just using the same variable name would helpful, even without using the portgroup.

tex/texlive-bin-extra/Portfile
drkp: I am working on this one.

drkp:
Fixed for texlive-bin-extra in [efd11f4](https://github.com/macports/macports-ports/commit/efd11f440038be5b3223f4f4c02bf2c486745818)

textproc/lilypond/Portfile
nerdling: Noted. Will dig in.

jmroot:
All the ports that were using a non-deterministic perl should be fixed now.

nerdling:
Fixed pspp-devel in abffd244278cb4e3935dab4a62cd3fea7956ba77

Fixed lilypond in 17c1aa6156851c50044e99ef88e068e6e768a171

mojca:
@ctreleaven: can you please take care of your ports?
@jmroot or anyone else: is anyone willing to go through the remaining ports that aren't using the specified version of perl?

sysutils/amttools/Portfile
```diff
@@ -38,7 +39,7 @@ checksums           rmd160  9fb55ade1f5ca3015be0794236ea3031fc137367 \
                     sha256  1242cea467827aa1e2e91b41846229ca0a5b3f3e09260b0df9d78dc875075590 \
                     size    40804
 
-set perl_branch     5.28
```
jmroot, line 41: There's another place where `perl_branch` was referenced.

mojca:
We are now down to:
* `littleutils`, `vislcg3`: perl version set by the port not being used
* `mythtv.27`, `mythweb.27`, `mythweb.28`

(I would be extremely happy if someone else would be willing to look into these.)

TinoDidriksen:
> We are now down to:
> 
>     * `vislcg3`: perl version set by the port not being used

Not sure what you mean, but the proposed changes look fine to me. I'd be even happier if the version was not mentioned at all - just require `perl5` and `p5-getopt-long` and let that be whatever version they refer to. vislcg3's helper scripts just need `perl5 >= 5.14` and the corresponding `getopt-long`.

jmroot:
> Not sure what you mean, but the proposed changes look fine to me. I'd be even happier if the version was not mentioned at all - just require `perl5` and `p5-getopt-long` and let that be whatever version they refer to. vislcg3's helper scripts just need `perl5 >= 5.14` and the corresponding `getopt-long`.

The problem is that the installed perl scripts run with `/usr/bin/env perl`, and there's no guarantee that whatever `perl` is found is the one for which `Getopt::Long` is installed. If you need a particular module, you have to depend on the port for that module in a particular perl version, and then make sure to use that same perl version at runtime.

I see however that `Getopt::Long` is shipped with perl, so if the version of that module that comes with perl 5.28 or later is sufficient, the dependency could be change to just the `perl5` port.

gnome/gnome-voice/Portfile
```diff
@@ -23,12 +25,14 @@ checksums           rmd160  f15cae481f168b181858f9b86bb3fc2f2fb923ee \
...
 
+perl5.major         5.34
+
 depends_build       port:autoconf \
                     port:automake \
                     port:gtk-doc \
                     port:intltool \
                     port:itstool \
-                    port:p5.28-xml-sax-expat \
+                    port:p${perl5.major}-xml-sax-expat \
```
jmroot, line 35: This one also doesn't seem to be doing anything to ensure the specified perl is used.

jmroot: Oh, I see we discussed this before. Even though it's only depends_build, it still needs to use the perl version that has `XML::SAX::Expat` installed.

mojca: I tried adding (based on investigating the sources, I didn't try to actually configure or build anything, too many dependencies are needed, maybe I'll try later):
```tcl
perl5.major 5.28
configure.env-append INTLTOOL_PERL=${perl5.bin}
```
The fun part is that `rdeps` now show a mixture of perl 5.28 and 5.34 due to `intltool` depending on 5.34. I don't know what that actually means (whether the version needs to be compatible or not).

The port `intltool` has the comment:
```tcl
# intltool only uses perl internally and publishes no public modules
# so there is no need for multiple perl variants
set pbranch         5.34
```

@dbevans

jmroot: Oh OK, if the dependency is only there as a workaround to the INTLTOOL_PERL bug, then that should just be fixed properly and the dependency removed.

mojca: Is this a widely known bug? How can one check the behaviour?

I tried removing the dependency, made sure that it wasn't installed, and destrooted with the `-vt` flag successfully. How do I diagnose whether it's still a bug or if the dependency can simply be removed?

```
checking for intltool >= 0.50.1... 0.51.0 found
checking for intltool-update... /opt/local/bin/intltool-update
checking for intltool-merge... /opt/local/bin/intltool-merge
checking for intltool-extract... /opt/local/bin/intltool-extract
checking for xgettext... (cached) /opt/local/bin/xgettext
checking for msgmerge... /opt/local/bin/msgmerge
checking for msgfmt... (cached) /opt/local/bin/msgfmt
checking for gmsgfmt... (cached) /opt/local/bin/msgfmt
checking for perl... /opt/local/bin/perl
checking for perl >= 5.8.1... 5.34.1
checking for XML::Parser... ok
checking that generated files are newer than configure... done
```

mojca: @oleaamot

mojca:
I now removed `mythtv`, `mythweb` from this PR, they are too messy for me to be willing to handle them in this batch. I'm counting on maintainer.

The `vislcg3` port uses tons of hardcoded `#!/usr/bin/perl` and `#!/usr/bin/env perl` that need some patching. I would prefer to leave those to the maintainer (@TinoDidriksen) as outside of scope of this ticket. I would assume that merging this as-is without fixing handling of the correct perl in that single port is still better than not proceeding towards perl 5.36 (that is: I don't consider `vislcg3` to be a blocker, we won't make the state worse, but it would still be nice to fix the port soon after that).

TinoDidriksen:
> The `vislcg3` port uses tons of hardcoded `#!/usr/bin/perl` and `#!/usr/bin/env perl` that need some patching. I would prefer to leave those to the maintainer (@TinoDidriksen) as outside of scope of this ticket. I would assume that merging this as-is without fixing handling of the correct perl in that single port is still better than not proceeding towards perl 5.36 (that is: I don't consider `vislcg3` to be a blocker, we won't make the state worse, but it would still be nice to fix the port soon after that).

See https://github.com/macports/macports-ports/pull/15426 for how I'd solve this for vislcg3. I removed all `#!/usr/bin/perl` in favour of `#!/usr/bin/env perl` upstream - the helper scripts are neither critical nor complex, so they will work with any Perl 5 version that happens to be picked up.

PR #15285, opened 2022-07-05
Make several ports explicitly depend on perl

dgilman:
#### Description

This should probably be merged into the branch of https://github.com/macports/macports-ports/pull/15265

The PR here changes several perl-dependent apps to explicitly depend on perl 5.34 and use the perl variant machinery. For testing:
- conf2struct was tested with upstream's example
- flamegraph was tested with upstream's regression test suite
- myrepos was tested with upstream's tests
- stow, swaks, and ddclient were just launched as I couldn't find a way to get their test suites to run. But they launched just fine, and were correctly building against and running with perl5.34.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H1922 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you


mojca (changes requested):
Thanks a lot for the extra effort.

There is no need to merge these changes anywhere. If you fix some of the ports here, we'll simply merge this to the master branch and remove the changes from the other PR.

But these changes are partially going a bit into the opposite direction of what we want to achieve. We do need to unify the way of handling perl, but there's no need to start supporting perl 5.30 at this point in time: I really want to get rid of anything older than perl 5.34 as soon as possible.

So the whole chunk:
```tcl
perl5.require_variant   yes
perl5.conflict_variants yes
perl5.branches          5.28 5.30 5.32 5.34
perl5.default_branch    5.34
perl5.create_variants   ${perl5.branches}
```
should ideally become just
```tcl
perl5.major 5.34
```
and verified whether it still works that way.

If you would be willing to test, I would be really grateful if you could change this to 5.32, for example, and check whether it actually uses perl 5.32. Checking whether it uses perl5.34 is a bit of a cheating as ports may use it just because it's the default already and it's a bit more difficult to check for correctness.

@jmroot

dgilman:
I also took a look at the shebangs for all of the installed programs and found that the ports were correctly rewriting it to perl5.34.

nerdling:
I do echo the suggestions from @mojca but I suspect it won't be the end of the world if we have to undo some of the PR after-the-fact. Would be nice to not have extra work though :)

dgilman:
I tested these out with the `perl5.major 5.34` and they were looking fine. Unfortunately I don't have time to try with perl 5.32.

devel/myrepos/Portfile
```diff
@@ -29,7 +29,7 @@ supported_archs     noarch
 use_configure       no
 build.target        build
 
-depends_lib         port:perl${perl5.major}
+perl5.major             5.34
```
mojca, line 32: Strange spacing

PR #15295, opened 2022-07-06
GCC12 Suite

cjones051073:
#### Description

Add GCC 12 and makes primary runtime provider.

###### Tested on

Numerous platforms between 10.6 and 12

Not completely all, and certainly not all Xcode/OS combinations. But enough to be hopefully of not too many issues.

ci tests will certainly timeout, so not to be relied upon.


Gcenx:
Going off my bump of `*-w64-mingw32-gcc` to gcc12 it failed to build on 10.6 i386

cjones051073:
can you post a link to the builder log for the failed build ?

Gcenx:
According to the buildbot it never even started the build but complained about the deps

https://build.macports.org/builders/ports-10.6_i386-builder/builds/75938

mojca:
Since the logs are no longer available, I restarted the build
* https://build.macports.org/builders/ports-10.6_i386-builder/builds/81238

Gcenx:
> Since the logs are no longer available, I restarted the build
> 
> * https://build.macports.org/builders/ports-10.6_i386-builder/builds/81238

We’ll want logs for `i686-w64-mingw32-gcc-bootstrap` as that stage started to build and failed.

mojca:
> We’ll want logs for `i686-w64-mingw32-gcc-bootstrap` as that stage started to build and failed.

Sure, that's precisely what I posted. The [previous failure](https://build.macports.org/builders/ports-10.6_i386-builder/builds/75936) happened after 22 minutes of compiling.

mojca:
It breaks at least for
```
insn-recog.cc:45196:ld: illegal text-relocation to ___stderrp in /usr/lib/libSystem.dylib from __ZN11source_info5debugEv in gcov.o for architecture i386
```

lang/gcc11/Portfile
```diff
@@ -7,13 +7,13 @@ PortGroup           active_variants              1.1
 PortGroup           conflicts_build              1.0
 PortGroup           cltversion                   1.0
 
-epoch               0
+epoch               1
```
mojca, line 10: Why do you need to increase the epoch?

cjones051073: it is how the automatic migration of libgcc works. see the activate hacks in the gcc12 Portfile.

cjones051073: https://trac.macports.org/wiki/PortfileRecipes#deactivatehack

Gcenx:
> 

Maybe we could use the `-read_only_relocs suppress` Linker flag as that seems to be a common answer for similar issues on legacy versions of macOS

cjones051073:
Hmm, both 10.6 builds failed for libgcc-devel. libgcc12 though built fine for me on 64bit 10.6…. I am busy the next few days but will take a look next week.

cjones051073:
the libgcc-devel ports builds fine on both 10.6 64 and 32 bit

https://build.macports.org/builders/ports-10.6_x86_64-builder/builds/114710
https://build.macports.org/builders/ports-10.6_i386-builder/builds/81347

as this port is almost identical to gcc12, and as I have alway already tested the 64 bit build, I am hopefully it will be fine.

I don't think there is much more I can do to test, so going to merge and see what happens...

PR #15360, opened 2022-07-13
openssh: add x509 variant

sstallion:
#### Description

This PR introduces a new `x509` variant to `openssh` to use upstream patches for X.509 v3 certificate support provided by [Roumen Petrov](https://roumenpetrov.info/secsh/). The upstream patch was massaged slightly to be compatible with all existing patches in the Portfile. This variant has been tested extensively against hosts that support X.509 and those that rely on traditional SSH keys.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/net/openssh/Portfile
+++ b/net/openssh/Portfile
@@ -42,7 +42,6 @@ if {${name} eq ${subport}} {
                         port:libedit \
                         port:ncurses \
                         port:zlib
-    depends_run         port:ssh-copy-id
 
     platform darwin 10 {
         # /usr/bin/ranlib: object: libopenbsd-compat.a(base64.o) malformed object (unknown load command 2)
@@ -148,6 +147,10 @@ if {${name} eq ${subport}} {
 
         file rename "${destroot}${prefix}/etc/ssh/sshd_config" "${destroot}${prefix}/etc/ssh/sshd_config.example"
         file rename "${destroot}${prefix}/etc/ssh/ssh_config" "${destroot}${prefix}/etc/ssh/ssh_config.example"
+
+        # install ssh-copy-id as it could have been patched
+        xinstall -m 755 ${worksrcpath}/contrib/ssh-copy-id ${destroot}${prefix}/bin
+        xinstall -m 644 ${worksrcpath}/contrib/ssh-copy-id.1 ${destroot}${prefix}/share/man/man1
     }
 
     post-activate {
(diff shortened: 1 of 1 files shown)


chrstphrchvz (changes requested):
I’m just another MacPorts user peeking around…

I don’t know whether the proposed approach for installing patched ssh-copy-id is appropriate. At a minimum the openssh and ssh-copy-id ports would need to be marked as conflicting with each other, and the revision of openssh be increased because its runtime dependencies have changed. I’m not aware what a better approach might be, other than possibly having ssh-copy-id remain a dependency of openssh, adding the x509 variant to ssh-copy-id as well, and adding logic to require that openssh and ssh-copy-id be installed with the x509 variant set for both or for neither.

I’m not aware whether a 4MB patchfile is unprecedented, but it does not sound ideal.

Pinging @l2dy (ssh-copy-id maintainer)

l2dy:
I don't think we should abuse variants this way. PKIX-SSH is an independent project with its own repository https://gitlab.com/secsh/pkixssh.

l2dy:
Could you turn this into a separate port that conflicts with openssh and ssh-copy-id?

sstallion:
> I don't think we should abuse variants this way. PKIX-SSH is an independent project with its own repository https://gitlab.com/secsh/pkixssh.

This is a tough one. Roumen distributes PKIX-SSH in two forms; one is a standalone tarball as you've pointed out, and the other is a patch against OpenSSH. Most distributions tend to favor the patch as I've done here. One of the main motivators behind this if a separate port were created, we'd need to duplicate the other patches as they are still desirable. I'm happy to create a separate port, but it could potentially create some headaches if it gets out of sync with openssh et al.

sstallion:
> I don’t know whether the proposed approach for installing patched ssh-copy-id is appropriate.

I wasn't very happy with that change either. One of the challenges is that `ssh-copy-id` is treated as a dependency to `openssh`, which means it is installed prior to openssh being patched (they're derived from the same sources). I agree that making them conflict might be the cleanest solution here - I understand the utility of wanting `ssh-copy-id` without installing `openssh`, but as it is, the dependency looks wrong to me.

l2dy:
> [...], but it could potentially create some headaches if it gets out of sync with openssh et al.

If someday PKIX-SSH development stagnates, should we stop updating `openssh` at all because of a variant, or maintain this 4 MB patch ourselves? Both aren't ideal, and it already happened to other variants of this port, which is in part why nobody wanted to maintain it. Testing and adapting all these patches for `openssh` upgrades is a tough responsibility.

You could use the subport mechanism to reuse parts of the Portfile, just like `ssh-copy-id`. This way PKIX-SSH won't get in the way of `openssh` upgrades.

sstallion:
That's a really good point. I'll close this PR and submit a new one once I get a chance to move this out into a new package. Thanks!

PR #15364, opened 2022-07-14
rxvt-unicode: update to 9.30 and use perl 5.34

exg:


x11/rxvt-unicode/Portfile
```diff
@@ -24,38 +24,39 @@ master_sites    http://dist.schmorp.de/rxvt-unicode/ \
...
 compiler.blacklist {clang < 318}
 
-configure.perl  ${prefix}/bin/perl5.30
+configure.perl  ${prefix}/bin/perl5.34
 
 patchfiles      patch-Makefile.in.diff \
-                patch-no-clock_gettime.diff \
                 patch-perl-embed-args.diff
 
+configure.cxxflags-append -std=c++11
```
chrstphrchvz, line 51: If a compiler supporting C++11 is required then `compiler.cxx_standard 2011` should also be set:

```suggestion
compiler.cxx_standard 2011
configure.cxxflags-append -std=c++11
```

chrstphrchvz (changes requested):

x11/rxvt-unicode/Portfile
exg: Added, thanks.

devel/libptytty/Portfile
```diff
@@ -0,0 +1,20 @@
...
+PortGroup           cmake 1.1
+
+name                libptytty
+version             2.0
+categories          devel
+license             GPL-2+
+maintainers         {gmail.com:emanuele.giaquinta @exg} openmaintainer
+description         library for OS-independent pseudo-tty management
+long_description    ${description}
+platforms           darwin
```
mascguy, line 13: While this is minor, `platforms darwin` is now the default. So you can remove that declaration.

exg: Removed, thanks.

x11/rxvt-unicode/Portfile
```diff
@@ -18,44 +18,41 @@ long_description \
...
 
 # urxvt also depends on xorg-libX11, xrender and fontconfig but all these ports are in the dependency tree of Xft2
 depends_lib \
+        port:libptytty \
         port:Xft2 \
         port:startup-notification \
         port:xorg-libXmu \
         port:ncurses \
-        port:perl5.30
+        port:perl5.34
```
mojca, line 39: Please see #15265 for the "correct" changes.

devel/libptytty/Portfile
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+
+name                libptytty
```
mojca, line 6: It's confusing to have your PR called "rxvt-unicode: ..." which adds another completely unrelated port to the PR.

It would make sense to split the other one into its own separate PR.

x11/rxvt-unicode/Portfile
exg: Fixed, thanks.

PR #15389, opened 2022-07-17
exiv2: enable BMFF support

mascguy:
### Description

Enable BMFF support, to improve functionality in dependent ports like `darktable`.

Thus far, upstream releases haven't enabled this by default, due to potential concerns related to patents. However, the topic has been thoroughly discussed and vetted, enough so that upstream has now enabled by default on Master.

Patent-related discussions: https://github.com/Exiv2/exiv2/issues/1679

### Tested on macOS Releases

* Intel x86_64: 10.6 through Monterey
* PPC: 10.5

### DIFF
--- a/graphics/exiv2/Portfile
+++ b/graphics/exiv2/Portfile
@@ -7,7 +7,7 @@ PortGroup           legacysupport 1.0
 PortGroup           compiler_blacklist_versions 1.0
 
 github.setup        Exiv2 exiv2 0.27.5 v
-revision            0
+revision            1
 checksums           rmd160  15486a2b4d13db80709318d3d0f70140de2de434 \
                     sha256  1da1721f84809e4d37b3f106adb18b70b1b0441c860746ce6812bb3df184ed6c \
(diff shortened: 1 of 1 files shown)


mascguy:
@jmroot Added you as a reviewer, so that you can also review the patent-related discussion.

mascguy:
@evanmiller @barracuda156 Tested on 10.5 PPC as well, and looks good.

barracuda156:
> @evanmiller @barracuda156 Tested on 10.5 PPC as well, and looks good.

I can check on 10.6.8 Rosetta, let me build `cmake`.

mascguy:
@ryandesign and @jmroot, thoughts on this?

mascguy:
The key takeaway - from the final comment in the patent discussion - is as follows:

> The first version of ISOBMFF was 2004 and the container format was first used by MPEG in the 1990s. So, even if a patent has been infringed, the holder's "ownership" has almost certainly expired.

Given that, I think we're safe.

PR #15400, opened 2022-07-18
tio: submission

rhapsodyv:
#### Description

New port submission for https://github.com/tio/tio

###### Type(s)

- [x] submission

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/comms/tio/Portfile
+++ b/comms/tio/Portfile
@@ -1,0 +1,27 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+PortGroup               meson 1.0
+
+github.setup            tio tio 1.45 v
+revision                0
+categories              comms
+maintainers             nomaintainer
+license                 GPL-2
+
+homepage                https://tio.github.io
+
+description             Simple TTY terminal I/O application
+long_description        tio is a simple serial device tool which \
+                        features a straightforward command-line \
+                        and configuration file interface to easily \
+                        connect to serial TTY devices for basic \
+                        I/O operations.
+
+checksums               rmd160  f60449da4edf2490720c4c32030cfd6fbd4f9511 \
+                        sha256  1f82c5c44f803da5f1e0c74e858b2a8a329a21b508e3eab4d87a6ffa2f8b5fd6 \
+                        size    3532357
+
+depends_build-append    port:pkgconfig
+


comms/tio/Portfile
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           meson 1.0
+
+name                tio
+version             1.45
+revision            1
```
herbygillot, line 8: Since this is a brand new port, `revision` should be starting at 0.

comms/tio/Portfile
```diff
@@ -0,0 +1,32 @@
...
+use_xz              yes
+
+description         Simple TTY terminal I/O application
+long_description    tio is a simple serial device tool which \
+                    features a straightforward command-line \
+                    and configuration file interface to easily \
+                    connect to serial TTY devices for basic \
+                    I/O operations.
+
+checksums           sha256  4e72a946792346f86504865ed2923688f4dd9281905bb81e76ba3fd6f3b96014 \
```
herbygillot, line 25: You also need to add the `rmd160` checksum.  You can use `port bump tio` to automatically fill in the correct checksum, or you can calculate it manually using `openssl`: `openssl dgst -rmd160 <file>`

rhapsodyv:
@herbygillot thanks for your review!

herbygillot:
Hey @rhapsodyv, this is looking ready to merge - there was one more thing I forgot to mention, and then we're good.  Thank you.

ryandesign:
> Doing this allows you to remove the name, version, master_sites, distfiles & use_xz fields as github.setup will automatically populate those implicitly.

@herbygillot The github portgroup will never set `use_xz` automatically. Its defaults are to use an automatically-generated tar.gz distfile from GitHub's "tarball" endpoint, which is deprecated. Ports that must use an automatically-generated tarball should request the "archive" endpoint instead of the "tarball" endpoint by setting `github.tarball_from archive`. This will change the checksums and would need to be handled as a [stealth update](https://trac.macports.org/wiki/PortfileRecipes#stealth-updates) if the filename is not changed at the same time, for example by updating the version.

However the developers of tio have offered a [manually-created tar.xz release download](https://github.com/tio/tio/releases/tag/v1.46) (look in the Assets section), and release downloads should always be preferred to automatically-generated tarballs when available. To do this, set `github.tarball_from releases` and `use_xz yes`. This will change the checksums but since the filename also changes it does not need to be treated as a stealth update.

rhapsodyv:
@ryandesign thanks for your feedback! I will fix the port and add the lib dependency.

PR #15431, opened 2022-07-20
qt6: Tcl syntax adjustment, drop mysql57 variant 

chrstphrchvz:
#### Description

The mysql57 variant of qt6-mysql-plugin has never built, because the desired shared library is actually named "libmysqlclient.dylib", not "libmysqlclient_r.dylib". I propose deleting the variant rather than fixing it, because MySQL 5.7 is no longer supported by upstream on macOS (even though it remains supported on other OSes until 2023-10, and MacPorts continues to update it).

###### Verification 
Have you

⚠️ **Suggest merging after** #15427 


mascguy (approved):
LGTM, we can merge after your broader qt6-related PR.

PR #15459, opened 2022-07-22
glib2-upstream: disable tests on Rosetta

barracuda156:
#### Description

Some tests fail to build on 10.6.8 Rosetta, so disable those: https://trac.macports.org/ticket/65517
In fact I had that fix long back, but somehow it got lost on the way to the port: https://github.com/macports/macports-ports/commit/7ffa0be646e8d1b28ceb038c2bc9961fd9a48fbf

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode 3.2.6

###### Verification 
Have you

P. S. `lint --nitpick` gives two warnings, but those are unrelated to my PR:
```
10:~ svacchanda$ port lint --nitpick glib2-upstream
--->  Verifying Portfile for glib2-upstream
Warning: Patchfile patch-gio_gcredentialsprivate.h does not follow the source patch naming policy "*.diff" or "*.patch"
Warning: Patchfile patch-gio_gsocket.h does not follow the source patch naming policy "*.diff" or "*.patch"
--->  0 errors and 2 warnings found.
```

### DIFF
--- a/devel/glib2-upstream/Portfile
+++ b/devel/glib2-upstream/Portfile
@@ -161,6 +161,11 @@ post-destroot {
 
 platform darwin {
     configure.args-append   -Ddtrace=false
+    
+    # some tests fail to build on 10.6.8 Rosetta: https://trac.macports.org/ticket/65517
+    if {${os.major} == 10 && ${build_arch} eq "ppc"} {
+        configure.args-append   -Dtests=false
+    }
 }
 
 platform darwin 8 {


devel/glib2-upstream/Portfile
```diff
@@ -161,6 +161,11 @@ post-destroot {
 
 platform darwin {
     configure.args-append   -Ddtrace=false
+    
+    # some tests fail to build on 10.6.8 Rosetta: https://trac.macports.org/ticket/65517
+    if {${os.major} == 10 && ${build_arch} eq "ppc"} {
```
mascguy, line 166: Would this also apply to 10.4 and 10.5, such that we'd update the check to `${os.major} <= 10`? Or is this truly specific to 10.6?

CC: @evanmiller

barracuda156: > Would this also apply to 10.4 and 10.5, such that we'd update the check to `${os.major} <= 10`? Or is this truly specific to 10.6?

@mascguy I cannot test at the moment on anything but Rosetta, but I will check my portfiles if I used to disable tests or not (that won’t guarantee anything though).

barracuda156: UPD. @mascguy So judging by my portfiles for `glib` 2.72.1 from Leopard and 10.6 PPC, I did not disable tests there, which I did for Rosetta. This likely implies that tests build fine on PPC natively, and only Rosetta has an issue. (Unless of course something changed in tests since then.)

P. S. By the way, while I will have trouble here pushing any 10.6 PPC-specific fixes, if a fix is actually specific to Rosetta (and to 10.6.8 by implication), and not needed for 10.6 PPC, can it be made more precise? So that it is applied only for 10.6.8 and `ppc` arch. Not a big deal, but just in case.

mascguy: > UPD. @mascguy So judging by my portfiles for `glib` 2.72.1 from Leopard and 10.6 PPC, I did not disable tests there, which I did for Rosetta. This likely implies that tests build fine on PPC natively, and only Rosetta has an issue. (Unless of course something changed in tests since then.)
> 
> P. S. By the way, while I will have trouble here pushing any 10.6 PPC-specific fixes, if a fix is actually specific to Rosetta (and to 10.6.8 by implication), and not needed for 10.6 PPC, can it be made more precise? So that it is applied only for 10.6.8 and `ppc` arch. Not a big deal, but just in case.

I'm fine with your changes as-is, since they only apply to 10.6 PPC.

mascguy (approved):

PR #15465, opened 2022-07-23
MacVim/vim: add ruby30 variant

sstallion:
#### Description

This PR adds a `ruby30` variant to the Portfile. This is especially useful for users that need Ruby scripting support (eg. Command-T) with version parity to the latest MacVim release.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/editors/MacVim/Portfile
+++ b/editors/MacVim/Portfile
@@ -169,45 +169,50 @@ variant python39 conflicts python36 python37 python38 description {Enable Python
 }
 
 variant ruby requires ruby18 description {Compatibility variant, requires +ruby18} {}
-variant ruby18 conflicts ruby19 ruby20 ruby21 ruby22 ruby23 ruby24 ruby25 description {Enable Ruby scripting} {
+variant ruby18 conflicts ruby19 ruby20 ruby21 ruby22 ruby23 ruby24 ruby25 ruby30 description {Enable Ruby scripting} {
     configure.args-append   --enable-rubyinterp
     configure.args-append   --with-ruby-command=${prefix}/bin/ruby1.8
     depends_lib-append      port:ruby
 }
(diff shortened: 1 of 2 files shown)


mascguy (approved):
LGTM. @raimue ?

raimue:
Thank you for the patch! :+1: 

However, as you already mention this is also of interest for users of MacVim for which we also have a port. I try to keep vim and MacVim as close as possible as they share the same configuration files. Can we please get the same change for a +ruby30 variant into `editors/MacVim/Portfile`?

mascguy:
> Thank you for the patch! 👍
> 
> However, as you already mention this is also of interest for users of MacVim for which we also have a port. I try to keep vim and MacVim as close as possible as they share the same configuration files. Can we please get the same change for a +ruby30 variant into `editors/MacVim/Portfile`?

Rainer, would you like to merge this as-is, and have a 2nd PR created? Or would you prefer that MacVim be updated too, as part of this PR?

raimue:
I think it would be better to do this within this PR. Then users can switch to `+ruby30` for both and rely upon it in their scripts.

sstallion:
> I think it would be better to do this within this PR. Then users can switch to `+ruby30` for both and rely upon it in their scripts.

Will do! At the moment MacVim seems to be broken on the latest version of Monterey (I haven't been able to get a successful build on amd64 since upgrading from Big Sur and I haven't attempted on M1 yet).

raimue:
The build failure on arm64 seems unrelated to this change. It might be due to the `XCODEFLAGS` being set in the Portfile, but I am not sure. In any case, as +ruby30 is not in the default variants, the CI does not help that much to test it.

raimue:
Thanks again for the additional work on MacVim! :+1:

PR #15484, opened 2022-07-25
wine|wine-devel: Update to 7.x

Gcenx:
#### Description

- Converted `wine` & `wine-devel` over to muniversal 1.1
- Allow 64Bit only installs
- Force 32Bit loader on supported systems
- `wine` Updated to 7.0
- `wine-devel` Updated to 7.13
- Dropped unsupported OS X versions (past wine-2.0 10.6 wasn't really supported see [old wine-packaging](https://github.com/wine-compholio/wine-packaging/blob/master/cross-osx/init.sh))

Modern wine requires a newer SDK to build (even with legacysupport) the Ports use MacOSX10.11.sdk at contains tbd stubs so rev-upgrade _shouldn't_ cry.

With this `wine` & `wine-devel` are much closer to what I'm doing for the official WineHQ macOS packages, those are not yet being pushed directly to winehq due to [bug 52354](https://bugs.winehq.org/show_bug.cgi?id=52354) (reverts are provided and marked)

Not much point dealing with `wine-crossover` until CX22 sources land but this version still requires the custom llvm/clang-8.0 from cw that supports the custom `-mwine32` target.

###### Type(s)

- [x] enhancement

###### Tested on

macOS 10.9 13F1911 x86_64
Command Line Tools 6.2.0.0.1.1424975374

macOS 10.14.6 18G9323 x86_64
Xcode 11.3.1 11C505
_(Forced 10.13 deployment target, it would be preferable to use [DevSDK](https://github.com/macports/macports-base/pull/223))_

macOS 11.6.7 20G630 x86_64
Xcode 13.2.1 13C100

###### Verification 


Gcenx:
Somethings to note;
- sdl2 use used for controller support
- wineusb would require libusb
- skipped bug 49199 workaround (see below)
```
# Standard dlopen() is used so wine(64) no longer finds dylibs, since wine-5.6
# https://bugs.winehq.org/show_bug.cgi?id=49199
configure.ldflags-append     -Wl,-rpath,${prefix}/lib
```

I’ve skipped the including of libusb due to the following report from a user
```
wine64-preloader claims usb interface for no reason.
```
This could be but behind a variant

Gcenx:
It’s preferable to first deal with https://github.com/macports/macports-ports/pull/14858 so El Capitan so would also have Vulkan support thus matching upstream.

Gcenx:
Seems there no interest in this so closing

PR #15500, opened 2022-07-26
kotlin: set fallback to openjdk17

breun:
#### Description

Change fallback Java to `openjdk17`, because that is the latest Long Term Support release and `openjdk8` is not available on arm64.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.5 21G72 arm64
Xcode 13.4 13F17a

###### Verification 
Have you

### DIFF
--- a/lang/kotlin/Portfile
+++ b/lang/kotlin/Portfile
@@ -29,7 +29,7 @@ checksums           rmd160  1a1899fdb1f4af50504ec9a374987c8f9e4cc6b8 \
                     size    72957814
 
 java.version        1.8+
-java.fallback       openjdk8
+java.fallback       openjdk17
 
 worksrcdir          kotlinc
 


mascguy:
It would be preferable to only do this for ARM, if Kotlin still supports Java 8.

breun:
Why would it be preferable to stick with Java 8 as the fallback on x86_64? Using the latest compatible LTS version sounds like a good fallback for any port that requires a JVM.

mascguy:
> Why would it be preferable to stick with Java 8 as the fallback on x86_64? Using the latest compatible LTS version sounds like a good fallback for any port that requires a JVM.

Sorry, I'm getting my wires crossed, in terms of required vs. fallback.

You're right, and I retract my earlier comment. Lol

mascguy (approved):

PR #15544, opened 2022-07-31
gobject-introspection: fix build on Rosetta

barracuda156:
#### Description

This PR only affects builds for `ppc` on 10.6.x. It fixes a failure on Rosetta: https://github.com/mesonbuild/meson/issues/10351

(In fact, _only_ Rosetta needs this change, while 10.6 PPC (10A190) does not.)

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/gnome/gobject-introspection/Portfile
+++ b/gnome/gobject-introspection/Portfile
@@ -77,7 +77,16 @@ configure.args      -Ddoctool=enabled \
 # when building this project. Identical logic already exists
 # in the gobject_introspection PortGroup, so dependent projects
 # should not require modification.
-build.env-append    CC=${configure.cc}
+#
+# 10.6.8 Rosetta requires a specific solution: https://github.com/mesonbuild/meson/issues/10351
+
+platform darwin {
+    if {${os.major} == 10 && ${build_arch} eq "ppc"} {
+        build.env-append    CC=${developer_dir}/usr/bin/powerpc-apple-darwin10-gcc-4.2.1
+    } else {
+        build.env-append    CC=${configure.cc}
+    }
+}
 
 platform darwin 8 {
     # Tiger does not support RPATHs at this time


barracuda156:
@mascguy I just checked, it is possible to restrict the condition to 10.6.8 (instead of 10.6.x):
```
platform macosx {
    if {[vercmp ${macos_version} 10.6.8] == 0} {
        build.env-append    CC=${developer_dir}/usr/bin/powerpc-apple-darwin10-gcc-4.2.1
    } else {
        build.env-append    CC=${configure.cc}
    }
}
```
Will this be preferable?

mascguy:
> @mascguy I just checked, it is possible to restrict the condition to 10.6.8 (instead of 10.6.x):
> ```
> platform macosx {
>     if {[vercmp ${macos_version} 10.6.8] == 0} {
>         build.env-append    CC=${developer_dir}/usr/bin/powerpc-apple-darwin10-gcc-4.2.1
>     } else {
>         build.env-append    CC=${configure.cc}
>     }
> }
> ```
> Will this be preferable? (How to add `ppc` here correctly?)

I'd be more comfortable if the check was generalized to 10.6, but could potentially be persuaded.

@kencu your thoughts on the best approach here?

kencu:
I don't really have an opinion, except to say that a large pile of spaghetti code to fix something as niche as building as powerpc on 10.6 using rosetta is not something I would in general be encouraging.

As we said months ago, general fixes to portfiles that "do the right thing" and in the process fix weird builds are fine, IMHO.

Specific fixes like "if 10.6 and ppc do this" are not something I would support in the general case, as the usage group is microscopic and the confusion is large.

IF the specific chunk of spaghetti does absolutely nothing to any other system, and if the maintainer of the port feels s/he can live with it, and if it is generally agreed that the fix is unmaintainable, unsupportable, tickets will be closed without any attempt at fixing them, and it is known that the fix will most likely break with the next port update, but that is OK -- well, that is not up to me.

But these hundreds of rosetta fixes belong in a custom overlay repo, IMHO.

kencu:
balanced against that is that @barracuda156 has become an enthusiastic and valuable member of the community, we have shepherded him along though the hard parts of learning about the intricacies of macports, and he now has a very good understanding of the process, and we don't want to do anything much to dissuade him from his support and participation for macports.

so if his interest is (pretty much) 10.6 on PPC, and that is what he wants to promote, and people feel it is worth biting the bullet of the 10.6 on PPC spaghetti code to keep him enthusiastic, that is also worth pondering!.. 

decisions, decisions...

kencu:
I will admit to not really having a strong opinion at present.

barracuda156:
> I'd be more comfortable if the check was generalized to 10.6, but could potentially be persuaded.

@mascguy I think restriction to 10.6.8 does no harm to anyone, since there are literally two versions of 10.6.x currently in use – 10.6.8 (where `ppc` means Rosetta, and my fix is required) and 10.6 PPC (10A96 or 10A190, in future maybe some other build, where the fix is unnecessary, since `cpu_arch` is PPC).
I do not think anyone realistically gonna use 10.6.2 etc.
At the same time, the fix won’t break anything on 10.6 PPC, so I am okay with either.

mascguy (approved):
I'm fine with this, let's merge.

PR #15551, opened 2022-08-01
playdate-sdk: new port

idleberg:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.15.7 19H2026 x86_64
No receipt for 'com.apple.pkg.CLTools_Executables' found at '/'.
Command Line Tools

###### Verification 
Have you

### DIFF
--- a/devel/playdate-sdk/Portfile
+++ b/devel/playdate-sdk/Portfile
@@ -73,9 +73,12 @@ destroot {
 
 
 notes {
-    For your convenience, it is recommended to set PLAYDATE_SDK_PATH environmental variable:
+    Setting the PLAYDATE_SDK_PATH environmental variable is recommended, but not required:
 
         export PLAYDATE_SDK_PATH=${prefix}/share/playdate-sdk
+
+    You may also want to add $PLAYDATE_SDK_PATH/bin to your shell $PATH variable. This allows running pdc,
+    pdutil and the Simulator from any location without a fully qualified path.
 
     See https://sdk.play.date/1.12.3/Inside%20Playdate.html#_set_playdate_sdk_path_environment_variable
 }
--- a/devel/playdate-sdk/Portfile
+++ b/devel/playdate-sdk/Portfile
@@ -71,8 +71,13 @@ destroot {
     xinstall -d -m 777 "${destroot}${prefix}/share/Disk/Data"
 }
 
-post-activate {
-    system "export PLAYDATE_SDK_PATH=${prefix}/share/${name}"
+
+notes {
+    For your convenience, it is recommended to set PLAYDATE_SDK_PATH environmental variable:
+
+        export PLAYDATE_SDK_PATH=${prefix}/share/playdate-sdk
+
+    See https://sdk.play.date/1.12.3/Inside%20Playdate.html#_set_playdate_sdk_path_environment_variable
 }
 
 test.run            yes
--- a/devel/playdate-sdk/Portfile
+++ b/devel/playdate-sdk/Portfile
@@ -3,7 +3,7 @@
 PortSystem          1.0
 
 name                playdate-sdk
-version             1.12.2
+version             1.12.3
 revision            0
 categories          devel
 maintainers         @idleberg
@@ -18,9 +18,9 @@ master_sites        https://download-keycdn.panic.com/playdate_sdk/
 distname            PlaydateSDK-${version}
 distfiles           ${distname}.zip
 
-checksums           rmd160  f0768a68c50b9d92029b562fff36e9235562de80 \
-                    sha256  42c793eb1e5d5251f3f1b8ac0cd1b1489c5f3e5a763b130c31205496d12389ce \
-                    size    30080744
+checksums           rmd160  b3312e20880832d5a41bd312d936ffe9440bde70 \
+                    sha256  97ff5005093b4b05af7f5a053d1d260de7821f5576105d749cc317f212c7b5af \
+                    size    30076091
 
 depends_extract     port:xar
 
--- a/devel/playdate-sdk/Portfile
+++ b/devel/playdate-sdk/Portfile
@@ -69,7 +69,7 @@ destroot {
     }
 
     # Set permissions for the Playdate Simulator's data store
-    xinstall -d -m 755 "${destroot}${prefix}/share/Disk/Data"
+    xinstall -d -m 777 "${destroot}${prefix}/share/Disk/Data"
 }
 
 post-activate {
--- a/devel/playdate-sdk/Portfile
+++ b/devel/playdate-sdk/Portfile
@@ -27,8 +27,8 @@ depends_extract     port:xar
 worksrcdir          ${name}-${version}
 
 extract {
-    system -W ${workpath} "mkdir -p ${worksrcpath}/pkg"
-    system -W ${workpath} "mkdir -p ${worksrcpath}/files"
+    xinstall -d "${worksrcpath}/pkg"
+    xinstall -d "${worksrcpath}/files"
     system -W ${distpath} "unzip ${distpath}/${distfiles} -d ${worksrcpath}/pkg"
     system -W ${worksrcpath}/pkg "${prefix}/bin/xar -xf ${worksrcpath}/pkg/PlaydateSDK.pkg -C ${worksrcpath}/files"
     system -W ${worksrcpath}/files/PlaydateSDK.pkg "cat Payload | gunzip -dc | cpio -i"
--- a/devel/playdate-sdk/Portfile
+++ b/devel/playdate-sdk/Portfile
@@ -7,8 +7,7 @@ version             1.12.2
 revision            0
 categories          devel
 maintainers         @idleberg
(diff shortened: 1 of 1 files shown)


aeiouaeiouaeiouaeiouaeiouaeiou:
@mascguy accidentally I started a review on this instead of commenting (thank you GitHub), so my suggestions are stuck on "pending" status.

aeiouaeiouaeiouaeiouaeiouaeiou (changes requested):

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,94 @@
...
+            system -W "${prefix}/share/examples/${name}/${example}" \
+                "${prefix}/bin/pdc -sdkpath ${prefix}/share/${name} --no-compress Source /tmp/${example}.pdx"
+            file delete -force "/tmp/${example}.pdx"
+        }
+    }
+}
+
+livecheck.type      regex
+livecheck.url       ${homepage}
+livecheck.regex     "Download Playdate SDK (\\d+(?:\\.\\d+)*) for macOS"
```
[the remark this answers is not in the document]
idleberg: I'm not sure what this means, should I check the anchor for the download instead? If so, the problem is that it uses aliases (`latest` will resolve to a specific version).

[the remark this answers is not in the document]
aeiouaeiouaeiouaeiouaeiouaeiou: I could be wrong, but it should look something like this:

```
livecheck.type      regex
livecheck.url       ${master_sites}
livecheck.regex     "PlaydateSDK-(\d+(?:\.\d+)*)"
```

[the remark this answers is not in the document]
idleberg: Oh yes, that would work

aeiouaeiouaeiouaeiouaeiouaeiou (approved):

aeiouaeiouaeiouaeiouaeiouaeiou:
@mascguy LGTM, any advice on this?

mascguy (approved):
While I haven't tested with this locally, it looks good overall.

I'd like to have @reneeotten or someone else review as well though, as I often miss minor issues with new submissions.

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
...
+license             Restrictive
+
+description         Playdate SDK
+long_description    This package installs includes a rich set of APIs, a \
+                    compiler, and documentation for Playdate game development.
+
+homepage            https://play.date/dev/
+master_sites        https://download-keycdn.panic.com/playdate_sdk/
+distname            PlaydateSDK-${version}
+distfiles           ${distname}.zip
```
reneeotten, line 19: if this uses a ZIP file, you probably just should set `use_zip yes` and then the `distfiles` field and manually `unzip` in the extract phase shouldn't be required

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
...
+
+checksums           rmd160  f0768a68c50b9d92029b562fff36e9235562de80 \
+                    sha256  42c793eb1e5d5251f3f1b8ac0cd1b1489c5f3e5a763b130c31205496d12389ce \
+                    size    30080744
+
+depends_extract     port:xar
```
reneeotten, line 25: this should likely be `depends_extract-append` as I suspect that the `port:zip` dependency gets added already when setting `use_zip` yes and you don't want to overwrite that

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
...
+
+worksrcdir          ${name}-${version}
+
+extract {
+    system -W ${workpath} "mkdir -p ${worksrcpath}/pkg"
+    system -W ${workpath} "mkdir -p ${worksrcpath}/files"
```
reneeotten, line 31: this should use Tcl code with `xinstall -d <dirname>`

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
...
+        system "cp -R \"${worksrcpath}/${name}/Examples/${dir}\" ${destroot}${prefix}/share/examples/${name}"
+    }
+
+    # Set permissions for the Playdate Simulator's data store
+    xinstall -d -m 777 "${destroot}${prefix}/share/Disk/Data"
+}
+
+post-activate {
+    system "export PLAYDATE_SDK_PATH=${prefix}/share/${name}"
+}
```
reneeotten, line 76: this isn't going to "stick" once the user restarts a shell or something. So you likely should add `notes` and state there that this should be added to the users shell init scripts?

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
...
+            "${destroot}${prefix}/share/doc/${name}/${dir}.html"
+    }
+    copy -force "${worksrcpath}/${name}/PlaydateSDK.docset" \
+        "${destroot}${prefix}/share/doc/${name}"
+
+    # Examples
+    foreach dir [glob -type d -nocomplain -tails -directory "${worksrcpath}/${name}/Examples/" *] {
+        xinstall -d "${destroot}${prefix}/share/examples/${name}/${dir}"
+        system "cp -R \"${worksrcpath}/${name}/Examples/${dir}\" ${destroot}${prefix}/share/examples/${name}"
+    }
```
reneeotten, line 68: can the `system` calls above be replaced with Tcl code instead?

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                playdate-sdk
+version             1.12.2
```
reneeotten, line 6: while this has been in the PR queue it appears a newer version is available, perhaps update to that one?

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,93 @@
...
+                    size    30080744
+
+depends_extract     port:xar
+
+worksrcdir          ${name}-${version}
+
+extract {
+    system -W ${workpath} "mkdir -p ${worksrcpath}/pkg"
+    system -W ${workpath} "mkdir -p ${worksrcpath}/files"
+    system -W ${distpath} "unzip ${distpath}/${distfiles} -d ${worksrcpath}/pkg"
```
reneeotten, line 32: this shouldn't be needed I think (see above)

reneeotten (changes requested):
@idleberg I left a few comments but must admit that I haven't tried to do anything locally. It's a bit of an unconventional Portfile with quite some overriding of the `extract` and `destroot` phases ; how does upstream want you to install this if download it directly from their website?

idleberg:
@reneeotten IIRC, the Portfile became so ”unconventional” because of the same problems I'm having now. Do you have any advice? Unfortunately, I currently don't have access to the computer on which I could test my initial commits locally.

devel/playdate-sdk/Portfile
```diff
@@ -0,0 +1,99 @@
...
+
+description             Playdate SDK
+long_description        This package installs includes a rich set of APIs, a \
+                        compiler    , and documentation for Playdate game development.
+
+homepage                https://play.date/dev/
+master_sites            https://download-keycdn.panic.com/playdate_sdk/
+distname                PlaydateSDK-${version}
+distfiles               ${distname}.zip
+use_zip                 yes
```
mascguy, line 20: Minor nit-pick: Move `use_zip yes` above the line for `distfiles`, then change the latter to:

```
distfiles         ${distname}${extract.suffix}
```

That's the preferred way to go, rather than hard-coding the file extension.

reneeotten: I don't think setting "distfiles" is needed at all - right?

devel/playdate-sdk/Portfile
mascguy: > this isn't going to "stick" once the user restarts a shell or something. So you likely should add `notes` and state there that this should be added to the users shell init scripts?

Alternatively, the typical approach is to add symlinks to items in your new bin directory, ensuring the names are unique enough to not clash with other ports. So ideally they'd all have a unique prefix, like `playdate-`. (The target scripts/utils don't need to be renamed, but the symlinks in `${prefix}/bin` should definitely have the prefix.)

With that being said, if the targets can't/don't infer the location of `PLAYDATE_SDK_PATH`, it might be good to create wrapper scripts [in `${prefix}/bin`] instead, which set that environment variable first. It's a convention we use elsewhere as well, and it significantly improves the user experience.

devel/playdate-sdk/Portfile
mascguy: > I don't think setting "distfiles" is needed at all - right? Isn't what you show the default, only setting "distname" should be enough I think.

Yes, you're right. As usual, I'm compartmentalizing a bit too much. Lol

PR #15553, opened 2022-08-01
inih: submission

rhapsodyv:
#### Description

inih (INI Not Invented Here) is a simple .INI file parser written in C. It's used by tio and I need it to fix tio port file, as commented in #15400 and https://trac.macports.org/ticket/65523.

###### Type(s)

- [x] submission

###### Tested on

macOS 12.4 21F79 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/inih/Portfile
+++ b/devel/inih/Portfile
@@ -1,0 +1,24 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+PortGroup               meson 1.0
+
+github.setup            benhoyt inih 56 r
+revision                0
+categories              devel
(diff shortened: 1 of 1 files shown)


rhapsodyv:
should this have a "lib" prefix? If so, I need to customize the port file because the name of the repository doesn't have "lib" in the name...

devel/inih/Portfile
```diff
@@ -0,0 +1,26 @@
...
+PortGroup               github 1.0
+PortGroup               meson 1.0
+
+github.setup            benhoyt inih 56 r
+revision                0
+categories              devel
+maintainers             nomaintainer
+license                 GPL-2
+
+homepage                https://github.com/benhoyt/inih
```
herbygillot, line 13: `homepage` can be removed.

It is being automatically set by `github.setup`

herbygillot:
> should this have a "lib" prefix? If so, I need to customize the port file because the name of the repository doesn't have "lib" in the name...

Not necessary - the port name can match the project name in this case, as long as there are no prior ports that already have the name in question.

PR #15558, opened 2022-08-01
isl: bump to 0.25

kencu:
and disable cxx17 on older darwin systems
the libc++ in these older systems does not
support sufficient c++17 for isl, but the
isl c++17 tests don't dig deeply enough
to find these failures

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.7

###### Verification 
Have you

This PR uses a rather blunt method to force the c++17 tests to fail, thereby disabling c++17 on systems where it is not fully supported. Proper tests that fail on macOS systems without full c++17 support would be better than this, but I leave this here as a concept of what might be done in the meantime perhaps.

[isl-0.25-tests-Lion.txt](https://github.com/macports/macports-ports/files/9235674/isl-0.25-tests-Lion.txt)

### DIFF
--- a/devel/isl/Portfile
+++ b/devel/isl/Portfile
@@ -10,8 +10,8 @@ name                isl
 # (e.g. port file all | sort -u | xargs grep -E ':isl( |$)' | cut -d / -f 13 | sort -u)
 # see https://lists.macports.org/pipermail/macports-dev/2019-May/040678.html
 epoch               4
-version             0.24
-revision            1
+version             0.25
+revision            0
 
 categories          devel math
 platforms           darwin
@@ -38,9 +38,9 @@ depends_lib         port:gmp
 master_sites        sourceforge:libisl
 use_bzip2           yes
 
-checksums           rmd160  b3bf8e1ad50207d4eebecc47cb4cffdac6581a57 \
-                    sha256  fcf78dd9656c10eb8cf9fbd5f59a0b6b01386205fe1934b3b287a0a1898145c0 \
-                    size    2261594
+checksums           rmd160  73b9005b8e7fe52eb9500b98ba65450bdfb9f7fb \
+                    sha256  4305c54d4eebc4bf3ce365af85f04984ef5aa97a52e01128445e26da5b1f467a \
+                    size    2304378
 
 configure.args      --disable-silent-rules
 
@@ -58,6 +58,12 @@ if {${subport} eq ${name}} {
 
     # ./include/isl/typed_cpp.h:2132:12: error: no matching conversion for functional-style cast from 'const isl::basic_set' to 'isl::typed::basic_set<>'
     compiler.blacklist {clang < 500}
+
+    if { ${os.platform} eq "darwin" && ${os.version} < 18 } {
+        # cxx17 support on these systems is incomplete when building against libc++
+        patchfiles-append   patch-configure-force-cxx17-to-fail.diff
+    }
+
 }
 
 subport isl14 {
--- a/devel/isl/files/patch-configure-force-cxx17-to-fail.diff
+++ b/devel/isl/files/patch-configure-force-cxx17-to-fail.diff
@@ -1,0 +1,22 @@
+--- configure.orig	2022-07-29 10:01:55.000000000 -0700
++++ configure	2022-07-29 10:02:40.000000000 -0700
+@@ -9283,6 +9285,9 @@
+ 
+ #else
+ 
++// MacPorts
++#error "C++17 support forced off"
++
+ #include <initializer_list>
+ #include <utility>
+ #include <type_traits>
+@@ -10105,6 +10110,9 @@
+ 
+ #else
+ 
++// MacPorts
++#error "C++17 support forced off"
++
+ #include <initializer_list>
+ #include <utility>
+ #include <type_traits>


kencu:
for consistency, might be best to have c++17 disabled on all systems < 10.14.

to fine-grain it, you would have to disable c++17 on all systems between 10.7 and 10.13 that are building with the cxx_stdlib set to "libc++" or "libstdc++" as those are the ones where it would fail.

The test to apply the patch could be adjusted to accomplish that if that level of detail is wanted. Of course, having inconsistency between systems makes the systems different, and that makes fixing errors all that much harder.

Now, if there was some big advantage to having c++17 support added to isl, then you might be able to make a case that it is critical that certain ancient systems have it enabled. As far as I can tell, it's pretty much meaningless, and as it can't be supported on 10.7 to 10.13 anyway, any fixes that come along will probably just apply to all systems < 10.14.

So I would leave it like this, but I'm not the maintainer of isl, just trying to help show one possible way forward.

barracuda156:
@kencu Honestly I don′t know if we get any benefits on OS <10.7 with c++17 enabled in `isl`, but apparently `isl` builds fine with it (and having `libstdc++` set). Maybe @iains can advise if that is of any consequence for `gcc`.

iains:
GCC  builds with C++11 (from GCC-11), so the only benefit of using a newer ```isl``` would be any bug fixes there.  I currently bootstrap and test with isl-0.24.

barracuda156:
Then I guess @kencu solution is fine, we can disable c++17 in `isl` for all old systems.

kencu:
in macports, having the cxx_stdlib set to libstdc++ means using the gcc libraries in /usr/lib.

those gcc libraries are from gcc-4.2, and they don’t support c++17.

I did not see any reports about exactly how isl 0.25 built on 10.5 before…perhaps c++17 was not detected and was disabled there anyway already.

iains:
> in macports, having the cxx_stdlib set to libstdc++ means using the gcc libraries in /usr/lib.
> 
> those gcc libraries are from gcc-4.2, and they don’t support c++17.

They do not support C++11 either (there are some entries in tr1, and some GNU extensions, but not general).

> I did not see any reports about exactly how isl 0.25 built on 10.5 before…perhaps c++17 was not detected and was disabled there anyway already.

If it's built with gcc-4.2, then it will be using C++98..

the compilers generally have a default standard version which increases as they support said standards ... so...

.. if it's built with a later GCC then it could be using C++11 or maybe 14 if you have gcc-7.5 (but then the libstdc++ library would have to be the compiler's version)

likewise, if it's clang > 3.4 then could be C++11.. etc etc. 

but, I'd assume that the build recipes are specific about the C++ standard they expect (regardless of the compiler used to do the build) so that any assumptions would result in a build error if the compiler did not support?)

barracuda156:
> in macports, having the cxx_stdlib set to libstdc++ means using the gcc libraries in /usr/lib.
> those gcc libraries are from gcc-4.2, and they don’t support c++17.

@kencu You mean when `gcc42` is being used or for everything? We certainly _can_ build ports that require newer C and CXX standards on Leopard (and SL PPC) using `gcc10` or later, so perhaps they use libs from `libgcc` port instead of /usr/lib?

kencu:
when a port sets

compiler.cxx_standard 2011

or higher, macports base changes the cxx_stdlib from libstdc++ to macports-libstdc++. This then tells everything that the gcc libraries in /opt/lical/lib/libgcc are to be used instead of the gcc libraries in  /usr/lib, amongst other changes.

part of the other changes is to use a newer gcc version to do the build in some cases. As the newer gcc will automatically use the newer libraries, it does not need to be instructed what to do. But if clang is being used, it does need to be told to use the new libraries explicitly, as by default it will not.

there are many thing that go on behind the scenes that make this seem simpler than it actually is… there are many movingnparts to keep coherent.

kencu:
macports, by the way, still uses gcc7 to build newer software on older systems.

gcc10 is not used, or needed. You made yourvown modifications to macports thinking gcc10 might be needed for something, but it is not needed yet.

Sometime, when gcc10-bootstrap gets sorted out properly, somebody will fix things up properly so gcc12/libgcc12 is used for all systems, including older systems. So far, there is no software that doesn’t build with gcc7 though, so nobody has bothered to do that work as yet.

barracuda156:
@kencu Thanks for the explanation!

> So far, there is no software that doesn’t build with gcc7 though, so nobody has bothered to do that work as yet.

Well, unless @iains backported his fix for `ppc64` into `gcc7`, I assume `icu` _needs_ `gcc10`+ on Leopard for `ppc64` or `+universal` build.
Then, if this bug gets fixed, we may fix `SDL2` for PPC too: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105522 (and if it does, it will perhaps be fixed in `gcc12` or later).

kencu:
i have the last libsdl2 that worked on ppc in my leopardports and tigerports repos now.

no newer version will likely ever build due to SDK issues.

barracuda156:
> i have the last libsdl2 that worked on ppc in my leopardports and tigerports repos now.
> no newer version will likely ever build on < 10.6 due to SDK issues.

I actually have `libsdl2` 2.0.9 built for PPC: https://trac.macports.org/ticket/64655
That is still WIP though.

kencu:
I believe what you built in that ticket would never work though, right?

You can certainly #ifdef things to the point they might compile, by cutting out the parts that don’t, but there is not much point in that. We had actually rewritten parts of libsdl2 to function properly on 10.4 and 10..5 using SDK features available on those systems, until it became practically impossible to go any farther without massive re-engineering.

barracuda156:
> I believe what you built in that ticket would never work though, right?
> 
> You can certainly #ifdef things to the point they might compile, by cutting out the parts that don’t, but there is not much point in that. We had actually rewritten parts of libsdl2 to function properly on 10.4 and 10..5 using SDK features available on those systems, until it became practically impossible to go any farther without massive re-engineering.

Once I go through the patches and fix what is remaining, I will try testing it.

kencu:
It would be very great news if libsdl2 could leap ahead to 2.0.9 on PPC! Good luck!

kencu:
I believe you are likely right, @barracuda156  -- building icu as ppc64 is likely not possible with gcc7. So indeed that is the beginning of the pressure to upgrade the default gcc/libgcc to something newer. Which should happen, certainly, as it is best to have all the systems as homogeneous as possible to ease the maintenance burden.

barracuda156:
@kencu So what happened to your PR, why closed?

kencu:
> @kencu So what happened to your PR, why closed?

It was just a proof-of-concept, for Marcus or whoever if helpful. It was a silly, bad fix, and Marcus could have come up with it in 10 seconds, probably something better.

Someone needs to work with upstream to get a proper c++17 test that covers MacOS into the configure script.

In the meantime, nobody cares if isl stays at version 0.24 for the next few years.

PR #15563, opened 2022-08-01
denemo: remove port

kakuhen:
#### Description
* 0 requested installs in the past 90 days
* Port health is red across the board
* No maintainer
* No dependents

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Verification 
Have you


barracuda156:
@kakuhen I can check if it builds, but that will take some time (need to build `guile`, which takes forever).

kakuhen:
As of now, this port (together with an unmaintained gstreamer plugins port) is preventing fluidsynth from being updated, which in turn is preventing dosbox-staging from being added to MacPorts. In the mean time, I am considering adding a variant to fluidsynth that allows outdated versions to be installed. The issue is that almost none of the fluidsynth dependents clearly state which versions of fluidsynth they support, and this library seems to introduce breaking API and ABI changes often.

barracuda156:
> unmaintained gstreamer plugins port

@kakuhen I think I have updated them and fixed for PPC. Will check it. If they build normally, I can make a PR.

kakuhen:
> > unmaintained gstreamer plugins port
> 
> @kakuhen I think I have updated them and fixed for PPC. Will check it. If they build normally, I can make a PR.

Sounds like good news. In particular, the dependent is `gstreamer1-gst-plugins-bad`. I was unable to determine if this can build with later versions of Fluidsynth, or even the version that's currently in MacPorts! Many of the plugins depending on Fluidsynth test for really outdated versions and it's not clear if the API is stable enough for `gstreamer1-gst-plugins-bad` to build without issue.

aeiouaeiouaeiouaeiouaeiouaeiou:
I see some misinformation here.

- The number of installations is sent only from Macs on which [`mpstats`](https://ports.macports.org/statistics/faq/) is installed
- Locally a port is installed [without any errors](https://github.com/macports/macports-ports/pull/14510#pullrequestreview-937146713)
- First of all a port should be removed if source program development is abandoned. I don't think that Denemo fits this criteria.

kakuhen:
The mpstats story is very unfortunate, since it makes the port look unrequested by seemingly anyone.



As for building, this was not a good sign to me


BUT given that there are reports of successful local builds, I will close this PR. With that said, @aeiouaeiouaeiouaeiouaeiouaeiou do you know if this software can build locally using FluidSynth 2.2.x or later? This information to me since the FluidSynth port maintainer needs information on the status of this port, and whether upstream has adapted to the newer APIs.

barracuda156:
@kakuhen One of the problems with `mpstats` is that `curl` is broken on old systems, so it literally cannot work, unless one takes a pain of bootstrapping a separate Macports tree in different prefix and then rebuilding the main Macports installation against that one `--with-curl-prefix=`.
Understandably, pretty much no one does that. (I have it on all my systems, and often if you see 1–2 installations of a port for PPC, that’s all mine.)

kakuhen:
Huh, I expected that was only an issue on PPC due to the age of Leopard. In any case, I agree, the workaround is lengthy and it doesn't even work on my G5! I wonder if we could suggest an option to allow HTTP-only option for older Mac OS versions. It may be insecure (e.g. MITM attacks), but ultimately, a JSON blob is sent to a server, and presumably validation can be performed on the JSON blob.

barracuda156:
> Huh, I expected that was only an issue on PPC due to the age of Leopard. In any case, I agree, the workaround is lengthy and it doesn't even work on my G5!

This should work: https://forums.macrumors.com/threads/building-with-macports-for-10-6-powerpc-10a190-and-10-6-8-rosetta.2332711/?post=30906371#post-30906371
`mpstats` is working properly on all my systems from Tiger to SL PPC and 10.6.8.

PR #15565, opened 2022-08-02
dosbox-x: update to 0.84.2

srirangav:
#### Description

Update to v0.84.2

Please note, the version tag for this version has changed from the format from the prior version which was 0.84.1 to 2022.08.0.  I wasn't sure how best to address this.  If the way I did this isn't proper, please let me know and I'll try to revise this.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.6.8 20G730 arm64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/emulators/dosbox-x/Portfile
+++ b/emulators/dosbox-x/Portfile
@@ -7,12 +7,12 @@ PortGroup           app 1.0
 PortGroup           compiler_blacklist_versions 1.0
 PortGroup           muniversal 1.0
 
-github.setup        joncampbell123 dosbox-x 0.84.1 dosbox-x-v
+github.setup        joncampbell123 dosbox-x 2022.08.0 dosbox-x-windows-v
 revision            0
 
-checksums           rmd160  c379c38b51e8ee01092839537e043b9bdca9f5e0 \
-                    sha256  d5ba5b3d87b0dc69d70f6c9701eec36772bbc3716e0b201b74ec73d4b3ff38cc \
-                    size    67617800
+checksums           rmd160  a97405c17b9f24c70855d5553f42f9e23122e947 \
+                    sha256  bac936ee198b607fb0cc55dfc4575e44b700dbc9138904d2e264792a6e6a941c \
+                    size    67809452
 
 categories          emulators
 platforms           darwin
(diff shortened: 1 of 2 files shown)


kakuhen:
The GitHub port group should manage the version numbers automatically. Luckily, 2022 is a much bigger number than 0, so I think the commit as-is should not present any problems for people wanting to update to the latest version. 

Moreover, it seems that [the latest release also refers to itself as 0.84.2](https://github.com/joncampbell123/dosbox-x/releases/tag/dosbox-x-windows-v2022.08.0)

srirangav:
> The GitHub port group should manage the version numbers automatically. Luckily, 2022 is a much bigger number than 0, so I think the commit as-is should not present any problems for people wanting to update to the latest version.
> 
> Moreover, it seems that [the latest release also refers to itself as 0.84.2](https://github.com/joncampbell123/dosbox-x/releases/tag/dosbox-x-windows-v2022.08.0)

Hi @kakuhen,

If dosbox-x reverts to using the 0.84.x tag format for its next minor release, will the fact that the Portfile sets the version number to 2022.08.0 cause problems for future upgrade?

Is there another way to download the correct distfile but leave the version number as 0.84.2?

I looked through Portfile reference for the GitHub port group:

https://guide.macports.org/chunked/reference.portgroup.html#reference.portgroup.github

But I couldn't figure out how to get the latest distfile without setting the version number to 2022.08.0.

Thanks,

-ranga

kakuhen:
> If dosbox-x reverts to using the 0.84.x tag format for its next minor release, will the fact that the Portfile sets the version number to 2022.08.0 cause problems for future upgrades?

This assumes the GitHub release does not identify itself as 2022.x.y. In that case, you can bump the `epoch` of the Portfile. I don't expect the versioning to be an issue as long as dosbox-x consistently uses their new versioning scheme.

srirangav:
> > If dosbox-x reverts to using the 0.84.x tag format for its next minor release, will the fact that the Portfile sets the version number to 2022.08.0 cause problems for future upgrades?
> 
> This assumes the GitHub release does not identify itself as 2022.x.y. In that case, you can bump the `epoch` of the Portfile. I don't expect the versioning to be an issue as long as dosbox-x consistently uses their new versioning scheme.

Makes sense to me.  Thanks.

neverpanic (approved):
Yeah, you may have to increase the epoch with the next update if they return to their previous numbering scheme. No biggie.

PR #15573, opened 2022-08-02
clang-11-bootstrap: fix build on macOS 10.9

catap:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.9.5 13F1911 x86_64
Xcode 6.2 6C131e

###### Verification 
Have you

### DIFF
--- a/lang/clang-11-bootstrap/Portfile
+++ b/lang/clang-11-bootstrap/Portfile
@@ -211,6 +211,11 @@ if {${configure.sdkroot} ne ""} {
     configure.args-append \
                     -DDARWIN_osx_SYSROOT="${configure.sdkroot}" \
                     -DDEFAULT_SYSROOT="${configure.sdkroot}"
+} else {
+    # compiler-rt does a broad search for an SDK it likes, but this
+    # search fails on older systems that don't have a MacOSX.sdk
+    configure.args-append \
+                    -DDARWIN_osx_SYSROOT=/
 }
 
 if {${os.major} <= 18} {


catap:
@kencu / @mascguy may I ask you to merge this hotfix? Thanks.

mascguy (approved):
You bet, thanks for the quick fix!

catap:
@mascguy thanks, I hope that now I have a green panel on this port and may move forward with https://github.com/macports/macports-ports/pull/13995 and https://github.com/macports/macports-ports/pull/13986 ;)

Gcenx:
After both of those get merged/built it would be nice to circular back around to this and have the port copy in libgcc so anyone installing the prebuilt package won’t also need to install gcc10-bootstrap.

I’ve found clang-11-bootstrap rather useful just the current requirement of gcc10-bootstrap as a runtime dep is off putting.

catap:
@Gcenx because all this ports is kind of bootstrap and should be used only to build the real toolchain... well.. I haven't see any benefit to do it :)

Gcenx:
@catap I’ve been using it to bootstrap/breaks some dependency cycles. I was hoping to not always need to carry a custom version but maybe that will be the case.

I still believe these bootstrap changes are a great addition!

catap:
> @catap I’ve been using it to bootstrap/breaks some dependency cycles. I was hoping to not always need to carry a custom version but maybe that will be the case.

kencu:
next will be to see if we can go from system roots right to libtapi + ld64-latest in one step on 10.6 through 10.13, skipping right over all the ld64 manual bootstrapping hassles.

catap:
@kencu maybe you know a way to see a build dependency tree for a port?

Gcenx:
> next will be to see if we can go from system roots right to libtapi + ld64-latest in one step on 10.6 through 10.13, skipping right over all the ld64 manual bootstrapping hassles.

I’ve only done this for 10.9, 10.14 and 11 but it’s possible just a little long winded due to the constant gcc10-bootstrap requirement even for systems that could easily build clang-11-bootstrap directly.

catap:
@kencu / @Gcenx I plan to try to simplify building of `cctools` via `clang-11-boostrap` and `libcxx` when it requries which should broke a lot of dependencies.

kencu:
cctools is pretty easy… just simple c code, IIRC, never had any trouble there.

ld64 is the headache.

catap:
@kencu yeah, but a few step is required before move on ;)

Gcenx:
@kencu are you actually referring to the actual latest version of cctools or the latest on Macports?

ld64 609 will build against clang-11-bootstrap just fine

kencu:
to find all the recursive deps I believe it is 

port -rdeps cctools

from memory. BUT it is not very accurate, I’m afraid, for various reasons, esp “bin” deps and similar. These totally mess you up trying to get accurate deps.

kencu:
> @kencu are you actually referring to the actual latest version of cctools or the latest on Macports?
> 
> ld64 609 will build against clang-11-bootstrap just fine

cctools is (was ) easy, all versions I updated. However, they stopped testing ppc and i386 and those have bitrotted pretty badly, and I had to fix several errors there. So be careful if you try to update cctools for that reason.

ld64 (all newer versions) are much harder to build. If clang11 can build ld64-609 using the system ld64 on 10.6 as a bootstrap that would make things much easier indeed.

Gcenx:
I’ve been using cctools 973.0.1 and ld64 609 for months without issue (outside of my own dumb mistake on an ld64 patch lol)

i386 binaries launched correctly unlike with the stock Xcode11 cctools that produced broken i386 binaries (as)

PR #15636, opened 2022-08-06
M2VDownsizer: new port

RobK88:
#### Description

* Create a new port, M2VDownsizer, for MPEG-2 (or m2v) video stream re-compression
* Convert the Xcode project into a simpler Makefile based project
* Add a custom Makefile to build M2VDownsizer
* Replace deprecated MPProcessors() function with sysctlbyname() function in main.c
* Use the system libmpeg2 library from Macports instead of the provided old FAT binary
  for libmpeg2
* Add include guards to config.h
* Fix the inline macro in config.h
* Simplify the code in config.h
* Change #includes "config.h" to #includes <config.h> in mjpeg_logging.c
* Add a man page (created by me for the Macports project)

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.7.5 11G63 i386
Xcode 4.6.3 4H1503

###### Verification 
Have you


multimedia/M2VDownsizer/Files/M2VDownsizer.1
```diff
@@ -0,0 +1,73 @@
...
+recompress the MPEG-2 video in the above example.  Time to take a long coffee
+break!
+.SH SEE ALSO
+\fBM2VRequantiser\fP(1)
+.SH AUTHOR
+The author is unkown.  \fBM2VDownsizer\fP was released as open source software
+on the website for Metakine Inc., the developer of DVDRemaster.
+.PP
+This manual page was written by Robert Kennedy <amtor@hotmail.com> for the
+Macports project (but may be used by others).
```
ryandesign, line 73:
```suggestion
MacPorts project (but may be used by others).
```

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+#
+
+# Define the "macports" PHONY target
+#
+macports: depend all
+
+# Define $(PRODUCT) real target
+#
+# Note:  Do not add phony targets to the $(PRODUCT) REAL target. 
+# Phony targets should not be a prerequisite  of a real target file!
```
ryandesign, line 160:
```suggestion
# Phony targets should not be a prerequisite of a real target file!
```

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+#FWORKDIRS = -F./ -F/System/Library/Frameworks/
+
+# Define any libraries to link into executable:
+#   Statement Format: LIBS := -llibname
+#   E.g. if I want to link the math library, use -lm as follows:
+#   LIBS := -lm
+#
+LIBS := -lmpeg2
+
+# Define any MacOS Frameworks to link into executable:
```
ryandesign, line 81:
```suggestion
# Define any macOS frameworks to link into executable:
```

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+OBJS-MAIN := $(SRCS-MAIN:.c=.o)
+
+# Define main object directory
+#
+OBJDIR := obj
+
+# Define object sub-directories
+#
+# Do not change the following directories!!
+# The name of the last directory  MUST match the name of the corresponding
```
ryandesign, line 121:
```suggestion
# The name of the last directory MUST match the name of the corresponding 
```

multimedia/M2VDownsizer/Files/1003-Patch-config.h.diff
```diff
@@ -0,0 +1,251 @@
+--- config.h.orig	2006-02-27 21:38:49.000000000 -0500
++++ config.h	2022-08-06 07:34:47.000000000 -0400
```
ryandesign, line 2: This patch is much larger than I would have expected to be necessary. All I would have expected to need is
```
#ifndef CONFIG_H
#define CONFIG_H
```
at the top and
```
#endif
```
at the bottom.

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+#CFLAGS ?= -mmacosx-version-min=10.7 -isysroot /Developer/SDKs/MacOSX10.7.sdk
+
+# Append additional architecture dependent compile-time flags
+#
+CFLAGS += -Xarch_ppc -faltivec
+CFLAGS += -Xarch_ppc64 -faltivec
+
+# Alternative code for appending additional architecture dependent compile-time flags
+# Note: Use as last resort
+#       It is better to use "-Xarch_ppc" etc as these CFLAGS will work
```
ryandesign, line 37:
```suggestion
#       It is better to use "-Xarch_ppc" etc. as these CFLAGS will work 
```

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+INCLUDES := -I./ -Iutils -Iutils/altivec -Impeg2dec -Impeg2enc \
+-I/System/Library/Frameworks/CoreServices.framework/Headers 
+
+# Define library directories other than the system library directory, typically /usr/lib
+#   Statement Format:  LFLAGS := -Llibrary_path
+#   E.g. LIBDIRS := -L/opt/local/lib
+#
+LIBDIRS := -Lobj  
+
+# Define any MacOS Framework directories other than the system Framework directory
```
ryandesign, line 68:
```suggestion
# Define any macOS frameworks directories other than the system Frameworks directory
```

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+#
+LDFLAGS += 
+
+# Define any directories containing header files other than directory containing 
+# the system header files, typically /usr/include
+#   Statement Format: INCLUDES := -Iinclude_path
+#   E.g. INCLUDES := -I./  -I/opt/local/include
+#
+INCLUDES := -I./ -Iutils -Iutils/altivec -Impeg2dec -Impeg2enc \
+-I/System/Library/Frameworks/CoreServices.framework/Headers
```
ryandesign, line 60: You should never need to write an include flag for framework headers like this. `-framework CoreServices` (already specified below) should take care of it, no?

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,78 @@
...
+
+PortSystem          1.0
+PortGroup           makefile 1.0
+
+name                M2VDownsizer
+version             1.1
+revision            0
+
+categories          multimedia
+platforms           darwin
```
ryandesign, line 11: This is the default.
```suggestion
```

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,78 @@
...
+long_description    ${name} is a tool that re-compresses MPEG-2 (or m2v) video \
+                    streams.  In other words, ${name} will perform a complete \
+                    decompression and recompression of MPEG-2 video streams for the \
+                    best possible video quality.\
+                    \n \
+                    \n${name}  accepts the raw MPEG-2 (or m2v) video data (not VOB) \
+                    from the standard input and writes the recompressed video frames \
+                    to the standard output.\
+                    \n \
+                    \nBe  forewarned!  Decompression  and  recompression  could take \
```
ryandesign, line 27:
```suggestion
                    \nBe forewarned!  Decompression and recompression could take \
```

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,78 @@
...
+                    openmaintainer
+
+description         MPEG-2 video stream recompressesion
+
+long_description    ${name} is a tool that re-compresses MPEG-2 (or m2v) video \
+                    streams.  In other words, ${name} will perform a complete \
+                    decompression and recompression of MPEG-2 video streams for the \
+                    best possible video quality.\
+                    \n \
+                    \n${name}  accepts the raw MPEG-2 (or m2v) video data (not VOB) \
```
ryandesign, line 23:
```suggestion
                    \n${name} accepts the raw MPEG-2 (or m2v) video data (not VOB) \
```

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,78 @@
...
+                    from the standard input and writes the recompressed video frames \
+                    to the standard output.\
+                    \n \
+                    \nBe  forewarned!  Decompression  and  recompression  could take \
+                    hours for large MPEG-2 video streams!  It is recommended that one \
+                    first tries transcoding MPEG-2 streams using M2VRequantiser since \
+                    transcoding is a lot faster and produces excellent results for most \
+                    MPEG-2 video streams. ${name} is only intended to be used for \
```
ryandesign, line 31:
```suggestion
                    MPEG-2 video streams.  ${name} is only intended to be used for \
```

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,78 @@
...
+                    large MPEG-2 video streams when transcoding does not produce \
+                    acceptable results.
+
+homepage            https://web.archive.org/web/20120508050929if_/http://www.metakine.com/
+master_sites        ${homepage}/files
```
ryandesign, line 36:
```suggestion
master_sites        ${homepage}files
```

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,78 @@
...
+                    1003-Patch-config.h.diff \
+                    1004-Fix-Includes-in-mjpeg_logging.diff
+
+#
+# Turning off parallel builds prevents garbled output from the compiler
+# caused by the following warning messages being generated at
+# the same time when non-ppc builds are being compiled:
+#
+# clang: warning: argument unused during compilation: '-Xarch_ppc -faltivec'
+# clang: warning: argument unused during compilation: '-Xarch_ppc64 -faltivec'
```
ryandesign, line 70: You could add `-Wno-unused-command-line-argument` to `configure.cflags` to suppress this warning.

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+#
+# 'make depend' uses makedepend to automatically generate dependencies 
+#               (dependencies are added to end of Makefile)
+# 'make'        build executable file 'M2VDownsizer'
+# 'make clean'  removes all .o and executable files
+#
+#  Note:    There is NO "make install"
+#           After running "make", just copy M2VDownsizer into the desired directory
+#           e.g.    sudo chmod +x ./M2VDownsizer
+#                   sudo cp M2VDownsizer /usr/local/bin
```
ryandesign, line 16: In the amount of space it took to write this comment, an install target could probably have been written. 😃

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+#               
+
+# Define the C compiler to use
+#
+CC := clang
```
ryandesign, line 21: Is there a reason why you want to specify clang here? The default CC for make is `cc` which is sufficient for most projects, and on Mac OS X 10.7 and later, `cc` is clang, and MacPorts will override CC anyway.

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
+
+# The following also works for the "all" PHONY target but will not display messages 
+# after each subdirectory has been processed (e.g. compiled).
+#
+# all: $(PRODUCT)
+#
+
+# Define the "macports" PHONY target
+#
+macports: depend all
```
ryandesign, line 155: Does the "macports" target work? In a parallel build scenario, the prerequisites "depend" and "all" will be built in parallel, unless for example the "all" target were to define that the "depend" target is a prerequisite, which it currently doesn't.

Is the "macports" target needed? The Portfile doesn't specify a target so it uses the "all" target.

ryandesign (commented):
The "Files" directory needs to be renamed to "files" or else the build will fail on case-sensitive filesystems (which our buildbot machines use, as well as some MacPorts users).

multimedia/M2VDownsizer/Portfile
RobK88: I do not know if this change will work.  I have to see if macports can find the source  after this change.  It was stored under the files sub-directory of the website.

multimedia/M2VDownsizer/Files/Makefile
RobK88: Yes.  You are right! :-)

multimedia/M2VDownsizer/Files/Makefile
RobK88: No reason other than I was having more trouble with gcc early on with this Project.  It could safely be changed to CC := cc if you so desire.

multimedia/M2VDownsizer/Files/Makefile
RobK88: Yes, the `macports` target works.  But if you use it, one will need to add to the Portfile:

```
depends_build       port:makedepend
build.target            macports
```

I elected not to bother running `make depend` since the generation of all the header dependencies are really not needed to build M2VDownsizer from scratch.

Of course, if you were a developer trying to hack or change the source code for M2VDownsizer, then running `make depend` makes sense.  Since running `make` or 'make all` would only re-compile the minimum number of source code files after one has changed a header file (or other file).

P.S.  Changing the build target from `all` to `macports` will not very much time to the build time in Macports, given the size of the project.

multimedia/M2VDownsizer/Portfile
RobK88: I will  have to try that first.  I did try `-Wno-unused` but that did not suppress the warnings.  

If `-Wno-unused-command-line-argument` does the trick, it should definitely be added to CFLAGS.  I will try it tomorrow.

multimedia/M2VDownsizer/Files/1003-Patch-config.h.diff
RobK88: Yes, it is overkill as I mentioned in one of my posts to the dev mailing list.  The first include guard is probably all that is needed.  But all the other include guards will not slow down the execution of the final binary given how the CPP works.

multimedia/M2VDownsizer/Portfile
RobK88: Ryan I do not understand this proposed change.  The download location is:
https://web.archive.org/web/20120512092147if_/http://www.metakine.com/files/M2VDownsizer.tgz

How could the port download the source from the above download location of the `/` is removed from ${master_sites}?

multimedia/M2VDownsizer/Portfile
RobK88: Thank-you Ryan!  I added `-Wno-unused-command-line-argument` to CFLAGS in the Makefile and it did eliminate all those pesky clang warnings!! 

So I propose that the following be changed in the Makefile:

`CFLAGS += -Wall -g`

be changed to:

`CFLAGS += -Wall -g -Wno-unused-command-line-argument`

And the comment in the Portfile regarding parallel builds should be removed.

multimedia/M2VDownsizer/Files/Makefile
RobK88: Ryan, If you are worried that others might get confused regarding the presence of the `macports` target in the Makefile, we could change the name of this target from `macports` to something like `dev-first-build`.  Only two changes would be needed.  In the .PHONY statement and in the target definition itself (and in the comment above).

Or we could just use the `macports` target in the port.  But that would add a build dependency.

multimedia/M2VDownsizer/Files/Makefile
RobK88: Verified.  I changed `CC := clang` to `CC := clang` in the Makefile and M2VDownsizer built successfully.  

The binary builds fine using Macports or using the Makefile outside the Macports build environment.

multimedia/M2VDownsizer/Files/1003-Patch-config.h.diff
RobK88: Ryan, I cleaned up `config.h` and eliminated most of the include guards.  M2VDownsizer build just fine.

See https://pastebin.com/K5ian88P

And I finally found the bug in config.h.  HAVE_STDINT_H was defined twice in config.h.   Once I cleaned up the original formatting, it was easier to spot.  I do not know how I missed it the first time!

I recommend that we go with the updated config.h.    I could EMAIL the file to you or you could just use the pastbin.  I will likely be away from my development computer for a few days so pushing a new update to the PR may be difficult.

RobK88: Here is the pastebin for the updated file - 1003-Patch-config.h.diff

https://pastebin.com/SFtxvnUu

multimedia/M2VDownsizer/Files/M2VDownsizer.1
RobK88: P.S.  I have also updated the man page for the sister port - `M2VRequantiser`.  I will submit it via a PR next week and will make the same change.

multimedia/M2VDownsizer/Portfile
RobK88: Yes.  Since I included `platforms` on past ports, I included it here.  If MacPorts no longer wants `platforms  darwin`, I have no problem with dropping it.

multimedia/M2VDownsizer/Files/Makefile
RobK88: It is whether a question whether the compiler (e.g clang) can find the headers for the macOS framework headers without them being explicitly specified.   On my Mac, I could not get the code to compile without specifying the location of the macOS framework headers in INCLUDES in the Makefile, if my memory serves me right.  (Please note that you do not need to specify the location of these headers when using XCode.  You just need to tell XCode that you want your code to link against a certain framework.  But that does not appear to be the case when you are not using XCode).   

INHO,  Apple wants developers to use XCode when using their frameworks.  

If you would like me to do some more testing, it will need to wait a couple of days util I get back home.

multimedia/M2VDownsizer/Files/1003-Patch-config.h.diff
ryandesign: You may use whatever filenames and capitalizations within the files directory that you wish, provided you reference them from the Portfile with the same names and capitalizations.

multimedia/M2VDownsizer/Portfile
ryandesign: On the preceding line you defined `homepage` to end with a slash:
```
homepage            https://web.archive.org/web/20120508050929if_/http://www.metakine.com/
```
Your proposed `master_sites` were:
```
master_sites        ${homepage}/files
```
After variable expansion that means `master_sites` will be set to `https://web.archive.org/web/20120508050929if_/http://www.metakine.com//files` which works but has an unnecessarily doubled slash before "files". My suggestion eliminates the unnecessary extra slash.

multimedia/M2VDownsizer/Files/Makefile
ryandesign: > I changed `CC := clang` to `CC := cc` in the Makefile

You can do that if you wish but it is pointless since `make` sets CC to `cc` by default.

multimedia/M2VDownsizer/Files/Makefile
ryandesign: > Yes, the `macports` target works.

You are certain it does what you want? I assume you wanted `make depend` to finish running before `make all` starts to run but I don't think that is guaranteed to be the case if parallel building is happening.

> Or we could just use the `macports` target in the port. But that would add a build dependency.

Since it works when using just the `all` target I would leave it that way, and in that case I'm not sure that the `macports` target needs to exist at all.

multimedia/M2VDownsizer/Files/Makefile
ryandesign: A zillion ports are able to link with macOS frameworks just fine without using Xcode and I've never seen one use an absolute path to the framework headers like that.

multimedia/M2VDownsizer/Files/Makefile
RobK88: Thanks Ryan.  Whenever I updated the port to use the `macports` build target (instead of the default `all` build target), I noticed that Macports always finished the `make depend` target before starting the `make all` target.  Perhaps, that was caused by the design of `make` which normally will not start the next target until the previous target was finished.  Or maybe it was just a fluke.

**In any event, probably the easiest resolution is to delete the `macports` target and the reference to `macports` from `.PHONY` from the Makefile.**   It may be too confusing for others.

This approach would be consistent with my earlier decision to omit the `make install` target from the Makefile since it was not needed for the Macports port package.

multimedia/M2VDownsizer/Portfile
RobK88: Please note the flag `"-Wno-unused-command-line-argument"` appears to be only available when using `clang`. This flag does not appear to be available when using `gcc`.

So adding the following to the Portfile is preferable since it may be possible to restrict its use to certain compilers:

`configure.cflags-append      "-Wno-unused-command-line-argument"`

multimedia/M2VDownsizer/Files/1003-Patch-config.h.diff
RobK88: This issue has been resolved in my last push.  The patch file - **1003-Patch-config.h.diff** - has been revised.  All the unnecessary include guards have been removed from config.h

multimedia/M2VDownsizer/Portfile
RobK88: This issue has been resolved in my last push.  The flag **-Wno-unused-command-line-argument** has been added to the Makefile.

multimedia/M2VDownsizer/Files/Makefile
RobK88: I changed `CC := clang` to `CC := cc` in my last push even though as pointed out it really does not make any difference since MacPorts will over-ride `CC:`  anyway.

multimedia/M2VDownsizer/Files/Makefile
RobK88: This issue has been resolved in my last push.  I removed the `macport` target from the Makefile since it is not needed and my just cause confusion to others reading the Makefile.

multimedia/M2VDownsizer/Files/Makefile
RobK88: This has been resolved in the my last push.  I removed the comment and added a `make install` target to the Makefile.

multimedia/M2VDownsizer/Files/Makefile
RobK88: This has been resolved in my last push.  I removed `-I/System/Library/Frameworks/CoreServices.framework/Headers` from `INCLUDES`.  It is not needed to build the project.   And this deletion will help make the port more resilient to any unexpected changes to the location of these headers by Apple.

(But please note that this include is helpful for the `make depend` target to help properly build the list of header dependencies.  But these header dependencies are not needed to build the project via MacPorts).

multimedia/M2VDownsizer/Files/Makefile
```diff
@@ -0,0 +1,257 @@
...
```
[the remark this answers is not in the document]
RobK88: Issued Resolved.  I deleted the `macports` target in my last push.  As noted, it is not necessary to build the project and may confuse others reading the Makefile.

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,77 @@
...
+#
```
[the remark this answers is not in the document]
RobK88: Issued resolved.  I decided the add the `-Wno-unused-command-line-argument` flag to the Makefile not the Portfile.

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,66 @@
...
+post-extract {
+     copy ${filespath}/Makefile ${worksrcpath}
+     copy ${filespath}/${name}.1 ${worksrcpath}
+     file mkdir ${worksrcpath}/obj/utils
+     file mkdir ${worksrcpath}/obj/altivec
+     file mkdir ${worksrcpath}/obj/mpeg2enc
+     file mkdir ${worksrcpath}/obj/main
+}
+
+patchfiles          1001-Replace-MPProcessors-Patch.diff \
```
mascguy, line 58: I recommend using `patchfiles-append`, rather than `patchfiles`. Reason being, the latter replaces/removes any patches that have already been added.

For a case like this, where you only have one `patchfiles` statement, it's not a hugh deal. But it's still good practice, and can save time (and hair-pulling) when diagnosing portfile issues.

Ditto for things like `depends_xxx`: It's important to use `depends_build-append`, `depends_lib-append`, etc, to prevent wiping of any existing things added by macports base, and/or portgroups!

mascguy: > Ditto for things like `depends_xxx`: It's important to use `depends_build-append`, `depends_lib-append`, etc, to prevent wiping of any existing things added by macports base, and/or portgroups!

p.s. I realize you don't have any of these statements presently, but just an FYI for future work!

RobK88: Thanks Christopher for the tips!  I am leaning something new every port I work on.

I tested the portfile using "patchfiles-append" in my local repo and it worked without any issues.
Do you want me to change `patchfiles` to `patchfiles-append` for this port and push the change?

mascguy: > Thanks Christopher for the tips! I am leaning something new every port I work on.
> 
> I tested the portfile using "patchfiles-append" in my local repo and it worked without any issues. Do you want me to change `patchfiles` to `patchfiles-append` for this port and push the change?

Sure, go ahead and push!

RobK88: The push is done!

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,66 @@
...
+
+checksums           rmd160 d662bfe195b59f5815bfdccaf3f81dae46d4e0d8 \
+                    sha256 316a931810116d70a07cb57a80ae8483fa949bd575beee222a9a8e375c4d8590 \
+                    size 371928
+
+distname            ${name}
+
+dist_subdir         ${name}/${version}
+
+extract.suffix      .tgz
```
mascguy, line 47: Since these three are all closely-related, you can group them together with no blank lines between.

multimedia/M2VDownsizer/Portfile
```diff
@@ -0,0 +1,66 @@
...
+                    first tries transcoding MPEG-2 streams using M2VRequantiser since \
+                    transcoding is a lot faster and produces excellent results for most \
+                    MPEG-2 video streams.  ${name} is only intended to be used for \
+                    large MPEG-2 video streams when transcoding does not produce \
+                    acceptable results.
+
+homepage            https://web.archive.org/web/20120508050929if_/http://www.metakine.com/
+master_sites        ${homepage}files
+
+depends_lib         port:libmpeg2
```
mascguy, line 37: Change this to `depends_lib-append`

multimedia/M2VDownsizer/Portfile
mascguy: > Since these three are all closely-related, you can group them together with no blank lines between.

I'm specifically referring to `distname`, `dist_subdir`, and `extract.suffix`.

multimedia/M2VDownsizer/Portfile
RobK88: Will do!

mascguy (approved):
Tested locally on a few older macOS releases - 10.9, 10.11, 10.13, and 10.15 - and looks good.

Great work, and thank you so much for the contribution!

mascguy:
Just a quick heads-up: I had to make a minor change to the preprocessor check for ARM, though it's easy enough to miss.

One tip: If you're adding additional platform checks - and you're not running on that hardware/platform - temporarily move them to the top of the list. That way they'll be evaluated/validated.

RobK88:
Thanks for catching the small bug in config.h.  I am glad to see that it builds and runs on the older macOS releases.  I tested it here locally on my old Mac Pro running Lion 10.7.  It seems to build and run fine on that old macOS release as well.

PR #15649, opened 2022-08-07
flint: update to 2.9.0 and move to Cmake build system

barracuda156:
#### Description

1. Update to 2.9.0.
2. Move to Cmake build system.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/math/flint/Portfile
+++ b/math/flint/Portfile
@@ -4,21 +4,20 @@ PortSystem              1.0
 PortGroup               muniversal 1.0
 
 name                    flint
-version                 2.8.0
+version                 2.9.0
 revision                0
 categories              math devel
 license                 LGPL-2.1+
 maintainers             {mcalhoun @MarcusCalhoun-Lopez} openmaintainer
-platforms               darwin
 description             Fast Library for Number Theory
 long_description        FLINT is a C library for doing number theory
 
 homepage                http://www.flintlib.org
 master_sites            ${homepage}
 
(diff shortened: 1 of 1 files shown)


herbygillot:
If the commit that moves `flint` to make produces any changes in the final built assets (new or additional files), then `revision` should be incremented in that commit.

barracuda156:
> If the commit that moves `flint` to CMake produces any changes in the final built assets (new or additional files), then `revision` should be incremented in that commit.

But is it ok that we jump then from 2.8.0 rev. 0 to 2.9.0 rev. 1?

herbygillot:
You have two options:

**Option 1:** Combine the version bump and the move to CMake into a single commit.

The commit message would be `flint: update to 2.9.0`, and you would mention the migration to CMake in the commit message body.  If you go this route, you do not need to touch `revision` since `version` is being updated.

**Option 2:** Keep the version update and CMake move as two _separate_ commits.

In this case, you would do as I describe in my earlier comment. Going from 2.9.0 to 2.9.1 rev 1 is totally fine.

The most important thing is to signal to MacPorts to update a port whenever a given commit introduces a change that ultimately changes the contents of the port.

When doing so in a commit that introduces changes without changing `version`, like the standalone CMake move does, then in that situation `revision` _must_ be incremented.

PR #15653, opened 2022-08-07
cctools: fix build on macOS 10.5, bootstrap subport

catap:
#### Description

It also fix build on macOS 10.5, add variant bootstrap, and enable +bootstrap as default variant on bootstrap subport.

Thus, it's the third PR in series which allows to bootstrap ld64-latest on macOS 10.5 as simple as:
```
macos-leopard:~ catap$ sudo port installed
  clang-11-bootstrap @11.1.0_1+emulated_tls (active)
  gcc10-bootstrap @10.3.0_5+universal (active)
macos-leopard:~ catap$ sudo port install ld64
--->  Computing dependencies for ld64
 cctools-bootstrap
 cmake-bootstrap
 ld64-latest
 legacy-support
 libblocksruntime
 libcxx
 libmacho-headers
 libtapi
 libunwind-headers
 python27-bootstrap
 xz-bootstrap
Continue? [Y/n]:
```

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

macOS 12.5 21G72 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/cctools/Portfile
+++ b/devel/cctools/Portfile
@@ -10,9 +10,11 @@ version                 949.0.1
 set ld64_version        530
 revision                2
 
+subport ${name}-bootstrap { revision 0 }
+
 categories              devel
 platforms               darwin
-maintainers             {jeremyhu @jeremyhu} openmaintainer
+maintainers             {jeremyhu @jeremyhu} {@catap korins.ky:kirill} openmaintainer
 license                 {APSL-2 GPL-2+}
 installs_libs           no
 description             Compiler Tools for Mac OS X and Darwin
@@ -46,7 +48,10 @@ patchfiles              cctools-829-lto.patch \
                         cctools-949-build-with-SDKs-older-than-1012.diff \
                         cctools-949-nm-allow-no-lto-support.diff \
                         cctools-949-write_object-fix.diff \
+                        cctools-949-missed-stdio.diff \
                         cctools-949-lipo-segalign-log2fix.diff
+
+set install_prefix      ${prefix}
 
 # lipo-segalign-log2fix https://trac.macports.org/ticket/63164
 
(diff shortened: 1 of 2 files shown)


mascguy:
In terms of ultimately merging this, let's ensure both @jmroot and @kencu review and approve first.

Going forward, this is something we should also do for all foundational ports like this.

catap:
@mascguy agreed. I do have a queue of such ports which allows me too build clan-14 and gcc12 on 10.5 with without any actions like rebuilding libcxx or deactivating some ports which is conflicting during the process.

kencu:
this will have to be built on every system, including universal, including Tiger -> wherever cctools +xcode starts. 

the links will have to be tested. LTO will have to be tested. 

unfortunately running the cctools test suite is hard...

kencu:
as there is a group of people doing this now, you can engage everyone who has an interest.

In previous years I would have done this all myself, but -- no more. And I can't commit anything on anyone's "say so" without documentation of actual building (and ideally testing) as you have seen, there is no patience for breaking core toolchain ports.

The libcxx was a trivial thing, and the fix was trivial, but it was not caught prior to pushing.

This is a wholesale difference in how this port is built, so will need that level of validation.

catap:
@kencu I see you point. Anyway, this PR is adding a subport which I will use in the next PRs.

The test that I've made: `cctools-bootstrap` -> `libtapi` -> `ld64-latest`

I've used this `ld64` to start building complicated software like `llvm-11`. Well, I can't say that I've archived that, but I'm near the end of the road where I have `clang-11` which is bootstrapet via a shortcut on 10.5

Gcenx:
To avoid headaches down the road I’d split cctools-bootstrap onto its own port locking it to the current version.

catap:
> To avoid headaches down the road I’d split cctools-bootstrap onto its own port locking it to the current version.

Very good idea. Also I may try to use https://github.com/iains/darwin-xtools instead

kencu:
I also think splitting cctools-bootstrap out into it’s own port makes more sense. Then it can be managed more easily and will be less confusing, less intrusive, and less fragile.

modifying the primary cctools port like this is too risky.

catap:
Ken, from another hand I'd like to use this opportunity to build toolchain based on xtools. Or at least to try it.

Let me mark it as draft for a while.

kencu:
xtools is right now being updated, so this is not a great moment to do that.

kencu:
xtools from Iain is mostly useful for his ppc port of llvm.

for all our intel systems, we should stick to ld64-latest, which is presently newer than xtools, and our cctools, which is also newer.

don’t PR a switch to xtools before a full discussion on macports-dev, please. At least, I won’t ever push it.

kencu:
also, we have already somewhat embraced in libtapi a different port of xtools, so some kind of coherent plan would be needed. We should focus on only one of those, for consistency, testing, etc.

catap:
Close it in favour of https://github.com/macports/macports-ports/pull/15684

PR #15660, opened 2022-08-08
clang-11-bootstrap, llvm-*: fix build on macOS 10.5

catap:
#### Description

On macOS 10.5 `__MAC_OS_X_VERSION_MAX_ALLOWED` has value `1060`, anyway, llvm uses different macros to detect availability of `pthread_setname_np`.

See: https://trac.macports.org/ticket/37418

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

###### Verification 
Have you

### DIFF
--- a/lang/clang-11-bootstrap/files/0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch
+++ b/lang/clang-11-bootstrap/files/0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch
@@ -12,19 +12,11 @@ diff --git llvm_master/lib/Support/Unix/Threading.inc macports_master/lib/Suppor
 index ed9a9656305..e8f9a13860f 100644
 --- llvm_master/lib/Support/Unix/Threading.inc
 +++ macports_master/lib/Support/Unix/Threading.inc
-@@ -14,6 +14,7 @@
- #include "llvm/ADT/Twine.h"
- 
- #if defined(__APPLE__)
-+#include <Availability.h>
- #include <mach/mach_init.h>
- #include <mach/mach_port.h>
- #endif
 @@ -153,8 +154,10 @@ void llvm::set_thread_name(const Twine &Name) {
    ::pthread_setname_np(::pthread_self(), "%s",
      const_cast<char *>(NameStr.data()));
  #elif defined(__APPLE__)
-+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
++#if HAVE_PTHREAD_SETNAME_NP
    ::pthread_setname_np(NameStr.data());
  #endif
 +#endif
--- a/lang/llvm-10/files/0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch
+++ b/lang/llvm-10/files/0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch
@@ -12,19 +12,11 @@ diff --git llvm_master/lib/Support/Unix/Threading.inc macports_master/lib/Suppor
 index ed9a9656305..e8f9a13860f 100644
 --- llvm_master/lib/Support/Unix/Threading.inc
 +++ macports_master/lib/Support/Unix/Threading.inc
-@@ -14,6 +14,7 @@
- #include "llvm/ADT/Twine.h"
- 
- #if defined(__APPLE__)
-+#include <Availability.h>
- #include <mach/mach_init.h>
- #include <mach/mach_port.h>
- #endif
 @@ -153,8 +154,10 @@ void llvm::set_thread_name(const Twine &Name) {
    ::pthread_setname_np(::pthread_self(), "%s",
      const_cast<char *>(NameStr.data()));
  #elif defined(__APPLE__)
-+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
++#if HAVE_PTHREAD_SETNAME_NP
    ::pthread_setname_np(NameStr.data());
  #endif
 +#endif
--- a/lang/llvm-11/files/0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch
+++ b/lang/llvm-11/files/0005-Threading-Only-call-pthread_setname_np-on-SnowLeopar.patch
@@ -12,19 +12,11 @@ diff --git llvm_master/lib/Support/Unix/Threading.inc macports_master/lib/Suppor
 index ed9a9656305..e8f9a13860f 100644
 --- llvm_master/lib/Support/Unix/Threading.inc
 +++ macports_master/lib/Support/Unix/Threading.inc
-@@ -14,6 +14,7 @@
- #include "llvm/ADT/Twine.h"
- 
- #if defined(__APPLE__)
-+#include <Availability.h>
- #include <mach/mach_init.h>
- #include <mach/mach_port.h>
- #endif
 @@ -153,8 +154,10 @@ void llvm::set_thread_name(const Twine &Name) {
    ::pthread_setname_np(::pthread_self(), "%s",
      const_cast<char *>(NameStr.data()));
  #elif defined(__APPLE__)
-+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 1060
++#if HAVE_PTHREAD_SETNAME_NP
    ::pthread_setname_np(NameStr.data());
  #endif
 +#endif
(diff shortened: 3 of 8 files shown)


catap:
The root cause of issue is this pice of code inside `/usr/include/available.h`:
```c
#ifndef __MAC_OS_X_VERSION_MAX_ALLOWED
    #if __MAC_OS_X_VERSION_MIN_REQUIRED > __MAC_OS_X_VERSION_10_5
        #define __MAC_OS_X_VERSION_MAX_ALLOWED __MAC_OS_X_VERSION_MIN_REQUIRED
    #else
        #define __MAC_OS_X_VERSION_MAX_ALLOWED __MAC_OS_X_VERSION_10_5
    #endif
#endif
```
which is works as:
```
macos-leopard:tmp catap$ cat test.cc 
#include <Availability.h>
#include <iostream>
using namespace std;
int main() {
	cout << "__MAC_OS_X_VERSION_MAX_ALLOWED: " << __MAC_OS_X_VERSION_MAX_ALLOWED << endl;
	cout << "__MAC_OS_X_VERSION_10_5: " << __MAC_OS_X_VERSION_10_5 << endl;
	cout << "__MAC_OS_X_VERSION_MIN_REQUIRED: " << __MAC_OS_X_VERSION_MIN_REQUIRED << endl;
        return 0;
}
macos-leopard:tmp catap$ g++ test.cc
macos-leopard:tmp catap$ ./a.out 
__MAC_OS_X_VERSION_MAX_ALLOWED: 1060
__MAC_OS_X_VERSION_10_5: 1050
__MAC_OS_X_VERSION_MIN_REQUIRED: 1058
macos-leopard:tmp catap$ 
```
on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

kencu:
It’s a known bug in the 10.5 SDK… one of several.

MacPorts has always recommended users follow this:

https://trac.macports.org/wiki/LeopardSDKFixes

which fixes this issue I believe.

catap:
@kencu yeah, but here is clear more direct approach since llvm has a macros which detects aviabillity of `pthread_setname_np` :)

kencu:
still, you should modify your sdk if you have not….everyone else has/should.

kencu:
you’re using getname to cover off setname. IDK…. is this 100% the case that testing for getname tests properly for setname?

I would leave it as it is, and fix your sdk as macports requires, myself.

catap:
@kencu thanks, this is a copy-and-paste typo (and lack of morning coffee). The correct macros name is `HAVE_PTHREAD_SETNAME_NP`.

catap:
> Otherwise you will stumble across many many other similar unnecessary “errors” too.

depends on the case. But a file `Threading.inc` is full of `HAVE_PTHREAD_SETNAME_NP` and it seems natural to use such macros without enforcing user to edit systemfiles until it is necessary

kencu:
> > Otherwise you will stumble across many many other similar unnecessary “errors” too.
> 
> depends on the case. 

don’t think so… the sdk is wrong. There are many places where this needs to be right.

> But a file `Threading.inc` is full of `HAVE_PTHREAD_SETNAME_NP` and it seems natural to use such macros without enforcing user to edit systemfiles until it is necessary

users **must** fix their sdk anyway, otherwise horrors will happen like this in many other places.

but, sure, this test came into llvm-5.0, long after we did the sdk test for it, so could use it instead. Makes no difference, if sdk is properly fixed.

I know, you didn’t know about broken sdk, it’s ok. Now you do, you can fix your system, and tell others.

catap:
> I know, you didn’t know about broken sdk, it’s ok. Now you do, you can fix your system, and tell others.

and make it a kind of hidden knowledge. I try to avoid as many hidden knowledge as possible :)

kencu:
also, see 

https://github.com/macports/macports-legacy-support/commit/8af62cfd2c9b97990df91400a6dcbfbd08ab97e0

where marcus added only one of these to legacysupport…

kencu:
> > I know, you didn’t know about broken sdk, it’s ok. Now you do, you can fix your system, and tell others.
> 
> and make it a kind of hidden knowledge.

I guess that’s what I’m still here for…. unhiding knowledge :)

catap:
> also, see
> 
> [macports/macports-legacy-support@8af62cf](https://github.com/macports/macports-legacy-support/commit/8af62cfd2c9b97990df91400a6dcbfbd08ab97e0)
> 
> where marcus added only one of these to legacysupport…

FYI `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` has a correct value on untached system :)

catap:
So, moving from `__MAC_OS_X_VERSION_MAX_ALLOWED >= 1060` to `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1060` inside MacPorts patches seems the right way at least.

kencu:
> > also, see
> > [macports/macports-legacy-support@8af62cf](https://github.com/macports/macports-legacy-support/commit/8af62cfd2c9b97990df91400a6dcbfbd08ab97e0)
> > where marcus added only one of these to legacysupport…
> 
> FYI `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__` has a correct value on untached system :)

sure, that is set by the compiler, nothing to do with sdk.

for this use though, it’s not the right macro to use, as we want sdk features test.

kencu:
> So, moving from `__MAC_OS_X_VERSION_MAX_ALLOWED >= 1060` to `__ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__ >= 1060` inside MacPorts patches seems the right way at least.

please don’t do that. it will break many other patches that have done fallback functions properly already.

just please fix your broken sdk…

barracuda156:
> It’s a known bug in the 10.5 SDK… one of several.
> MacPorts has always recommended users follow this:
> https://trac.macports.org/wiki/LeopardSDKFixes
> which fixes this issue I believe.

That approach gonna break `ppc64`: https://trac.macports.org/ticket/64397

kencu:
> > It’s a known bug in the 10.5 SDK… one of several.
> > MacPorts has always recommended users follow this:
> > https://trac.macports.org/wiki/LeopardSDKFixes
> > which fixes this issue I believe.
> 
> That approach gonna break `ppc64`: https://trac.macports.org/ticket/64397

not on my systems it didn’t.

please fix broken sdk, as Jeremy said for many years now, and (almost) all of us have done…

kencu:
@barracuda156 … I recall 5 or 7 years ago compiling a new libgcc_s.10.5.dylib with all 4 archs in it, of the right version, and using that. IIRC to do that I did a universal build of apple-gcc42 on Intel and ppc, and lipo’d the library together manually. It had the needed 128 bit functions in it, I believe.

All this was years ago now, so it’s a bit fuzzy now.

barracuda156:
> @barracuda156 … I recall 5 or 7 years ago compiling a new libgcc_s.10.5.dylib with all 4 archs in it, of the right version, and using that. IIRC to do that I did a universal build of apple-gcc42 on Intel and ppc, and lipo’d the library together manually. It had the needed 128 bit functions in it, I believe.

 If you recall an outline of the procedure or have some code for that left, it is very interesting.

barracuda156:
> edit: oh, you’re talking about the ppc64 code in the libs… yes, different issue, indeed, but on that page.

Yeah, I added it there back then.

kencu:
> > @barracuda156 … I recall 5 or 7 years ago compiling a new libgcc_s.10.5.dylib with all 4 archs in it, of the right version, and using that. IIRC to do that I did a universal build of apple-gcc42 on Intel and ppc, and lipo’d the library together manually. It had the needed 128 bit functions in it, I believe.
> 
> If you recall an outline of the procedure or have some code for that left, it is very interesting.

that is the outline :)

kencu:
this change to use llvm’s internal mechanism is optional, now that catap fixed his sdk properly.

I haven’t tested it, but I think at least one of the functions will now be found in legacysupport and linked agaist that was previously blocked. so the llvms on < 10.6 will be trivially different. Some would say that requires revbumping them all…. technically true. on the other hand the function is basically a no-op anyway and likely is optimized out.

other than that, this could be pushed, or not. It won’t fix anything, but seems somewhat more tidy to use the llvm test that was added later, long after we first added the sdk blocker (before the test existed).

PR #15678, opened 2022-08-09
google-glog: fix build on old systems using gcc

barracuda156:
#### Description

Changes concern older systems <10 plus 10.6 PPC/Rosetta. Newer MacOS that use Clang aren’t affected.
For `google-glog` port:
1. GCC earlier than `gcc-4.6` are blacklisted, because they do not support `#pragma GCC diagnostic` inside functions.
2. For Rosetta case, whitelist of GCC versions is added, because it ignores blacklist and still tries to use `llvm-g++-4.2`, which fails (blacklisting it explicitly fails to work).
3. When GCC is used as a compiler, `-latomic` ld flag is added, otherwise build fails on undefined atomics symbols.
4. Finally, identical compiler choice settings (not relevant for Clang-using systems) are added to `gflags` port, because while `gflags` itself does build fine with `gcc-4.2`, `google-glog` build then fails with undefined symbols `__GLOBAL__sub_I_logging.cc in logging.cc.o` and ` __GLOBAL__sub_I_vlog_is_on.cc in vlog_is_on.cc.o`. Apparently identical compiler is to be used. Similar issue: https://github.com/gflags/gflags/issues/203
Once `gflags` is built with the same compiler as used for `google-glog`, the latter builds fine too.
```
10:~ svacchanda$ port -v installed gflags
  gflags @2.2.2_0 (active) requested_variants='' platform='darwin 10' archs='ppc' date='2022-08-10T00:26:32+0545'
10:~ svacchanda$ port -v installed google-glog
  google-glog @0.6.0_0+gflags (active) requested_variants='' platform='darwin 10' archs='ppc' date='2022-08-10T00:49:00+0545'
```
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/devel/gflags/Portfile
+++ b/devel/gflags/Portfile
@@ -39,4 +39,13 @@ platform darwin 10 powerpc {
 configure.args-append \
         -DBUILD_SHARED_LIBS=ON \
         -DBUILD_STATIC_LIBS=ON \
+        -DGFLAGS_BUILD_gflags_LIB=ON \
         -DREGISTER_INSTALL_PREFIX=OFF
+
+platform darwin {
+    # Needed to support building dependents. See: https://kumasento.github.io/2020-06-12-glog-gflags-and-c-abi/
+    if {${os.major} < 16} {
+        configure.args-append \
+                    -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=0"
+    }
+}
--- a/devel/gflags/Portfile
+++ b/devel/gflags/Portfile
@@ -5,7 +5,7 @@ PortGroup           github 1.0
 PortGroup           cmake 1.1
 
 github.setup        gflags gflags 2.2.2 v
-revision            0
+revision            1
 
 categories          devel
 maintainers         nomaintainer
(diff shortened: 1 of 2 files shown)


devel/gflags/Portfile
```diff
@@ -5,26 +5,47 @@ PortGroup           github 1.0
...
 maintainers         nomaintainer
 
 description         commandline flags module for C++
 long_description    The gflags package contains a library that implements commandline flags processing. \
-                    As such it's a replacement for getopt(). It has increased flexibility, \
+                    As such it’s a replacement for getopt(). It has increased flexibility, \
                     including built-in support for C++ types like string, and \
                     the ability to define flags in the source file in which \
-                    they're used.
+                    they’re used.
```
mascguy, line 18: Sergey, don't change single quotes to these characters. The standard ones are clean single-byte ASCII characters, but the curly variants are multi-byte.

You can see the result by viewing the file via `less`:

```
long_description    The gflags package contains a library that implements commandline flags processing. \
                    As such it<E2><80><99>s a replacement for getopt(). It has increased flexibility, \
                    including built-in support for C++ types like string, and \
                    the ability to define flags in the source file in which \
                    they<E2><80><99>re used.
```

Please revert.

mascguy: In short, avoid using curly apostrophes anywhere in portfiles.

mascguy: Not a big deal though, as it doesn't cause any problems. So no worries if this change has been made elsewhere!

barracuda156: > Not a big deal though, as it doesn't cause any problems. So no worries if this change has been made elsewhere!

@mascguy I changed it for a reason, in BBEDit (and I suspect other code editors) it breaks code coloring, since the editor understand that as a quote character of the code. My change not just makes typography correct (a matter of æsthetics), but fixes code coloring in code editor.

barracuda156: @mascguy Dropped the commit for BBEdit code coloring fix.

barracuda156:
@mascguy Should be good to merge. We need this for `folly` and friends to build.

mascguy (approved):

PR #15687, opened 2022-08-10
*-mingw32-binutils: Update to 2.39

Gcenx:
#### Description

Modify `crossbinutils-1.0.tcl` to no longer rename infopages, just install into a subdirectory. This just seems avoids the current headaches we've encountered when updating binutils

*-mingw32-binutils were testing building wine-staging-7.14

###### Type(s)

- [x] enhancement

###### Tested on

macOS 11.6.7 20G630 arm64
Xcode 13.2.1 13C100

###### Verification 


mojca (approved):
I'm less familiar with the patches than the author, but this looks ok.

It doesn't need to be done now, but should we perhaps remove 2.38 from the list, to potentially prevent other cross binutils from using it unpatched?

Gcenx:
I’m not in-favor of removing versions of binutils just because they may require patches, this could also be the case for crossgcc.

Adding notes to versions that require patches including a url to said patch, possibly having this information pushed to the port?

It would be nice if there were a way to automatically provide known required patches into affected versions.

PR #15696, opened 2022-08-11
unxip: submission

i0ntempest:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.5 21G72 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/archivers/unxip/Portfile
+++ b/archivers/unxip/Portfile
@@ -1,0 +1,55 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           xcodeversion 1.0
+
+github.setup        saagarjha unxip 1.1.1 v
+revision            0
+github.tarball_from archive
+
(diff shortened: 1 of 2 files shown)


i0ntempest:
@saagarjha Since this is your code, do you want to take maintainership?

saagarjha:
Sorry, I was busy the last couple weeks, which are probably not the best circumstances to come back from to mention I'd be willing to take maintainership :) I should probably upstream your patch so you don't have to carry it here. I'd also like to point out that in the interests of being able to bootstrap a full Xcode installation, it is kind of useful to not have this port depend on Xcode itself. Can we switch the Portfile to using swiftc directly, which let it be installed via the command line tools?

i0ntempest:
> Sorry, I was busy the last couple weeks, which are probably not the best circumstances to come back from to mention I'd be willing to take maintainership :) I should probably upstream your patch so you don't have to carry it here. I'd also like to point out that in the interests of being able to bootstrap a full Xcode installation, it is kind of useful to not have this port depend on Xcode itself. Can we switch the Portfile to using swiftc directly, which let it be installed via the command line tools?

No prob, you can add youself in in the next commit. I'm not an swift expert, this port was actually mostly copied from `macos-trash` port, and that port required Xcode (and made universal variant easy). You may change it by all means.

PR #15716, opened 2022-08-14
afscompress: enable LZVN/LZFSE support

catap:
#### Description

Support of this compression algorithms was made via required submodule.

Unfortunately GitHub PG doesn’t support such feature :( So I implemented it by hand.

###### Type(s)
- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.5 21G72 x86_64
Xcode 13.4.1 13F100

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

###### Verification
Have you

### DIFF
--- a/sysutils/afscompress/Portfile
+++ b/sysutils/afscompress/Portfile
@@ -5,8 +5,10 @@ PortGroup            cmake 1.1
 PortGroup            github 1.0
 
 github.setup         RJVB afsctool 1.7.2
+set lzfse_version    e634ca58b4821d9f3d560cdc6df5dec02ffc93fd
 
 name                 afscompress
+revision             1
 categories           sysutils
 platforms            darwin
 license              GPL-3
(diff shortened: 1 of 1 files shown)


ylluminarious:
@catap Thanks, I don't see any problems in this change. Looks good to me 👍

sysutils/afscompress/Portfile
```diff
@@ -45,6 +55,11 @@ pre-fetch {
     }
```
catap, line 55: @ylluminarious are you sure that it won't work before 10.6?

I see that it has some code to make it works: https://github.com/RJVB/afsctool/blob/master/src/afsctool.c#L73-L77

catap:
I've marked it as draft because I need some time to check it on macOS 10.5

catap:
It requires a lot of patches to be build on 10.5; and it also absolutely useless :)

let forgot about 10.5

sysutils/afscompress/Portfile
mascguy: > @ylluminarious are you sure that it won't work before 10.6?
> 
> I see that it has some code to make it works: https://github.com/RJVB/afsctool/blob/master/src/afsctool.c#L73-L77

I'll test on my old 10.5 machines - both Intel and PPC - when I can find some time.

For now though, let's go ahead and merge, as this is a nice addition. Thanks as always Kirill!

mascguy:
> It requires a lot of patches to be build on 10.5; and it also absolutely useless :)
> 
> let forgot about 10.5

Ah, I hadn't seen this before, when replying to the other comment. Works for me!  :-)

mascguy (approved):

PR #15735, opened 2022-08-15
ports: simplify via pg stub

mascguy:
### Description

For ports that define a stub subport, simplify via pg `stub`.

### DIFF
--- a/cross/mingw-w64/Portfile
+++ b/cross/mingw-w64/Portfile
@@ -150,16 +150,11 @@ if {${subport} ne ${name}} {
     destroot.violate_mtree  yes
     livecheck.type          none
 } else {
+    PortGroup               stub 1.0
+
+    supported_archs         noarch
     description             "GCC cross-compiler for Windows 64 & 32 bits (meta port)"
 
     depends_run             port:i686-${mingw_name}-gcc \
                             port:x86_64-${mingw_name}-gcc
-    distfiles
-    use_configure           no
-    build                   {}
-    destroot {
-        xinstall -d -m 755 "${destroot}${prefix}/share/doc/${name}"
-        system "echo $name is a stub port > ${destroot}${prefix}/share/doc/${name}/README"
-    }
-    supported_archs         noarch
 }
--- a/devel/qscintilla/Portfile
+++ b/devel/qscintilla/Portfile
@@ -122,17 +122,11 @@ foreach qt_major {4 5} {
     }
 
     subport py-${name}-qt${qt_major} {
+        PortGroup stub 1.0
+
+        supported_archs   noarch
         categories-append python
-        use_configure no
-        distfiles
-        patchfiles
-        supported_archs noarch
         depends_lib port:py27-qscintilla-qt${qt_major}
-        build {}
-        destroot {
-            xinstall -d -m 0755 ${destroot}${prefix}/share/doc/${subport}
-            system "echo $name is a stub port > ${destroot}${prefix}/share/doc/${subport}/README"
-        }
     }
 
     set python_versions {27 35 36 37 38 39 310}
@@ -224,17 +218,10 @@ foreach pver ${python_versions} {
 }
 
 if {${subport} eq ${name}} {
-    revision 2
+    PortGroup   stub 1.0
 
-    distfiles
-    patchfiles
-    use_configure no
+    revision    2
     supported_archs noarch
-    build {}
-    destroot {
-        xinstall -d -m 0755 ${destroot}${prefix}/share/doc/${subport}
-        system "echo $name is a stub port > ${destroot}${prefix}/share/doc/${subport}/README"
-    }
 
     variant qt4 description "build Qt4 version of ${name}" {
         depends_lib-append \
--- a/devel/ccls/Portfile
+++ b/devel/ccls/Portfile
@@ -1,9 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim: fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem                  1.0
-PortGroup                   cmake 1.1
 PortGroup                   github 1.0
-PortGroup                   compiler_blacklist_versions 1.0
 
 name                        ccls
 github.setup                MaskRay ccls 0.20201025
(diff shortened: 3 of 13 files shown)


breun:
Looks good to me for `openjdk-distributions`.

lpsinger (approved):

l2dy (approved):
LGTM

nerdling (approved):

michaelld (approved):
LGTM for Qscintilla

cross/mingw-w64/Portfile
```diff
@@ -150,16 +150,11 @@ if {${subport} ne ${name}} {
     destroot.violate_mtree  yes
     livecheck.type          none
 } else {
+    PortGroup               stub 1.0
+
+    stub.noarch             yes
```
mojca, line 155: Why is this not simply the default, and ports that actually need to depend on architecture would explicitly set this to `no` inestad? Isn't the default for most stub ports to just install a README file?

mascguy: > Why is this not simply the default, and ports that actually need to depend on architecture would explicitly set this to `no` inestad? Isn't the default for most stub ports to just install a README file?
> 
> Similar question for `install_libs`. I didn't explicitly set it to `false` because I never thought of it.

Per discussion with @jmroot, we felt it would be safer to use defaults that don't materially change how dependencies and such work. Does that make sense?

mascguy:
Reviewers: As an FYI, the previous version of pg `stub` didn't support a `noarch` option. Now that it does, I've updated these ports where it's currently applicable. (Though I haven't changed what you folks were doing before, so everything should match with what we already have today.)

devel/qscintilla/Portfile
```diff
@@ -224,17 +218,9 @@ foreach pver ${python_versions} {
 }
 
 if {${subport} eq ${name}} {
-    revision 2
+    PortGroup stub 1.0
 
-    distfiles
-    patchfiles
-    use_configure no
-    supported_archs noarch
```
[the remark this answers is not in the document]
mascguy: > It looks like I missed the `noarch` case here, I'll fix ASAP.

Fixed.

cross/mingw-w64/Portfile
mojca: Not too much if the portgroup is actually meant to handle ports that are fundamentally different from any other ports anyway.

We have just a handful of ports for now, so we can easily make sure that the variables are set correctly on all of them.

I'm not sure what's the exact intent of the portgroup, but my first impression is that it should handle any ports that install just a README file and maybe add a dependency or two. Nothing else. For those cases I would also expect reasonable defaults for this scenario, and an easy way out (setting a single variable to `true` or `false`) in case the port needs something else.

What I do find problematic are portgroups that provide 90 % of the needed functionality, but then set one fatal variable that cannot easily be reverted when you don't want that part. (One example that comes to mind is using the `python` PG for compiling some C++ software. They provide tons of nice helpers, but setting the configure/build to the default port behaviour is almost impossible. But that's different from this case.)

mascguy: > Not too much if the portgroup is actually meant to handle ports that are fundamentally different from any other ports anyway.
> 
> We have just a handful of ports for now, so we can easily make sure that the variables are set correctly on all of them.

For sure! And I'd be fine with enabling `noarch` by default, given that it's enabled for the vast majority of our stubs.

But we need consensus from everyone. So I'll let @jmroot and others chime in.

mascguy:
Observers: Let's keep this open for another week or two, in order to provide ample time for folks to review.

cross/mingw-w64/Portfile
jmroot: > I'm not sure what's the exact intent of the portgroup, but my first impression is that it should handle any ports that install just a README file and maybe add a dependency or two. Nothing else. For those cases I would also expect reasonable defaults for this scenario, and an easy way out (setting a single variable to `true` or `false`) in case the port needs something else.

The exact intent of the portgroup is an important question. There are two main types of stub port:
 1. The kind that exists for user convenience, depends on one or more other ports, and has no dependents. Examples include `gimp`, `texlive`, and all the `p5-*` stubs.
 2. The kind that stands in for another port and has dependents, like `msgpack`.

Type 1 can always be noarch. Type 2 _must not_ be noarch unless all its dependencies are also noarch, because arch checking will not work correctly for the dependents otherwise. (The indirection provided by this kind of stub breaks the rule that ports should depend directly on everything that they use directly.)

If the same portgroup is to be used for both types of stub port, then it must default to not-noarch. Forgetting to make a type 2 stub not-noarch leads to fairly serious breakage. Forgetting to make a stub noarch on the other hand has minimal consequences (two archives per OS version might unnecessarily be built.)

I don't see any benefit to having an extra portgroup option controlling whether the stub is noarch; you can just as easily write `supported_archs noarch` as `stub.noarch yes`.

I can't actually think of any situation where it would be useful to change `installs_libs` from its default value in a stub port. So `stub.libs` can also go.

mascguy: > The exact intent of the portgroup is an important question.

The goals of the pg are two-fold:
* Eliminate the copy-pasta related to stub creation, making it trivial to create on. (Kinda obvious, but...)
* Provide a formal semantic marker - akin to PG `obsolete` - ensuring we can quickly find such ports in a simple and full-proof manner.

> I don't see any benefit to having an extra portgroup option controlling whether the stub is noarch; you can just as easily write `supported_archs noarch` as `stub.noarch yes`.
> 
> I can't actually think of any situation where it would be useful to change `installs_libs` from its default value in a stub port. So `stub.libs` can also go.

IMHO, I prefer to abstract such details via the two formal PG options, `stub.noarch` and `stub.libs`: If there is ever a need/desire to refine the behavior of those, this ensures we can easily do so via the PG. And while I can't think of a concrete example at this point, there may be in the future.

So why not leave them in?

mascguy: > IMHO, I prefer to abstract such details via the two formal PG options, `stub.noarch` and `stub.libs`: If there is ever a need/desire to refine the behavior of those, this ensures we can easily do so via the PG. And while I can't think of a concrete example at this point, there may be in the future.

It might be useful to add a third option like `port.leaf` though, which provides a shortcut to setting `stub.noarch yes` and `stub.libs no`. Previously I had that, via `stub.lib_dep`... but `stub.leaf` might be a better name.

Thoughts?

jmroot: > So why not leave them in?

If you can't think of a single use for them, the better question is why add them in the first place?

mascguy: > > So why not leave them in?
> 
> If you can't think of a single use for them, the better question is why add them in the first place?

If folks don't think they're beneficial, I'm fine with eliminating options `stub.noarch` and `stub.libs`.

> I can't actually think of any situation where it would be useful to change `installs_libs` from its default value in a stub port. So `stub.libs` can also go.

So just to confirm, are you advocating that stubs never set `installs_libs no` any longer?

jmroot: > So just to confirm, are you advocating that stubs never set `installs_libs no` any longer?

Yes. In the cases where it would be OK, it has no effect, because the stub has no dependents or is noarch.

mascguy: > > So just to confirm, are you advocating that stubs never set `installs_libs no` any longer?
> 
> Yes. In the cases where it would be OK, it has no effect, because the stub has no dependents or is noarch.

Works for me. I'll update the PG, and reconcile this PR.

mascguy: > > > So just to confirm, are you advocating that stubs never set `installs_libs no` any longer?
> > 
> > Yes. In the cases where it would be OK, it has no effect, because the stub has no dependents or is noarch.
> 
> Works for me. I'll update the PG, and reconcile this PR.

Changes pushed.

mascguy:
Re-added `php`, as CI isn't going to succeed anyway.

mascguy:
Now that we have consensus regarding our direction - along with approvals from a number of maintainers - let's merge.

PR #15737, opened 2022-08-15
krew: add installation instructions

ajhall:
#### Description
I installed `krew` and had to look up the [setup guide](https://krew.sigs.k8s.io/docs/user-guide/setup/install/) to get it working as a `kubectl` plugin.

This change adds a note to help with initial setup.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.5 21G72 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/sysutils/krew/Portfile
+++ b/sysutils/krew/Portfile
@@ -49,3 +49,11 @@ destroot {
     xinstall -d ${destroot}${krew_doc_dir}
     copy {*}[glob ${worksrcpath}/docs/*] ${destroot}${krew_doc_dir}/
 }
+
+notes "
+To initialize ${name} as a kubectl plugin, run:
+
+    ${name} install krew
+
+Then add the \$HOME/.krew/bin directory to your PATH environment variable.
+"


herbygillot:
Thanks @ajhall.

Do you know if `krew install krew` copies the same `krew` binary from the MacPorts prefix (`/opt/local/bin`) into `$HOME/.krew/bin`, or is it downloading some completely new and different binary from the Internet into `$HOME/.krew/bin`?

Also, you do not have to bump `revision` since you are only making changes to `notes`.  Since these are not changes that would alter the files and config in the built package itself (like modifications to `configure.*`, `build.*`, and/or `destroot.*` would), the installed package is unaffected by this change, and as such, you don't need to increase this port's `revision`.

ajhall:
Thanks! I reverted the `revision` change.

Now that I look more closely at the [setup guide](https://krew.sigs.k8s.io/docs/user-guide/setup/install/), the `krew` executable is sort of an installer for another copy of itself. The guide never installs it into a common `bin` dir, it just extracts it and runs `./krew install krew` to set up `$HOME/.krew` and download another copy of the same `krew` binary from the internet, symlinked as `kubectl-krew` so it can be used as a subcommand of `kubectl`.

```bash
$ ls ~/.krew/bin/
kubectl-krew

$ readlink ~/.krew/bin/kubectl-krew
/Users/andy.hall/.krew/store/krew/v0.4.3/krew
```

That means that the `/opt/local/bin/krew` installed by MacPorts only gets used for initializing `$HOME/.krew`, which is where all the plugin binaries (including `krew`, which is itself a plugin) get installed.

herbygillot:
Makes sense.  I was asking as I was curious whether the `krew` provisioned in `$HOME` could have a version that differs from the one installed by MacPorts.

PR #15746, opened 2022-08-16
dnscontrol: update to 3.19.0

ajhall:
#### Description

Update to dnscontrol 3.19.0, which no longer hard-codes a version number into its source code. Instead, set version info at build time.

```bash
$ dnscontrol version
dnscontrol 3.19.0 (MacPorts)
```

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.5 21G72 arm64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/sysutils/dnscontrol/Portfile
+++ b/sysutils/dnscontrol/Portfile
@@ -3,11 +3,11 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/StackExchange/dnscontrol 3.18.1 v
+go.setup            github.com/StackExchange/dnscontrol 3.19.0 v
 
-checksums           rmd160  d718f8be18ec2d14d721b69e39964363dfe59ea8 \
-                    sha256  cfb0e74331ab27d43adc68f99f2ddb384c2d718869dd66a9b39c93351d73b620 \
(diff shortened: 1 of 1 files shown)


sysutils/dnscontrol/Portfile
```diff
@@ -29,7 +29,11 @@ maintainers         {@ajhall} \
 # Allow deps to fetched at build time
 build.env-delete    GO111MODULE=off GOPROXY=off
 
-build.args-append   -ldflags \"-s -w\"
+build.args-append   -ldflags \"-s -w \
+                                -X main.Version=${version} \
+                                -X main.BuildTime=[clock seconds] \
```
herbygillot, line 34: MacPorts prefers that build time fields usually be set to some constant value like or "unset", "", "unknown", or "n/a" for the sake of reproducible builds.  We still have this behavior in quite a few ports still (and also within the makefiles and build processes of many projects), but the preference is to reduce instances of setting build time dynamically across ports.

ajhall: Ah, makes sense. It works fine if that value isn't specified, so I just removed it.

PR #15748, opened 2022-08-16
bloaty: new port

nickdowell:
#### Description

Adds [bloaty](https://github.com/google/bloaty) - a size profiler for binaries.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.5 21G72 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/bloaty/Portfile
+++ b/devel/bloaty/Portfile
@@ -1,0 +1,24 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake   1.1
+PortGroup           github  1.0
+
+github.setup        google bloaty 1.1 v
+github.tarball_from releases
+revision            0
+
+description         A size profiler for binaries
+
+long_description    Ever wondered what's making your binary big? Bloaty will show you a size \
+                    profile of the binary so you can understand what's taking up space inside.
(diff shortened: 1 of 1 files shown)


devel/bloaty/Portfile
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           github 1.0
+
+name                bloaty
```
herbygillot, line 7: You can start your Portfile off with `github.setup`.

It will implicitly set `name`, `version`, `homepage`, and `master_sites`.

So the following line:
```
github.setup        google bloaty 1.1 v
```

...means you can then delete the `name`, `version`, `homepage` and `master_sites` lines.

nickdowell: Updated as suggested 👍

herbygillot:
Your changes were correct, but when you removed `github.tarball_from releases`, that invalidated the checksum, as the checksum you had there was for the `releases` bzip, and with that line removed, MacPorts was trying to apply that checksum for the `releases` bzip against the default, which is the default tarball.

I made some minor edits and restored `github.tarball_from releases` and re-added `use_bzip2 yes`.

Thank you for your submission, @nickdowell.

herbygillot (approved):

PR #15750, opened 2022-08-16
pugixml: add cxx 2011 standard

barracuda156:
#### Description

The build fails with `gcc-4.2` with:
```
CMake Error in CMakeLists.txt:
  Target "pugixml-shared" requires the language dialect "CXX11" (with
  compiler extensions).  But the current compiler "GNU" does not support
  this, or CMake does not know the flags to enable it.
```

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/textproc/pugixml/Portfile
+++ b/textproc/pugixml/Portfile
@@ -21,4 +21,6 @@ description         C++ Library for XML processing
 long_description    C++ library for creating and manipulating XML DOMs. Features: \
                     efficient parsing of XML, XPath 1.0 support, full Unicode support.
 
+compiler.cxx_standard 2011
+
 configure.args-append -DBUILD_SHARED_LIBS=ON


kencu:
leave this one for @ryandesign , who specifically removed this a few months ago.

barracuda156:
> leave this one for @ryandesign , who specifically removed this a few months ago.

I am also okay with just blacklisting `*gcc-4.*` instead, that would fix the build on PPC systems.

kencu:
However, if cxx11 really is needed, then adding the proper cxx_standard also does other important things in the background (like adding that _GLIBCXX_USE_CXX11_ABI=0 flag you manually added to another port recently.

So you can't just blacklist and expect things to work properly, because they won't robustly work properly, even if they seem to work on your one-off local build.

we put a lot of effort into making this all work seamlessly :>

https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

kencu:
For this port, I think it is likely that cxx11 really is needed. 

We just need @ryandesign to say so / agree as he removed it.

Or show us why it is not really needed, if that is the case.

barracuda156:
> However, if cxx11 really is needed, then adding the proper cxx_standard also does other important things in the background (for example, adding that _GLIBCXX_USE_CXX11_ABI=0 flag you manually added to another port recently.)
> 
> So you just blacklisting won't robustly work properly.

From the source code it is quite evident that C++11 _is_ expected. But if @ryandesign removed it before, perhaps there was some non-obvious reason.
(I mean, I can restrict a fix to PPC, be it needed.)

kencu:
it’s not a ppc issue, so restricting it to ppc is not helpful.

let’s let ryan have a chance to take another look. give him a few weeks…there is no hurry.

barracuda156:
@kencu Yes, sure.

chrstphrchvz:
I believe the maintainer’s handle is @thinkoid (#15909); pinging just in case.

thinkoid:
> I believe the maintainer’s handle is @thinkoid (#15909); pinging just in case.

Pong. No longer involved w/ Macs, if anyone wants to take over please do so. Thanks.

barracuda156:
@mascguy I guess this can be merged now.

mascguy (approved):
No concerns here, given that this is benign.

PR #15763, opened 2022-08-17
flac: fix for Rosetta

barracuda156:
Fixes: https://trac.macports.org/ticket/64610

#### Description

Fixes a build failure on Rosetta due to Intel assembler being used.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/audio/flac/Portfile
+++ b/audio/flac/Portfile
@@ -53,6 +53,11 @@ platform darwin i386 {
     configure.args-append   --disable-asm-optimizations
 }
 
+platform darwin 10 powerpc {
+    # Otherwise Intel assembler is used on Rosetta:
+    configure.args-append   --build=powerpc-apple-darwin${os.major}
+}
+
 # -fstack-protector-strong introduced in GCC 4.9
 if {[string match *gcc-4.* ${configure.compiler}]} {
     configure.args-append   --disable-stack-smash-protection


kencu:
you’ve been disabling doxygen all over the place, but I don’t see that you ever tried the proper fix we suggested of binwrapping the doxygen binaries instead. You should try that first.

It’s simple. See:

https://github.com/macports/macports-ports/blob/f3ef4cd21fd307e941e7deadd638fb81516a5589/devel/cmake/Portfile#L58

if it doesn’t work, fine. But it almost always does work, and you can revert all your doxygen disabling patches.

barracuda156:
@kencu Thank you! I will try that.

kencu:
I just installed ppc doxygen on my 10.4 Tiger PPC system and it works perfectly, and passes 99% of the test suite as well.

So whatever is wrong on your system with doxygen, it is not a ppc issue generally, nor is it apparently an issue with building with gcc7.

kencu (changes requested):
doxygen works fine on ppc, Tiger at least verified.

barracuda156:
@kencu Thank you for checking. I will check on Leopard soon, once I’m back to TW.
I have seen doxygen failures quite a number of times, but you are right, it may be specific to either OS or compiler, not all PPC.

barracuda156:
@kencu Removed Doxygen part here, since the thing is fixed. Rosetta fix still needed (it tries using Intel assembler due to physical cpu).

barracuda156:
@mascguy Safe to merge.

barracuda156:
@kencu Can this be merged? I removed Doxygen part, but specifying `--build` still needed, otherwise it fails:
```
ld: warning: in ia32/.libs/libFLAC-asm.a, file was built for unsupported file format which is not the architecture being linked (ppc)
Undefined symbols:
…
ranlib: archive member: .libs/libFLAC-static.a(cpu_asm.o) cputype (7) does not match previous archive members cputype (18) (all members must match)
…
```

PR #15768, opened 2022-08-17
sparsehash: do not build arch-dependent libtcmalloc test

barracuda156:
Fixes: https://trac.macports.org/ticket/65685

#### Description

Add arch flags, fixes Rosetta build.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/devel/sparsehash/Portfile
+++ b/devel/sparsehash/Portfile
@@ -30,18 +30,16 @@ post-patch {
     reinplace "s|@VERSION@|${version}|g" ${worksrcpath}/configure.ac
 }
 
-# While the port installs only headers, it builds tests in the process.
-# Using noarch setting breaks the build: https://github.com/macports/macports-ports/pull/15768
-installs_libs       no
-
 use_autoreconf      yes
 autoreconf.cmd      ./autogen.sh
 autoreconf.args
 depends_build-append \
                     port:autoconf \
                     port:automake \
-                    port:gperftools \
                     port:libtool
+
+# This is a header-only library.
+supported_archs     noarch
 
 # Setting compiler.cxx_standard (to any value) is necessary so that MacPorts
 # sets configure.cxx_stdlib, which the port needs in order to build for the
@@ -49,5 +47,9 @@ depends_build-append \
 # requires, so I'll just guess.
 compiler.cxx_standard 1998
 
-test.run
+# https://github.com/macports/macports-ports/pull/15768
+configure.args-append \
+                    ac_cv_header_google_malloc_extension_h=no
+
+test.run            yes
 test.target         check
--- a/devel/sparsehash/Portfile
+++ b/devel/sparsehash/Portfile
@@ -17,7 +17,6 @@ long_description    {*}${description}. 2 bits/entry overhead! The SparseHash \
                     characteristics. It's easy to replace hash_map by \
                     sparse_hash_map or dense_hash_map in C++ code.
 
-platforms           darwin
 license             BSD
 
 checksums           rmd160  8156dead55d2cf8fc3af2cd0017d494397ee812f \
@@ -31,16 +30,24 @@ post-patch {
     reinplace "s|@VERSION@|${version}|g" ${worksrcpath}/configure.ac
 }
 
+# While the port installs only headers, it builds tests in the process.
+# Using noarch setting breaks the build: https://github.com/macports/macports-ports/pull/15768
+installs_libs       no
+
 use_autoreconf      yes
 autoreconf.cmd      ./autogen.sh
 autoreconf.args
-depends_build-append port:autoconf port:automake port:libtool
-
-# This is a header-only library.
-supported_archs     noarch
+depends_build-append \
+                    port:autoconf \
+                    port:automake \
+                    port:gperftools \
+                    port:libtool
 
 # Setting compiler.cxx_standard (to any value) is necessary so that MacPorts
 # sets configure.cxx_stdlib, which the port needs in order to build for the
 # right stdlib. I don't know what version of the C++ standard this port
 # requires, so I'll just guess.
 compiler.cxx_standard 1998
+
+test.run
+test.target         check


kencu:
this fix is too fragile. won't work universal, won't work with some gcc versions, etc.

barracuda156:
> this fix is too fragile. won't work universal, won't work with some gcc versions, etc.

So should I make it Rosetta-only?

kencu:
> So should I make it Rosetta-only?

No, it's not a Rosetta-only problem (and I won't commit those anyway, although others seem to sometimes do it.

I realize that you can sneak in "bad" fixes by wrapping them in a Rosetta block that nobody seems to care about, but -- really -- at this point in the game you need to figure out proper fixes and submit those.

I know it takes longer to figure out what proper fixes are, and submit them. But hacking in a bunch of Rosetta-only garbage is really not making you any fans. You are close, at least closer, to doing the right thing. Let's get this sorted out properly.

For one -- MacPorts ALREADY sets up all the proper archs for the build. Why is this build not using them? THAT is what needs fixing... but with no log, nobody can help you see why the build is not using the proper settings MacPorts has already taken great pains to set up.

barracuda156:
@kencu Well, if the problem is not specific to Rosetta, of course I want to make a proper fix. Here is the log (unmodified portfile):
[sparsehash.log](https://github.com/macports/macports-ports/files/9365927/sparsehash.log)

kencu:
we're missing in this log all the stuff that macports sets.

please do this:

```
sudo port clean sparsehash
sudo port -v install sparsehad
```

and when it fails, please upload the resulting log.

barracuda156:
> we're missing in this log all the stuff that macports sets.
> 
> please do this:
> 
> ```
> sudo port clean sparsehash
> sudo port -v install sparsehash
> ```
> 
> and when it fails, please upload the resulting log.

OK, give me a min.

kencu:
at first glance, it appears that this build is not seeing the environment set up the way it usually is set up. There are no settings for CPPFLAGS, CXXFLAGS, CFLAGS, LDFLAGS, etc.

```
:debug:build Executing org.macports.build (sparsehash)
:debug:build Environment: 
:debug:build CC_PRINT_OPTIONS='YES'
:debug:build CC_PRINT_OPTIONS_FILE='/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_sparsehash/sparsehash/work/.CC_PRINT_OPTIONS'
:debug:build CPATH='/opt/local/include'
:debug:build DEVELOPER_DIR='/Developer'
:debug:build LIBRARY_PATH='/opt/local/lib'
:debug:build MACOSX_DEPLOYMENT_TARGET='10.6'
:info:build Executing:  cd "/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_devel_sparsehash/sparsehash/work/sparsehash-2.0.4" && /usr/bin/make -j6 -w all 
```

so a bit of looking into the build will probably show why that is...

and we see this, in the Portfile -- AHA:

```
# This is a header-only library.
supported_archs     noarch
```

so -- this is not entirely correct, it appears. if you don't mind, please just try removing or commenting out this line:
```
supported_archs     noarch
```

and then see what happens to the build environment and your build.

barracuda156:
@kencu Awesome,  that worked! Thank you very much.

P. S. Wonder why someone added that to the portfile at all.

kencu:
yeah, now we have to sort that part out. Does this port only install headers? But use some software tool to generate them, and that then requires the archs to be set properly?

That would be odd, but not unheard of. Perhaps we just have to remove the no-arch business, and then perhaps (if there are not libs installed) set ```installs_libs no```.

These parts of MacPorts can be tricky. Often when somebody does something like what we are proposing, it turns out to be "wrong" for some subtle use case and one of the admins will jump on it. But, there is nothing we can do about that except let it happen and try to remember whatever the odd subtlety was for next time. MacPorts has a lot of subtleties, and pretty much nobody except for one or two people know them all, so ... have to have thick skin sometimes.

By the way, so here, your Rosetta setup let you to push this build system a bit and you found a flaw. This led to a proper fix that helps everyone. That is a win-win.

barracuda156:
@kencu Apparently here is what gets installed:
```
a ./+COMMENT
a ./+CONTENTS
a ./+DESC
a ./+PORTFILE
a ./+STATE
a ./opt
a ./opt/local
a ./opt/local/include
a ./opt/local/lib
a ./opt/local/share
a ./opt/local/share/doc
a ./opt/local/share/doc/sparsehash
a ./opt/local/share/doc/sparsehash/AUTHORS
a ./opt/local/share/doc/sparsehash/ChangeLog
a ./opt/local/share/doc/sparsehash/COPYING
a ./opt/local/share/doc/sparsehash/dense_hash_map.html
a ./opt/local/share/doc/sparsehash/dense_hash_set.html
a ./opt/local/share/doc/sparsehash/designstyle.css
a ./opt/local/share/doc/sparsehash/implementation.html
a ./opt/local/share/doc/sparsehash/index.html
a ./opt/local/share/doc/sparsehash/INSTALL
a ./opt/local/share/doc/sparsehash/NEWS
a ./opt/local/share/doc/sparsehash/performance.html
a ./opt/local/share/doc/sparsehash/README
a ./opt/local/share/doc/sparsehash/README_windows.txt
a ./opt/local/share/doc/sparsehash/sparse_hash_map.html
a ./opt/local/share/doc/sparsehash/sparse_hash_set.html
a ./opt/local/share/doc/sparsehash/sparsetable.html
a ./opt/local/share/doc/sparsehash/TODO
a ./opt/local/lib/pkgconfig
a ./opt/local/lib/pkgconfig/libsparsehash.pc
a ./opt/local/include/google
a ./opt/local/include/sparsehash
a ./opt/local/include/sparsehash/dense_hash_map
a ./opt/local/include/sparsehash/dense_hash_set
a ./opt/local/include/sparsehash/internal
a ./opt/local/include/sparsehash/sparse_hash_map
a ./opt/local/include/sparsehash/sparse_hash_set
a ./opt/local/include/sparsehash/sparsetable
a ./opt/local/include/sparsehash/template_util.h
a ./opt/local/include/sparsehash/type_traits.h
a ./opt/local/include/sparsehash/internal/densehashtable.h
a ./opt/local/include/sparsehash/internal/hashtable-common.h
a ./opt/local/include/sparsehash/internal/libc_allocator_with_realloc.h
a ./opt/local/include/sparsehash/internal/sparseconfig.h
a ./opt/local/include/sparsehash/internal/sparsehashtable.h
a ./opt/local/include/google/dense_hash_map
a ./opt/local/include/google/dense_hash_set
a ./opt/local/include/google/sparse_hash_map
a ./opt/local/include/google/sparse_hash_set
a ./opt/local/include/google/sparsehash
a ./opt/local/include/google/sparsetable
a ./opt/local/include/google/template_util.h
a ./opt/local/include/google/type_traits.h
a ./opt/local/include/google/sparsehash/densehashtable.h
a ./opt/local/include/google/sparsehash/hashtable-common.h
a ./opt/local/include/google/sparsehash/libc_allocator_with_realloc.h
a ./opt/local/include/google/sparsehash/sparsehashtable.h
```
I have seen `installs_libs no` it a few portfiles, but dunno if that actually does smth or is rather an informational thing.

kencu:
```installs_libs no``` tells all the ports that depend on this port to ignore it's arch when sorting out what archs are required during building.

So that would be appropriate here (as there are no libs installed) and replaces the previous noarch setting, as clearly this port does require arch knowledge to build properly, so noarch doesn't work robustly.

**Comment by kencu** [comment]
one thing to check -- if there are any "arch" flag settings in the *.pc file, then those either need to be stripped out, or we have to remove ```installs_libs no``` -- because those arch settings, if they exist in that *.pc files, would be important to match.

like I said, lots of subtleties -- I only know a few of the many

**Comment by kencu** [comment]
And here's another issue you have uncovered:

You found that sparsehash tried to link with ```/opt/local/lib/libtcmalloc.dylib```.

But I don't see that there is a dependency listed in the sparsehash portfile for anything that looks like it might provide ```/opt/local/lib/libtcmalloc.dylib```.

Please do this:

```port provides /opt/local/lib/libtcmalloc.dylib```

and see what port is providing that library. Then we either need to add a dependency in sparsehash for that port (whatever it is), or we need to find a way to prevent sparsehash from trying to link against it.

**Comment by kencu** [comment]
if you look at the trail of commits for this port

https://github.com/macports/macports-ports/commits/master/devel/sparsehash/Portfile

you will see all the problems over the years that people have had tried to sort this port out properly

**Comment by kencu** [comment]
so the deal seems to be….

the port is only headers. therefore noarch.

by default, running configure then make builds a bunch of test binaries. these are not needed for the installation. these test binaries will look for and use tcmalloc if it exists. they are not noarch, but not part of the installed port either.

best thing to do might be to not build the test binaries, just install the header files.

if the port wasn’t from 2009 with no updates since it might be worth setting up a test variant, but it’s a dinosaur port.

see https://github.com/sparsehash/sparsehash/blob/master/README

**Comment by barracuda156** [comment]
> But I don't see that there is a dependency listed in the sparsehash portfile for anything that looks like it might provide `/opt/local/lib/libtcmalloc.dylib`.
> 
> Please do this:
> 
> `port provides /opt/local/lib/libtcmalloc.dylib`

@kencu 
```
10:libhangul-libhangul-313b888 svacchanda$ port provides /opt/local/lib/libtcmalloc.dylib
/opt/local/lib/libtcmalloc.dylib is provided by: gperftools
```
There is a related issue about it too: https://github.com/sparsehash/sparsehash/issues/143

**Comment by kencu** [comment]
if we want a test variant, we can add all that other stuff in… but for a 2009-era unmaintained port, I’m not motivated to spend much effort.

**Comment by barracuda156** [comment]
> so perhaps we skip the build stage, and just destroot only.

So just add `build {}` and bring back `noarch`?

**Comment by kencu** [comment]
yes, maybe so.

does that work?

**Comment by barracuda156** [comment]
> yes, maybe so.
> 
> does that work?

Well, no, it does not. Whether I keep autotools block or remove, it still tries to build cc objects.

**Comment by kencu** [comment]
the thing about tmalloc will also need to be addressed, if we are building the tests, so that needs to get sorted out still.

Or maybe someone smarter than me will figure out how to build it without the tests, or perhaps just manually install the headers and forget about the (problematic) makefile, dunno.

**Comment by barracuda156** [comment]
> the thing about tmalloc will also need to be addressed, if we are building the tests, so that needs to get sorted out still. Or someone might point out that if it ignored tmalloc, it would build as it is, perhaps (build wrongly, for the wrong arch, but few would care about that at this point)…

Should we declare dependency on `gperftools` then?

P. S. Thank you very much!

**Comment by kencu** [comment]
> Should we declare dependency on `gperftools` then?

given that the tests are never installed, I would think that would be undesirable for a port that just installs headers. What we might better do is patch the build to ignore tcmalloc if it is found.

@ryandesign will likely not agree with changing this port from noarch, as he specifically changed it to noarch recently -- although it is inherently broken unless we do, unless we can disable the building of the test binaries.

File: devel/sparsehash/Portfile [hunk 1]
```diff
@@ -31,16 +30,24 @@ post-patch {
     reinplace "s|@VERSION@|${version}|g" ${worksrcpath}/configure.ac
 }
 
+# While the port installs only headers, it builds tests in the process.
+# Using noarch setting breaks the build: https://github.com/macports/macports-ports/pull/15768
+installs_libs       no
```
ryandesign, line 35: Per the comment that was in the Portfile that this commit removes (and also per the `long_description`), sparsehash is a header-only library, so you cannot declare that it does not install a library; it does. True, it doesn't install a dynamic library or a shared library but it does install a library of functionality implemented in headers. `installs_libs no` disables both architecture checking and license checking. We would not want the license checking to be disabled in this case; we must verify license compliance when a port uses a library.
```suggestion
```

devel/sparsehash/Portfile
```diff
@@ -31,16 +30,24 @@ post-patch {
...
+                    port:autoconf \
+                    port:automake \
+                    port:gperftools \
+                    port:libtool
 
 # Setting compiler.cxx_standard (to any value) is necessary so that MacPorts
 # sets configure.cxx_stdlib, which the port needs in order to build for the
 # right stdlib. I don't know what version of the C++ standard this port
 # requires, so I'll just guess.
 compiler.cxx_standard 1998
```
ryandesign, line 50: `compiler.cxx_standard` only needed to be set when `supported_archs` was `noarch`; if that's is removed, this should be too.

ryandesign (commented):
> P. S. Wonder why someone added that to the portfile at all.

`git` has a "[blame](https://git-scm.com/docs/git-blame)" feature (and it's even [available on the web on GitHub](https://github.com/macports/macports-ports/blame/202a47af4a6662c50ea5a0c777980eb26aae5a5c/devel/sparsehash/Portfile#L40), GitLab, and other similar services) that shows you [the commit that last modified a line](https://github.com/macports/macports-ports/commit/aab03dad4bf66c0589c8829916043b78017bc55c); from there you can read that commit's message to discover why the change was made.

> @ryandesign will likely not agree with changing this port from noarch, as he specifically changed it to noarch recently -- although it is inherently broken unless we do, unless we can disable the building of the test binaries. -- which Ryan can probably do, with his knowledge of Makefiles

I changed it to `noarch` because what it installs does not vary by architecture so it seemed the correct thing to do. At the time I was not aware of the issue being presented here now. I agree that possible solutions are to remove `supported_archs noarch` as currently in this PR or don't compile the tests, or at least don't compile the tests that use libtcmalloc. I haven't tried to figure out if all the tests can be disabled, but at least the libtcmalloc tests can be prevented from building by adding `ac_cv_header_google_malloc_extension_h=no` to `configure.args`; if we just did that, we could probably keep the port `noarch` which seems more correct to me but I don't particularly mind either way.

barracuda156:
@ryandesign Thank you, that worked! I returned `noarch` and `compiler.cxx_standard`, adding `ac_cv_header_google_malloc_extension_h=no`. It builds fine now for me on Rosetta.

barracuda156:
@mascguy @kencu @ryandesign Are we good with this? Ryan’s fix worked.

devel/sparsehash/Portfile
```diff
@@ -44,3 +46,10 @@ supported_archs     noarch
...
+
+# https://github.com/macports/macports-ports/pull/15768
+configure.args-append \
+                    ac_cv_header_google_malloc_extension_h=no
+
+test.run
```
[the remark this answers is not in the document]
barracuda156: @mascguy Yes, gimme a sec :)

mascguy (approved):

PR #15783, opened 2022-08-19
libssh2: apply pragma patch for gcc versions, not OS

barracuda156:
#### Description

Rosetta also needs existing patch, otherwise build fails with:
```
x11.c: In function ‘main’:
x11.c:328: error: #pragma GCC diagnostic not allowed inside functions
x11.c:329: error: #pragma GCC diagnostic not allowed inside functions
x11.c:333: error: #pragma GCC diagnostic not allowed inside functions
make[1]: *** [x11.o] Error 1
```

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/devel/libssh2/Portfile
+++ b/devel/libssh2/Portfile
@@ -28,9 +28,11 @@ checksums           rmd160  b3af89cc5974dbffbcba0dd80dcd521e6e60b39d \
                     sha256  2d64e90f3ded394b91d3a2e774ca203a4179f69aebee03003e5a6fa621e41d51 \
                     size    965044
 
-if {${os.platform} eq "darwin" && ${os.major} < 10} {
-    # remove errant pragmas inside functions no supported on older gcc versions
-    patchfiles-append patch-libssh2-pragmas-older-gcc.diff
+platform darwin {
+    if {${os.major} < 10 || (${os.major} == 10 && ${build_arch} eq "ppc")} {
+    # Remove errant pragmas inside functions not supported on older gcc versions. Rosetta also needs this patch.
+        patchfiles-append patch-libssh2-pragmas-older-gcc.diff
+    }
 }
 
 depends_lib-append  port:zlib
--- a/devel/libssh2/Portfile
+++ b/devel/libssh2/Portfile
@@ -10,7 +10,6 @@ github.tarball_from releases
 revision            0
 
 categories          devel net
-platforms           darwin
 maintainers         {wohner.eu:normen @Gminfly} openmaintainer
 
 homepage            https://www.libssh2.org/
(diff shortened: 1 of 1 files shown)


kencu:
I really don't like all this 10.6-PPC spaghetti fixes. They are both fragile and cumbersome and I believe we (you) can do better than this. 

A better fix, that would happen to fix rosetta and 10.6 on PPC too, would be to match against the gcc versions that won't support these pragmas. For example, something that would match *gcc-4.* would probably do it. That is a fair, general, correct fix without the 10.6 on PPC specific hacks in there.

However, A=an even better, possibly upstreamable fix, is to use some predefined compiler definitions to sort out which compilers can support pragmas inside functions, and then only use them if supported, like I did here:

https://github.com/macports/macports-ports/blob/master/devel/libuv/files/patch-libuv-unix-core-close-nocancel.diff

I would much prefer if your fixes were generic and not 10.6-on-PPC specific. I think everyone would.

barracuda156:
> I really don't like all this 10.6-PPC spaghetti fixes. They are both fragile and cumbersome and I believe we (you) can do better than this.
> A better fix, that would happen to fix rosetta and 10.6 on PPC too, would be to match against the gcc versions that won't support these pragmas. For example, something that would match `*gcc-4.*` would probably do it. That is a fair, general, correct fix without the 10.6 on PPC specific hacks in there.
> However, an even better, possibly upstreamable fix, is to use some predefined compiler definitions to sort out which compilers can support pragmas inside functions, and then only use them if supported, like I did here:
> 
> https://github.com/macports/macports-ports/blob/master/devel/libuv/files/patch-libuv-unix-core-close-nocancel.diff

This makes sense. Matching against `gcc` version is certainly doable (and indeed better than against OS – after all, gcc-4.* can be used on much later OS that 10.6), and I will look into compiler definitions alternative too.

barracuda156:
@kencu By the way, speaking about more correct and general fixes: what could be done with ports which fail to use Macports build_arch settings, forcing instead cpu_arch (which breaks some ports on Rosetta)? Who is at fault here – why in many cases triple is correct but in some we get x86 or i386 when building for PPC? (To be clear, I refer to configure/makefiles-based ports here, not cmake/meson-based ones.)

kencu:
> @kencu By the way, speaking about more correct and general fixes: what could be done with ports which fail to use Macports build_arch settings, forcing instead cpu_arch (which breaks some ports on Rosetta)? Who is at fault here – why in many cases triple is correct but in some we get x86 or i386 when building for PPC? (To be clear, I refer to configure/makefiles-based ports here, not cmake/meson-based ones.)

There may well be some inconsistencies uncovered in your copious cross-compiling using rosetta that have never been uncovered before. All these should be fixed, if you flag them as tickets they will get attention, and if base needs something Josh or Ryan will weigh in too.

kencu:
@barracuda156 -- if you confirm this works, no issues, we can push

barracuda156:
> @barracuda156 -- if you confirm this works, no issues, we can push

@kencu Yes, it works.

kencu:
thank you, very nice

PR #15787, opened 2022-08-20
glib2: Fix build with pre-C++11 compilers

evanmiller:
#### Description

`glib_typeof` relies on C++11 which isn't available on ancient systems. So, fall back to regular `typeof`.

The patch should be safe everywhere, but I've chosen to add it to the 10.6 and earlier patch list so it's unobtrusive.

Tested through install.

Fixes: https://trac.macports.org/ticket/65663

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11

###### Verification 
Have you

### DIFF
--- a/devel/glib2/Portfile
+++ b/devel/glib2/Portfile
@@ -59,6 +59,7 @@ patchfiles-append           patch-meson-build-broken_poll.diff
 platform darwin {
     if {${os.major} < 11} {
         patchfiles-append   patch-gio_gcredentialsprivate.h.diff \
+                            patch-gio_gcredentials.c.diff \
                             patch-gio_gsocket.h.diff
     }
 }
--- a/devel/glib2/files/patch-gio_gcredentials.c.diff
+++ b/devel/glib2/files/patch-gio_gcredentials.c.diff
@@ -1,0 +1,14 @@
+glib_typeof relies on C++11, so we need to fall back to regular typeof
+(supported by GCC and Clang) to build with older compilers.
+
+--- gio/gcredentials.c.orig	2022-08-20 08:42:34.000000000 -0400
++++ gio/gcredentials.c	2022-08-20 08:46:40.000000000 -0400
+@@ -225,7 +225,7 @@
+ {
+   GString *ret;
+ #if G_CREDENTIALS_USE_APPLE_XUCRED
+-  glib_typeof (credentials->native.cr_ngroups) i;
++  typeof (credentials->native.cr_ngroups) i;
+ #endif
+ 
+   g_return_val_if_fail (G_IS_CREDENTIALS (credentials), NULL);


kencu:
here’s the commit that brought it in

https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1575/diffs?commit_id=0cbf5f89b614c6e2477982b5ad2fe22e74c47019

they were using the underscored typeof before, not the bare one…dunno if that makes a difference.

if we required glib2 to use a c++11 compiler, would that also be a fix for this issue? Would that not be better if so, as they seem to have gone all-in on the c++11 version….

evanmiller:
> if we required glib2 to use a c++11 compiler, would that also be a fix for this issue? Would that not be better if so, as they seem to have gone all-in on the c++11 version….

This patch is the only change needed to get GLib 2.68 compiling with GCC 4.2. One other small change is needed for GLib 2.70 (the old pragmas-in-functions chestnut). So I don't think C++11 should be required just yet.

mascguy:
> > if we required glib2 to use a c++11 compiler, would that also be a fix for this issue? Would that not be better if so, as they seem to have gone all-in on the c++11 version….
> 
> This patch is the only change needed to get GLib 2.68 compiling with GCC 4.2. One other small change is needed for GLib 2.70 (the old pragmas-in-functions chestnut). So I don't think C++11 should be required just yet.

We can certainly consider C++11 for the future. But a simple fix is definitely preferred at the moment, so let's go with this.

Thanks so much Evan!

mascguy (approved):

kencu:
sounds OK, but I'm going to go with c++11 and glib_typeof, so we can see how things diverge.

As all the glib2 headers have changed to delivering up glib_typeof, and given all the many other ports that call the glib2 headers, seems to me they will soon be broken without this anyway. And I might as well keep my older systems as similar to the newer ones as possible to avoid issues.

We can see what happens.

mascguy:
> sounds OK, but I'm going to try to go with c++11 and glib_typeof, so we can see how things diverge.
> 
> As all the glib2 headers have changed to delivering up glib_typeof, and given all the many other ports that call the glib2 headers, seems to me they will soon be broken without this anyway. And I might as well keep my older systems as similar to the newer ones as possible to avoid issues.
> 
> We can see what happens.

For sure Ken! And you might want to focus on `glib2-upstream`, as it coincides with the latest stable release. (`glib2` will be at parity with that, after two more update cycles. The next update will occur in the next month-ish, so we're only 2-3 months away from that.)

kencu:
I guess everything that uses glib2 (which is almost the entire gnome family) will also now need to use c++11, otherwise this issue will keep arising and need patching everywhere it is called from, which is kinda beyond my tedium threshold. We'll have to see.

That is trivial for 10.6 Intel which already uses a c++11-capable compiler to build everything. And that is what I mostly use, so that is a no-op unless people start patching away to fix 10.6 unnecessarily (Rosetta happily excluded).

But for all the family of systems and compilers below that, we might expect to see c++11 needed in all gnome ports that stumble across this.

evanmiller:
glib2 is used widely beyond gnome, so I don't think gnome's C++ policy should prevail here. I think a lowest common denominator approach makes sense for core libraries (glib2 is used by 347 ports). If any of those ports need `glib_typeof` then by all means we can require C++11 for them.

PR #15800, opened 2022-08-21
gtk4: new port, version 4.6.7

kencu:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.5.1 21G83 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/gnome/gtk4/Portfile
+++ b/gnome/gtk4/Portfile
@@ -1,0 +1,140 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           meson 1.0
+PortGroup           active_variants 1.1
+
+#PortGroup           legacysupport 1.1
+
+
+name                gtk4
+#conflicts           gtk4-devel
+set my_name         gtk4
+
+# GNOME project name
+set gname           gtk
+
+version             4.6.7
+revision            0
+
+categories          gnome
+platforms           darwin
+license             GPL-3+
+maintainers         nomaintainer
+
+description         ${my_name} is a companion library to GObject and Gtk+.
+long_description    ${description}  It includes a collection of dazzling Gtk widgets, \
+                    data structures, search engines, a shortcut engine, panels, \
+                    desktop integration, and those missing pieces from common libraries \
+                    that help you write cleaner and safer code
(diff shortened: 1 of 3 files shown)


kencu:
this port works really quite well using the **quartz** variant. The demos work nicely, the majority of the tests pass, and the introductory applications here build from source and work well:

https://docs.gtk.org/gtk4/getting_started.html

you build them like this, FYI:

```
clang `pkg-config --cflags gtk4 glib-2.0` testgtk.c `pkg-config --libs gtk4 glib-2.0`
```

Using the **x11** variant also works, but at the moment needs some surgery to some other ports to support EGL. 

It seems gnome has gone fairly strongly towards using egl, at least from my limited browsing with looking to update webkit2-gtk and looking at other ports. It is at first attempt at least difficult to make this port, gtk4, not try to use <EGL/egl.h>.

So what I did instead on my local machine was to adjust our mesa port to install EGL (which required a toggle and one minor patch in one file), and then adjust libepoxy to use EGL too (which required a toggle and two minor patches). Once mesa and libepoxy were built with EGL support, this built x11 easily.

It works as x11, in xorg-server, running the gtk4-demo application. It is slow, however. There are many messages displayed in the terminal window indicating displeasure with this and that. Some things didn't work quite the same way as they did in the quartz version.

This may be the port then that forces MacPorts to default to quartz instead of x11 -- dunno. Maybe someone knows a lot about EGL on MacOS, and can fix it better than I could.

As it is now, if you want gtk4 and you want the quartz variant of it, this works great.

kencu:
I have only tested this (so far) on the most current macOS system, 12.5.1, with the latest Xcode, on an Intel system.

kencu:
[gtk4-quartz-testlog.txt](https://github.com/macports/macports-ports/files/9388629/gtk4-quartz-testlog.txt)

kencu:
the bot failures are expected, as the MacPorts-default x11 variant requires the mentioned changes to mesa and libepoxy. 

the quartz variant should build right through, if that were the Macports-default for the bot.

kencu:
I (temporarily perhaps) made the quartz variant the default for this port only, at least to run it through the bots, until EGL is supported in mesa and libepoxy, or (perhaps) someone sorts out how to avoid EGL completely better than I could.

mascguy:
Awesome, this is a great starting point Ken. Thank you!

Also, you can add me as the official maintainer for this port.

kencu:
sure

mascguy:
I didn't expect this to be merged so quickly, and my plate is already full. So I'll let you tackle fixing this for older macOS releases.

PR #15807, opened 2022-08-21
xrootd: fix build with gcc

barracuda156:
#### Description

The PR:
1. Replaces blacklist of `*gcc*` with `*gcc-4.*`, to allow using new GCC. (Do we actually need those at all, given portfile has `compiler.cxx_standard  2014`? Maybe do away with blacklists altogether?)
2. Fixes an error with `byteswap.h` missing on MacOS. (To be honest I wonder how Clang does not err out on that – FWIU that header is missing on all version of MacOS or at least on majority of them.)

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/science/xrootd/Portfile
+++ b/science/xrootd/Portfile
@@ -20,7 +20,6 @@ description         Generic suite for fast, low-latency and scalable data access
 long_description    {*}${description}
 
 license             LGPL-3+
-platforms           darwin
 
 homepage            http://xrootd.org/
 
@@ -35,12 +34,14 @@ depends_lib-append  port:curl \
                     path:lib/libuuid.dylib:ossp-uuid \
                     port:zlib
 
+patchfiles-append   patch-byteswap.diff
+
 default_variants    +ssl +readline +kerberos
 
 cmake.out_of_source yes
 
 # Force a compatible compiler
-compiler.blacklist-append *gcc* {clang < 900} {macports-clang-3.[0-9]} {macports-clang-[4-5].0}
+compiler.blacklist-append *gcc-4.* {clang < 900} {macports-clang-3.[0-9]} {macports-clang-[4-5].0}
 # Blacklist Xcode clang on 10.12
 # error: template argument for template template parameter must be a class template or type alias template
 compiler.blacklist-append {clang < 1000}
--- a/science/xrootd/files/patch-byteswap.diff
+++ b/science/xrootd/files/patch-byteswap.diff
@@ -1,0 +1,22 @@
+--- src/XrdOssCsi/XrdOssCsiTagstoreFile.hh.orig	2022-06-09 17:56:41.000000000 +0800
++++ src/XrdOssCsi/XrdOssCsiTagstoreFile.hh	2022-08-21 21:15:49.000000000 +0800
+@@ -37,7 +37,19 @@
+ 
+ #include <memory>
+ #include <mutex>
++
++#if defined(__GLIBC__) || defined(__BIONIC__) || defined(__CYGWIN__)
+ #include <byteswap.h>
++#elif defined(__APPLE__)
++// Make sure that byte swap functions are not already defined.
++#if !defined(bswap_16)
++// Mac OS X / Darwin features
++#include <libkern/OSByteOrder.h>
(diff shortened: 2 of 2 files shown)


kencu:
clang uses the builtin_bswap compiler intrinsic.

newer gcc versions also have such an intrinsic.

barracuda156:
> clang uses the builtin_bswap compiler intrinsic.
>
> newer gcc versions also have such an intrinsic.

@kencu Okay, but how to include it? Existing code does not build with gcc12 (otherwise I would not made a patch for that case).

cjones051073:
I would rather avoid introducing a patch file to address this…

cjones051073:
Have you discussed this with upstream ?

barracuda156:
> I would rather avoid introducing a patch file to address this…

What alternative would you recommend?

cjones051073:
> > I would rather avoid introducing a patch file to address this…
> 
> What alternative would you recommend?

Well a start would be to determine why current build with clang have no problems.

I would start by discussing the issue with upstream.

cjones051073:
b.t.w. is the issue with all gcc versions, or just gcc12 ?

If only gcc12, or some of the newer ones, and older ones are OK I would prefer as a temporary fix (until upstream fixes support with the newer gcc versions) to just black the specific gcc versions that have problems, and thus use the newest that is OK, without needing a patch.

kencu:
it's a missing header, so it would affect any build on macos that tries to include this file (this module, really).

What is actually going on here is that this gcc-only module is usually never built on macOS (which pretty much never tries to build this with gcc).

https://github.com/xrootd/xrootd/blob/7cc1184aebff87384483ab376d67adbad2405cb2/src/CMakeLists.txt#L63

So upstream would never expect to see or test this.

The patch is OK, I think. I haven't tested it but similar things are seen around the web to work around this issue of the "missing" header on macOS. I kinda doubt upstream would be interested in such a patch (that fixes the build on a totally unsupported system) but you never know. @barracuda156 can be very persuasive :>

kencu:
examples:

https://github.com/sgan81/apfs-fuse/issues/6#issuecomment-363601678

https://debugjournal.tumblr.com/post/99661010362/missing-byteswaph-and-endianh-on-mac-os-x

https://bugs.freedesktop.org/show_bug.cgi?id=8882

kencu:
@barracuda156 -- I suggest you try selling this upstream. You have a fair case, although they may not accept as very few people build this with gcc on darwin.

barracuda156:
@kencu @cjones051073 I will deal with this in 1–2 days.

barracuda156:
> b.t.w. is the issue with all gcc versions, or just gcc12 ?
> 
> If only gcc12, or some of the newer ones, and older ones are OK I would prefer as a temporary fix (until upstream fixes support with the newer gcc versions) to just black the specific gcc versions that have problems, and thus use the newest that is OK, without needing a patch.

The issue is that there is no `byteswap.h` on MacOS, at least on many versions of it.
Example with 10.15: https://github.com/Xilinx/qemu/issues/47
Fix (similar to mine here): https://github.com/Xilinx/qemu/commit/0d8efa472c074fc340a360a9bd74099a5030d7aa

barracuda156:
> @barracuda156 -- I suggest you try selling this upstream. You have a fair case, although they may not accept as very few people build this with gcc on darwin.

@kencu I have an open PR there: https://github.com/xrootd/xrootd/pull/1766

barracuda156:
UPD. Do not merge this yet, upstream suggests possibly moving the definitions to `XrdSysPlatform.hh` for the sake of consistency, and I wait for an update in that PR thread.

kencu:
nice job with upstream, looks like they will accept soon.

barracuda156:
> nice job with upstream, looks like they will accept soon. 

@kencu @cjones051073 Merged in upstream in: https://github.com/xrootd/xrootd/commit/8e1e528232d8ead08468eefcbf991737eb5751da

kencu:
thanks!

PR #15855, opened 2022-08-23
qscintilla: work around failure to find SDK

amake:
#### Description

See https://trac.macports.org/ticket/65293

I only need/use qscintilla-qt5, but according to the Port Health on ports.macports.com (e.g. https://ports.macports.org/port/py37-pyqt5-scintilla/details/) all of the Qt5 subports of this portfile are affected, so I applied the workaround to all of them.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.5.1 21G83 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/qscintilla/Portfile
+++ b/devel/qscintilla/Portfile
@@ -67,6 +67,13 @@ foreach qt_major {4 5} {
 
             # fixup @rpath in the library's install_name
             patchfiles-append patch-src-qscintilla.pro.diff
+
+            # https://trac.macports.org/ticket/65293
+            if { ${os.platform} eq "darwin" && (( ${os.major} >= 15 && ${os.major} <= 16 ) || ${os.major} >= 20 ) } {
+                use_xcode   yes
+            }
+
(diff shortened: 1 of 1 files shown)


michaelld:
See also: https://trac.macports.org/ticket/64713 https://trac.macports.org/ticket/62309 https://trac.macports.org/ticket/63558 https://trac.macports.org/ticket/65107

michaelld:
[Per @kencu 's note here](https://trac.macports.org/ticket/64713#comment:18), while this PR will work for most folks, it seems like it still doesn't fix the underlying issue. That said, given the complexity of the issue, I don't know if there is a better solution ... one that will fix _all_ situations.

michaelld (approved):
Per my note: LGTM ... I think this is the best way forward until we find a more robust solution. I think this tweak will be required for a few other ports as well -- `py*-pyqt5*` for example. That said, one step at a time!

PR #15856, opened 2022-08-23
gtk4: add system blockers

kencu:
block +quartz on darwin < 15
block +x11 on all systems

change default to x11 to match macports
this presently gives a message to users to install with +quartz

revbump to enforce new defaults

### DIFF
--- a/gnome/gtk4/Portfile
+++ b/gnome/gtk4/Portfile
@@ -15,7 +15,7 @@ set my_name         gtk4
 set gname           gtk
 
 version             4.6.7
-revision            0
+revision            1
 
 categories          gnome
 platforms           darwin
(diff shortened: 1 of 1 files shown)


kencu:
@mascguy, now that we know where the quartz variant builds and where it does not, it seems logical to move this to the (current) macports default of using x11. Leaving it as is to default as +quartz is likely to just cause confusion.

This will give users an error and tell them to install gtk4 manually as +quartz.  

There is **no timeline** for when this might support x11 on macOS however - that is not apparently a target upstream is supporting at present, as far as I can see - although the artists on MacPorts may fix that.

One question -- should it enforce glib2 as matching the variant? I notice gtk3 does not do that...It seems it probably should enforce a matching glib2, but I notice it appears to build fine as quartz no matter how glib2 has been installed (x11 or quartz), which is a bit surprising -- although there could be subtle differences in the build of gtk4 that I did not notice.

mascguy (approved):

mascguy:
> One question -- should it enforce glib2 as matching the variant? I notice gtk3 does not do that...It seems it probably should enforce a matching glib2, but I notice it appears to build fine as quartz no matter how glib2 has been installed (x11 or quartz), which is a bit surprising -- although there could be subtle differences in the build of gtk4 that I did not notice.

Since we don't have any dependents to worry about yet, let's keep it less strict for now.

And thanks again for all of the work on this Ken!

PR #15886, opened 2022-08-24
gitlab-runner: don't use hardcoded REVISION and BRANCH in ld_flags

tsabirgaliev:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.5.1 21G83 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/gitlab-runner/Portfile
+++ b/devel/gitlab-runner/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           golang 1.0
 
 go.setup            gitlab.com/gitlab-org/gitlab-runner 15.3.0 v
-revision            0
+revision            1
 
 homepage            https://docs.gitlab.com/runner/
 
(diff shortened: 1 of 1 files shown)


devel/gitlab-runner/Portfile
```diff
@@ -29,14 +29,18 @@ checksums           rmd160  9f61ee71f3578a182ef17751ae458a1371ab521d \
 # Allow Go to fetch deps at build time
 build.env-delete    GO111MODULE=off GOPROXY=off
 
-# Reproduce the "build_simple" target from the upstream Makefile
+# Reproduce the "runner-bin-host" target from the upstream Makefile
 # Lookup REVISION from https://gitlab.com/gitlab-org/gitlab-runner/-/releases/v${version}
+set release_info [exec curl -s "https://gitlab.com/api/v4/projects/gitlab-org%2Fgitlab-runner/releases/v${version}"]
```
herbygillot, line 34: ❗ This is a bad idea.

If you do this, then every single time this Portfile is even **_parsed_**, MacPorts will have to do a `curl` exec.

This can happen when MacPorts is doing a routine `portindex`, someone does a `port info gitlab-runner`, etc.

We should stick to hardcoding the `REVISION` and `BRANCH` id for now.  What you propose here will have too many negative impacts on workflows that may not even have anything to do with `gitlab-runner`.

tsabirgaliev: The main reason behind this was avoid errors like in https://github.com/macports/macports-ports/commit/5a7f5d10128e19e4f8857237d8eab77ef395d7f8.

tsabirgaliev: Is it better to calculate that at build phase?

tsabirgaliev:
@herbygillot here is the configuration is enclosed in pre-build phase, please review

herbygillot:
> @herbygillot here is the configuration is enclosed in pre-build phase, please review

OK yes, this makes more sense.

herbygillot:
Thanks @tsabirgaliev

herbygillot:
There is also the question of what would happen should that `curl` call fail for whatever reason, but I'm going to guess those variables just get set to empty...

tsabirgaliev:
> There is also the question of what would happen should that `curl` call fail for whatever reason, but I'm going to guess those variables just get set to empty...

@herbygillot when curl or regexp fails the build phase fails with error message

herbygillot:
Ok, makes sense. Thanks.

PR #15917, opened 2022-08-26
erlang: do not use *gcc-4.*

barracuda156:
#### Description

Existing compliler.blacklist leads to `llvm-gcc-4.2` being picked on 10.6: https://trac.macports.org/ticket/52507
This is a wrong choice when the build is for `ppc` (such as Rosetta case): to begin with, Rosetta starts building for Intel, but even when that is fixed (adding `--build=`), build still fails.

PR adds:
```
platform darwin 10 powerpc {
    compiler.blacklist-append *gcc-4.*
}
```
No other systems are affected.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Rosetta
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/lang/erlang/Portfile
+++ b/lang/erlang/Portfile
@@ -77,7 +77,7 @@ depends_lib         port:ncurses
 # Erlang seems to have an on-again/off-again relationship with Clang.
 # As of Xcode 4.2, it's off again.
 # GCC 4.2 also fails: https://trac.macports.org/ticket/52507
-compiler.blacklist  {clang < 300} gcc-4.2
+compiler.blacklist  {clang < 300} *gcc-4.*
 
 post-destroot   {
     system "tar -C ${destroot}${prefix}/lib/erlang -zxvf [shellescape ${distpath}/otp_doc_html_${version}${extract.suffix}]"


kencu:
don’t make a powerpc-snowleopard specific block, as this has nothing to donwith powerpc?

gcc-4.2 is already blacklisted two lines up as not working, so just extend that to cover your failing llvm-gcc-4.2.

barracuda156:
> don’t make a powerpc-snowleopard specific block, as this has nothing to do with powerpc.
> 
> gcc-4.2 is already blacklisted two lines up as not working, so just extend that to cover your failing llvm-gcc-4.2.

I am happy with this. (I just thought that it was done on purpose to let `llvm-gcc` be used for 10.6 Intel.)
Made a generic case now.

kencu:
good, thanks

barracuda156:
> good, thanks

@kencu Thank you!

P. S. On a side note, does Macports ever use Xcode clang on 10.6.x? There are cases when Rosetta tries pulling it for PPC (when gcc-4.2 is blacklisted), which of course never works and build fails. If that particular Clang is never used normally, then perhaps it can be added (for those ports only) to blacklist, otherwise a specific powerpc-apple-darwin10 condition needed.

PR #15965, opened 2022-09-01
mesa, libepoxy: add EGL support, fix mesa 19.0.8 for 10.6

barracuda156:
#### Description

This PR:
1. Adds EGL support to `mesa` and `libepoxy`: needed for `gtk4`.
2. Allows `mesa` 19.0.8 to be built on 10.6.
3. Fixes the patch error: https://trac.macports.org/ticket/65750
4. Adds needed fixes for old systems.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/graphics/libepoxy/Portfile
+++ b/graphics/libepoxy/Portfile
@@ -3,13 +3,13 @@
 PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           meson 1.0
+PortGroup           compiler_blacklist_versions 1.0
 
 github.setup        anholt libepoxy 1.5.10
-revision            0
+revision            1
 license             MIT permissive
 categories          graphics
 maintainers         nomaintainer
-platforms           darwin
 
 description         Epoxy is a library for handling OpenGL function \
                     pointer management for you
@@ -43,6 +43,27 @@ patchfiles          prefix.patch \
 #   in /opt/local/lib/libgdk-3.0.dylib
 configure.args-append \
                     -Dglx=yes
+
+if {${os.platform} eq "darwin" && ${os.major} > 9} {
+    PortGroup       legacysupport 1.1
+
+    # strndup
+    legacysupport.newest_darwin_requires_legacy 10
+
+    compiler.blacklist-append \
+                    *gcc-4.* {clang < 400}
+
+    # enable EGL, needed for gtk4
+    configure.args-append \
+                    -Degl=yes
+
+    # https://github.com/anholt/libepoxy/issues/278
+    configure.cppflags-append \
+                    -D_DARWIN_C_SOURCE
+
+    patchfiles-replace \
+                    prefix.patch patch-egl.diff
+}
 
 # https://trac.macports.org/ticket/64468
 platform darwin 8 {
--- a/graphics/libepoxy/files/patch-egl.diff
+++ b/graphics/libepoxy/files/patch-egl.diff
@@ -1,0 +1,24 @@
+--- src/dispatch_common.h.orig	2022-02-17 19:56:12.000000000 +0700
++++ src/dispatch_common.h	2022-09-01 21:11:53.000000000 +0700
+@@ -28,7 +28,7 @@
+ #define PLATFORM_HAS_GLX ENABLE_GLX
+ #define PLATFORM_HAS_WGL 1
+ #elif defined(__APPLE__)
+-#define PLATFORM_HAS_EGL 0 
++#define PLATFORM_HAS_EGL ENABLE_EGL
+ #define PLATFORM_HAS_GLX ENABLE_GLX
+ #define PLATFORM_HAS_WGL 0
+ #elif defined(ANDROID)
+
+--- src/dispatch_common.c.orig	2022-02-17 19:56:12.000000000 +0700
++++ src/dispatch_common.c	2022-09-01 21:11:35.000000000 +0700
+@@ -174,7 +174,8 @@
+ #include "dispatch_common.h"
+ 
+ #if defined(__APPLE__)
+-#define GLX_LIB "/opt/X11/lib/libGL.1.dylib"
++#define GLX_LIB "@PREFIX@/lib/libGL.1.dylib"
++#define EGL_LIB "@PREFIX@/lib/libEGL.1.dylib"
+ #define OPENGL_LIB "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL"
+ #define GLES1_LIB "libGLESv1_CM.so"
+ #define GLES2_LIB "libGLESv2.so"
--- a/x11/mesa/Portfile
+++ b/x11/mesa/Portfile
@@ -5,12 +5,13 @@ PortGroup           compiler_blacklist_versions 1.0
 
 # May need clock_gettime()
 PortGroup           legacysupport 1.0
+
 legacysupport.newest_darwin_requires_legacy 15
 
 name                mesa
 epoch               1
 version             19.0.8
-revision            1
+revision            2
 categories          x11 graphics
 maintainers         {jeremyhu @jeremyhu} openmaintainer
 license             MIT
@@ -78,9 +79,9 @@ configure.env-append \
 compiler.blacklist gcc-3.3 gcc-4.0 gcc-4.2 llvm-gcc-4.2 {clang < 800}
 
 platform darwin {
-    if {${os.major} < 11} {
+    if {${os.major} < 10} {
 
-        # versions > 17 do not presently compile on systems prior to 10.7
+        # versions > 17 do not presently compile on systems prior to 10.6
         version             17.1.6
         revision            2
         checksums           sha1    2acc201e24ea67c5231074d6746a42a747228ed6 \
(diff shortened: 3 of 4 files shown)


kencu:
this should probably be discussed with upstream mesa before being pushed into macports. The mesa test suite would need to be run and examined. nobody so far uses EGL on macOS.

my patch is slightly different than this patch. That would have to be looked at.

But, enabling EGL in mesa may turn out to be OK, after all that has been done. And then after that, a similar project for libepoxy.

To get gtk4 x11 working, it’s probably better to just force gtk4 to use GLX instead. I’ve already done that. So going the EGL path urgently is not needed.

barracuda156:
@kencu Can we first wait for checks result and then convert to draft? I would like to see if both ports build fine on current Intel systems.

kencu:
your EGL fix for libepoxy is briefer than mine was (I was on Monterey)… that’s cool, if it works!

barracuda156:
> also, if EGL is enabled in mesa, no variant needed. Just adds more pathways for build failures and tickets. It either works or doesn’t work.

I am okay with that. I just followed existing arrangement (`mesa` portfile sets 2 variants and adds both as default ones).

> And then after that, a similar project for libepoxy.

Done in the second commit here, just need to see if it also builds with Clangs on new MacOS.

kencu:
the build is in progress…does making it a draft stop that? If so, I had not noticed that previously.

barracuda156:
> your EGL fix for libepoxy is briefer than mine was (I was on Monterey)… that’s cool, if it works!

If it works here, we can fix it with upstream, I guess. These defines should fix one half of the problem (in place of our cppflag):
```
#elif defined(__APPLE__)
#define _DARWIN_C_SOURCE
```
And then `strndup` error, which is fixed by PG here, perhaps can be also fixed in the source by borrowing from PG code.

barracuda156:
@kencu I get one warning when building `libepoxy` with `gcc12` on 10.6.8:
```
WARNING: ['nm'] does not work. Relinking will always happen on source changes.
error: nm: invalid argument --
Usage: nm [-agnopruUmxjlfAPL[s segname sectname] [-] [-t format] [[-arch <arch_flag>] ...] [file ...]
```
Dunno how serious is that.

kencu:
No, this issue needs to back up 10 steps. Nobody but you cares if this builds with gcc12 on 10.6.8, so let's start somewhere realistic: the current macOS, on Intel and M1. THAT needs to work.  No just build, work.

Upstream mesa should be be involved before this can go ahead. 

What is the status of EGL support in mesa on Apple systems? We know it is almost never enabled, but is it reasonably complete, and does it work at all? How does the mesa test suite perform (esp with respect to EGL)?

We're not pushing junk here.

kencu:
Having said that -- it might turn out in the end that enabling EGL support in mesa across-the-board for all systems is OK. But the proper legwork and testing needs to be done.

MacPorts has hundreds of thousands of users, I'm told. Many of these users actually use it for something important. You can't just push things and hope they stick against the wall. You have to test and verify them first.

What is the effect of that MONOTONIC patch? Is that setting needed? If so, what is the APPLE equivalent setting to use? If not needed, what is the least impactful way of excluding it?

barracuda156:
> What is the effect of that MONOTONIC patch? Is that setting needed? If so, what is the APPLE equivalent setting to use? If not needed, what is the least impactful way of excluding it?

There are some related cases due to missing `pthread_condattr_setclock`:

https://github.com/python/cpython/commit/001fee14e0f2ba5f41fb733adc69d5965925a094

https://github.com/farsightsec/fstrm/commit/bedf826e458f34f24f1be628383217dde8d9397a

barracuda156:
So Intel builds are good on 10.6.8:
```
10:bin svacchanda$ ./port -v installed mesa
  mesa @19.0.8_2+egl+osmesa+python27 (active) requested_variants='' platform='darwin 10' archs='x86_64' date='2022-09-01T23:26:25+0700'
10:bin svacchanda$ ./port -v installed libepoxy
  libepoxy @1.5.10_0 (active) requested_variants='' platform='darwin 10' archs='x86_64' date='2022-09-01T23:33:40+0700'
```

kencu:
so the next step is to take this to the mesa dev mailing list, and ask for
opinions and help there. For now, IMHO, nobody should push this any time
soon.

(the "nm" warning is commonly seen building with meson, I know it came with
libepoxy, and it is nothing to worry about).

Regarding getting gtk4 working as x11, I believe I will force to gtk4 to
build with GLX on Apple x11 systems, and not use EGL, at least for now,
until the EGL issue is all sorted out someday. GLX is more reliable and
actually tested on Apple systems, and the performance with the EGL enabled
in mesa/libepoxy seemed  poor when I tried that a month ago.

If someday EGL is enabled on Apple systems, it might be via ANGLE instead
of mesa, but here we need some folks who actually know about deep OpenGL
issues. Too bad Jeremy moved on.

To be noted that OpenGL is deprecated on MacOS, and it is said support will
be dropped from M1 / arm64 Macs soon enough. I don't actually know what
that means for x11 on MacOS going forward. Some of the mesa devs are trying
to reverse-engineer the video driver subsystem on M1 macs to see if they
can come up with an alternate long-term support plan.

kencu:
I’m going to take up this EGL discussion on the meas mailing list, and see where it leads, so pls don’t merge this yet.

I have a patchset for gtk4 that avoids EGL on MacOS and uses GLX instead that we might better wind up usingbif EGL is too fragile…

barracuda156:
> I’m going to take up this EGL discussion on the mesa mailing list, and see where it leads, so pls don’t merge this yet.
> 
> I have a patchset for gtk4 that avoids EGL on MacOS and uses GLX instead that we might better wind up using if EGL is too fragile…

@kencu If you or others decide not to have EGL at all or have it as a non-default variant, I will modify PR accordingly.
The wrongly used patch should be moved in any case (that is a bug), and perhaps building 19.0.8 should be allowed on 10.6, regardless of EGL decision outcome.

kencu:
> > I’m going to take up this EGL discussion on the mesa mailing list, and see where it leads, so pls don’t merge this yet.
> > I have a patchset for gtk4 that avoids EGL on MacOS and uses GLX instead that we might better wind up using if EGL is too fragile…
> 
> @kencu If you or others decide not to have EGL at all or have it as a non-default variant, I will modify PR accordingly. The wrongly used patch should be moved in any case (that is a bug), and perhaps building 19.0.8 should be allowed on 10.6, regardless of EGL decision outcome.

Agree 100%. 

Maybe mesa will say EGL is pretty good, needs a bit of work, etc. I don't know yet. I hope it's fairly good, because manually fixing ports to build ignoring it is hard.

I think there is an alternate EGL in ANGLE that is used by Apple for Webkit, and by us for Webkit2-gtk, that may be a better option for EGL support. This stuff is complicated, to be honest, and it's hard to unbuild many ports once it's out there and you find out it's all wrong.

kencu:
request for information posted:

https://lists.freedesktop.org/archives/mesa-dev/2022-September/225848.html

barracuda156:
> request for information posted:
> https://lists.freedesktop.org/archives/mesa-dev/2022-September/225848.html

@kencu Thank you! Let us see what they say.

kencu:
So I have been reading through the mesa source code, in particular here:

https://github.com/Mesa3D/mesa/commit/ad503c41557606d15b0420c824369456f6d20a8f

and in the glx subdirectories, and looking at some other references such as I put in the gtk4 PR yesterday.

It looks like what went on here is that in 2010 Apple made all their GLX code available to mesa, and Jeremy integrated it. That gave mesa hardware support for GLX on Apple systems, using Apple's OpenGL. If that is right, and I think it seems right looking through what is in the egl subsystems on mesa, then on darwin, the only performant OpenGL implementation will be GLX. I think the EGL support on darwin via mesa is through their software renderers only.

If EGL support is brought in with this PR, or with one like it, then these EGL headers will be found by various software builds, and perhaps used. But OpenGL performance will probably be dismal -- which is more or less what I saw when I used this approach building gtk4.

So I would not presently be in favour of committing this without some indication of it being workable and performant.

I will tweak in @jeremyhu here, although I know he is way too busy to comment usually.

kencu:
the main driver for this PR was to get gtk4 on x11 for older systems, and I took the approach of forcing GLX for that

https://github.com/macports/macports-ports/commit/a1bf4959c541fe8a00e6ef016c23a9fea810bbe8

if we can get that to build through on all interesting systems, then that might well be the better approach.

No doubt what I did in that commit could be improved; I'm not an OpenGL engineer. I look forward to improvements.

kencu:
btw, there is a PR to update mesa to the latest release here https://github.com/macports/macports-ports/pull/16260 as well.

barracuda156:
@kencu I am certainly no expert here, though I see no harm in adding EGL as an optional variant (why not make it available for testing?). As for `gtk4`, I trust your solution will be better than mine :)

kencu:
If <EGL/egh.h> is found by software (like gtk4 would have found it) then it would often be preferentially used because of how upstream is angling these days. But because it seems so poorly performant, the outcome could be very poor.

On the other hand, if <EGL/egh.h> is not found then attention would be drawn to that fact, and a workaround developed, like happened with gtk4.

So that is the downside of having it there.

kencu:
I am hoping that someone with some deeper knowledge of how mesa implements EGL on macOS might step forward to comment at some point here. But if it is just left to me, as far as I can tell from the source code, Apple's hardware OpenGL implementation only applies to GLX.

kencu:
I am hoping that someone with some deeper knowledge of how mesa implements EGL on macOS might step forward to comment at some point here. But if it is just left to me, as far as I can tell from the source code, Apple's hardware OpenGL implementation only applies to GLX.

kencu:
not going to to enable egl at present.

kencu:
recent info from the guru

https://lists.freedesktop.org/archives/mesa-dev/2023-January/225907.html

barracuda156:
> recent info from the guru
> 
> https://lists.freedesktop.org/archives/mesa-dev/2023-January/225907.html

Thank you.

PR #15993, opened 2022-09-04
opencv4-devel: Turn on QUIRC support

essandess:
### Description

* Enable QUIRC support
* Depends on PR: https://github.com/macports/macports-ports/pull/15994

### Type(s)
- [ ] bugfix
- [x] enhancement
- [ ] security fix

### Tested on
macOS 12.5.1 21G83 x86_64
Xcode 13.4.1 13F100

### Verification 
Have you

### DIFF
--- a/graphics/opencv4-devel/Portfile
+++ b/graphics/opencv4-devel/Portfile
@@ -57,7 +57,7 @@ if {${os.major} > 12} {
 
 if {${opencv_latest}} {
     github.setup    opencv opencv 4.6.0
-    revision        2
+    revision        3
     epoch           1
 
     checksums-append \
@@ -71,7 +71,7 @@ if {${opencv_latest}} {
     livecheck.regex {/archive/([0-9.]+)[a-z]?\.[tz]}
 } else {
     github.setup    opencv opencv 4.5.0
-    revision        8
+    revision        9
 
     checksums-append \
                     ${distname}${extract.suffix} \
@@ -139,6 +139,7 @@ depends_lib-append  \
                     port:libpng \
                     path:include/turbojpeg.h:libjpeg-turbo \
                     port:openjpeg \
+                    port:libquirc \
                     port:tiff \
                     port:webp \
                     port:openexr \
@@ -283,6 +284,7 @@ configure.args-append \
                     -DWITH_IMGCODEC_PFM:BOOL=ON \
                     -DWITH_QUIRC:BOOL=OFF \
                     \
+                    -DHAVE_QUIRC:BOOL=ON \
                     -DHAVE_COCOA:BOOL=ON \
                     \
                     -DBUILD_opencv_aruco:BOOL=ON \


mascguy:
Steve, many of my ports have Devel equivalents, which should be updated instead. And such is the case for this one.

Also, the reason I haven't updated the non-Devel version yet, is that some features are being disabled due to ultra-strict version checks for dependencies. And I haven't had a chance to fix those yet, which is why `opencv4` hasn't been updated as well.

Net-Net, this PR is premature at best.

However, if you simply want to enable QUIRC support, please change this to update `opencv4-devel` instead.

mascguy:
Also, when you have multiple interdependent PRs like this - which require one or more other PR to be merged first - please clearly state that in the description. (And ensure dependent PRs like this one, are submitted as Draft.)

essandess:
> many of my ports have Devel equivalents, which should be updated instead

Okay. What I really care about here is the added cmake flag, so I'll modify to just turn that on.

As for the `-devel` ports, I tend to stay off the bleeding edge within my own workflows, so I'll simply wait for you or someone else to migrate to 4.6.0 at the appropriate time.

mascguy:
> As for the `-devel` ports, I tend to stay off the bleeding edge within my own workflows, so I'll simply wait for you or someone else to migrate to 4.6.0 at the appropriate time.

Well, for ports that I maintain, the Devel equivalents generally aren't bleeding edge. Rather, they're simply the latest release. (I depend on Devel ports to pre-validate buildbot builds, as well as pre-test new releases before rolling out to everyone.)

So you're safe to use Devel ports under my stewardship.

mascguy:
This PR is absolutely NOT ready for review, as the upstream PR for QUIRC hasn't been merged yet. Please don't change from draft status, until that's been done.

mascguy:
Okay, let me repeat this one more time: Please do NOT update the non-Devel PR.

Please drop that commit.

mascguy:
And update the PR description, to clearly call out the dependency on the QUIRC submission PR.

mascguy:
I think you meant to drop the commit for `opencv4`, rather than `opencv4-devel`?

essandess:
> I think you meant to drop the commit for `opencv4`, rather than `opencv4-devel`?

Um, no. I have a whole workflow with `opencv4` in the dependency tree. I'd simply like to add QUIRC support. May we please turn on this cmake option in `opencv4`?

mascguy:
> > I think you meant to drop the commit for `opencv4`, rather than `opencv4-devel`?
> 
> Um, no. I have a whole workflow with `opencv4` in the dependency tree. I'd simply like to add QUIRC support. May we please turn on this cmake option in `opencv4`?

The reason `opencv4` is non-openmaintainer, as that it's incredibly easy to break. That's why the Devel port is there, to pre-validate such changes.

Furthermore, all dependents are setup to support either `opencv4` or `opencv4-devel`, depending on what the user has installed. (I'll have to double check whether that's also the case for the Python subports.)

Net-Net, you can uninstall `opencv4`, and install `opencv4-devel` in its place. And everything will work automagically.

mascguy:
> Furthermore, all dependents are setup to support either `opencv4` or `opencv4-devel`, depending on what the user has installed. (I'll have to double check whether that's also the case for the Python subports.)

Python ports need to be fixed to support a path-style dep, and I'll take care of that ASAP. Will let you know when it's been merged to master.

mascguy:
> Python ports need to be fixed to support a path-style dep, and I'll take care of that ASAP. Will let you know when it's been merged to master.

Done.

You also need to re-add the dependency for `quirc`.

essandess:
> > Python ports need to be fixed to support a path-style dep, and I'll take care of that ASAP. Will let you know when it's been merged to master.
> 
> Done.
> 
> You also need to re-add the dependency for `quirc`.

I do not believe that there is a `quirc` dependency. The cmake file adds its own dependency, and doesn't use system libraries:

https://github.com/opencv/opencv/blob/5bb098ba5d87ee224bc92e20d7e864dcd3596c55/CMakeLists.txt#L834-L836

mascguy:
> I do not believe that there is a `quirc` dependency. The cmake file adds its own dependency, and doesn't use system libraries:
> 
> https://github.com/opencv/opencv/blob/5bb098ba5d87ee224bc92e20d7e864dcd3596c55/CMakeLists.txt#L834-L836

Shouldn't we use your new port, rather than the internal version? That would be highly preferable, even if it requires patching the CMake file.

mascguy:
> > I do not believe that there is a `quirc` dependency. The cmake file adds its own dependency, and doesn't use system libraries:
> > https://github.com/opencv/opencv/blob/5bb098ba5d87ee224bc92e20d7e864dcd3596c55/CMakeLists.txt#L834-L836
> 
> Shouldn't we use your new port, rather than the internal version? That would be highly preferable, even if it requires patching the CMake file.

And shame on upstream for not supporting an external version. Silliness...

mascguy:
> > > I do not believe that there is a `quirc` dependency. The cmake file adds its own dependency, and doesn't use system libraries:
> > > https://github.com/opencv/opencv/blob/5bb098ba5d87ee224bc92e20d7e864dcd3596c55/CMakeLists.txt#L834-L836
> > 
> > 
> > Shouldn't we use your new port, rather than the internal version? That would be highly preferable, even if it requires patching the CMake file.
> 
> And shame on upstream for not supporting an external version. Silliness...

Hmmm, presumably that's supported, if we also enable `HAVE_QUIRC`. Lol

mascguy:
> Shouldn't we use your new port, rather than the internal version? That would be highly preferable, even if it requires patching the CMake file.

Okay, so here's the bottom line: We don't want to use/build internal subprojects here, unless absolutely necessary. And no exception is needed in this case.

So please update to use the external version.

essandess:
> So please update to use the external version.

I have done this by setting the upstream flag `-DWITH_QUIRC:BOOL=ON` and confirm that it works.

mascguy:
> > So please update to use the external version.
> 
> I have done this by setting the upstream flag `-DWITH_QUIRC:BOOL=ON` and confirm that it works.

Please ensure you also add `-DBUILD_QUIRC:BOOL=OFF`, to enforce using the external version.

essandess:
@mascguy Thanks, I think I get it now—I didn't know I could pass `-DHAVE_QUIRC:BOOL=ON`.

Would you please advise on how to fix this `opencv` linking issue within the `opencv4-devel` Portfile?

I install #15994, then install `py310-opencv4-devel` using the flag `-DHAVE_QUIRC:BOOL=ON`.

But upon `import cv2`, python shows this linking error—it's looking in `/usr/local/lib` and `/usr/lib`, not `/opt/local/lib`.

I could always hack in an `install_name_tool` to fix this, but there must be a standard approach within the `opencv4` Portfile that's preferred.

Or perhaps somehow I haven't set an absolute path within `libquirc.1.0.dylib` when I built it? (Please see the `otool` output below.) That looks like the most likely explanation based on below, but I'm not sure what the link command error is that caused this.

```python
ImportError: dlopen(/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/cv2/python-3.10/cv2.cpython-310-darwin.so, 0x0002): Library not loaded: 'libquirc.1.0.dylib'
  Referenced from: '/opt/local/lib/opencv4/libopencv_objdetect.4.6.0.dylib'
  Reason: tried: 'libquirc.1.0.dylib' (no such file), '/usr/local/lib/libquirc.1.0.dylib' (no such file), '/usr/lib/libquirc.1.0.dylib' (no such file), '/Users/username/current-directory/libquirc.1.0.dylib' (no such file), '/usr/local/lib/libquirc.1.0.dylib' (no such file), '/usr/lib/libquirc.1.0.dylib' (no such file)
```

Also note that:
```sh
$ port contents quirc
Port quirc contains:
  /opt/local/bin/quirc-demo
  /opt/local/bin/quirc-scanner
  /opt/local/include/quirc.h
  /opt/local/lib/libquirc.1.0.dylib
  /opt/local/lib/libquirc.a
  /opt/local/lib/libquirc.dylib
```

```sh
$ otool -L /opt/local/lib/libquirc.1.0.dylib
/opt/local/lib/libquirc.1.0.dylib:
	libquirc.1.0.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

mascguy:
> Or perhaps somehow I haven't set an absolute path within `libquirc.1.0.dylib` when I built it? (Please see the `otool` output below.) That looks like the most likely explanation based on below, but I'm not sure what the link command error is that caused this.
> 
> ```shell
> $ otool -L /opt/local/lib/libquirc.1.0.dylib
> /opt/local/lib/libquirc.1.0.dylib:
> 	libquirc.1.0.dylib (compatibility version 0.0.0, current version 0.0.0)
> 	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
> ```

That doesn't look right to me either.

mascguy:
> > Or perhaps somehow I haven't set an absolute path within `libquirc.1.0.dylib` when I built it? (Please see the `otool` output below.) That looks like the most likely explanation based on below, but I'm not sure what the link command error is that caused this.
> > ```shell
> > $ otool -L /opt/local/lib/libquirc.1.0.dylib
> > /opt/local/lib/libquirc.1.0.dylib:
> > 	libquirc.1.0.dylib (compatibility version 0.0.0, current version 0.0.0)
> > 	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
> > ```
> 
> That doesn't look right to me either.

Please see my latest comment in your PR for `quirc`: My guess is that you're missing some compiler/link flags. And that's why you'll want to use portgroup `makefile`, to avoid the headaches with trying to get all of those right!

essandess:
Headaches is right with the Linux/BSD-oriented code base in https://github.com/dlbeer/quirc.

The only reason to consider merging #15994 is get get a dylib that can be used to provide `opencv` with QUIRC capabilities.

Given that upstream has already solved this issue with a simple `-DWITH_QUIRC` flag that integrates the C code within Python calls, it doesn't seem worthwhile to jump through all the hoops of converting upstream's Linux-oriented code into a macOS shared library.

@mascguy Isn't this an obvious case of just using upstream's `thirdparty` code in  #15993 and calling it day?

If anyone else needs a quirc shared library outside of `import cv2`, then they'll be motivated to sort through these issues.

I'm ready to throw in the towel on ##15994 and just use `-DWITH_QUIRC`  in `opencv4`.

essandess:
I fixed #15994 and verify that it works with this PR.

mascguy:
> The reason I haven't updated the non-Devel version yet, is that some features are being disabled due to ultra-strict version checks for dependencies. And I haven't had a chance to fix those yet, which is why `opencv4` hasn't been updated as well.

The final lingering issue has been fixed, via the following commit: https://github.com/macports/macports-ports/commit/6f5838653ab6794b6e8dca15181991f8790d4c61

So I'll be updating `opencv4` to 4.6.0 soon, over the next day or two. (This doesn't impact your PR, but just a heads-up.)

graphics/opencv4-devel/Portfile
```diff
@@ -283,6 +284,7 @@ configure.args-append \
                     -DWITH_IMGCODEC_PFM:BOOL=ON \
                     -DWITH_QUIRC:BOOL=OFF \
                     \
+                    -DHAVE_QUIRC:BOOL=ON \
```
mascguy, line 287: You need to change this to `-DBUILD_QUIRC:BOOL=OFF` and `-DWITH_QUIRC:BOOL=ON`.

And move to the earlier section, where we pair these together.

mascguy (changes requested):

graphics/opencv4-devel/Portfile
essandess: Are you sure about that? It’s true that I hacked in a `BUILD_QUIRC` variable into `libquirc`‘s Makefile, but that is unrelated and I’m pretty sure that this is the correct setting the the `cmake` build in this repo.

mascguy: > Are you sure about that? It’s true that I hacked in a `BUILD_QUIRC` variable into `libquirc`‘s Makefile, but that is unrelated and I’m pretty sure that this is the correct setting the the `cmake` build in this repo.

I'll double-check tomorrow.

mascguy (approved):
LGTM, and validated locally.

This is also a nice addition for our base `opencv4` installation. Thanks for the work on this Steve!

mascguy:
Oh, we also need a rev-bump. Just took care of that.

PR #15994, opened 2022-09-04
libquirc: new port

essandess:
### Description

* Add new port `quirc`
* NOTE: Needs to be merged before: https://github.com/macports/macports-ports/pull/15993

### Type(s)
- [x] bugfix
- [x] enhancement
- [ ] security fix

### Tested on
macOS 12.5.1 21G83 x86_64
Xcode 13.4.1 13F100

### Verification 
Have you


graphics/quirc/Portfile
```diff
@@ -0,0 +1,80 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        dlbeer quirc 1.1 v
+revision            0
+categories-append   math
```
mascguy, line 8: This should be changed to: `categories graphics math`

Per lint errors:

```
port lint quirc: Error: Portfile parent directory graphics does not match primary category math
```

essandess: Thanks done.

mascguy:
Steve, relative to your new port, have you tried using portgroup `makefile`? It takes care of the vast majority of what you're doing, particularly in terms of setting environment variables, compile/link flags, etc, etc.

kencu:
you need a bit more than this when you make a library:
```
$(CC) -dynamiclib -o $@ $(LIB_OBJ) $(LDFLAGS) -lm
```

you need to also set an install_name, and some versions for current_version and compatibilty_version, and headerpad_max_install_names.

see, for example, the flags here (but don't copy those as they are not right for this project):

https://github.com/macports/macports-legacy-support/blob/03a02c1fb64e1ca254cb7c563ec367b12bd7fb18/Makefile#L39

mascguy:
> you need a bit more than this when you make a library:
> 
> ```
> $(CC) -dynamiclib -o $@ $(LIB_OBJ) $(LDFLAGS) -lm
> ```

Yep, that's why I'd strongly recommend using pg `makefile`. LOL

kencu:
> Yep, that's why I'd strongly recommend using pg `makefile`. LOL

Nope, the Makefile PG will provide exactly none of that.

mascguy:
> > Yep, that's why I'd strongly recommend using pg `makefile`. LOL
> 
> Nope, the Makefile PG will provide exactly none of that.

Maybe not, but it will still simplify this all significantly.

devel/sdl12-compat/Portfile
```diff
@@ -26,13 +26,15 @@ depends_lib     port:libsdl2
...
+                SDL_stdinc.h.patch \
+                patch-SDL_endian_h.diff
+
 post-patch {
     reinplace "s|__PREFIX__|${prefix}|" ${worksrcpath}/src/SDL12_compat.c
 }
 
 cmake.out_of_source yes
-configure.args-append   -DSDL2_PATH=${prefix}
+configure.args-append   -DSDL2_PATH=${prefix} -D__OpenBSD__=
```
jmroot, line 37: Defining a reserved identifier that indicates that we're on a different OS than we actually are is going to cause nothing but trouble. What problem are you trying to solve?

jmroot (changes requested):

devel/sdl12-compat/Portfile
essandess: > Defining a reserved identifier that indicates that we're on a different OS than we actually are is going to cause nothing but trouble. What problem are you trying to solve?

This problem:

https://github.com/dlbeer/quirc/blob/9c0f555acb2531d818f2c405044418fc75379fb2/demo/camera.c#L26-L30

Don't define the variable, get a Linux include. Define the variable and get something that compiles. It's the only appearance of this variable in upstream.

essandess:
Arguably, the only reason to consider merging this PR is get get a dylib that can be used to provide `opencv` with QUIRC capabilities.

Given that upstream has already solved this issue with a simple `-DWITH_QUIRC` flag that integrates the C code within Python calls, it doesn't seem worthwhile to jump through all the hoops of converting upstream's Linux-oriented code into a macOS shared library.

@mascguy Isn't this an obvious case of just using upstream's `thirdparty code in  #15993 and calling it day?

If anyone else needs a quirc shared library outside of `import cv2`, then they'll be motivated to sort through these issues.

I'm ready to throw in the towel on this PR and just use `-DWITH_QUIRC`  in `opencv4`.

devel/sdl12-compat/Portfile
essandess: Whoops, this problem too:

https://github.com/libsdl-org/sdl12-compat/blob/6d68bff0e950be9c550f3b3bc60e6fa4a86edd7d/include/SDL/SDL_endian.h#L41-L45

macOS uses `<machine/endian.h>`, which isn't represented upstream. Again, this is the only place this variable is used in compilation.

If it's not defined, then no `endian.h` can be found, and build like this one that use `libsdl_gfx` don't compile.

Whatever we do with the `quirc` port, some version of the fix this commit addresses should be merged.

essandess: I used the `__APPLE__` variable, used elsewhere in upstream.

essandess:
> > > Yep, that's why I'd strongly recommend using pg `makefile`. LOL
> > 
> > 
> > Nope, the Makefile PG will provide exactly none of that.
> 
> Maybe not, but it will still simplify this all significantly.

I didn't see how using the makefile PG would simplify any of this. Upstream's Makefile is very hardcoded and one must patch in explicit commands by hand, and pass in flags by hand.

mascguy:
> I didn't see how using the makefile PG would simplify any of this. Upstream's Makefile is very hardcoded and one must patch in explicit commands by hand, and pass in flags by hand.

Bear in mind that there are layers in play - pg makefile adds things on top of sensible defaults set earlier in the chain - so it may not be immediately apparent based on review of the pg itself.

Instead, try including it, and test with all of the following removed. And review `main.log`, to see what args and env vars are being set. I think you might be pleasantly surprised...

```
compiler.cpath-append \
                    ${workpath}/include

configure.cflags-append \
                    -I${workpath}/include

build.post_args-append \
                    CC="${configure.cc}" \
                    CFLAGS="${configure.cflags} [get_canonical_archflags cc]" \
                    CPPFLAGS="${configure.cppflags}" \
                    LDFLAGS="${configure.ldflags} [get_canonical_archflags ld]"
```

mascguy:
> Instead, try including it, and test with all of the following removed. And review `main.log`, to see what args and env vars are being set. I think you might be pleasantly surprised...

I could always be completely wrong though - hey, it happens! - so perhaps not. But perhaps I'll grab your latest updates, and do some testing myself.

kencu:
using the Makefile PG will undoubtedly simplify the Portfile, as Chris described.

But you still need to make the library properly. Setting the name wasn’t so hard! The other bits I mentioned are equally trivial.

And then someone won’t have to come along 10 minutes after you commit this and fix it properly…

devel/sdl12-compat/Portfile
```diff
@@ -5,7 +5,7 @@ PortGroup       cmake 1.0
 PortGroup       github 1.0
 
 github.setup    libsdl-org sdl12-compat 1.2.52 release-
-revision        1
+revision        2
```
mascguy, line 8: With the latest fixes just added by @jmroot, the revision is now at 2. So you'll need to bump this to 3.

mascguy: > With the latest fixes just added by @jmroot, the revision is now at 2. So you'll need to bump this to 3.

Ditto for rebasing against master, to pick up the rest of the changes.

mascguy:
Based on some quick local testing on 10.15 - and please verify this yourself, on your setup - all of the following can be removed:

```
compiler.cpath-append \
                    ${workpath}/include

configure.cflags-append \
                    -I${workpath}/include

build.post_args-append \
                    CC="${configure.cc}" \
                    CFLAGS="${configure.cflags} [get_canonical_archflags cc]" \
                    CPPFLAGS="${configure.cppflags}" \
                    LDFLAGS="${configure.ldflags} [get_canonical_archflags ld]"
```

Unless I'm missing something...?

mascguy:
Testing on older macOS releases - which we absolutely need to support, if you want to enable QUIRC by default for `opencv4` - the following failure is seen:

```
:info:build libquirc/work/compwrap/cc/usr/bin/clang -dynamiclib -install_name /opt/local/lib/libquirc.1.0.dylib -o libquirc.1.0.dylib lib/decode.o lib/identify.o lib/quirc.o lib/version_db.o -Wl,-syslibroot, -arch x86_64 -lm
:info:build ld: library not found for -lSystem
```

mascguy:
> Testing on older macOS releases - which we absolutely need to support, if you want to enable QUIRC by default for `opencv4` - the following failure is seen:
> 
> ```
> libquirc/work/compwrap/cc/usr/bin/clang -dynamiclib -install_name /opt/local/lib/libquirc.1.0.dylib -o libquirc.1.0.dylib lib/decode.o lib/identify.o lib/quirc.o lib/version_db.o -Wl,-syslibroot, -arch x86_64 -lm
> ld: library not found for -lSystem
> ```

@kencu, can you provide some tips/pointers for Steve, to get this working correctly across-the-board?

devel/sdl12-compat/Portfile
essandess: Thanks done.

mascguy:
Was the addition of `hidapi` intentional...?

essandess:
> Based on some quick local testing on 10.15 - and please verify this yourself, on your setup - all of the following can be removed:
> 
> ```
> compiler.cpath-append \
>                     ${workpath}/include
> 
> configure.cflags-append \
>                     -I${workpath}/include
> 
> build.post_args-append \
>                     CC="${configure.cc}" \
>                     CFLAGS="${configure.cflags} [get_canonical_archflags cc]" \
>                     CPPFLAGS="${configure.cppflags}" \
>                     LDFLAGS="${configure.ldflags} [get_canonical_archflags ld]"
> ```
> 
> Unless I'm missing something...?

Yes. I've done this by using PG `makefile`.

However, note that I've added a variant that builds the full upstream, and this variant quires the OpenBSD include file that is added to `${workpath}/incldue`, so those lines have to stay.

essandess:
> 



> you still need to make the library properly. Setting the name wasn’t so hard! The other bits I mentioned are equally trivial.

Okay thanks! Added those too.

mascguy:
> Testing on older macOS releases - which we absolutely need to support, if you want to enable QUIRC by default for `opencv4` - the following link failure is seen:
> 
> ```
> libquirc/work/compwrap/cc/usr/bin/clang -dynamiclib -install_name /opt/local/lib/libquirc.1.0.dylib -o libquirc.1.0.dylib lib/decode.o lib/identify.o lib/quirc.o lib/version_db.o -Wl,-syslibroot, -arch x86_64 -lm
> ld: library not found for -lSystem
> ```

Since I had to update ports across all of my VMs anyway, I also tested your latest changes. And looks good across-the-board, from 10.6 thru Monterey.

Nice work!

mascguy (approved):
Now builds across-the-board, and the latest streamlining looks good.

@jmroot, are you good with the changes for `sdl12-compat`?

essandess:
> > Testing on older macOS releases - which we absolutely need to support, if you want to enable QUIRC by default for `opencv4` - the following link failure is seen:
> > ```
> > libquirc/work/compwrap/cc/usr/bin/clang -dynamiclib -install_name /opt/local/lib/libquirc.1.0.dylib -o libquirc.1.0.dylib lib/decode.o lib/identify.o lib/quirc.o lib/version_db.o -Wl,-syslibroot, -arch x86_64 -lm
> > ld: library not found for -lSystem
> > ```
> 
> Since I had to update ports across all of my VMs anyway, I also tested your latest changes. And looks good across-the-board, from 10.6 thru Monterey.
> 
> Nice work!

Thanks is due to @kencu and you—I just did what you told me to do. Thanks!

graphics/libquirc/Portfile
```diff
@@ -0,0 +1,91 @@
...
+
+    destroot {
+        xinstall -m 0644 \
+                    ${worksrcpath}/${build_dylib} \
+                    ${destroot}${prefix}/lib/${build_dylib}
+        xinstall -m 0644 \
+                    ${worksrcpath}/lib/${github.project}.h \
+                    ${destroot}${prefix}/include
+    }
+}
```
mascguy, line 81: I believe this line has trailing whitespace, which you'll want to fix.

And be sure to run `port lint --nitpick` one final time, to ensure there's nothing else.

essandess:
An intriguing question on a possible capability of this port:

The default video path is `/dev/video0`:

https://github.com/dlbeer/quirc/blob/9c0f555acb2531d818f2c405044418fc75379fb2/demo/demo.c#L31

MacPorts docs for `ffmpeg` explain how to list the supported, connected capture devices: https://trac.ffmpeg.org/wiki/Capture/Webcam#OSX. For example, one of my devices is:
```
[AVFoundation indev @ 0x7fb06200a191] [2] OWC Thunderbolt 3 Audio Devic
```

How can this code be adapted to use one of the attached cameras on the box?

graphics/libquirc/Portfile
essandess: Thanks, all done.

devel/sdl12-compat/Portfile
jmroot: As I said earlier, the changes to sdl12-compat don't seem to be necessary at all.

jmroot (changes requested):

jmroot:
> @jmroot, are you good with the changes for `sdl12-compat`?

No, see review comments. (Sorry about not getting those posted earlier; GitHub helpfully accepted them but marked them as "pending", which I didn't notice at first…)

devel/sdl12-compat/files/patch-SDL_endian_h.diff
```diff
@@ -0,0 +1,13 @@
+diff -NaurdwB ./include/SDL/SDL_endian.h
+--- ./include/SDL/SDL_endian.h	2022-09-02 18:20:22.000000000 -0400
++++ ./include/SDL/SDL_endian.h	2022-09-05 08:24:53.000000000 -0400
+@@ -41,6 +41,9 @@
+ #ifdef __linux__
+ #include <endian.h>
+ #define SDL_BYTEORDER  __BYTE_ORDER
++#elif defined(__APPLE__)
```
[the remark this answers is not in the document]
essandess: Doh that’s right. I’ll fix it. Thanksyou.

mascguy:
> An intriguing question on a possible capability of this port:
> 
> The default video path is `/dev/video0`:
> 
> https://github.com/dlbeer/quirc/blob/9c0f555acb2531d818f2c405044418fc75379fb2/demo/demo.c#L31
> 
> MacPorts docs for `ffmpeg` explain how to list the supported, connected capture devices: https://trac.ffmpeg.org/wiki/Capture/Webcam#OSX. For example, one of my devices is:
> 
> ```
> [AVFoundation indev @ 0x7fb06200a191] [2] OWC Thunderbolt 3 Audio Devic
> ```
> 
> How can this code be adapted to use one of the attached cameras on the box?

Steve, is this still an open question? Or does the addition of `videoio.h` take care of it?

mascguy:
One other comment: We should rename the variant `quirc` to something else, like `full`. Or something more akin to what it accomplishes.

essandess:
> Steve, is this still an open question? Or does the addition of `videoio.h` take care of it?

This is a hacker's open question for another day. I'll push the new variant name.

mascguy (approved):
LGTM

PR #16007, opened 2022-09-05
imlib2: disable building docs on Rosetta

barracuda156:
Fixes: https://trac.macports.org/ticket/65770

#### Description

Doxygen is broken on Rosetta, so apply default variant `+doc` in all cases except for that one.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/graphics/imlib2/Portfile
+++ b/graphics/imlib2/Portfile
@@ -66,7 +66,7 @@ post-destroot {
         TODO ${destroot}${docdir}
 }
 
-variant doc {
+variant doc description {Build documentation} {
     depends_lib-append \
                     path:bin/doxygen:doxygen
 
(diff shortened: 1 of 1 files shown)


kencu:
1. I would like to see evidence that using legacysupports binwrapping fix is ineffective for doxygen before any more of these doxygen blockers are committed. Maybe it doesn't work, but we have yet to see that I believe. If it has been already proven that it won't work, my apologies, I must have missed it of forgot about it. Also, it would be nice for at least one other person to verify that it doesn't work before that is taken as a certainty.
2. if doxygen does trigger the free-after-malloc error on gcc/libgcc builds, it should be an across-the-board issue for gcc builds, and having a "10.6 on PPC" specific fix for it in the port is never an ideal plan.

barracuda156:
> * I would like to see evidence that using legacysupports binwrapping fix is ineffective for doxygen before any more of these doxygen blockers are committed. Maybe it doesn't work, but we have yet to see that I believe. If it has been already proven that it won't work, my apologies, I must have missed it or forgot about it. Also, it would be nice for at least one other person to verify that it doesn't work before that is taken as a certainty to prevent Heisenbergs.

@kencu Converted to draft, gonna try that tomorrow and update.

PR #16008, opened 2022-09-05
libproxy: update to 0.4.18

barracuda156:
#### Description

There are no changes to the code.
First commit updates the port to current version. Second makes a few minor non-functional improvements to portfile (adds -append, sorts dependencies alphabetically).

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/net/libproxy/Portfile
+++ b/net/libproxy/Portfile
@@ -64,6 +64,9 @@ configure.args-append \
 # Fix an issue with clang-900 (Xcode 9) using the Sierra SDK, <rdar://problem/31263056>
 configure.cppflags-append -D_DARWIN_C_SOURCE=1
 
+# Set standard via cxx flags, as cmake.set_cxx_flags is not working properly
+configure.cxxflags-append -std=c++11
+
 variant python27 conflicts python38 python36 python37 python39 description {Build Bindings for Python 2.7} {
     set python_prefix ${frameworks_dir}/Python.framework/Versions/2.7
     depends_lib-append port:python27
--- a/net/libproxy/Portfile
+++ b/net/libproxy/Portfile
@@ -4,12 +4,11 @@ PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           cmake 1.1
 
-github.setup        libproxy libproxy 0.4.17
+github.setup        libproxy libproxy 0.4.18
 revision            0
 epoch               1
 categories          net
 maintainers         {devans @dbevans} openmaintainer
-platforms           darwin
 license             LGPL-2.1+
 homepage            https://libproxy.github.io/libproxy
 
@@ -19,9 +18,9 @@ long_description    Libproxy exists to answer the question: Given a \
                     network resource, how do I reach it? It handles all \
                     the details, enabling you to get back to programming.
 
-checksums           rmd160  4440c0024eaf3b958eba049d0b180a872bb84045 \
-                    sha256  1a40d344d61eee9d962cbdcf9b382255a63ed5a15c487384cf94174503dfffdb \
-                    size    95576
+checksums           rmd160  18680c62196d6b1765864ac8d8cfb0da19e96fb3 \
+                    sha256  712f5965b57dc329fd5700f906d9943dbdc379e547e31dc8b11c77332a652d52 \
+                    size    98063
 
 patchfiles          patch-libproxy-cmake.diff \
                     patch-libproxy-test-CMakeLists.txt.diff \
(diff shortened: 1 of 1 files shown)


barracuda156:
Hmm, bot fails on:
```
/opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_net_libproxy/libproxy/work/libproxy-0.4.18/libproxy/modules/config_kde.cpp:211:31: error: expected expression
          auto getenv_default = [](const char *name, const std::string &def) {
                                ^
  1 warning and 1 error generated.
  make[2]: *** [libproxy/CMakeFiles/config_kde.dir/modules/config_kde.cpp.o] Error 1
```

barracuda156:
@mascguy Fixed now. Apparently `cxx_standard` 2011 does not add standard flag itself. Adding `-std=c++11` solves the problem, like here: https://clay-atlas.com/us/blog/2021/09/18/cpp-en-warning-auto-extension/

mascguy:
> @mascguy Fixed now. Apparently `cxx_standard` 2011 does not add standard flag itself. Adding `-std=c++11` solves the problem, like here: https://clay-atlas.com/us/blog/2021/09/18/cpp-en-warning-auto-extension/

For projects using the CMake 1.1 portgroup - like this one - you should use the following instead:

```
cmake.set_cxx_standard  yes
```

barracuda156:
> > @mascguy Fixed now. Apparently `cxx_standard` 2011 does not add standard flag itself. Adding `-std=c++11` solves the problem, like here: https://clay-atlas.com/us/blog/2021/09/18/cpp-en-warning-auto-extension/
> 
> For projects using the CMake 1.1 portgroup - like this one - you should use the following instead:
> 
> ```
> cmake.set_cxx_standard  yes
> ```

@mascguy Perhaps I misunderstood. This still does not add `-std=c++11` and we get same error as before. Let me add back the flag.

mascguy:
> @mascguy Perhaps I misunderstood. This still does not add `-std=c++11` and we get same error as before. Let me add back the flag.

No, you understood correctly. And `cmake.set_cxx_standard` is the preferred approach.

You may need to patch upstream's CMake file, as it looks like they're also specifying it.

barracuda156:
> > @mascguy Perhaps I misunderstood. This still does not add `-std=c++11` and we get same error as before. Let me add back the flag.
> 
> No, you understood correctly. And `cmake.set_cxx_standard` is the preferred approach.
> 
> You may need to patch upstream's CMake file, as it looks like they're also specifying it.

@mascguy Yeah, `cmake.set_cxx_standard` fails with or without the flag added. Let me see what can be done with CMakeLists.

mascguy:
> @mascguy Yeah, `cmake.set_cxx_standard` fails with or without the flag added. Let me see what can be done with CMakeLists.

Does it work when you don't use `cmake.set_cxx_standard`, combined with setting `-std=c++11`?

barracuda156:
> > @mascguy Yeah, `cmake.set_cxx_standard` fails with or without the flag added. Let me see what can be done with CMakeLists.
> 
> Does it work when you don't use `cmake.set_cxx_standard`, combined with setting `-std=c++11`?

@mascguy Yes, all checks passed before. Let me revert, and we hopefully get a reproducible success.

barracuda156:
@mascguy So yeah, like this it builds fine.

mascguy:
> @mascguy So yeah, like this it builds fine.

I'm not seeing anything immediately-obvious in the CMake files, so that's fine. But add the following comment, so folks know that you tried doing the right thing:

```
# Set standard via cxx flags, as cmake.set_cxx_flags not working properly
configure.cxxflags-append -std=c++11
```

mascguy:
Ah, I think (?) this might be due to one of the policies that upstream is enabling, via their top-level `CMakeLists.txt`. Didn't confirm, but might be this:

https://cmake.org/cmake/help/latest/policy/CMP0067.html

mascguy:
> Ah, I think (?) this might be due to one of the policies that upstream is enabling, via their top-level `CMakeLists.txt`. Didn't confirm, but might be this:
> 
> https://cmake.org/cmake/help/latest/policy/CMP0067.html

But we probably don't want to patch that out, unless we're absolutely certain it doesn't cause other issues. But just for awareness.

mascguy:
> > Ah, I think (?) this might be due to one of the policies that upstream is enabling, via their top-level `CMakeLists.txt`. Didn't confirm, but might be this:
> > https://cmake.org/cmake/help/latest/policy/CMP0067.html
> 
> @mascguy Should we leave it for now like this (I added the comment re cxx flags) or try to do smth about `CMakeLists.txt`?

Nope, let's leave the CMake files alone, because it's possible that we could break something else.

mascguy (approved):

PR #16042, opened 2022-09-08
openjdk{11,17}-graalvm: add native-image subport

kakuhen:
#### Description
This port is part of an effort to bring [clojure-lsp](https://github.com/clojure-lsp/clojure-lsp/) into MacPorts. While it is possible to build clojure-lsp into a standalone JAR file; this is not recommended by upstream, and they plan to remove any JAR releases soon. Rather than make a future clojure-lsp port deviate from upstream releases, I have decided to attempt bringing GraalVM's native-image tool as naturally as possible to MacPorts.

I have been able to install and uninstall each subport locally. I have also successfully invoked `native-image` from the java17 subport to produce a native executable of a Java hello world program.

###### Tested on

macOS 10.15.7 19H2026 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/java/openjdk11-graalvm/Portfile
+++ b/java/openjdk11-graalvm/Portfile
@@ -86,3 +86,48 @@ by adding the following line to your shell profile:
 
     export JAVA_HOME=${target}/Contents/Home
 "
+
+subport ${name}-native-image {
+    depends_lib        port:${name}
+    description        Native Image component for GraalVM
+    long_description   ${description}
+
+    if {${configure.build_arch} eq "x86_64"} {
+        set jar_file native-image-installable-svm-java11-darwin-amd64-${version}.jar
+        distfiles    ${jar_file}
+        checksums    rmd160  b484f8c9b78afe5230a1c712a2c2e83728994906 \
+                     sha256  de9bf830d000a54934a01149691a9a8d4ef6e33414776abd14f95c65b149c908 \
+                     size    110346461
+    } elseif {${configure.build_arch} eq "arm64"} {
+        set jar_file native-image-installable-svm-java11-darwin-aarch64-${version}.jar
+        distfiles    ${jar_file}
+        checksums    rmd160  30811497def96c32a75940616a98efd4f14870ca \
+                     sha256  aba76d671017f93cdaae5102607d0bc7a1398adc5de8a4b1e308fa366d5983f9 \
+                     size    27717768
+    }
+
+    set java_home ${target}/Contents/Home
+
+    extract {}
+
+    destroot {
+        xinstall -d -m 0755 ${destroot}${prefix}/share/java/${subport}
+        file copy ${distpath}/${jar_file} ${destroot}${prefix}/share/java/${subport}
+    }
+
+    post-activate {
+        # Graal Updater doesn't signal errors if the component is already installed.
+        # Unfortunately, we require root privileges to invoke Graal Updater.
+        system "sudo ${java_home}/bin/gu -L install ${prefix}/share/java/${subport}/${jar_file}"
+    }
+
+    post-deactivate {
+        # This helps prevent breakage if the user removed native-image themselves
+        # and wants to deactivate or uninstall this port.
+        if {[regexp {(?i)native-image} [exec ${java_home}/bin/gu list] match]} {
+            system "sudo ${java_home}/bin/gu remove native-image"
+        }
+    }
+
+    notes "The Native Image component of GraalVM has been installed for you"
+}
--- a/java/openjdk17-graalvm/Portfile
+++ b/java/openjdk17-graalvm/Portfile
@@ -86,3 +86,48 @@ by adding the following line to your shell profile:
 
     export JAVA_HOME=${target}/Contents/Home
 "
+
+subport ${name}-native-image {
+    depends_lib        port:${name}
+    description        Native Image component for GraalVM
+    long_description   ${description}
+
+    if {${configure.build_arch} eq "x86_64"} {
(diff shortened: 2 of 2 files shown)


catap:
Cc: @breun

breun:
I'd like to suggest adding variants to the existing `openjdk*-graalvm` ports for GraalVM components that 'normally' (outside MacPorts) can be installed via `gu install <component>` (see [docs](https://www.graalvm.org/22.2/reference-manual/graalvm-updater/#install-components-on-graalvm-community)), like `native-image`, `js`, etc.

```
❯ /Library/Java/JavaVirtualMachines/openjdk17-graalvm/Contents/Home/bin/gu available                    
Downloading: Component catalog from www.graalvm.org
ComponentId              Version             Component name                Stability                     Origin 
---------------------------------------------------------------------------------------------------------------------------------
espresso                 22.2.0              Java on Truffle               Experimental                  github.com
espresso-llvm            22.2.0              Java on Truffle LLVM Java librExperimental                  github.com
js                       22.2.0              Graal.js                      Supported                     github.com
llvm                     22.2.0              LLVM Runtime Core             Experimental                  github.com
llvm-toolchain           22.2.0              LLVM.org toolchain            Supported                     github.com
native-image             22.2.0              Native Image                  Early adopter                 github.com
nodejs                   22.2.0              Graal.nodejs                  Supported                     github.com
python                   22.2.0              Graal.Python                  Experimental                  github.com
R                        22.2.0              FastR                         Experimental                  github.com
ruby                     22.2.0              TruffleRuby                   Experimental                  github.com
visualvm                 22.2.0              VisualVM                      Experimental                  github.com
wasm                     22.2.0              GraalWasm                     Experimental                  github.com
```

mascguy:
> I'd like to suggest adding variants to the existing `openjdk*-graalvm` ports for GraalVM components that 'normally' (outside MacPorts) can be installed via `gu install <component>` (see [docs](https://www.graalvm.org/22.2/reference-manual/graalvm-updater/#install-components-on-graalvm-community)), like `native-image`, `js`, etc.

So would that be done too, after the addition of this port? Or are you suggesting it as an alternative? (It isn't clear what the implications are...)

breun:
I was suggesting it as an alternative, but it could also be implemented like proposed in this MR. I don't know what would be considered the 'best' or 'nicest' way to add support for the additional optional GraalVM components (`native-image` is one of the 12 that are currently available): variants or separate subports.

kakuhen:
> I don't know what would be considered the 'best' or 'nicest' way to add support for the optional GraalVM components

Me neither.  As long as there is a way for ports to require specific variants, I think your approach is better than mine, however. For instance, a user with `openjdk17-graalvm +native-image +js` should be able to install ports requiring `openjdk17-graalvm +native-image`. Users without the variant should be prompted to install the port with this variant. I know some GTK dependent ports already do this by requiring e.g. `gtk3 +x11` instead of `gtk3 +quartz`

The end goal of this PR is to be able to submit clojure-lsp as a port, whose port makes use of native-image to build a native executable. As long as the situation I described above is possible, then I think rolling out GraalVM components as variants is the way to go.

mascguy:
There is one key advantage to subports, though: Dependency management is much easier, eliminating variant checks... and the user headaches that come along with it.

Instead, you can simply depend on the subport you need, and that's it. Clean and simple!

breun:
I guess subports are the way to go then.

I think it would be good to add these subports to the existing `openjdk*-graalvm` Portfiles and to also align their names with their parent ports, e.g. `openjdk*-graalvm-native-image`.

kakuhen:
OK. I have modified the commit message and created subports for the existing, non-obsolete GraalVM ports. You may notice I have appended the subports to the end of the Portfiles. This is to allow for easier expansion later on for e.g. GraalJS, Python, Truffle, etc.

catap:
@kakuhen is is possible to install both `openjdk11-graalvm` and `openjdk11-graalvm-native-image`?

kakuhen:
Yes. This is what I have on my machine:
```
% port installed | grep graalvm
  openjdk11-graalvm @22.2.0_0 (active)
  openjdk11-graalvm-native-image @22.2.0_0 (active)
  openjdk17-graalvm @22.2.0_0 (active)
  openjdk17-graalvm-native-image @22.2.0_0 (active)
```

Installing the `openjdk11-graalvm-native-image` subport will store a local copy of a JAR installer and use it to install Native Image when you activate the `openjdk11-graalvm-native-image` port. When you deactivate the subport, Native Image is removed. The installer JARs are stored in /opt/local/share/java/${subport}, so you can have multiple JDK versions together with local copies of the Native Image component.

breun:
This looks good to me.

kakuhen:
> Installation by name requires network access, so installing from a locally available file (provided by the subport) might indeed be nicer.

The rationale behind doing this is to keep the distribution under the Portfile's control. It also let's us verify checksums of the installer and ensure an updated component does not accidentally introduce breakage.

>it is not clear to me yet how the native image installation JAR is put in the local /opt/local/share/java/${subport} directory.

In short, I create a folder in ${prefix}/share/java named ${subport} then copy the downloaded JAR to that folder. If you would like, we can refactor the Portfile to make the process easier to understand.

breun:
Looking at the diff I didn’t see where the JAR was coming from, but the subport of course just reuses the `master_sites` from the ‘parent’ port. My bad. I think it’s fine like this.

mascguy:
Are you folks ready for merge, or is there further discussion needed?

breun:
I think it's ready to get merged.

mascguy (approved):

PR #16050, opened 2022-09-08
coreutils*: fix build on old macOS

catap:
#### Description

Used patch was backported to upstream of gnulib: https://lists.gnu.org/archive/html/bug-gnulib/2022-09/msg00045.html

See: https://trac.macports.org/ticket/62994

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

###### Verification 
Have you

### DIFF
--- a/sysutils/coreutils-devel/Portfile
+++ b/sysutils/coreutils-devel/Portfile
@@ -48,6 +48,9 @@ depends_lib-append \
 compiler.blacklist-append \
                 {clang < 900}
 
+# See: https://trac.macports.org/ticket/62994
+patchfiles      patch-getcwd.m4.diff
+
 configure.args-append \
                 --disable-silent-rules \
                 --program-prefix=g
--- a/sysutils/coreutils-devel/files/patch-getcwd.m4.diff
+++ b/sysutils/coreutils-devel/files/patch-getcwd.m4.diff
@@ -1,0 +1,13 @@
+Backport of https://lists.gnu.org/archive/html/bug-gnulib/2022-09/msg00045.html
+
+--- ./configure
++++ ./configure
+@@ -42907,7 +42907,7 @@
+ 
+   gl_abort_bug=no
+   case "$host_os" in
+-    mingw*)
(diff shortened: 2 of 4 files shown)


catap:
Used patch was backported to upstream of gnulib: https://lists.gnu.org/archive/html/bug-gnulib/2022-09/msg00045.html

mascguy:
Kirill, can you update `coreutils-devel` instead? Once it's been merged, I'll add the fix to `coreutils`.

catap:
@mascguy the path from `gnulib` to `coreutils` is a bit longer, and `coreutils-devel` isn't affected by this bug.

But I feel that this patch should be adde to both ports.

mascguy:
> @mascguy the path from `gnulib` to `coreutils` is a bit longer, and `coreutils-devel` isn't affected by this bug.
> 
> But I feel that this patch should be adde to both ports.

Sounds good. I'll test on my 10.5 machines - both Intel and PPC - either today or tomorrow. Once that's done, we can merge. (And thank you so much for the fix!)

Would this also apply to other ports affected by the same bug, like Bison?

catap:
> Would this also apply to other ports affected by the same bug, like Bison?

It should work, but the right place should be decided :)

You may see how I've backported another patch from gnulib to macports:
 - https://github.com/macports/macports-ports/blob/master/textproc/libunistring/files/patch-clang-before-34.diff
 - https://github.com/macports/macports-ports/blob/master/textproc/texinfo/files/patch-clang-before-34.diff
 - https://github.com/macports/macports-ports/blob/master/devel/m4/files/patch-clang-before-34.diff
 
 This is the same patch, but it is applied in the different place :(

catap:
Thus, this can be applied to `configure` that removed requirements of autoreconf. Let me make a force push.

catap:
Converted it to draft because upstream suggested much better way to solve it.

Will do it in coming days.

PR #16077, opened 2022-09-10
glfw: add v3.2.1 as fallback for 10.7-10.8

kencu:
closes: https://trac.macports.org/ticket/65506

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.7
Xcode 4.6.3 / 4H1503

patches from git history 20190214 prior to 3.3 update
installed, library names and links verified
not actually tested (is there a way)?

I believe from the build history this is also the last version that built on 10.8


michaelld (approved):
LGTM ; thanks!

PR #16085, opened 2022-09-11
cups: new port

usersxx:
#### Description
CUPS: open source printing system that supports printing to local and network printers

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/print/cups/Portfile
+++ b/print/cups/Portfile
@@ -65,4 +65,4 @@ build.target        all
 build.args          
 
 destroot.target     install
-destroot.post_args  BUILDROOT=${destroot}
+destroot.post_args  BUILDROOT=${destroot} 
--- a/print/cups/Portfile
+++ b/print/cups/Portfile
@@ -1,0 +1,68 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        OpenPrinting cups 2.4.2 v
+github.tarball_from releases
+distname            ${github.project}-${github.version}-source
+revision            0
+categories          print
+license             Apache-2
+maintainers         nomaintainer
+description         CUPS is the current standards-based, open source printing system developed by \
+                    OpenPrinting for Unix-like operating systems and uses IPP Everywhere™ to support \
+                    printing to local and network printers.
+long_description    OpenPrinting CUPS is the most current version of CUPS, a standards-based, \
+                    open source printing system for Unix-like operating systems. CUPS supports \
+                    printing to AirPrint™ and IPP Everywhere™ printers, \
+                    Network and local (USB) printers with Printer Applications, and \
+                    Network and local (USB) printers with (legacy) PPD-based printer drivers. \
+                    CUPS provides the System V ("lp") and Berkeley ("lpr") command-line interfaces, \
+                    a configurable web interface, a C API, and common print filters, drivers, and backends \
+                    for printing. The cups-filters project provides additional filters and drivers.
+homepage            https://openprinting.github.io/cups/
+
+checksums           rmd160  58a24fe177cfb66157899b6a53a90972a6080e74 \
+                    sha256  f03ccb40b087d1e30940a40e0141dcbba263f39974c20eb9f2521066c9c6c908 \
+                    size    8128785
+
+depends_build       port:gmake \
+                    port:autoconf \
+                    port:pkgconfig
+depends_lib         port:zlib
+
+patch {
+    reinplace "s|/System/Library/LaunchDaemons||g" configure
+}
+
+configure.cflags-append     -I${prefix}/include
+configure.cppflags-append   -I${prefix}/include
+configure.cxxflags-append   -I${prefix}/include
+configure.ldflags-append    -L${prefix}/lib
+configure.env-append    DSOFLAGS=-L${prefix}/lib
+configure.args      --disable-libusb \
+                    --disable-pam \
+                    --without-dnssd \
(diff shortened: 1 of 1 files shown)


print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+github.setup        OpenPrinting cups 2.4.2 v
+github.tarball_from releases
+distname            ${github.project}-${github.version}-source
+revision            0
+categories          print
+license             Apache-2
+maintainers         nomaintainer
+description         CUPS is the current standards-based, open source printing system developed by \
+                    OpenPrinting for Unix-like operating systems and uses IPP Everywhere™ to support \
+                    printing to local and network printers.
```
ryandesign, line 15: This short description is much longer than most ports' short descriptions which rarely are longer than one line.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+                    CUPS provides the System V ("lp") and Berkeley ("lpr") command-line interfaces, \
+                    a configurable web interface, a C API, and common print filters, drivers, and backends \
+                    for printing. The cups-filters project provides additional filters and drivers.
+homepage            https://openprinting.github.io/cups/
+
+checksums           rmd160  58a24fe177cfb66157899b6a53a90972a6080e74 \
+                    sha256  f03ccb40b087d1e30940a40e0141dcbba263f39974c20eb9f2521066c9c6c908 \
+                    size    8128785
+
+depends_build       port:gmake \
```
ryandesign, line 30: Is the gmake port really required? Doesn't the Xcode- or CLT-supplied gmake work?

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+                    port:autoconf \
```
ryandesign, line 31: Is autoconf really required? The port appears to be using the pre-generated configure script in the tarball, not regenerating the configure script with autoconf.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+                    port:pkgconfig
+depends_lib         port:zlib
+
+patch {
+    reinplace "s|/System/Library/LaunchDaemons||g" configure
```
ryandesign, line 36: Why is /System/Library/LaunchDaemons being removed here? Probably good to add a comment explaining this so someone reading the Portfile later will understand.

Why is it being done as a `reinplace` and not as a patchfile?

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+}
+
+configure.cflags-append     -I${prefix}/include
+configure.cppflags-append   -I${prefix}/include
+configure.cxxflags-append   -I${prefix}/include
```
ryandesign, line 41: Is `-I${prefix}/include` really needed in CFLAGS and CXXFLAGS? Normally it's only needed in CPPFLAGS.

You don't need to add it to CPPFLAGS yourself; MacPorts does it automatically.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+configure.ldflags-append    -L${prefix}/lib
```
ryandesign, line 42: You don't need to add `-L${prefix}/lib` to LDFLAGS yourself; MacPorts does it automatically.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+                    --libexecdir=${prefix}/libexec \
+                    --datarootdir=${prefix}/share/print/cups \
+                    --sysconfdir=${prefix}/share/print/cups/etc \
+                    --localstatedir=${prefix}/share/cups/var \
+                    --includedir=${prefix}/include \
+                    --oldincludedir=${prefix}/include \
+                    --docdir=${prefix}/share/doc/cups
+
+build.type          gnu
+build.target        all
```
ryandesign, line 64: MacPorts sets `build.target` to `all` by default.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+build.args
```
ryandesign, line 65: `build.args` are empty by default.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+configure.cflags-append     -I${prefix}/include
+configure.cppflags-append   -I${prefix}/include
+configure.cxxflags-append   -I${prefix}/include
+configure.ldflags-append    -L${prefix}/lib
+configure.env-append    DSOFLAGS=-L${prefix}/lib
+configure.args      --disable-libusb \
+                    --disable-pam \
+                    --without-dnssd \
+                    --without-tls \
+                    --with-bundledir="" \
```
ryandesign, line 48: Setting a `--with-...` flag to `""` is unusual. Maybe this build system requires that, but normally we would expect something like `--without-...` instead. Would `--without-...` have the same effect here?

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+                    --localstatedir=${prefix}/share/cups/var \
+                    --includedir=${prefix}/include \
+                    --oldincludedir=${prefix}/include \
+                    --docdir=${prefix}/share/doc/cups
+
+build.type          gnu
+build.target        all
+build.args          
+
+destroot.target     install
```
ryandesign, line 67: MacPorts sets `destroot.target` to `install` by default.

print/cups/Portfile
```diff
@@ -0,0 +1,68 @@
...
+destroot.post_args  BUILDROOT=${destroot}
```
ryandesign, line 68: MacPorts sets `destroot.destdir` to `DESTDIR=${destroot}` by default. I presume this line here this means this build system does not honor `DESTDIR`? If so, it is customary to override `destroot.destdir` with whatever this build system requires (i.e. `BUILDROOT=${destroot}`) rather than use another variable like `destroot.post_args`.

PR #16088, opened 2022-09-11
poetry: update to 1.2.0

dgilman:
#### Description

Two outstanding issues on this draft:
- py-cleo is creating invalid bash autocomplete scripts. They have a fix, they just need to make a release.
- It's not possible to upgrade an existing install of poetry right now. Going to write to the mailing list on this.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H2026 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you


python/poetry/Portfile
```diff
@@ -45,29 +43,37 @@ foreach pv {310 39 38 37} {
     }
 }
 
+python.pep517           yes
+python.pep517_backend
```
jmroot, line 47: I guess it doesn't matter that much since there's a dependency on poetry-core added below, but should this be set to `poetry`?

python/py-poetry-core/Portfile
```diff
@@ -5,17 +5,17 @@ PortGroup           python 1.0
 PortGroup           select 1.0
 
 name                py-poetry-core
-version             1.0.8
+version             1.1.0
```
jmroot, line 8: This version no longer works with python 3.6. It looks like nothing is using the 36 subport, so it can just be dropped.

python/py-poetry-plugin-export/Portfile
```diff
@@ -0,0 +1,42 @@
...
+maintainers         {gmail.com:davidgilman1 @dgilman} openmaintainer
+
+homepage            https://github.com/python-poetry/poetry-plugin-export
+description         A plugin that allows the export of locked packages to various formats.
+long_description    This plugin provides the same features as the existing export command \
+                    of Poetry which it will eventually replace.
+
+python.versions     37 38 39 310
+python.pep517       yes
+python.pep517_backend
```
jmroot, line 26: Needs to be set to `poetry`.

dgilman: This causes a dependency cycle if you do an install, though, as that flag sets a build requirement of py-poetry-core on py-poetry-core.

jmroot: I doesn't look like poetry-core depends on anything except (sometimes) importlib-metadata?

dgilman: Python portgroup does it https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/python-1.0.tcl#L350

jmroot: Yes, a dependency will be added if you do this, but it won't result in a cycle because py-poetry-core does not depend on py-poetry-plugin-export. The build of the latter is failing in CI because it does need py-poetry-core.

dgilman: My bad, ignore all of this, I thought we were discussing py-poetry-core and not this package.

PR #16094, opened 2022-09-12
py-undetected-chromedriver: Bugfix run dependency, clean up tests

essandess:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/python/py-undetected-chromedriver/Portfile
+++ b/python/py-undetected-chromedriver/Portfile
@@ -5,6 +5,7 @@ PortGroup           python 1.0
 
 name                py-undetected-chromedriver
 version             3.1.5.post4
+revision            1
 categories-append   python www
 maintainers         nomaintainer
 license             GPL-3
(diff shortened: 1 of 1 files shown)


reneeotten (changes requested):
when changing `depends_lib` dependencies the revision should be increased. 

However, the dependencies as declared currently in the Portfile are correct based on what upstream does - they have all of these listed as `install_requires` which implies build dependencies. Are they for sure not needed to build - in other words, does it build in trace mode for you after this change? 

If indeed all correct and these are not build dependencies, please report this upstream so that they can correct their dependency declaration.

essandess:
`websockets` is both a build and run dependency, so I added it as lib—is this correct? 

I observed this issue when trying to update `chromedriver +undetected`.

mascguy:
> `websockets` is both a build and run dependency, so I added it as lib—is this correct?

Yep, then a lib dep is correct.

mascguy (approved):

PR #16101, opened 2022-09-12
libnet: rename libnet11 to libnet and update

nerdling:
* remove libnet(10) and dependents
 * libnet11 replaced_by libnet
 * update the new libnet


nerdling:
While syslog-ng previously didn't build for macos 12, it seems macos 11 breaks with the same error after the libnet switch.

Unclear if syslog-ng is worth keeping but is the only package breaking now.

PR #16110, opened 2022-09-13
datree: new portfile in version 1.6.29

Ci7rix:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.5.1 21G83 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you

### DIFF
--- a/devel/datree/Portfile
+++ b/devel/datree/Portfile
@@ -1,0 +1,27 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            github.com/datreeio/datree 1.6.29
+revision            0
+
+categories          devel
+maintainers         {@Ci7rix unset.ch:max+github} openmaintainer
+license             Apache-2
+
+description         CLI tool to run policies against Kubernetes manifests YAML files or Helm charts
+long_description    Datree automatically validates Kubernetes objects for rule violations, ensuring no misconfigurations reach production
+
+checksums           ${distname}${extract.suffix} \
+                    rmd160  10a06b5dfa664a21d6ca803a574030b461bef1cb \
+                    sha256  54c373f9898fbbd6c6ab1dcce703f3816cf8c7cbb30605915e11a70c6d9ede4e \
+                    size    5448369
+
+set go_ldflags      "-s -w -X ${go.package}/cmd.CliVersion=${version}"
+build.args          -tags main -ldflags \"${go_ldflags}\"
+build.env-delete    GO111MODULE=off GOPROXY=off
+
+destroot {
+    xinstall -m 0755 ${worksrcpath}/${name} ${destroot}${prefix}/bin/
+}


devel/datree/Portfile
```diff
@@ -0,0 +1,414 @@
...
+
+description         CLI tool to run policies against Kubernetes manifests YAML files or Helm charts
+long_description    Datree automatically validates Kubernetes objects for rule violations, ensuring no misconfigurations reach production
+
+checksums           ${distname}${extract.suffix} \
+                    rmd160  cbe94d55f7651a32031983d0e6d6dd8d999e0554 \
+                    sha256  a31dd4a3d304c13ea0a0c40fa6f53614a61c708894dba709251ca135ac553f03 \
+                    size    5444221
+
+set go_ldflags      "-s -w -X github.com/datreeio/datree/cmd.CliVersion=${version}"
```
herbygillot, line 20: What you can do here is use `${go.package}` to avoid repetition:

```suggestion
set go_ldflags      "-s -w -X ${go.package}/cmd.CliVersion=${version}"
```

Ci7rix: Done ! Thank you for your feedback 👍

herbygillot:
Thanks - can you squash this change back into the original commit?

Ci7rix:
> Thanks - can you squash this change back into the original commit?

I think it should be good @herbygillot

devel/datree/Portfile
```diff
@@ -0,0 +1,414 @@
...
+long_description    Datree automatically validates Kubernetes objects for rule violations, ensuring no misconfigurations reach production
+
+checksums           ${distname}${extract.suffix} \
+                    rmd160  cbe94d55f7651a32031983d0e6d6dd8d999e0554 \
+                    sha256  a31dd4a3d304c13ea0a0c40fa6f53614a61c708894dba709251ca135ac553f03 \
+                    size    5444221
+
+set go_ldflags      "-s -w -X ${go.package}/cmd.CliVersion=${version}"
+build.args          -ldflags \"${go_ldflags}\"
+build.env           datree_build_env=main
```
herbygillot, line 22: One more issue.  `go` is actually downloading the packages, and ignoring all the vendored Go packages below in `go.vendor`.

I think this is happening because you're overriding `build.env` here, which normally defaults to environment variables that prohibit Go from downloading packages on its own, and makes it use the local packages in `go.vendor` instead.

We can fix this by changing your statement to **append** to the environment variables, instead of **overwriting them**:

```suggestion
build.env-append    datree_build_env=main
```

herbygillot: Evidence of this is in the CI check for this change: https://github.com/macports/macports-ports/actions/runs/3048411381/jobs/4913445882#step:8:2934

herbygillot:
OK, the build is failing now that we are actually using vendored libs.

Let's make the following changes to remove the vendored modules, and having Go build the way it normally would by default:

1. Add the following to remove the environment variables that force Go to use vendored libs:
  ```
  build.env-delete       GO111MODULE=off GOPROXY=off
  ```
  Keep the `build.env-append` line you had.
2. Delete that entire `go.vendors` section.

If the port builds with these change, we'll squash it back into the original commit and then we should be good.

Ci7rix:
@herbygillot This time it should be good. Thanks for taking the time to explain how to get a good Portfile.

I finally integrated the environment variable to the args as it is not used elsewhere.

herbygillot:
Thanks @Ci7rix, looks good.  Let's also add a `revision` field after the `go.setup` and we're good.

```suggestion
go.setup            github.com/datreeio/datree 1.6.28
revision            0
```

Ci7rix:
@herbygillot should be OK ! Thanks

herbygillot:
> @herbygillot should be OK ! Thanks

Great - merged! Thank you @Ci7rix.

Ci7rix:
@herbygillot thank you to you for taking time to explain me.

PR #16116, opened 2022-09-13
icu-devel: new port for testing

catap:
#### Description

I've also clean icu port up, and fixed livecheck.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.5.1 21G83 x86_64
Xcode 14.0 14A309

###### Verification 
Have you

### DIFF
--- a/devel/icu-devel/Portfile
+++ b/devel/icu-devel/Portfile
@@ -1,9 +1,8 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           clang_dependency 1.0
 
-name                icu
+name                icu-devel
 set my_name         icu
 
 # Please keep the icu and icu-devel ports as similar as possible.
@@ -29,7 +28,7 @@ epoch               1
 subport             ${name}-docs         { revision 0 }
 subport             ${name}-lx           { revision 0 }
 
-conflicts           icu-devel icu-devel-lx icu-devel-docs
+conflicts           icu icu-lx icu-docs
 
 categories          devel textproc
 platforms           darwin freebsd
@@ -188,7 +187,7 @@ if {${subport} eq ${name}} {
     livecheck.name      ${my_name}
     livecheck.version   [string map {. -} ${version}]
     livecheck.url       https://api.github.com/repos/unicode-org/icu/releases?per_page=100
-    livecheck.regex     \".+releases/tag/release-(\\d+\-\\d+)\"
+    livecheck.regex     \".+releases/tag/release-(\\d+(?:\-\\d+)?)(?:-rc)?\"
 } else {
     livecheck.type      none
 }
--- a/devel/icu-devel/files/patch-common-putil.cpp.diff
+++ b/devel/icu-devel/files/patch-common-putil.cpp.diff
@@ -1,0 +1,17 @@
+http://bugs.icu-project.org/trac/ticket/9367
+--- common/putil.cpp.orig	2016-10-19 12:20:56.000000000 -0500
++++ common/putil.cpp	2017-02-11 10:01:49.000000000 -0600
+@@ -124,6 +124,13 @@
+ #endif
+ 
+ /*
++ * Mac OS X 10.4 doesn't use its localtime_r() declaration in <time.h> if either _ANSI_SOURCE or _POSIX_C_SOURCE is #defined.
++ */
++#if defined(U_TZNAME) && U_PLATFORM_IS_DARWIN_BASED && (defined(_ANSI_SOURCE) || defined(_POSIX_C_SOURCE))
++U_CFUNC struct tm *localtime_r(const time_t *, struct tm *);
++#endif
++
++/*
+  * Only include langinfo.h if we have a way to get the codeset. If we later
+  * depend on more feature, we can test on U_HAVE_NL_LANGINFO.
+  *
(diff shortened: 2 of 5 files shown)


mascguy:
> @mascguy right now it is the copy of `icu` port, and I plan to switch it to the new version as soon as new `icu` is release.

Sounds good. Are you ready for merge?

catap:
@mascguy yes, let's do it.

mascguy:
One thing you'll want to add, is a top-level declaration for `dist_subdir`:

```
dist_subdir   icu
```

This specifies where MacPorts keeps the distribution files, on https://distfiles.macports.org. (By default, it's the name of the parent port.) And by specifying it, that ensures that everything will be stored under https://distfiles.macports.org/icu.

That saves space, otherwise the identical upstream files will also be stored under https://distfiles.macports.org/icu-devel, which you don't want.

mascguy:
> More generally, you'll want to ensure that everything is stored in the same place as `icu`, so that it's a true drop-in replacement.

And to save yourself maintenance headaches, use the same convention in your non-Devel port. That way, when you diff the two, you aren't ripping your hair out trying to differentiate between the two!  :-)

catap:
@macportsbot here it is. Differences between Portfiles are:
```
--- devel/icu/Portfile	2022-09-17 02:00:40.000000000 +0200
+++ devel/icu-devel/Portfile	2022-09-17 02:07:10.000000000 +0200
@@ -1,9 +1,8 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           clang_dependency 1.0
 
-name                icu
+name                icu-devel
 set my_name         icu
 
 # Please keep the icu and icu-devel ports as similar as possible.
@@ -29,7 +28,7 @@
 subport             ${name}-docs         { revision 0 }
 subport             ${name}-lx           { revision 0 }
 
-conflicts           icu-devel icu-devel-lx icu-devel-docs
+conflicts           icu icu-lx icu-docs
 
 categories          devel textproc
 platforms           darwin freebsd
@@ -188,7 +187,7 @@
     livecheck.name      ${my_name}
     livecheck.version   [string map {. -} ${version}]
     livecheck.url       https://api.github.com/repos/unicode-org/icu/releases?per_page=100
-    livecheck.regex     \".+releases/tag/release-(\\d+\-\\d+)\"
+    livecheck.regex     \".+releases/tag/release-(\\d+(?:\-\\d+)?)(?:-rc)?\"
 } else {
     livecheck.type      none
 }
```

You may notice that `livecheck.regex` are differnet. The `-devel` version is also tracking `-rc` releases.

catap:
rebased to master to trigger one more CI cycle

mascguy:
> rebased to master to trigger one more CI cycle

Kirill, would you expect Ci to fail...?

mascguy:
I'm seeing the following for `icu-devel`, which looks like something was missed:

```
--->  Building icu-devel                                 
Warning: reinplace s|host_alias=\".*\"|host_alias=\"\"|g didn't change anything in icu-devel/work/icu/source/config/icu-config
```

Also, when attempting to build `icu-devel-lx`, it's trying to pull in both `icu` and `icu-devel` is dependencies. This is likely why the CI build is failing:

```
$ sudo port destroot icu-devel-lx
--->  Computing dependencies for icu-devel-lx
 icu
 icu-devel
 icu-le-hb
Continue? [Y/n]: n
```

mascguy:
The issue with deps for `icu-devel-lx` was an error on my part, as I didn't have `icu-devel` installed already. When that's in place, `icu-devel-lx` builds fine. Lol

mascguy:
> I'm seeing the following for `icu-devel`, which looks like something was missed:
> 
> ```
> --->  Building icu-devel                                 
> Warning: reinplace s|host_alias=\".*\"|host_alias=\"\"|g didn't change anything in icu-devel/work/icu/source/config/icu-config
> ```

This is the only outstanding question/concern. Otherwise all three of your devel ports install fine locally.

catap:
@mascguy it builds fine on my machine — I have no idea what’s wrong

mascguy:
> @mascguy it builds fine on my machine — I have no idea what’s wrong

See my later comments, that was my issue.

But can you confirm why `reinplace` isn't changing anything?

mascguy (approved):

mascguy:
I'm not overly concerned with the `reinplace`, so I went ahead and merged.

But one thing you might want to change, relates to `icu-devel-lx`: Since it depends on `icu-le-hb` - and the non-Devel version of `icu` is installed by default - which causes failures on the buildbots. (As well as for for CI.)

Even though this all works fine locally - once a user has installed `icu-devel`, that will fulfill the `icu` dependency for `icu-le-hb` - it's still not ideal. (And since the buildbots can't build it, users will always have to build from source.)

To fix, you'd need to add a new port `icu-le-hb-devel`, which depends on your `icu-devel` port. (And then depend on `icu-le-hb-devel`, where necessary.) Not absolutely required, but it's nice to be able to publish binaries for all of these...

catap:
@mascguy it is good idea, indeed.

catap:
@mascguy a naive approach doesn't work https://github.com/catap/macports-ports/tree/icu

why? because `icu-le-hb` depends on `icu` via dependencies also.

mascguy:
> @mascguy a naive approach doesn't work https://github.com/catap/macports-ports/tree/icu
> 
> why? because `icu-le-hb` depends on `icu` via dependencies also.
> 
> This makes impossible to make prebuild version of `icu-devel-lx` when I'll switch icu-devel to the next release candidate.

I'll take a closer look, later today if possible. More to follow.

PR #16141, opened 2022-09-16
cabal ghc Bugfixes

essandess:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 x86_64
Xcode 14.0 14A309

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/haskell_cabal-1.0.tcl
+++ b/_resources/port1.0/group/haskell_cabal-1.0.tcl
@@ -32,6 +32,11 @@ proc haskell_cabal.add_dependencies {} {
         depends_build-append \
             port:cabal \
             port:ghc
+        depends_lib-append \
+            port:gmp \
+            port:libiconv \
+            port:ncurses \
+            port:zlib
     }
 }
 port::register_callback haskell_cabal.add_dependencies
@@ -133,7 +138,7 @@ default haskell_cabal.build_dir     {${workpath}/dist}
 # use to install prebuilt binaries for bootstrapping
 default haskell_cabal.use_prebuilt  {no}
 
-post-patch {
+post-extract {
     if {[tbool haskell_cabal.use_prebuilt]} {
         xinstall -d ${haskell_cabal.cabal_root}/bin
         # bootstrap from *-prebuilt
@@ -161,7 +166,7 @@ post-patch {
              } {
              ln -s   ${prefix}/bin/${f}-prebuilt \
                      ${haskell_cabal.cabal_root}/bin/${f}
-         }
+        }
 
         haskell_cabal.env-append \
                     "GHC=${haskell_cabal.cabal_root}/bin/ghc" \
--- a/lang/cabal/Portfile
+++ b/lang/cabal/Portfile
@@ -143,7 +143,7 @@ if {${name} eq ${subport}} {
     depends_lib-append \
                     path:bin/ghc:ghc
 
-    post-patch {
+    post-extract {
         haskell_cabal.env-replace \
                     "GHC=${haskell_cabal.cabal_root}/bin/ghc" \
                     "GHC=${prefix}/bin/ghc"
--- a/lang/ghc/Portfile
+++ b/lang/ghc/Portfile
@@ -191,12 +191,20 @@ if {${name} eq ${subport}} {
                     happy \
                     HsColour
 
-    post-patch {
+    post-extract {
         xinstall -d ${workpath}/bin
         ln -s       ${python3_bin} \
                     ${workpath}/bin/python3
         ln -s       ${prefix}/bin/sphinx-build-${python3_branch} \
                     ${workpath}/bin/sphinx-build
+
+        haskell_cabal.env-replace \
+                    "PATH=${haskell_cabal.cabal_root}/bin:$env(PATH)" \
+                    "PATH=${workpath}/bin:${haskell_cabal.cabal_root}/bin:$env(PATH)"
+        foreach phase {configure build destroot test} {
+            ${phase}.env \
+                    {*}${haskell_cabal.env}
+        }
     }
 
     post-patch {
@@ -220,19 +228,9 @@ if {${name} eq ${subport}} {
         }
     }
 
-    post-patch {
-        haskell_cabal.env-replace \
-                    "PATH=${haskell_cabal.cabal_root}/bin:$env(PATH)" \
-                    "PATH=${workpath}/bin:${haskell_cabal.cabal_root}/bin:$env(PATH)"
-        foreach phase {configure build destroot test} {
-            ${phase}.env \
-                    {*}${haskell_cabal.env}
-        }
-    }
-
     pre-configure {
         system -W ${worksrcpath} \
-                    "./boot"
+            "env ${haskell_cabal.env} ./boot"
     }
 
     use_autoconf    yes
@@ -246,7 +244,7 @@ if {${name} eq ${subport}} {
 
     post-build {
         system -W ${build.dir} \
-                    "${build.cmd} docs [haskell_cabal.build_getjobsarg]"
+            "env ${haskell_cabal.env} ${build.cmd} docs [haskell_cabal.build_getjobsarg]"
     }
 
     pre-destroot {


mascguy (approved):

essandess:
@mascguy Thanks.

Please note: This will not affect any of the buildbot or installed binaries; therefore, I did not revbump any ports. The hope based on comments in #16115 is that this will now build prior to macOS 11, but presumably the builedbot must be kicked off again because there is no revbump.

mascguy:
> Please note: This will not affect any of the buildbot or installed binaries; therefore, I did not revbump any ports. The hope based on comments in #16115 is that this will now build prior to macOS 11, but presumably the builedbot must be kicked off again because there is no revbump.

When a port is updated without a rev-bump, the buildbots will still check for any builds that failed previously. And for those cases where it did fail, a rebuild will occur.

mascguy:
> > Please note: This will not affect any of the buildbot or installed binaries; therefore, I did not revbump any ports. The hope based on comments in #16115 is that this will now build prior to macOS 11, but presumably the builedbot must be kicked off again because there is no revbump.
> 
> When a port is updated without a rev-bump, the buildbots will still check for any builds that failed previously. And for those cases where it did fail, a rebuild will occur.

In other words, this is the correct approach.  :-)

mascguy:
One thing to be aware of: On our 10.6 buildbot, the following failure occurred for `ghc-prebuilt`:

```
# Work around #17418 on Darwin
if [ -e "/usr/bin/xattr" ]; then "/usr/bin/xattr" -c -r "ghc-prebuilt/work/destroot/opt/local/lib/ghc-9.4.2/bin"; fi
option -c not recognized

usage: xattr [-l] [-r] [-v] [-x] file [file ...]
       xattr -p [-l] [-r] [-v] [-x] attr_name file [file ...]
       xattr -w [-r] [-v] [-x] attr_name attr_value file [file ...]
       xattr -d [-r] [-v] attr_name file [file ...]
```

I'm not sure which macOS release introduced option `-c` - maybe 10.7 or 10.8? - but it might be better to explicitly delete the attribute(s) you're concerned with.

mascguy:
On a separate note, apparently I'm not quite caffeinated enough this morning, as I missed one critical thing: Since additional dependencies are being added, we probably need to rev-bump affected ports. Otherwise, the dependencies for currently-installed ports [on users' machines] won't be properly reflected.

So even though it may seem like we don't technically need to rebuild everywhere, we probably do.

I'm happy to take care of though. Is it just `cabal` that's affected, or...?

essandess:
> On a separate note, apparently I'm not quite caffeinated enough this morning, as I missed one critical thing: Since additional dependencies are being added, we probably need to rev-bump affected ports. Otherwise, the dependencies for currently-installed ports [on users' machines] won't be properly reflected.
> 
> So even though it may seem like we don't technically need to rebuild everywhere, we probably do.
> 
> I'm happy to take care of though. Is it just `cabal` that's affected, or...?

Thanks. Anything that uses PG `haskell_cabal` *without* the `haskell_cabal.use_prebuilt` flag is affected. There's only about a half dozen of these or fewer so far.

mascguy:
> Thanks. Anything that uses PG `haskell_cabal` _without_ the `haskell_cabal.use_prebuilt` flag is affected. There's only about a half dozen of these or fewer so far.

Can you humor me and my impaired un-caffeinated state, and provide a list of those?  ;-)

essandess:
```sh
port echo depends:cabal | egrep -iv -e 'alex|cabal|ghc|happy|hadrian|hscolour'
adblock2privoxy                 
aeson-pretty                    
cpphs                           
HaXml                           
hlint                           
lhs2tex                         
matterhorn                      
pure-gen                        
shellcheck                      
```

The filtered ports are necessary to bootstrap this whole thing.

mascguy:
I'm seeing the following errors for port `pure-gen`. Was there some cleanup/reconciliation missed at some point? (Perhaps not by you, but just asking in-general for guidance on this...)

```
$ port lint pure-gen
--->  Verifying Portfile for pure-gen
Error: Unknown dependency: hs-cabal-install
Error: Unknown dependency: hs-alex
Error: Unknown dependency: hs-happy
```

essandess:
> I'm seeing the following errors for port `pure-gen`. Was there some cleanup/reconciliation missed at some point? (Perhaps not by you, but just asking in-general for guidance on this...)
> 
> ```
> $ port lint pure-gen
> --->  Verifying Portfile for pure-gen
> Error: Unknown dependency: hs-cabal-install
> Error: Unknown dependency: hs-alex
> Error: Unknown dependency: hs-happy
> ```

That Portfile code is all many generations obsolete. I have no idea what `pure` is. I'df just ignore that one until there's some demands-side signal.

essandess:
Once upon a time, MacPorts tried to recreate the hackage world, and the dependency hell that then reigned in hackage was imported and exponentiated within MacPorts. Nothing built and everyone gave up (see the `bazel`-based ports). We migrated to `stack`, and everything built again using `stack` as the root. But `stack` hasn't yet migrated to native `arm` builds, and even when it does the older `ghc`/`cabal` compilers/builders it downloads will never build `arm`, so the expected timeline to native `arm` via `stack` is long. `cabal` and `ghc` now both support `arm`, and with native MacPorts library linking (hence my several MacPorts dependency bugs!), so we now can and should migrate to PG `haskell_cabal`-based builds when they work. The next thing on the TODO is `pandoc`, but it's waiting for a Haskell `memory` library update to be able to work with the latest `ghc` compilers.

essandess:
Also, all `hs-*` ports should be obsoleted, as well as the old nonworking `ghc-bootstrap` port.

_resources/port1.0/group/haskell_cabal-1.0.tcl
```diff
@@ -32,6 +32,11 @@ proc haskell_cabal.add_dependencies {} {
         depends_build-append \
             port:cabal \
             port:ghc
+        depends_lib-append \
+            port:gmp \
+            port:libiconv \
+            port:ncurses \
+            port:zlib
```
jmroot, line 39: It can't be right to add all of these to every single port that uses cabal, can it? `shellcheck` for example only used gmp and libiconv.

_resources/port1.0/group/haskell_cabal-1.0.tcl
```diff
@@ -133,7 +138,7 @@ default haskell_cabal.build_dir     {${workpath}/dist}
 # use to install prebuilt binaries for bootstrapping
 default haskell_cabal.use_prebuilt  {no}
 
-post-patch {
+post-extract {
```
jmroot, line 141: Using post-extract for this has exactly the same problems as post-patch, as indeed would putting it inside any single phase. Any time the phase is already marked as complete in the statefile, it will be skipped and this code won't run. If these variables can't be set up at the global level, then you need to ensure that they are set up in every phase that might use them (preferably by calling a common proc that sets them up only if it hasn't already been done.)

_resources/port1.0/group/haskell_cabal-1.0.tcl
essandess: It's right from the *[Caedite eos](https://en.wikipedia.org/wiki/Caedite_eos._Novit_enim_Dominus_qui_sunt_eius.)* school of coding.

Okay, `cabal` builds all almost certainly will link to `libiconv` and `libgmp`. Some (`adblocktoprivoxy`) link to `zlib`. Some could link to `libncurses`. Arguably it's simpler just to include all these small libraries rather then making special cases every time.
```sh
port contents ghc | egrep -e '\.dylib' | xargs otool -L | egrep -E -e '^[[:space:]]+/opt/local' | sort -u
	/opt/local/lib/libgmp.10.dylib (compatibility version 15.0.0, current version 15.1.0)
	/opt/local/lib/libncurses.6.dylib (compatibility version 6.0.0, current version 6.0.0)
otool -L `which cabal` | egrep -E -e '^[[:space:]]+/opt/local' | sort -u
	/opt/local/lib/libgmp.10.dylib (compatibility version 15.0.0, current version 15.1.0)
	/opt/local/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.12)
```

But if you prefer special cases, I'll issue a modification.

_resources/port1.0/group/haskell_cabal-1.0.tcl
essandess: It does not appear that MacPorts provides a method to modify paths or environment variables globally. Is there such a Portfile method that avoids the issue you describe?

essandess:
@jmroot  @mascguy I'll incorporate these comments in a new PR:
* https://github.com/macports/macports-ports/pull/16115#issuecomment-1250137066
* https://github.com/macports/macports-ports/pull/16141#discussion_r973626299
* https://github.com/macports/macports-ports/pull/16141#discussion_r973616866

But before I act on that I need to better understand how to modify paths or environment variables in a Portfile.

The only method I'm area of is in a `pre-` or `post-` block that occurs before the phase for which you're trying to set the `PATH`. It seems a little crazy that it's this complicated. Is this really the only method? This is a common problem that has taken a lot of cycles to address as it stands. Is there a representative Portfile that performs `PATH` modification in this way?

mascguy:
> But before I act on that I need to better understand how to modify paths or environment variables in a Portfile.
> 
> The only method I'm area of is in a `pre-` or `post-` block that occurs before the phase for which you're trying to set the `PATH`. It seems a little crazy that it's this complicated. Is this really the only method? This is a common problem that has taken a lot of cycles to address as it stands.

I'd need to better understand the logic, before providing any definitive suggestions.

But one potential route, is to encapsulate some of this via a wrapper script(s)... with all of the key values passed via parameters, or custom environment variables.

That said, @jmroot and @ryandesign can provide the best guidance. So I'll defer to them.

All of this would also be worth discussing via our Dev mailing list. If nothing else, it will help others if/when they encounter similar challenges!

essandess:
Unexpected observed behavior (not in current code, which was written to sidestep this issue):

PG `haskell_cabal`:
> ```
> haskell_cabal.env PATH=FOO
> ```

some port:
> ```
> PortGroup haskell_cabal 1.0
> haskell_cabal.env-replace PATH=FOO PATH=BAR
> ```

Observed behavior: `PATH=FOO`.

Wrap all these in `pre-*` blocks that are executed in the right sequence, then it works: `PATH=BAR`.

Why is that?

The application is swapping in paths to the `-prebult` Haskell ports at the appropriate places.

essandess:
I just took a different approach and redefined all environments from scratch, also following @jmroot's guidance to use a `proc` to set the environment. This is much cleaner—thanks. Please see #16154.

ryandesign:
> > I'm seeing the following errors for port `pure-gen`.

> That Portfile code is all many generations obsolete.

Yes. I tried to update it but it doesn't work with stack. The author of Pure doesn't have a solution yet, so I'm just waiting for that. See https://github.com/agraef/pure-lang/issues/37.

> I have no idea what `pure` is.

It's a programming language; see https://agraef.github.io/pure-lang/ if you want to learn more.

_resources/port1.0/group/haskell_cabal-1.0.tcl
jmroot: Other ports specify their dependencies individually, and I don't really see any reason why these should be an exception.

_resources/port1.0/group/haskell_cabal-1.0.tcl
jmroot: Unfortunately, if you change `::env` it is changed for the entire process, not just the subinterpreter running your Portfile. So any PATH modification has to be done in a temporary way like `command_exec` does it, changing it before doing the thing that needs it and then restoring it to its original value afterwards.

Since you're already defaulting the env option for each phase to `${haskell_cabal.env}`, perhaps lazily evaluating that is all that's needed. Something like this:
```tcl
proc haskell_cabal.get_env {} {
    global haskell_cabal.cabal_root haskell_cabal.use_prebuilt
    set myenv [list CABAL_CONFIG=${haskell_cabal.cabal_root}/config]
    if {[tbool haskell_cabal.use_prebuilt]} {
        lappend myenv \
                    GHC=${haskell_cabal.cabal_root}/bin/ghc \
                    PATH=${haskell_cabal.cabal_root}/bin:$::env(PATH)
    }
    return $myenv
}
default haskell_cabal.env \
        {[haskell_cabal.get_env]}
```

essandess:
> Yes. I tried to update it but it doesn't work with stack. The author of Pure doesn't have a solution yet, so I'm just waiting for that. See [agraef/pure-lang#37](https://github.com/agraef/pure-lang/issues/37).

Please see #16171.

PR #16146, opened 2022-09-17
shared-mime-info: clean up cache on deactivate + auto-update watcher/daemon

RJVB:
#### Description
Clean the mime cache after deactivating the port, as requested in https://trac.macports.org/ticket/45396 .

_Note that in my testing and experience the `post-activate` step is only performed after an install. So this change means that the mime cache is removed after a `port deactivate shared-mime-info` so also after a `port activate shared-mime-info` to activate a different version - in which case a new cache is not built automatically.

I see no way to warn the user of this fact other than after an install/upgrade, which is why I am not in favour of this new behaviour personally._

This PR also proposes to replace the post-activate call to `update-mime-database` with a launchd daemon watching `$prefix/share/mime/packages`. To make this change as transparent as possible the daemon's plist is loaded in the `post-activate` and unloaded in the `pre-deactivate`; the user is informed of this via `port info` and `port notes` (which also contains instructions related to the re-activation issue mentioned above).

If this change is accepted all dependent Portfiles could lose their post-activate call to `update-mime-database`.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
Mac OS X 10.9.5

###### Verification 
Have you

### DIFF
--- a/devel/shared-mime-info/Portfile
+++ b/devel/shared-mime-info/Portfile
@@ -17,7 +17,9 @@ platforms           darwin
 description         Database of common types.
 
 long_description    The core database of common types and the \
-                    update-mime-database command used to extend it.
+                    update-mime-database command used to extend it. \n\
+                    Installs and starts a watcher daemon that keeps the \
+                    cache up-to-date when other ports add or remove mime definitions.
 
 homepage            https://www.freedesktop.org/wiki/Software/shared-mime-info/
 master_sites        https://gitlab.freedesktop.org/xdg/shared-mime-info/-/archive/${version}/
@@ -43,11 +45,45 @@ depends_lib         path:lib/pkgconfig/glib-2.0.pc:glib2 \
 #     xinstall -m 644 ${filespath}/ObjCpp.xml \
 #         ${destroot}${prefix}/share/mime/packages
 # }
+post-destroot {
+    xinstall -d -m 755 ${destroot}${prefix}/Library/LaunchDaemons
+    xinstall -m 644 ${filespath}/org.macports.shared-mime-info-updater.plist \
+        ${destroot}${prefix}/Library/LaunchDaemons
+    reinplace -locale C "s|@PREFIX@|${prefix}|g" \
+        ${destroot}${prefix}/Library/LaunchDaemons/org.macports.shared-mime-info-updater.plist
+}
 
 post-activate {
-    ui_debug "Updating MIME database..."
-    system "env XDG_DATA_DIRS=${prefix}/share ${prefix}/bin/update-mime-database -V ${prefix}/share/mime"
+    if {${os.platform} eq "darwin"} {
+        ui_msg "Activating MIME database auto-updater..."
+        system "sudo launchctl load ${prefix}/Library/LaunchDaemons/org.macports.shared-mime-info-updater.plist"
+    } else {
+        ui_msg "Updating MIME database..."
+        system "env XDG_DATA_DIRS=${prefix}/share ${prefix}/bin/update-mime-database -V ${prefix}/share/mime"
+    }
 }
+
+pre-deactivate {
+    if {${os.platform} eq "darwin"} {
+        ui_debug "Deactivating MIME database auto-updater..."
+        system "sudo launchctl unload ${prefix}/Library/LaunchDaemons/org.macports.shared-mime-info-updater.plist"
+    }
+}
+
+post-deactivate {
+    # cleanup
+    foreach f [glob -nocomplain ${prefix}/share/mime/*] {
+        if {${f} ne "${prefix}/share/mime/packages"} {
+            file delete -force ${f}
+        }
+    }
+    ui_msg [join ${notes}]
+}
+
+notes-append "Please check if the auto-updater is still active after reactivating (another) version of) this port: \n\
+    `sudo launchctl list | fgrep shared-mime-info-updater` \n\
+    If the updater is not active, load it manually with \n\
+    `sudo launchctl load ${prefix}/Library/LaunchDaemons/org.macports.shared-mime-info-updater.plist`"
 
 pre-pkg {
     xinstall -m 0755 ${filespath}/postinstall ${package.scripts}/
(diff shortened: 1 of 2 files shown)


mascguy:
Hey Rene, I completely forgot about this PR! But finally trying to test with it locally.

I did encounter one problem, which is that the launchd script appears to be missing: `org.macports.shared-mime-info-updater.plist`. I could probably cobble one together easily enough, as presumably (?) it simply runs:

- `env XDG_DATA_DIRS=${prefix}/share ${prefix}/bin/update-mime-database -V ${prefix}/share/mime`

But was hoping you still have that file in your repo? If so, would you mind adding that?

RJVB:
I did have the file and added it to my fork via the web interface but it doesn't seem the PR notited that change...

mascguy:
> I did have the file and added it to my fork via the web interface
> (note to self: this works if you just watch to which fork you're adding)

Thanks Rene! I'll test this later today.

mascguy:
> > I did have the file and added it to my fork via the web interface
> > (note to self: this works if you just watch to which fork you're adding)
> 
> Thanks Rene! I'll test this later today.

Looks good overall. One thought though: As you know, it's preferable to utilize MacPorts' `startupitem` functionality, rather than directly installing a Launch Agent/Daemon.

Is there a specific reason you didn't follow that approach?

devel/shared-mime-info/Portfile
mascguy, line 1: You need to rev-bump this, of course...

RJVB:
>Is there a specific reason you didn't follow that approach?

TBH I didn't know it was preferable, am not particularly fond of black boxes ;) but apart from that I see no reason to avoid it if it can do the same thing. Note that having the mime info updating done correctly means the agent has to be loaded automatically at the activation stage and not depend on the user doing it. If the startupitem approach has a way of doing that without calling an external application in the post-activation phase (and reliably, not just in the 1st one after an install) then that would in fact be preferable.

mascguy:
> TBH I didn't know it was preferable, am not particularly fond of black boxes ;) but apart from that I see no reason to avoid it if it can do the same thing. Note that having the mime info updating done correctly means the agent has to be loaded automatically at the activation stage and not depend on the user doing it. If the startupitem approach has a way of doing that without calling an external application in the post-activation phase (and reliably, not just in the 1st one after an install) then that would in fact be preferable.

While users can technically disable automatic loading of startup items - by setting `startupitem_autostart no` in `${prefix}/etc/macports/macports.conf`, very few users do. And we can add notes, to display a message to reiterate the importance of enabling the item for this port.

So I'd like to suggest we use MacPorts' startupitem functionality, rather than installing the agent directly.

RJVB:
>While users can technically disable automatic loading of startup items - by setting `startupitem_autostart no` in `${prefix}/etc/macports/macports.conf`, very few users do. And we 

I updated my DBus port not long ago, and IIRC I did have to activate it manually. For full disclosure I have to admit I upgraded it from a very old version that didn't use startupitem functionality.

mascguy:
> > While users can technically disable automatic loading of startup items - by setting `startupitem_autostart no` in `${prefix}/etc/macports/macports.conf`, very few users do.

> I updated my DBus port not long ago, and IIRC I did have to activate it manually. For full disclosure I have to admit I upgraded it from a very old version that didn't use startupitem functionality.

Yep, that's the usual case. So are you willing to switch to using `startupitem`?

RJVB:
>Yep, that's the usual case. So are you willing to switch to using `startupitem`?

If we're clear that the usual case isn't good enough here and that it can be made automatic... and maybe if you show me how ;) (I'm guess it'd take you less than than me...)

mascguy:
> > Yep, that's the usual case. So are you willing to switch to using `startupitem`?

> If we're clear that the usual case isn't good enough here and that it can be made automatic... and maybe if you show me how ;) (I'm guess it'd take you less than than me...)

Sure, I'm happy to make the modifications. But you'll need to enable `Allow edits and access to secrets by maintainers` for this PR, otherwise I can't push to your branch.

RJVB:
So that's what that's for :) I thought you'd just copy the snippet here, but this is more efficient :)

mascguy:
> So that's what that's for :) I thought you'd just copy the snippet here, but this is more efficient :)

Yep! Just pushed the changes, after testing and validating locally.

Let me know what you think!

mascguy:
Actually, reviewing the port notes, we have some redundancy:

```
--->  Some of the ports you installed have notes:
    shared-mime-info has the following notes:
  Please check if the auto-updater is still active after reactivating (another) version of) this port: 
   `sudo launchctl list | fgrep shared-mime-info-updater` 
   If the updater is not active, load it manually with 
   `sudo port load shared-mime-info`
  
  A startup item has been generated that will start shared-mime-info with launchd, and will be enabled automatically on activation.
  Execute the following command to manually _disable_ it:
  
      sudo port unload shared-mime-info
```

Since MacPorts base already provides default notes for the startupitem, that might be sufficient. What do you think?

mascguy (approved):

mascguy:
Rene, let's merge. I'm happy to refine the notes later, if you have any strong preferences.

Otherwise, thank you so much for taking this on. Great stuff, and much appreciated!

RJVB:
Well, thanks for the assistance! You basically caught me offline due to time differences so sorry for not reacting in more timely fashion.

FWIW, how reliable is the startupitem activation after activation? On my end the post-activation phase only seems to trigger after a port install, not when activating another already installed version. Never yet got around to looking into the reasons for that.

mascguy:
> Well, thanks for the assistance! You basically caught me offline due to time differences so sorry for not reacting in more timely fashion.

No worries, glad to help. You did the hard work, so the simplification was easy!

> FWIW, how reliable is the startupitem activation after activation? On my end the post-activation phase only seems to trigger after a port install, not when activating another already installed version. Never yet got around to looking into the reasons for that.

It looks like `post-activate` doesn't run after a deactivation/activation, though perhaps there's a reason for that behavior. (Or perhaps it's a known issue, that simply hasn't been addressed yet. Would have to search the open Trac tickets to say for sure.)

Thankfully this isn't an issue for Darwin/macOS though, as the startup item is consistently re-added/started on activation. So no worries there at least.

RJVB:
>Thankfully this isn't an issue for Darwin/macOS though

Are you suggesting the issue isn't moot elsewhere?

I *have* been tinkering with/on a reverse-engineered launchctl/launchd clone for *NIX with someone in the beginning of the year, but as so often I got completely side-tracked and the other person probably lost interest by now :-/

PR #16148, opened 2022-09-17
pg app 1.1: new version, utilizing callback mechanism

mascguy:
### Description

The background behind this, is available in the associated ticket:
* [65721 - pg app: create new version 1.1, utilizing callback mechanism](https://trac.macports.org/ticket/65721)

Tested with port Gramps, included with this PR, to validate.


mascguy:
Reviewers: I tried to keep everything as similar as possible between the old and new pg, with some minor exceptions:
* All procs are now scoped to namespace `app`.
* For cases where there's notable copy-pasta - such as for paths - that was cleaned up.
* The trace on `app.icon` was removed. If there's an interested in maintaining something similar, we can do so via an option proc for that.

PR #16227, opened 2022-09-26
update to libunistring: remove perl dependency

lukaso:
#### Description

The perl dependency is only required for "Hacking". It should not be required for someone to use this low level library.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0 14A309

###### Verification 
Have you

Test result:

```
Error: Failed to test libunistring: libunistring has no tests turned on. see 'test.run' in portfile(7)
```

### DIFF
--- a/textproc/libunistring/Portfile
+++ b/textproc/libunistring/Portfile
@@ -30,9 +30,6 @@ long_description    Text files are nowadays usually encoded in Unicode, and may 
 homepage            https://www.gnu.org/software/libunistring/
 master_sites        gnu
 
-depends_build       bin:perl:perl5 \
-                    port:texinfo
-
 # use gmake on 10.4, see:
 # https://trac.macports.org/ticket/63790
 platform darwin 8 {


mascguy:
While this seems reasonable, it's not clear from the Git history - and related tickets - whether these changes might cause issues for some cases. (We still support this back to 10.5 and 10.4, and we don't want to inadvertently break folks' systems.)

Adding a few others who have a better understanding of the history, and potential issues:
* @ryandesign
* @jmroot
* @Schamschula
* @danchr

textproc/libunistring/Portfile
```diff
@@ -30,9 +30,6 @@ long_description    Text files are nowadays usually encoded in Unicode, and may
 homepage            https://www.gnu.org/software/libunistring/
 master_sites        gnu
 
-depends_build       bin:perl:perl5 \
-                    port:texinfo
```
jmroot, line 34: The commit message should mention texinfo, because that's the main dependency being removed (perl is only a bin: dependency which is satisfied by the system perl), and also explain that the reason it isn't needed is that the release tarball already has the docs built.

jmroot (changes requested):

mascguy:
> It doesn't matter how low level the library is, what matters is whether it needs these dependencies to build. And it looks like it would only need perl and texi2html to regenerate the HTML docs, which doesn't happen because the release tarball already contains a built copy.

Agreed, and trace mode looks fine when those deps are removed. (At least based on a quick naive test, on 10.15.)

So are we safe to merge this then?

textproc/libunistring/Portfile
lukaso: I'll fix the commit message.

lukaso: Done in 80f2a89286a

jmroot (approved):

barracuda156:
```
Could not open /opt/local/lib/libunistring.2.dylib: Error opening or reading file (referenced from /opt/local/lib/clisp-2.49.92/base/lisp.run)
```

Looks like `clisp` has to be revbumped.

Schamschula:
I only revision bumped direct dependents. I guess a second wave is in order.

Schamschula:
However, I'm not going to revision bump 8445 ports! Not all of theses will actually need to be touched.

PR #16228, opened 2022-09-26
haskell_cabal PG: Fix arm builds

essandess:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0 14A309

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/haskell_cabal-1.0.tcl
+++ b/_resources/port1.0/group/haskell_cabal-1.0.tcl
@@ -273,8 +273,8 @@ post-destroot {
                 continue
             }
             xinstall -m 0755 \
-                    ${binfile} \
-                    ${binfile}.slash_hack
+                ${binfile} \
+                ${binfile}.slash_hack
             foreach build_datadir ${build_datadirs} {
                 set extra_slashes \
                     [expr [string length ${build_datadir}] - [string length ${prefix}/${haskell_cabal.datadir}]]
@@ -293,11 +293,16 @@ post-destroot {
                             ${binfile}.slash_hack"
                 }
             }
-            delete  ${binfile}
-            xinstall -m 0755 \
+            if {[file size ${binfile}.slash_hack] == [file size ${binfile}]} {
+                delete  ${binfile}
+                xinstall -m 0755 \
                     ${binfile}.slash_hack \
                     ${binfile}
+            }
             delete  ${binfile}.slash_hack
+            if {${configure.build_arch} eq {arm64}} {
+                system "codesign -f -s - ${binfile}"
+            }
         }
     }
 }
--- a/devel/HaXml/Portfile
+++ b/devel/HaXml/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                HaXml
 version             1.25.9
-revision            2
+revision            3
 categories          devel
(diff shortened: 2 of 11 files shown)


neverpanic:
So, is this not required for arm64, or will this just produce broken binaries on arm64?

If it's the latter, what do we actually gain from merging this now?

neverpanic:
Ah, I've found https://trac.macports.org/ticket/65890 now. I guess we could just re-sign the binaries after this modification to prevent the crash, and the hack would still work on arm64?

essandess:
> So, is this not required for arm64, or will this just produce broken binaries on arm64?
> 
> If it's the latter, what do we actually gain from merging this now?

I discovered that the existing `datadir` hack breaks `arm` binaries, so I disabled this hack for `arm`. (Upstream `cabal` has acknowledged and prioritized this issue.)

We want to merge ASAP because MacPorts is now distributing broken `arm` binaries for these ports, as well as breaking the buildbot builds for the ports that rely upon them.

neverpanic (changes requested):
Please replace this patch with this instead:

```diff
diff --git a/_resources/port1.0/group/haskell_cabal-1.0.tcl b/_resources/port1.0/group/haskell_cabal-1.0.tcl
index dfa56f1b8c0..9501b157e84 100644
--- a/_resources/port1.0/group/haskell_cabal-1.0.tcl
+++ b/_resources/port1.0/group/haskell_cabal-1.0.tcl
@@ -297,6 +297,9 @@ post-destroot {
             xinstall -m 0755 \
                     ${binfile}.slash_hack \
                     ${binfile}
+            if {${configure.build_arch} eq "arm64"} {
+                system "codesign -f -s - ${binfile}"
+            }
             delete  ${binfile}.slash_hack
         }
     }
```

essandess:
> Ah, I've found https://trac.macports.org/ticket/65890 now. I guess we could just re-sign the binaries after this modification to prevent the crash, and the hack would still work on arm64?

Is there a way to get this hack to work? My preference is to throw in the towel unless it’s easy. And just wait for upstream to fix this. Practically, there is a limited number of ways to hit this issue.

neverpanic:
> Is there a way to get this hack to work? My preference is to throw in the towel unless it’s easy. And just wait for upstream to fix this. Practically, there is a limited number of ways to hit this issue.

Yes, see my proposed patch. I verified that this works on arm64.

essandess:
> see my proposed patch. I verified that this works on arm64.

Holy smokes it works—thanks! I didn't know about `codesign`.
```sh
^@/opt/local///////////////////////////////////////////////////////////////////////\
///////////////////////////////share/alex^@^@^@^@^@^@alex_datadir^@^@^@^@

$ alex --help
Usage: alex [OPTION...] file
```

essandess:
> Please replace this patch with this instead:

Done, thanks again.

neverpanic:
Unfortunately you now modified the last commit `lhs2tex: revbump` to contain the fixes, but the original `haskell_cabal PG: Fix arm builds` still just disables the datadir handling on arm64.

Can you fix that?

essandess:
Oops, fixed, thanks.

mascguy (approved):

PR #16235, opened 2022-09-26
Fix gpsd when global -python27

lukaso:
#### Description

This means that gpsd won't fail to load when `-python27` is set globally (in variants). Without this fix, `port clean --all all` will fail on `gpsd`.

Alternative would be to disable all preset variants when doing `port clean`.

This is the only port that fails when doing `port clean` and `-python27` is set globally.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0 14A309

###### Verification 
Have you

I did test that `port clean --all all` worked properly in this situation.

### DIFF
--- a/net/gpsd/Portfile
+++ b/net/gpsd/Portfile
@@ -71,8 +71,9 @@ foreach s ${pythons_suffixes} {
     }
 }
 if {${pyver_no_dot} eq ""} {
-    ui_error "\n\nYou must select one of the Python variants (+pythonXY).\n"
-return -code error "Invalid Python variant selection"
+    set pyver_no_dot [lindex ${pythons_suffixes} [llength ${pythons_suffixes}]-1]
+    ui_error "\n\nA Python variant must be selected (+pythonXY).\nForcing +python${pyver_no_dot}.\n\n"
+    default_variants +python${pyver_no_dot}
 }
 
 set pyver_dotted [dotted_ver ${pyver_no_dot}]


fhgwright:
The `gpsd` port is currently quite outdated, and I've been holding off on updating it until I sort out the brokenness of the (currently unstated) dependency on `asciidoctor` (from `ruby30`) on older platforms.  I'm in the process of trying to sort out that mess right now.  After I do that, I'll take a look at this issue.  I'd prefer not to have the extra churn in the meantime.

It's possible to build `gpsd` with no Python support at all, though it never seemed worthwhile to bother implementing that in the port.

lukaso:
The issue for me is that there is no other way to clean all ports, than `port clean --all all`, even though all I want to do is clean the ports I intended to install. This is likely a flaw in the way that variants are handled, but regardless, it's blocking me. Even if you just remove the error when no python is set, that is good enough for me (I'm not actually installing this port).

fhgwright:
There are multiple bugs in `port clean`, but if `gpsd` isn't actually installed, you should be able to use `port clean --all installed`.  In theory, `port clean` should be a NOP for any port that isn't installed, but that may not be true in practice.

One of the bugs in `port clean` is that it sometimes misses things, so if thorough cleaning is important, you should follow up by looking for remnants in `${prefix}/var/macports/build/`, which may need to be removed manually.

kencu:
> The issue for me is that there is no other way to clean all ports, than `port clean --all all`

```sudo port -v reclaim``` finishes up with an offer to delete all build directories, which accomplishes this task.

fhgwright:
@kencu 
> `sudo port -v reclaim` finishes up with an offer to delete all build directories, which appears to accomplish this task for me when I use it.

Well `port reclaim` is so extreme that I can't imagine ever wanting to use it. :-)

kencu:
> @kencu
> 
> > `sudo port -v reclaim` finishes up with an offer to delete all build directories, which appears to accomplish this task for me when I use it.
> 
> Well `port reclaim` is so extreme that I can't imagine ever wanting to use it. :-)

I guess you’re out of luck. 

of course, you can skip the other parts of reclaim and only clean the build dirs, but I feel certain you would know that and find it also lacking for you. 

For others though, it might work well, as it’s a 5 second fix 😬😬😬😬😬😦

lukaso:
> port clean --all installed

The fatal flaw with this command is that the ports I need to clean are the
ones that *failed* to install. Because those are the ones that are breaking
the build.

What I really need something like this:

port clean --all failed_installs incomplete_installs

It's the same problem with dependencies. I'm installing something, it
fails, and then I can't figure out the dependencies because it's not
installed. Classic chicken and egg.

michaelld:
I'd say this is more of an issue with `port` itself rather than the port here, but maybe I'm missing something? Anyway if @fhgwright wants to hold off then let's hold off. As a co-maintainer of `gpsd` we need his approval to move forward unless the issue is critical & in this case it's not. My proposal is to close this issue & let @fhgwright deal with it with the next `gpsd` update, whenever than happens.

ryandesign:
> There are multiple bugs in `port clean`,

Oh? First I've heard of it.

> In theory, `port clean` should be a NOP for any port that isn't installed,

No, it shouldn't. It's perfectly reasonable to want to clean a port that isn't installed.

> One of the bugs in `port clean` is that it sometimes misses things,

I've never seen that happen.

fhgwright:
I guess there's no "Closes" syntax for a competing PR. :-)

@lukaso
Your original problem could have been worked around with `port -p clean --all all`.

@ryandesign 
> > In theory, `port clean` should be a NOP for any port that isn't installed,
> 
> No, it shouldn't. It's perfectly reasonable to want to clean a port that isn't installed.

Ah, you're right, though it should be a NOP for any port that was never attempted to be installed.

> > One of the bugs in `port clean` is that it sometimes misses things,
> 
> I've never seen that happen.

One source of such troubles is that it insists on deriving the build-tree location from the _exact_ source of the Portfile, even down to caring about which mirror was used.  Unless the current source of the Portfile is _exactly_ the same as it was at the time of the build, then the build tree isn't removed by cleaning.

PR #16274, opened 2022-10-03
Update ports rbenv and ruby-build: Break rbenv dependency from ruby-build

halostatue:
#### Description

rbenv: update to version 1.2.0

Add a notes block advising the installation of port ruby-build if the
subcommand `rbenv install` is desired.

Closes: https://trac.macports.org/ticket/65329
Signed-off-by: Austin Ziegler <austin@zieglers.ca>

ruby-build: remove incorrect dependency

- Removes the dependency on `rbenv`, as ruby-build does not require
  rbenv to work (but rbenv can integrate ruby-build as a plug-in).

- Add a notes block that this port no longer installs port rbenv
  automatically, and advising manual installation of port rbenv.

Closes: https://trac.macports.org/ticket/65329
Closes: https://github.com/macports/macports-ports/pulls/16273
Closes: https://github.com/macports/macports-ports/pulls/16274
Signed-off-by: Austin Ziegler <austin@zieglers.ca>

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 13.4.1 13F100

###### Verification

Have you

### DIFF
--- a/ruby/rbenv/Portfile
+++ b/ruby/rbenv/Portfile
@@ -4,6 +4,7 @@ PortSystem          1.0
 PortGroup           github 1.0
 
 github.setup        rbenv rbenv 1.2.0 v
+revision            0
 license             MIT
 categories          ruby
 platforms           any
@@ -29,4 +30,8 @@ destroot {
     xinstall -d ${destroot}${prefix}/share/zsh/site-functions
     xinstall ${worksrcpath}/completions/rbenv.zsh \
         ${destroot}${prefix}/share/zsh/site-functions/rbenv.zsh_completion
+}
+
+notes {
+    To enable the rbenv subcommand 'install', install port ruby-build.
 }
--- a/ruby/ruby-build/Portfile
+++ b/ruby/ruby-build/Portfile
@@ -3,7 +3,7 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        rbenv ruby-build 20221225 v
+github.setup        rbenv ruby-build 20230208 v
 categories          ruby
 license             MIT
 platforms           any
@@ -14,13 +14,16 @@ maintainers         {mojca @mojca} openmaintainer
 description         Compile and install Ruby
 long_description    {*}${description}
 
-checksums           rmd160  d6b29467efcbfbc3cb289bf2b5e24c0b3fd51074 \
-                    sha256  18cf31c9081a90f00d8810a48db8caf6f190926f9a9736130f1e90c9f483f49e \
-                    size    79455
+checksums           rmd160  720cc0c4b8ee5f0e835d3b358afe04f784e7b29f \
+                    sha256  db5597e6beea530ed9b6b97f0bbdf4395ac909a04e9f552b45c05b90d172e419 \
+                    size    78320
 
 use_configure       no
 build {}
 destroot.cmd        ./install.sh
 destroot.env        PREFIX=${destroot}${prefix}
 
-depends_lib         port:rbenv
+notes {
+    The ruby-build port no longer installs rbenv automatically. If required, please
+    install the port rbenv manually.
+}


halostatue:
This is a potentially *breaking* fix to `ruby-build` and `rbenv` in that the dependency is broken between `ruby-build` to `rbenv`  and added as an *optional* (variant-based) dependency *from* `rbenv` to `ruby-build`. The "safer" fix (but arguably still incorrect because the default behaviour still installs `rbenv`) is #16273.

halostatue:
I split this version into two commits.

ruby/ruby-build/Portfile
```diff
@@ -3,24 +3,23 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        rbenv ruby-build 20220713 v
+github.setup        rbenv ruby-build 20221225 v
+revision            1
```
amake, line 7: Since you are bumping the version, you don't need to bump the revision.

halostatue: I had bumped the revision here because this is a *change* to the way that this port works. I‘m not sure what sort of signal that would require otherwise. I can change it back.

amake: Thanks for clarifying, but that's not what `revision` means in MacPorts. Bumping `revision` is for when `version` is *un*changed but something else about that port has changed (dependencies, installed files, default variants, etc.) such that the port needs to be rebuilt. It is expected to reset `revision` to 0 when bumping the `version`.

amake: Now that #17090 has been merged and this PR will no longer bump the `version`, you _should_ bump the `revision` due to the change in dependencies.

halostatue:
Updated so that there’s no conflict. I also added Closes lines for both this PR and the alternate PR (#16273) to both this and the other, since only *one* of the two should be chosen. I think that this is the *better* of the two fixes, but it changes the (optional) dependency direction (correctly).

ruby/rbenv/files/patch-libexec-rbenv-init.diff
```diff
@@ -0,0 +1,27 @@
+This patch is needed to make
+    eval "$(rbenv init -)"
+work without complaining about
+    -bash: /opt/local/bin/../completions/rbenv.bash: No such file or directory
```
amake, line 4: This patch isn't actually being used by the port—you need to specify `patchfiles patch-libexec-rbenv-init.diff`.

Also I don't see the error you describe. Could you provide instructions for reproducing the issue?

ruby/rbenv/Portfile
```diff
@@ -4,9 +4,10 @@ PortSystem          1.0
 PortGroup           github 1.0
 
 github.setup        rbenv rbenv 1.2.0 v
+revision            1
 license             MIT
 categories          ruby
-platforms           any
+platforms           darwin
```
amake, line 10: Is there a reason for this change? Since rbenv is just a bash script, I believe it should run on other platforms as well.

amake:
I think this approach makes more sense than #16273, but I would suggest one of the following changes:

- I think users of rbenv will probably usually want ruby-build, so the `rubybuild` variant should be a default variant
- Or, alternatively, you could remove the variant entirely and add a `notes` block to rbenv advertising ruby-build

ruby/rbenv/Portfile
halostatue: The change here was related to `port lint`. I have started following more MacPorts messages, and it seems that `port lint` has not been updated to recognize `any` as a valid value, although it is the correct choice. I will revert it (I will resolve after I have made the change).

ruby/rbenv/files/patch-libexec-rbenv-init.diff
halostatue: This patch was previously included. It may not be necessary any longer. I will need to check.

halostatue:
I’ll look at the suggested changes and try to get them in place over the next day or so. I don’t use `rbenv`, preferring `chruby` (and now I mostly don’t even use that because I have a `chruby`-like `direnv` handler function). It’s also somewhat difficult because I now use fish instead of bash or zsh, so…

I think that what’s better will be to switch to adding a `notes` block suggesting the installation of `ruby-build` to get the plug-in capability. Should we also add a `notes` block to `ruby-build` so that when someone upgrades to this revision they will see that there is no longer an automatic connection to `rbenv`?

halostatue:
I believe that I have resolved most of the issues. I have reworded the commit messages and updated the description of the PR as well.

amake:
I am happy with this, but since it's a significant change (and may go against the intention of the original port authors) I think we should wait for @mojca.

amake (approved):

halostatue:
Also updated ruby-build to version `20230208`.

I’m wondering if there’s any way to get action on this PR and its associated ticket, because the original dependency between `ruby-build` and `rbenv` was incorrect (if anything, the dependency should have been *reversed* as most people that I know who use `rbenv` do so with `ruby-build`, whereas `ruby-build` is one of two general Ruby installers). I’ve subscribed to release notifications for `ruby-build`, so I’m also happy to become a maintainer for this port.

Please note that I have closed #16273 as I don’t believe that is the correct fix, and do not wish to keep updating it for new `ruby-build` releases.

amake:
I think this indisputably the right approach and given the maintainer's inaction I will just go ahead and merge it.

PR #16359, opened 2022-10-15
gflags: blacklist gcc-4.0, fixes build on 10.5.8

barracuda156:
Fixes: https://trac.macports.org/ticket/65994

#### Description

On Leopard gcc-4.0 gets invoked, which fails. Blacklisted.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8
Xcode 3.1.4

###### Verification 
Have you

### DIFF
--- a/devel/gflags/Portfile
+++ b/devel/gflags/Portfile
@@ -26,8 +26,9 @@ checksums           rmd160  9748753d2cf4d581cd0881bd5f5c1e3dac365896 \
 # While gflags builds fine with gcc-4.2, its dependent google-glog does not.
 # However they both should be built by the same compiler, otherwise google-glog build fails.
 # https://trac.macports.org/ticket/65645
+# https://trac.macports.org/ticket/65994
 compiler.blacklist-append \
-                    *gcc-4.2 *gcc-4.3 *gcc-4.4 *gcc-4.5
+                    *gcc-4.0 *gcc-4.2 *gcc-4.3 *gcc-4.4 *gcc-4.5
 
 # Rosetta ignores blacklist and still tries to use llvm-g++-4.2.
 # Prioritize newer gcc; list versions that are realistically used and confirmed to build and work.


kencu:
this Portfile also needs a cleanup…

no port like this should be whitelisting gccs like that.

no port should be forcing this: GLIBCXX_USE_CXX11_ABI=0

barracuda156:
@kencu Need to be careful with these dependencies of `folly`, it was a painful and time-consuming process to fix them all. But of course we can change syntax to a better one.

kencu:
I hear you -- these were early fixes you did.

In general, there should be no whitelisting of compilers in a Portfile. it is extremely fragile. I would only do that for something that would never be expected to change, like building clang11-bootstrap with gcc10-bootstrap.

The ABI flag was done because of wacky errors seen in an issue upstream I believe. That was never the right thing to do for a MacPorts build, and the MacPorts issue, if there was one, should have been handled a different way other than forcing that ABI flag in the Portfile.

barracuda156:
@kencu By the way, a relevant question: is there a way to ensure that given ports (two or more) are built with identical compiler? Preferably without setting up variants and then making specific one required.

kencu:
> @kencu By the way, a relevant question: is there a way to ensure that given ports (two or more) are built with identical compiler? Preferably without setting up variants and then making specific one required.

There are several sets of ports that have this issue in the ports tree mix. What is currently done is to set them up with the same compiler.cxx_standard and compiler.c_standard, so they will then build with the same compilers. Also, set a note in each Portfile indicating such was needed, so that someone doesn't come along later, not understand why that was done, and undo it without knowing why it was set that way.

ryandesign (approved):
While cleaning up the Portfile further would be great, the simple change made in this PR seems completely correct and mergeable as is.

barracuda156:
@ryandesign I will do cleanup as @kencu suggested, but it has to be done for all `folly` dependencies and dependents alike, which will require some testing so that nothing breaks.

So let me do it in a separate PR – and after this one, the one with fixes to `folly` for Leopard and the one updating `fb303` are merged.

PR #16381, opened 2022-10-17
pg debug: add variant debugoptimized

lukaso:
#### Description

This port group allows builders to create ports where debugging symbols are created but the code is still optimized. This is primarily useful to packagers of applications such as GIMP ([issue](https://gitlab.gnome.org/Infrastructure/gimp-macos-build/-/issues/4)) where the code needs to be optimized, but having debug symbols greatly improves the debuggability of crashes in the application.

This port group conflicts with the debug port group but is based on it.

Another approach was considered, which was to override the `debug.flags.delete` and `debug.flags.add` variables, but it was not possible to determine how this could be done in a new variant built into a port.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0.1 14A400

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/debug-1.0.tcl
+++ b/_resources/port1.0/group/debug-1.0.tcl
@@ -14,6 +14,10 @@ if { [variant_exists debug] } {
     error "pg_debug: variant 'debug' already exists"
 }
 
+if { [variant_exists debugoptimized] } {
+    error "pg_debug: variant 'debugoptimized' already exists"
+}
+
 default debug.configure \
     [list \
         cflags \
@@ -28,19 +32,24 @@ default debug.configure \
 
 default debug.flags.delete \
     [list -O1 -O2 -O3 -Os -mtune=native -DNDEBUG -DNDEBUG=1]
-
 default debug.flags.add \
     [list -g -O0]
 
-ui_debug "pg_debug: adding variant"
-variant debug description {Enable debug flags and symbols} {}
+default debugoptimized.flags.delete \
+    [list]
+default debugoptimized.flags.add \
+    [list -g]
 
-proc debug::setup_debug {} {
-    ui_debug "debug::setup_debug: configuring for debug build"
+ui_debug "pg_debug: adding variants"
+variant debug conflicts debugoptimized description {Enable debug flags and symbols} {}
+variant debugoptimized conflicts debug description {Enable debug flags and symbols, while building optimized code} {}
+
+proc debug::setup_debug {p_debug_mode} {
+    ui_debug "debug::setup_debug: configuring for debug mode: ${p_debug_mode}"
 
     set conf_names   [option debug.configure]
-    set flags_delete [option debug.flags.delete]
-    set flags_add    [option debug.flags.add]
+    set flags_delete [option ${p_debug_mode}.flags.delete]
+    set flags_add    [option ${p_debug_mode}.flags.add]
 
     foreach c ${conf_names} {
         foreach f ${flags_delete} {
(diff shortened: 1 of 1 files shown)


lukaso:
cc @mascguy, this would really help with building GIMP using MacPorts. Otherwise, I'll have to create copies of the relevant ports where I need debugging symbols to be included in the release.

The relevent GIMP ticket: https://gitlab.gnome.org/Infrastructure/gimp-macos-build/-/issues/4

kencu:
So great you’re using MacPorts to build and bundle gimp.

I’ve been helping with it for years, at it has been building for us all the back to MacOS 10.5, and also on PPC!

I just want to mention here an active user who carries a few patches for it who I’m. sure would like to contribute. @rmottola , here’s the MacOS gimp upstream for you! 

Re: your debug symbols, I’m sure you know, you quite easily maintain a ports tree “overlay” that makes a few modest mods you might need to the existing ports tree. Lots of us have a need for that, for various reasons.

lukaso:
Ken,

Thanks. I know I can overlay a copy of a complete port, but not how to
overlay just a few attributes of a port. I’d like to maintain as few ports
as possible 😀

I also couldn’t figure out how to create options/variants in a port group
otherwise I would have done that.

Any help on either count gratefully accepted.

kencu:
To accomplish what you are after today, you could use the existing debug Portgroup, and override the default flags to strip and add what you want, as you indicated above.

https://github.com/macports/macports-ports/blob/44e3fdf88906f4c5f403c28fe0785119d0e1fca3/_resources/port1.0/group/debug-1.0.tcl#L29

So add this to gimp, for example:

```
PortGroup debug 1.0

# leave optimization flags alone
debug.flags.delete -g -DNDEBUG -DNDEBUG=1
debug.flags.add -g
```

and I think you might well be done, and have what you are after.

It is also possible to add a new option to the debug PortGroup, something like:

```
options debug.optimized
default debug.optimized no
```
and then down in the body of the PortGroup, you would branch on whether that option is set or not:

```
if {[tbool debug.optimized]} then {
 # do some flag modification stuff
} else {
 # do some other flag modification stuff
}
```

but then you might wonder if that is really any easier than doing it as above, or if it just adds needless confiusion, and you have to also handle properly what to do if the user has modified several of the variables, for example setting ```debug.optimized yes``` but then also modifying ```debug.flags.delete``` -- to make sure the user will get what is expected...

I often think MacPorts shoots itself in the foot by trying to be too clever about options, etc, rather than just let the user set the needed flags directly.

kencu:
To have a Macports' port tree overlay with some custom mods in it, you would use a git checkout of the ports tree as your primary source of ports, and then make your mods on top of that.

When you do a port selfupdate or port sync, macports uses git autostash so your changes will be rebased on top of the pulled update.

If your changes are adding a new portgroup file and adding that portgroup to a few ports, for example, conflicts are quite unlikely.

I have to check what happens if you solidify your changes as an actual commit on top of the git checked-out ports tree. I think your commit(s) are rebased on top of the pull, but I'm not certain about that as I haven't done it.

lukaso:
> So add this to gimp, for example:
> 
> ```
> PortGroup debug 1.0
> 
> # leave optimization flags alone
> debug.flags.delete -g -DNDEBUG -DNDEBUG=1
> debug.flags.add -g
> ```

I couldn't figure this out for the life of me! Thank you.

On thinking further, I want to be able to set a variant globally, that I can then pick up when building on CI, which is why `+debugoptimized` does the trick really well. If I can add that variant to the debug PortGroup, then that would work...which this trick should allow me to try.

lukaso:
> To have a Macports' port tree overlay with some custom mods in it, you would use a git checkout of the ports tree as your primary source of ports, and then make your mods on top of that.
> 
> When you do a port selfupdate or port sync, macports uses git autostash so your changes will be rebased on top of the pulled update.
> 
> If your changes are adding a new portgroup file and adding that portgroup to a few ports, for example, conflicts are quite unlikely.
> 
> I have to check what happens if you solidify your changes as an actual commit on top of the git checked-out ports tree. I think your commit(s) are rebased on top of the pull, but I'm not certain about that as I haven't done it.

Hmm. Not a huge fan of doing that kind of thing. I have to stand up various installations of MacPorts with diverse caching mechanisms, so I feel there are a lot of things that could go wrong. But let's see if I can get `+debugoptimized` in and then it won't be needed :)

lukaso:
@mascguy any chance to have a look at this. Would be much appreciated! (Hope your holiday was a good one.)

mascguy:
Lukas, just pushed a new commit, to add this support to our existing pg `debug`: It doesn't make sense to create a new portgroup, given that the vast majority of the logic is identical.

I've tested locally, and looks good.

And thank you so much for the idea/contribution, it's definitely a good one!

mascguy (approved):

lukaso:
@mascguy thanks so much for doing this.

PR #16385, opened 2022-10-18
mplayer: update to 1.5

kencu:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.6, 10.7 and 10.14

Proposing a massive simplification of the MPlayer port. The current version is nearly unmaintainable. This version has no variants at all -- one default supported build.

It builds and works nicely on the above systems, and appears maintainable.


kencu:
I removed Jeremy as maintainer as he hasn't touched this in years, and he's not around much.

If someone wants his name back on, no problem. That is not the focus here.

kencu:
ok, let’s go with this a see how this plays out.

PR #16398, opened 2022-10-18
emacs{,-app}-devel: update to 20221107

catap:
###### Type(s)

- [x] enhancement

###### Tested on

macOS 12.6 21G115 x86_64
Xcode 14.0.1 14A400

###### Verification 
Have you

### DIFF
--- a/editors/emacs/Portfile
+++ b/editors/emacs/Portfile
@@ -10,7 +10,7 @@ legacysupport.newest_darwin_requires_legacy 13
 name            emacs
 categories      editors
 license         GPL-3+
-maintainers     {dports @drkp} openmaintainer
+maintainers     {dports @drkp} {@catap korins.ky:kirill} openmaintainer
 description     The GNU Emacs text editor
 long_description \
     GNU Emacs is a self-documenting, customizable, extensible real-time \
@@ -101,19 +101,17 @@ if {$subport eq $name || $subport eq "emacs-app"} {
 }
 
 if {$subport eq "emacs-devel" || $subport eq "emacs-app-devel"} {
-    set date        2022-09-20
+    PortGroup       github 1.0
+
+    github.setup    emacs-mirror emacs 802671493d7c18569dac47ca91a4b7d6e693aff6
     epoch           5
-    version         [string map {- {}} ${date}]
-    revision        0
+    version         20221107
 
-    fetch.type      git
-    git.url         --shallow-since=${date}T00:00:00 https://github.com/emacs-mirror/emacs.git
-    git.branch      e5de29aa47e96f675b7f4ef3a7f84bd30d68ea0f
+    master_sites    ${github.master_sites}
 
-    # --shallow-since needs a newer version of git than on some older systems
-    # So use MacPorts version
-    depends_fetch-append port:git
-    git.cmd         ${prefix}/bin/git
+    checksums       rmd160  cea40725dd2593dcb375210c95f1c0aca4961df7 \
+                    sha256  a318cc46ea0b143136693e539d3f031039e03a3361ced1877963eab8ae5aa59f \
+                    size    46262466
 
     configure.args-append \
         --with-sqlite3 \
@@ -274,7 +272,13 @@ variant nativecomp description {Builds emacs with native compilation support} {
     compiler.cpath-prepend         ${prefix}/include/gcc${gcc_v}
     compiler.library_path-prepend  ${prefix}/lib/gcc${gcc_v}
     configure.ldflags-append       "-Wl,-rpath ${prefix}/lib/gcc${gcc_v}"
-    build.args-append              NATIVE_FULL_AOT=1
+
+    if {$subport eq "emacs-devel" || $subport eq "emacs-app-devel"} {
+        configure.args-replace     --with-native-compilation \
+                                   --with-native-compilation=aot
+    } else {
+        build.args-append          NATIVE_FULL_AOT=1
+    }
 }
 
 default_variants-append +nativecomp


catap:
@drkp is it fine for you I'd like to be a co-mainater for this port.

catap:
Just made one more update

catap:
@amake do you think is it possible to merge?

editors/emacs/Portfile
```diff
@@ -101,19 +101,16 @@ if {$subport eq $name || $subport eq "emacs-app"} {
...
     epoch           5
-    version         [string map {- {}} ${date}]
+    version         20221029
+    set git_commit  ce527464440ca070e752ca18a68ebf2d369a14df
     revision        0
 
-    fetch.type      git
-    git.url         --shallow-since=${date}T00:00:00 https://github.com/emacs-mirror/emacs.git
-    git.branch      e5de29aa47e96f675b7f4ef3a7f84bd30d68ea0f
+    master_sites    https://github.com/emacs-mirror/emacs/archive/${git_commit}.tar.gz?dummy=
```
amake, line 109: My understanding is that fetching from git is desirable because we don't really want to waste mirror space on these ephemeral development versions. Is there a particular reason to switch to tarball?

catap: @amake to enjoy checksums! I may add `NoMirror` to exclude it. Or revert this changes.

BTW `-devel` port was introduced here: https://github.com/macports/macports-ports/commit/fb4f094976b6fe31f22263492f5348f8e1e380b5 and it was made with `fetch.type git` because git.savannah.gnu.org hasn't got a nice way to download an archive from it.

The switch to github was made here https://github.com/macports/macports-ports/commit/f62654eb19904b3bdc34336d94cb7f799a01ebe2

And I've introduced `--depth` to make things faster at https://github.com/macports/macports-ports/commit/68c42a1a1bff35daea513a761bdf9da6cf95192e

Which was converted to `--shallow-since` with dependency to git.

=> I haven't found any reason to not use github archive's to be honest.

amake: I have no problem pulling the archive from GitHub. My concern was first about mirroring, and second about the fact that the distfile will now sit on my machine unless I go to delete it (admittedly a small cost).

But if we are going to go this route, wouldn't it make more sense to either use `PortGroup github` or at least follow the same conventions? This is almost `github.tarball_from archives` but a little different. Why `archives` and not `tarball`?

amake:
I am running this version of Emacs.app and it's working fine so far.

drkp:
OK to merge on my end (sorry, have been away on vacation)

catap:
@drkp, @amake may I ask anyone to merge it? Or I may update it one more time :)

catap:
here it is

editors/emacs/Portfile
amake: @catap You requested my review again; maybe you missed my questions above?

catap: @amake sorry, I've missed your question.

To be honest I like an idea with `github` PG and just rebased my PR to this idea. It makes things cleaner.

editors/emacs/Portfile
```diff
@@ -101,19 +101,17 @@ if {$subport eq $name || $subport eq "emacs-app"} {
...
+    github.setup    emacs-mirror emacs 802671493d7c18569dac47ca91a4b7d6e693aff6
     epoch           5
-    version         [string map {- {}} ${date}]
-    revision        0
+    version         20221107
 
-    fetch.type      git
-    git.url         --shallow-since=${date}T00:00:00 https://github.com/emacs-mirror/emacs.git
-    git.branch      e5de29aa47e96f675b7f4ef3a7f84bd30d68ea0f
+    master_sites    ${github.master_sites}
```
amake, line 110: Thanks for moving to `PortGroup github`. It's less surprising this way.

Do we need to set `master_sites` here? I think the call to `github.setup` should set it, but maybe there's some magic that it does to not overwrite values set earlier?

catap: `github.setup` defines `default master_sites`. This Portfile is specify at begining `master_sites gnu` => I must switch to use PG's master_sites.

amake: OK thanks.

amake:
I'm building the new version and I'll merge when I've smoke tested it.

Or if others are confident they can go ahead and merge when they feel like it.

editors/emacs/Portfile
```diff
@@ -274,7 +272,13 @@ variant nativecomp description {Builds emacs with native compilation support} {
...
     compiler.library_path-prepend  ${prefix}/lib/gcc${gcc_v}
     configure.ldflags-append       "-Wl,-rpath ${prefix}/lib/gcc${gcc_v}"
-    build.args-append              NATIVE_FULL_AOT=1
+
+    if {$subport eq "emacs-devel" || $subport eq "emacs-app-devel"} {
+        configure.args-replace     --with-native-compilation \
+                                   --with-native-compilation=aot
+    } else {
+        build.args-append          NATIVE_FULL_AOT=1
+    }
```
amake, line 281: On first launch I'm seeing a lot of internal packages being natively compiled. Are we sure that `NATIVE_FULL_AOT` is no longer needed?

amake: I don't fully understand this, but it looks like `--with-native-compilation=aot` and `NATIVE_FULL_AOT=1` do the opposite thing (the former appears wrong)?

https://github.com/emacs-mirror/emacs/blob/80c3fadfd57ee797404c80b8e88371c53edecfe5/lisp/Makefile.in#L35-L42

amake: `NATIVE_SKIP_NONDUMP` is used [like this](https://github.com/emacs-mirror/emacs/blob/68e2a9cd92996dd55ff9599dc1bf99c70070cddb/lisp/Makefile.in#L377) so actually I guess the flags are OK.

catap:
@amake thanks!

PR #16407, opened 2022-10-19
boost: fix context for PPC

barracuda156:
[skip ci] – because changes concern _only_ PPC, which is not built anyway.

#### Description

PR fixes `context` for PPC.
Upstream commits:
https://github.com/boostorg/context/commit/a8f60b95767383a9d3f13c36287fc84d14be9ed3
https://github.com/boostorg/context/commit/10310f56642929d78bb4a8621145472fbee49982

P. S. We do not need to revbump 1.80, since it does not build with GCC at the moment: https://github.com/boostorg/json/issues/774

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

macOS 10A190
Xcode 3.2

macOS 10.5.8
Xcode 3.1.4

###### Verification 
Have you

### DIFF
--- a/devel/boost169/Portfile
+++ b/devel/boost169/Portfile
@@ -6,7 +6,7 @@ PortGroup       active_variants 1.1
 PortGroup       compiler_blacklist_versions 1.0
 
 version         1.69.0
-revision        8
+revision        9
 
 set branch      [join [lrange [split ${version} .] 0 1] .]
 set tag         [string map {. {}} ${branch}]
@@ -94,6 +94,11 @@ post-patch {
     reinplace "s|__MACPORTS_CXX__|${configure.cxx}|g" ${worksrcpath}/tools/build/src/tools/clang-darwin.jam
 }
 
+# Fix context for Darwin PPC
+# https://github.com/boostorg/context/commit/a8f60b95767383a9d3f13c36287fc84d14be9ed3
+# https://github.com/boostorg/context/commit/10310f56642929d78bb4a8621145472fbee49982
+patchfiles-append patch-darwin-ppc.diff
+
 proc write_jam s {
     global worksrcpath
     set config [open ${worksrcpath}/user-config.jam a]
@@ -129,15 +134,6 @@ configure.args      --without-libraries=python \
 # boost build scripts default to clang on darwin
 if {[string match *gcc* ${configure.compiler}]} {
     configure.args-append --with-toolset=gcc
-}
-
-# makecontext/swapcontext introduced in 10.6
-# In 10.6.x Rosetta they do not work: https://trac.macports.org/ticket/64978
-platform darwin {
-    if {${os.major} < 10 || (${os.major} == 10 && ${build_arch} eq "ppc")} {
-        configure.args-append   --without-libraries=context \
-                                --without-libraries=coroutine
-    }
 }
 
 configure.universal_args
--- a/devel/boost169/files/patch-darwin-ppc.diff
+++ b/devel/boost169/files/patch-darwin-ppc.diff
@@ -1,0 +1,660 @@
+--- libs/context/src/asm/jump_ppc32_sysv_macho_gas.S.orig	2021-04-14 00:21:41.000000000 +0800
++++ libs/context/src/asm/jump_ppc32_sysv_macho_gas.S	2022-10-06 09:41:54.000000000 +0800
+@@ -80,122 +80,122 @@
+     ; reserve space on stack
+     subi  r1, r1, 244
+ 
+-    stfd  f14, 0(r1)  # save F14
+-    stfd  f15, 8(r1)  # save F15
+-    stfd  f16, 16(r1)  # save F16
+-    stfd  f17, 24(r1)  # save F17
+-    stfd  f18, 32(r1)  # save F18
+-    stfd  f19, 40(r1)  # save F19
+-    stfd  f20, 48(r1)  # save F20
+-    stfd  f21, 56(r1)  # save F21
+-    stfd  f22, 64(r1)  # save F22
+-    stfd  f23, 72(r1)  # save F23
+-    stfd  f24, 80(r1)  # save F24
+-    stfd  f25, 88(r1)  # save F25
+-    stfd  f26, 96(r1)  # save F26
+-    stfd  f27, 104(r1)  # save F27
+-    stfd  f28, 112(r1)  # save F28
+-    stfd  f29, 120(r1)  # save F29
+-    stfd  f30, 128(r1)  # save F30
+-    stfd  f31, 136(r1)  # save F31
+-    mffs  f0  # load FPSCR
+-    stfd  f0, 144(r1)  # save FPSCR
+-
+-    stw  r13, 152(r1)  # save R13
+-    stw  r14, 156(r1)  # save R14
+-    stw  r15, 160(r1)  # save R15
+-    stw  r16, 164(r1)  # save R16
+-    stw  r17, 168(r1)  # save R17
+-    stw  r18, 172(r1)  # save R18
+-    stw  r19, 176(r1)  # save R19
+-    stw  r20, 180(r1)  # save R20
+-    stw  r21, 184(r1)  # save R21
+-    stw  r22, 188(r1)  # save R22
+-    stw  r23, 192(r1)  # save R23
+-    stw  r24, 196(r1)  # save R24
+-    stw  r25, 200(r1)  # save R25
+-    stw  r26, 204(r1)  # save R26
+-    stw  r27, 208(r1)  # save R27
+-    stw  r28, 212(r1)  # save R28
+-    stw  r29, 216(r1)  # save R29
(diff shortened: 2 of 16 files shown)


michaelld (approved):
LGTM; thx!

michaelld:
Excellent work @barracuda156 !

barracuda156:
@michaelld Thank you!

barracuda156:
@michaelld Good to merge? Or we do it after `icu` is updated, to avoid revbumping twice?

michaelld:
hoping for a review from @mascguy  ...

michaelld:
what is the PR for the ICU update?

barracuda156:
> what is the PR for the ICU update?

@michaelld We have to ask @catap perhaps, when he is gonna update the regular `icu` port (`icu-devel` already updated).

evanmiller:
Have you tested the context functionality? I see that in one of the PRs that you built `folly`, but I just wanted to make sure that this works as intended at run-time. I believe there are some tests and examples included in the Boost context source tree.

barracuda156:
@michaelld So `icu` done, and I have rebased against master with subsequent revbumps. (No need to revbump `boost180`, which does not build with GCC anyway, so cannot have been built for PPC.)

@evanmiller If you could help here, it would be awesome.

evanmiller:
@barracuda156 I tested previously by running

```
g++-mp-7 fibonacci.cpp -lboost_context-mt -L/opt/local/lib
```

on [fibonacci.cpp](https://github.com/boostorg/context/blob/develop/example/fiber/fibonacci.cpp). If that prints the Fibonacci sequence I will be satisfied :-).

barracuda156:
@evanmiller Well, it compiles, but does not run. I am not sure if it is a context-specific failure though or Boost-general failure.

Here Boost atomics fail with Bus error (context is _not_ used): https://github.com/PurpleI2P/i2pd/issues/1726?notification_referrer_id=NT_kwDOBXwLlrMzMDg4MjkzMTYzOjkyMDE1NTEw#issuecomment-1299053779

Compare with what I got:
```
sergey-fedorovs-power-mac-g5:Dev svacchanda$ /opt/local/bin/g++-mp-11 fibonacci.cpp -I/opt/local/libexec/boost/1.76/include -lboost_context-mt -L/opt/local/libexec/boost/1.76/lib -o boost_test
sergey-fedorovs-power-mac-g5:Dev svacchanda$ /Users/svacchanda/Dev/boost_test 
Bus error
```
```
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_PROTECTION_FAILURE at address: 0x00000000
0x0001a200 in jump_fcontext ()
(gdb) where
#0  0x0001a200 in jump_fcontext ()
Cannot access memory at address 0x0
#1  0x00000000 in ?? ()
```
Something seems to be broken with memory access. Any ideas welcome.

kencu:
build it like this, perhaps:

```
g++-mp-7 fibonacci.cpp -I/opt/local/libexec/boost/1.76/include  -std=c++17 -lboost_context-mt -L/opt/local/libexec/boost/1.76/lib -g -O0
```

then load it into the debugger:
```
ggdb ./a.out
```
then set a breakpoint at main:
```
b main
```
then run it:
```
run
```
when it hits the breakpoint at main, you should be able to "step" one step at a time through the source code until you see exactly where it bus errors.

The few steps before that is usually the problem.

You can also look at the backtraces in the crashlog folder. Once it is build with debugging symbols and no optimization, some combination of the crashlogs and what you see stepping through will give you, hopefully, some hint about what is going wrong.

Next step is for you to sort out how to examine the frame variables in the debugger and see what they look like just before the call that causes the bus error.

evanmiller:
@barracuda156 The context error and the atomics error look like two separate issues.

You might also try kencu's advice using the `gdb-apple` port, which I have used successfully on PPC.

barracuda156:
@kencu @evanmiller Here is what I get when trying `b main` with OS `gdb`:
```
(gdb) b main
Die: DW_TAG_unspecified_type (abbrev = 5, offset = 71)
	has children: FALSE
	attributes:
		DW_AT_name (DW_FORM_strp) string: "decltype(nullptr)"
Dwarf Error: Cannot find type of die [in module /Users/svacchanda/Dev/boost/a.out.dSYM/Contents/Resources/DWARF/a.out]
(gdb) run
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_PROTECTION_FAILURE at address: 0x00000000
0x0001a200 in jump_fcontext ()
```
Same from `gdb-apple`:
```
(gdb) b main
Breakpoint 1 at 0x2318: file fibonacci.cpp, line 14.
Breakpoint 2 at 0x2278: file fibonacci.cpp, line 17.
warning: Multiple breakpoints were set.
Use the "delete" command to delete unwanted breakpoints.
(gdb) run
Starting program: /Users/svacchanda/Dev/boost/a.out 
Unable to find Mach task port for process-id 14993: (os/kern) failure (0x5).
```

barracuda156:
So with `gdb-apple`, once run with `sudo`, this is the output:
```
Breakpoint 1, main () at fibonacci.cpp:27
27	        }};
(gdb) c
Continuing.

Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_PROTECTION_FAILURE at address: 0x00000000
0x0001a200 in jump_fcontext ()
```

kencu:
there is a learning curve to debugging, and you will enjoy it!

Of course, if you set a breakpoint at main(), and then just hit "c" to continue once you stop there, you will crash without much further information than you had before.:> 

Look back at what I wrote above. You need to step, etc. 

There are many tutorials and books about how to debug with gdb available on the internet and web and at bookstores and online classes. It would frankly be just too hard for me to type that all in here, even if I did have the time to do that.

barracuda156:
@kencu Sure, sorry :) I will deal with this in coming days.

barracuda156:
I will update with some notes first, to keep track.

One thing that we have wrong is “hidden”, which was inherited from Linux and should not be used on Darwin. Found some info here: https://www.mail-archive.com/ports@openbsd.org/msg89321.html

barracuda156:
@evanmiller I have also tried using versions from here with no changes besides correcting a typo: https://github.com/twlostow/libcontext
And they do not work either.

We need help with magic code :)

olk:
Does this happen on a PPC32 or PPC64 system?

`Cannot access memory at address 0x0` - means NULL-pointer.
Probably the assembler instructions / memory alignment isn't correct (check the calling convetions ABI) and requires debugging - stepping through the assembler instructions and comparing the content of the registers.
Unfortunately I don't own a PPC system and I'm too busy.

barracuda156:
@olk I have currently tested only ppc32, since at the moment building Boost as ppc+ppc64 is broken in Macports. Yeah, I will try to debug and figure out what goes wrong. Alignments might be the issue indeed.

evanmiller:
I have an upstream PR going here if any PPCheads want to take it for a spin: https://github.com/boostorg/context/pull/215

Something I noticed though was that while Context and Coroutine now compile on PPC, the Fiber module fails with assembly errors using GCC7 (G++7). So even with the proposed fixes, we might have to keep Fiber excluded on these platforms, at least for `boost181` (haven't tried previous versions).

barracuda156:
> I have an upstream PR going here if any PPCheads want to take it for a spin: https://github.com/boostorg/context/pull/215
> 
> 
> 
> Something I noticed though was that while Context and Coroutine now compile on PPC, the Fiber module fails with assembly errors using GCC7 (G++7). So even with the proposed fixes, we might have to keep Fiber excluded on these platforms, at least for `boost181` (haven't tried previous versions).

@evanmiller I think we a) need to test boost176, which is the default as of now, and b) try building with newer GCC.
If that fails, we should fix fiber rather than disabling it, IMHO.

evanmiller:
> @evanmiller I think we a) need to test boost176, which is the default as of now, and b) try building with newer GCC. If that fails, we should fix fiber rather than disabling it, IMHO.

Let's get Context working then we can worry about enabling Fiber. Since Fiber depends on Context, it's currently implicitly disabled on PPC.

barracuda156:
> > @evanmiller I think we a) need to test boost176, which is the default as of now, and b) try building with newer GCC. If that fails, we should fix fiber rather than disabling it, IMHO.
> 
> Let's get Context working then we can worry about enabling Fiber. Since Fiber depends on Context, it's currently implicitly disabled on PPC.

@evanmiller My concern is that it will take another forever to get the additional fix merged. However, this is your PR, I just expressed my opinion. (It is still better to have something fixed than nothing at all, of course.)

barracuda156:
P. S. It may be worth adding this too: https://github.com/boostorg/smart_ptr/commit/97c9204a952f60d13def0e321c0a92ec6dec148b

Not context-related, but ppc32-related.

barracuda156:
Closing in favor of this: https://github.com/macports/macports-ports/pull/17086

PR #16434, opened 2022-10-21
icu-devel: clean used python version

catap:
#### Description

See: https://trac.macports.org/ticket/65870

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

macOS 12.6 21G115 x86_64
Xcode 14.0.1 14A400

###### Verification 
Have you

### DIFF
--- a/devel/icu-devel/Portfile
+++ b/devel/icu-devel/Portfile
@@ -1,6 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:filetype=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
+PortGroup           clang_dependency 1.0
 
 name                icu-devel
 set my_name         icu
@@ -116,10 +117,20 @@ if {${subport} eq ${name} || ${subport} eq "${name}-lx"} {
     test.target     check
     test.args       VERBOSE=1
 
+    # icu4c requires python2.7+ regardless of configure:
+    #   checking for python3...
+    # libxml2 which is used to compile llvm/clang depends on icu,
+    # and adding MacPorts' python may introduce a dependency loop.
+    # See: https://unicode-org.atlassian.net/browse/ICU-20301
+    # See: https://trac.macports.org/ticket/65870
     if {${os.platform} eq "darwin" && ${os.major} < 11} {
         depends_build-append    port:python27
         license_noconflict      python27
         configure.python        ${prefix}/bin/python2.7
+    } elseif {${os.platform} eq "darwin" && ${os.major} >= 19} {
+        configure.python        /usr/bin/python3
+    } else {
+        configure.python        /usr/bin/python
     }
 
     if {${build_arch} eq "ppc" || ${build_arch} eq "ppc64"} {


catap:
@ryandesign may I ask you to review this changes? I'll plan to synchronise it with `icu` port when I'll update it.

devel/icu-devel/Portfile
```diff
@@ -116,10 +116,20 @@ if {${subport} eq ${name} || ${subport} eq "${name}-lx"} {
     test.target     check
     test.args       VERBOSE=1
 
+    # icu4c requires python2.7+ regardless of configure:
+    #   checking for python3...
+    # libxml which is used to compile llvm/clang depends on icu,
```
ryandesign, line 121: It's libxml2, not libxml. The libxml port is for the ancient 1.x version.

devel/icu-devel/Portfile
```diff
@@ -116,10 +116,20 @@ if {${subport} eq ${name} || ${subport} eq "${name}-lx"} {
...
+    # and adding MacPorts' python may introduce a dependency loop.
```
ryandesign, line 122: Do you really think it will cause a dependency loop? The port has been depending on python27 already for a long time without (as far as I know) any problems. The libxml2 and icu and python27 ports already include the clang_dependency portgroup, which instructs MacPorts to avoid using clang 5 or later, which should avoid a dependency cycle. It would be good not to make the user install yet another python port (python27-bootstrap) unless that's really needed.

I note however that icu-devel does not include the clang_dependency portgroup, and although icu-devel is not a clang dependency, icu is, and the purpose of icu-devel is to test the suitability of a new version of icu prior to updating the main port, so it might be a good idea to make icu-devel use the clang_dependency portgroup too so that it uses the same (older) compiler that icu eventually will, to ensure there are no problems with that.

devel/icu-devel/Portfile
```diff
@@ -116,10 +116,20 @@ if {${subport} eq ${name} || ${subport} eq "${name}-lx"} {
...
+    # See: https://unicode-org.atlassian.net/browse/ICU-20301
+    # See: https://trac.macports.org/ticket/65870
     if {${os.platform} eq "darwin" && ${os.major} < 11} {
-        depends_build-append    port:python27
+        depends_build-append    port:python27-bootstrap
         license_noconflict      python27
-        configure.python        ${prefix}/bin/python2.7
+        configure.python        ${prefix}/libexec/python27-bootstrap/bin/python2.7
+    } elseif {${os.platform} eq "darwin" && ${os.major} < 19} {
+        configure.python        /usr/bin/python3
```
ryandesign, line 130: /usr/bin/python3 does not exist on Darwin < 19. Did you mean:
```suggestion
    } elseif {${os.platform} eq "darwin" && ${os.major} >= 19} {
        configure.python        /usr/bin/python3
```

devel/icu-devel/Portfile
catap: thanks for catching such typo

devel/icu-devel/Portfile
catap: Right now python27 is enough to build it. As you read in the ticket some users are confused by this configure line and I'd like to keep it clean.

Anyway, in the future it might change.

catap:
@ryandesign thanks for review!

devel/icu-devel/Portfile
ryandesign: So again, why would switching from python27 to python27-bootstrap be good?

catap: @ryandesign to minimise dependency tree.

But, after some thinking I feel that I'm overengenering here. Let me revert this changes.

catap:
@pmetzger I guess yes, thanks!

PR #16467, opened 2022-10-25
libomp: Update to 15.0.3; drop -devel version.

eborisch:
Maintainer update, but worth testing since every clang-* depends upon it...


kencu (approved):
builds successfully for me  +universal on 10.6.8 using clang-3.7.

PR #16482, opened 2022-10-27
nodejs16: Make python a build dependency

Robotex:
#### Description

Makes python a build dependency instead of a library dependency because seems like it is not needed at runtime (used only for gyp), just like how the brew counterpart formula already declares

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0.1 14A400

###### Verification 
Have you

### DIFF
--- a/devel/nodejs16/Portfile
+++ b/devel/nodejs16/Portfile
@@ -13,7 +13,7 @@ compiler.cxx_standard   2014
 
 name                    nodejs16
 version                 16.18.1
-revision                0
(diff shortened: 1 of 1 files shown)


reneeotten (changes requested):
since you are changing a "library" dependency you'll need to increase the "revision"

reneeotten (approved):

PR #16486, opened 2022-10-28
mosh: Update to 1.4.0

neverpanic:
#### Description

Switch to GitHub and run autoreconf, because the website does not have a tarball for 1.4.0 (yet?).

Drop support for older perl branches. I doubt users care which version of perl is used for mosh as long as it works, so we might as well use a recent one.

Remove patches that have been merged upstream.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.6.1 21G217 x86_64
Xcode 13.4.1 13F100

###### Verification 
Have you


Grounded0:
Reported tarball missing at mosh.org to #mosh on Libera.

quentinmit:
There's some discussion on IRC/the mailing list about how they're going to do the release, so I'd like to hold off for a few days until the dust settles if that's okay.

neverpanic:
Sure, just let me know.

net/mosh/Portfile
```diff
@@ -38,12 +37,7 @@ depends_lib             port:ncurses \
                         port:p${perl5.major}-getopt-long \
                         port:p${perl5.major}-io-socket-ip
 
-patchfiles              network-bind.patch \
-                        patch-version-subdir.diff
-
-# Remove patch-version-subdir.diff and autoreconf when mosh > 1.3.2 is released.
 use_autoreconf          yes
```
quentinmit, line 40: `use_autoreconf yes` should no longer be necessary (and if it is, you need to not get rid of the args either)

net/mosh/Portfile
```diff
@@ -4,8 +4,8 @@ PortSystem              1.0
 PortGroup               perl5 1.0
 
 name                    mosh
-version                 1.3.2
-revision                18
+version                 1.4.0
+revision                0
```
quentinmit, line 8: Nit: revision 0 is the default, so you can just omit it now.

quentinmit (approved):

net/mosh/Portfile
neverpanic: I'm aware of that, but I think there was some discussion on the macports-dev list recently where we decided that keeping it in and set to 0 is not recommended over removing it.

kencu: because the revision inevitably increases for so many ports, and it can be tricky to put it in the right spot, and we all hoped for some automation to manipulate the revision when revbumping often large groups of ports, we decided to leave it in, even when resetting to zero…

quentinmit: Gotcha, I didn't see that thread go by. *shrug*

PR #16489, opened 2022-10-28
zig: fix depends_lib: clang-XX for libc++

eborisch:
bin/zig requires `libexec/llvm-{ver}/lib/libc++.1.dylib`, which is provided by `clang-{ver}`, not `llvm-{ver}`.

cc: @felix

### DIFF
--- a/lang/zig/Portfile
+++ b/lang/zig/Portfile
@@ -25,7 +25,7 @@ homepage            https://ziglang.org/
 
 set llvm_version    13
 
-depends_lib-append  port:llvm-${llvm_version}
+depends_lib-append  port:clang-${llvm_version}
 
 checksums           rmd160  0bee29c8a68f0bfb2bb0454783c8be8ff12f7080 \
                     sha256  2a6ba6a72b9619b83dab77f5b6e2b6f0958bb0f85cded055be2c632386e0ff2d \


i0ntempest (approved):

kencu:
If zig is linking that libc++.dylib by accident, when it is really going after the other libraries in llvm/clang, that is a potential problem we might have to deal with.

eborisch:
Perhaps; I'm certainly not a zig expert, but I (rev-upgrade, actually) noticed it broke when removing leaves due to the missing depends_lib. Changing lifrom llvm-X to clang-X still marks llvm-X as needed (via clang-X's depends_lib) so this was a low-risk change.

Working out if it shouldn't be linking against it at all is left as an exercise for the maintainer or those interested. ;)

cjones051073:
Indeed, the change here fxes the deps so is fine as is. *if* zig should be linking against clang's libc++ lib at all is another question.

kencu:
other pkg managers have had to move libc++.dylib out of the general llvm libraries folder because ports linking against those libraries were accidentally linking against the new libc++, and things were breaking.

zig should undoubtedly not be linking against the new libc++. The zig maintainer likely would be unaware of issues like this related to MacPorts’ packaging.

Yes, Chris, I know this narrowly fixes the deps… you know me, though… always looking for the deeper real issue.

PR #16496, opened 2022-10-28
sbcl: hardcode MacPorts XDG_DATA_DIRS

catap:
#### Description

Suggest for sbcl where it should also looks for installed packages.

This hardcoded path allows to start packagining lisp code to MacPorts.

###### Type(s)

- [x] bugfix
- [x] enhancement

###### Tested on

macOS 12.6 21G115 x86_64
Xcode 14.0.1 14A400

right now MacPorts has only one lisp port: cl-ppcre, so, I've used it to test it as:
```lisp
(require :asdf)
(require :cl-ppcre)
```

###### Verification 
Have you

### DIFF
--- a/lang/sbcl/Portfile
+++ b/lang/sbcl/Portfile
@@ -9,7 +9,7 @@ name            sbcl
 # Please bump the revision of math/maxima (and when it exists
 # math/maxima-devel) when this port changes.
 version         2.2.9
-revision        0
+revision        1
 
 categories      lang
 license         BSD
@@ -34,6 +34,7 @@ use_bzip2       yes
 
 patchfiles \
     patch-contrib-sb-posix-posix-tests.lisp.diff \
+    patch-macports-xdg-data-dir.diff \
     patch-sbcl-realtime.diff
 
 distfiles       ${name}-${version}-source${extract.suffix}:sbcl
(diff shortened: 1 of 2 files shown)


kakuhen:
Speaking from experience, packaging ASDF systems with package managers is very strange to me because virtually every Common Lisp user I know of will not be using this, and instead opt for manually compiling and loading ASDF systems or use Quicklisp (typically the official Quicklisp dist or Ultralisp).

With that said, this is just a strong personal opinion of mine. I do have a question that is relevant to this PR: is there any particular reason we patch the lisp file within MacPorts instead of e.g. making a patch with the same effect and submitting upstream? Unlike the other patches we apply (e.g. the MONOTONIC_CLOCK one), it is not derived from any copyrighted material that would conflict with the license of SBCL (mostly public domain and 2-clause BSD as far as I understand).

All of this aside, the patch seems ok to me, but the end goal is just hard for me to understand

catap:
@kakuhen I agree with your point, but my goal is packaging enough dependencies to bring [jfricas](https://github.com/macports/macports-ports/pull/16488) :)

Regarding upstream: I not sure how may I send such patch to upstream because MacPorts can be installed to different prefix, not only `/opt/local`.

kakuhen:
> Regarding upstream: I not sure how may I send such patch to upstream because MacPorts can be installed to different prefix, not only `/opt/local`.

I know that SBCL has a Makefile for various stages of bootstrapping, and there is likely a way to do some fancy text macro stuff there. But I agree, this will be more complicated than patching within MacPorts. I may try out my idea and see if I can do what I'm suggesting.

In any case, your patch seems good to go. It's pretty nice seeing maintenance on ports like these :)

catap:
@pmetzger may I ask you to review this PR? It was opened for more than a week and it is seems like maintainer timeout :)

PR #16517, opened 2022-10-30
crossbinutils: add build dep texinfo

Gcenx:
#### Description

Resolves "'makeinfo' is missing on your system" on macOS Ventura

###### Type(s)

- [x] bugfix

###### Tested on

macOS 13.0 22A380 x86_64
Xcode 14.1 14B47b

###### Verification 

### DIFF
--- a/_resources/port1.0/group/crossbinutils-1.0.tcl
+++ b/_resources/port1.0/group/crossbinutils-1.0.tcl
@@ -139,6 +139,9 @@ proc crossbinutils.setup {target version} {
             ${worksrcpath}/libiberty/Makefile.in
     }
 
+    depends_build \
+        bin:makeinfo:texinfo
+
     depends_lib \
         port:gettext \
         port:zlib


mojca:
Shouldn't these changes go to the portgroup?

(If this is only an issue on Ventura, maybe only add the dependency there? I'm not really sure about the latter though, I would like a second opinion. For runtime dependencies we strive for the same experience on all operating systems.)

Gcenx:
> Shouldn't these changes go to the portgroup?

Yeah I should have put this into the portgroup.

> (If this is only an issue on Ventura, maybe only add the dependency there? I'm not really sure about the latter though, I would like a second opinion. For runtime dependencies we strive for the same experience on all operating systems.)

I'd think it makes more sense to just use it on all versions.

mojca:
The update in `crossbinutils` looks ok to me. I didn't yet upgrade, so I cannot easily test, but I trust the code.

(I first thought that `texinfo` pulls in some `texlive` packages, but I was wrong, so it's not so important.)

@kencu or anyone else: does it make sense to use conditional build-time dependencies (for `texinfo`) or not?

There's another point though. If we add a new build-time dependency: do we need to revbump all the ports that use this portgroup? (I hope not, but I'm not entirely sure either.)

mojca (approved):

mojca:
Btw: commit message says `textinfo`, but the dependency is `texinfo`.

kencu:
> @kencu or anyone else: does it make sense to use conditional build-time dependencies (for `texinfo`) or not?

I suppose the convention is to use a bin: type dependency, if any version of the bin works.

Gcenx:
I wasn't aware `bin:` searched for the binary anywhere in `$PATH`, tested on 10.9 doesn't install `texinfo` on Ventura it does.

Gcenx:
> Btw: commit message says `textinfo`, but the dependency is `texinfo`. The commit message could also include a short sentence explaining that this was needed in order to support Ventura.

There also added a comment to the commit to _why_ texinfo Port is needed

_resources/port1.0/group/crossbinutils-1.0.tcl
```diff
@@ -139,6 +139,9 @@ proc crossbinutils.setup {target version} {
             ${worksrcpath}/libiberty/Makefile.in
     }
 
+    depends_build \
```
mojca, line 142: Do we need an `-append` (to avoid removing basic stuff such as compiler on an older OS from the list)?

mojca: (I'm asking, not yet suggesting a change. Most other portgroups are combined abd need to append, this one is a bit more special.)

Gcenx: > Do we need an `-append` (to avoid removing basic stuff such as compiler on an older OS from the list)?

This won't affect compilers being added if required.

> (I'm asking, not yet suggesting a change. Most other portgroups are combined abd need to append, this one is a bit more special.)

These other Portfiles|groups have multiple subports so `-appended` is required to not overrider the common base deps.

PR #16560, opened 2022-11-02
tinyxml2: blacklist compilers which fail to build consistently

barracuda156:
Fixes: https://trac.macports.org/ticket/66144

#### Description

Fix for 32-bit platforms.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 (i386, ppc, x86_64)
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/textproc/tinyxml2/Portfile
+++ b/textproc/tinyxml2/Portfile
@@ -2,6 +2,7 @@
 
 PortSystem          1.0
 PortGroup           cmake 1.1
+PortGroup           compiler_blacklist_versions 1.0
 PortGroup           github 1.0
 
 name                tinyxml2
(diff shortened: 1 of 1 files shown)


barracuda156:
@kencu Mergeable?

kencu:
Yes, but -- there are issues with forcing two different compilers for different archs on the same system. 

Let's say we went ahead with this. Someone trying to build this +universal on a certain system would build it once with gcc-4.2 and the second time with gcc7 (at present). And then if that worked, the muniversal PortGroup would try to merge these two builds.

So the way I do things, to keep things as sane, and consistent, and reproducible as they possibly can be, if gcc-4.2, for example, won't build the port on XYZ system in XYZ arch, I blacklist it, even if some of the systems can build some of the archs.

That keeps the build more consistent, and avoids further confusion later on trying to sort out why a port works OK when run as i386 but not when run as x86_64, or vice versa.

So, IMHO, I would keep it simple and just blacklist ```*gcc-4.* *gcc-3.*``` even though sometimes some gccs in that blacklist can build the port for some archs.

barracuda156:
> So, IMHO, I would keep it simple and just blacklist `*gcc-4.* *gcc-3.*` even though sometimes some gccs in that blacklist can build the port for some archs.

@kencu Okay, sure. Gimme a min.

PR #16562, opened 2022-11-02
llvm-10: fix build problem, Apple removed Python 2.x from the system

kistlin:
according to https://www.macrumors.com/2022/01/28/apple-removing-python-2-in-macos-12-3/ Python 2.x got removed from macOS 12.3 (Monterey) https://support.apple.com/en-us/HT201260

Make it work with Ventura, based off of https://trac.macports.org/wiki/PortfileRecipes#compare-osx-darwin-version it should be Darwin version 22.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.0
Xcode Version 14.1 (14B47b) / Command Line Tools 14.1.0.0.1.1666437224

###### Verification 
Have you

### DIFF
--- a/lang/llvm-10/Portfile
+++ b/lang/llvm-10/Portfile
@@ -287,7 +287,9 @@ if {[lsearch -exact $PortInfo(depends_build) port:cctools] != -1} {
 }
 
 # use the system python27 if present
-if {${os.platform} eq "darwin" && ${os.major} >= 11} {
+if {${os.platform} eq "darwin" && ${os.major} >= 22} {
+    set pythonfullpath   /usr/bin/python3
+} elseif {${os.platform} eq "darwin" && ${os.major} >= 11} {
     set pythonfullpath   /usr/bin/python2.7
 } else {
     set pythonfullpath   ${prefix}/bin/python2.7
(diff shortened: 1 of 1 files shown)


lang/llvm-10/Portfile
```diff
@@ -287,7 +287,9 @@ if {[lsearch -exact $PortInfo(depends_build) port:cctools] != -1} {
 }
 
 # use the system python27 if present
-if {${os.platform} eq "darwin" && ${os.major} >= 11} {
+if {${os.platform} eq "darwin" && ${os.major} >= 22} {
+    set pythonfullpath   /usr/bin/python3
+} elseif {${os.platform} eq "darwin" && ${os.major} >= 11} {
```
pmetzger, line 292: Would it be better to depend on MacPorts python3 and not worry about this?

pmetzger: Old OSes can happily run Python 3 so there's no need for them to have a build dependency on Python 2.

kencu:
if all goes well with this, we can do the same to llvm/clang-9.0

pmetzger:
@kencu I still say that just depending on MacPorts python3 would be best.

PR #16629, opened 2022-11-09
hdf5: Advertise correct stdlib with gcc flavors.

eborisch:
Not rev-bumping as no change to built-bot (clang) results, and local installs would fail previously on rev-upgrade.

Fixes https://trac.macports.org/ticket/66208

Maintainer update.

### DIFF
--- a/science/hdf5/Portfile
+++ b/science/hdf5/Portfile
@@ -115,6 +115,9 @@ variant cxx description {
 } {
     configure.args-delete       --disable-cxx
     configure.args-append       --enable-cxx
+    if {[gcc_variant_isset]} {
+        configure.cxx_stdlib        libstdc++
+    }
 }
 
 variant fortran description {


science/hdf5/Portfile
```diff
@@ -115,6 +115,9 @@ variant cxx description {
 } {
     configure.args-delete       --disable-cxx
     configure.args-append       --enable-cxx
+    if {[gcc_variant_isset]} {
+        configure.cxx_stdlib        libstdc++
```
jmroot, line 119: Should be `macports-libstdc++` if it's using the libstdc++ provided by the libgcc port.

jmroot (changes requested):

jmroot:
This actually seems like a bug in the compilers portgroup. If you are using gcc to build it should be setting the stdlib appropriately: https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/compilers-1.0.tcl#L338

cjones051073:
> This actually seems like a bug in the compilers portgroup. If you are using gcc to build it should be setting the stdlib appropriately: https://github.com/macports/macports-ports/blob/master/_resources/port1.0/group/compilers-1.0.tcl#L338

Yeah, it always confuses me a bit that setting `configure.cxx_stdlib` doesn't automatically make sure the right flags are then passed to `cxxflags` etc.

cjones051073:
This MR is no longer required.

https://github.com/macports/macports-ports/commit/6c7e20407ccc5f48c8a89b6cfef7561fea0002b0

Fixes the issue with the compiler PG not properly setting the stdlib compiler flags, and 

https://github.com/macports/macports-ports/commit/8e79a8d54070c069be638be5f399deea5ca80d07

rev bumps hdf5 to pick up those fixes.

With that, building `hdf5 +gcc12` correctly uses gcc12 *and* `stdlib=libc++`. e.g.

```
Oberon ~/Projects/MacPorts/ports > otool -L  /opt/local/lib/libhdf5_cpp.dylib

/opt/local/lib/libhdf5_cpp.dylib:
	/opt/local/lib/libhdf5_cpp.200.dylib (compatibility version 203.0.0, current version 203.0.0)
	/opt/local/lib/libhdf5.200.dylib (compatibility version 203.0.0, current version 203.0.0)
	/opt/local/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.13)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	/usr/lib/libc++abi.dylib (compatibility version 1.0.0, current version 1300.23.0)
	@rpath/libgcc_s.1.1.dylib (compatibility version 1.0.0, current version 1.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

jmroot:
> Yeah, it always confuses me a bit that setting `configure.cxx_stdlib` doesn't automatically make sure the right flags are then passed to `cxxflags` etc.

It does, but not for macports-gcc prior to ebd2261, which is not yet in any release.

cjones051073:
> > Yeah, it always confuses me a bit that setting `configure.cxx_stdlib` doesn't automatically make sure the right flags are then passed to `cxxflags` etc.
> 
> It does, but not for macports-gcc prior to [macports/macports-base@ebd2261](https://github.com/macports/macports-base/commit/ebd226139d5ba8389367606761658ad0e3bb5cce), which is not yet in any release.

Right, I forgot that was not yet in a release.

once it is, in principle the work around I added to the PG is not needed any mire and could be removed.

PR #16642, opened 2022-11-11
diffutils-for-muniversal: v3.8

kencu:
a sequestered diffutils, no macports deps
change muniversal to use it on Ventura+

### DIFF
--- a/sysutils/diffutils-for-muniversal/Portfile
+++ b/sysutils/diffutils-for-muniversal/Portfile
@@ -1,0 +1,84 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+
+name                diffutils-for-muniversal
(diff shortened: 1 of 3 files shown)


sysutils/diffutils-for-muniversal/Portfile
```diff
@@ -0,0 +1,85 @@
...
+        if {[string match "*gz*" ${f}]}  {
+            system -W ${workpath} "/usr/bin/tar xzf ${distpath}/${f}"
+        }
+        if {[string match "*xz*" ${f}]}  {
+            system -W ${workpath} "/usr/bin/tar xJf ${distpath}/${f}"
+        }
+    }
+}
+
+# sterilize MacPorts build environment; we want nothing picked up from MP prefix
```
catap, line 52: Maybe extract it to PG?

kencu: perhaps, at some future point. although there are so few ports that use this, maybe best to just keep it like this...

kencu:
Yoops -- this is being invoked on Monterey, because 21.6.0 is > 21 I suppose.

I will have to change the test to >= 22 to trigger only on Ventura.

PR #16651, opened 2022-11-12
luajit: fix PPC build

barracuda156:
Fixes: https://trac.macports.org/ticket/65996

#### Description

1. Sources explicitly require at least `gcc-4.3`, so `gcc-4.0` must be added to blacklist, otherwise it is pulled and build fails.
2. Build with GCC, at least on PPC (both Leopard and SL), fail with undefined unwinder symbols. I have tried a number of ways to fix that, what worked was adding `-lgcc_eh` ldflag. Closes my own ticket referred to above.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8
Xcode 3.1.4

macOS 10A190
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/lang/luajit/Portfile
+++ b/lang/luajit/Portfile
@@ -35,10 +35,16 @@ post-patch {
 
 use_configure       no
 
-compiler.blacklist  {clang < 700} *gcc-4.2 macports-clang-3.3 macports-clang-3.4
+compiler.blacklist  {clang < 700} *gcc-4.0 *gcc-4.2 macports-clang-3.3 macports-clang-3.4
 
 # changes to compiler flags must be made before `CFLAGS=...`
 xcode_workaround.type append_to_compiler_flags
+
+# https://trac.macports.org/ticket/65996
+if {[string match *gcc* ${configure.compiler}]} {
+    configure.ldflags-append \
+                    -lgcc_eh
+}
 
 build.target        amalg
 build.args-append   CC="${configure.cc}" \


barracuda156:
@kencu Could you please review this? :)

mojca:
Out of curiosity: does this work now?
* https://www.freelists.org/post/luajit/LuaJIT-on-PowerPC-Macs

I managed to compile LuaJIT on PPC many many years ago (I didn't try since), but the above example in fact didn't work and `luajittex` produced useless PDFs on PPC.

None of that should prevent you from merging the patch, I'm just curious.

barracuda156:
@mojca Unfortunately, as of now, it is indeed quite broken. I could not figure out how to fun commands via GDB, but I tried to run this test suite now: https://github.com/openresty/luajit2-test-suite
It performs miserably: 134 tests fail, 30 tests pass (unless I do something wrong, which is not impossible).

We are probably still better off having it build on PPC rather than not, but it looks like fixing it will require some effort.

barracuda156:
@mojca I actually wonder how can it be so bad. We normally get a much better results with software with zero PPC support in the sources. Maybe assembler is wrong? And it is huge in `luajit`, hard to verify.

barracuda156:
@kencu @mojca If we are okay with this for now, perhaps merging is due. Old Apple GCC must be blacklisted, and that is relevant for Intel no less than for PPC.

Fixing unwinder linking should be fine too, though it does not make `luajit` work correctly. The solution seems okay though, and used elsewhere. Random example from google: http://www.tau.uoregon.edu/tau-wiki/GNU_Linking_issues_on_XT3_%28May_2009%29

kencu:
Short story -- it is usually pretty simple to try a few things until you get a port to build.

Actually understanding exactly what is going on, and why a given fix is needed, leads you to very solid, unassailable, "one true fix" that is really the proper exact solution.

By and large, that kind of fix is the only kind of fix MacPorts wants in the ports tree.

barracuda156:
@kencu Ok, fair enough. Let me close it for now.

barracuda156:
> I have trouble committing something like this as it's an odd fix, and there are many unanswered questions that come to mind...
> 
> is this just because the wrong compiler is being used for linking? Should it be the gcc CXX compiler?

@kencu As a general question: what possible reasons have you encountered with these unwinder failures? I have seen it is not limited to PPC or old macOS.
Presence of active `libunwind-headers` aside.

PR #16711, opened 2022-11-17
vim, MacVim: Update to 9.0.0472

dylanbr:
#### Description

- Updated both Vim and MacVim to 9.0.0472.
- Removed `--with-developer-dir` from Vim configure because it was failing when `-t` (trace) was enabled. I am not sure if there is a better way to fix this?
- Changed all references of `snapshot` to `release` for MacVim, to match the new naming convention.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6.1 21G217 x86_64
Xcode 14.1 14B47b

###### Verification 
Have you

  - There is an [open PR](https://github.com/macports/macports-ports/pull/16632) to update Vim to a later patchlevel (852). This PR instead uses the same patchlevel as MacVim 174 (472), as suggested there.

### DIFF
--- a/editors/MacVim/Portfile
+++ b/editors/MacVim/Portfile
@@ -8,11 +8,11 @@ PortGroup           compiler_blacklist_versions 1.0
 # is because they both share the same set of configuration files and ensures
 # that the user's configuration files will always work with both.
 set vim_version     9.0
-set snapshot        173
-github.setup        macvim-dev macvim ${snapshot} snapshot-
-revision            1
+set release         174
+github.setup        macvim-dev macvim ${release} release-
+revision            0
 name                MacVim
-version             ${vim_version}.snapshot${snapshot}
+version             ${vim_version}.release${release}
 categories          editors
 platforms           darwin
 license             Vim GPL-2+
@@ -25,9 +25,9 @@ long_description \
 
 homepage            https://macvim-dev.github.io/macvim/
 
-checksums           rmd160  88e64d4c6c5352a998f92a759b2a64118f93f844 \
-                    sha256  edf8cb0cd4215281374af95585830721d7200020a081e213ecaf92c19710381a \
-                    size    24213601
+checksums           rmd160  5db536887934411f93238f1db1c4537b7ba17b1b \
+                    sha256  bd8cfba37a7dfc9eba30fdfb9fd3b2082ca3bbc6740eac9f61964bb802415ea0 \
+                    size    23694113
 
 depends_lib         port:ncurses \
                     port:gettext \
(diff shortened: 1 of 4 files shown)


raimue:
> - Updated both Vim and MacVim to 9.0.0472.

Thank you very much for the work on this patch! Especially for adapting the patches. :+1:

I still need to check whether the patches for +pythonXY and other variants still apply. Non-default variants are not checked by CI.

> - Removed `--with-developer-dir` from Vim configure because it was failing when `-t` (trace) was enabled. I am not sure if there is a better way to fix this?

Hm, this might indicate we are missing some paths in trace mode that should actually be allowed? But I am pretty sure `developer_dir` was always in the list of allowed paths.

The `DEVELOPER_DIR` was definitely necessary on old legacy installations with universal builds (ppc, i386). Maybe we would only need it on old versions by now...

> - Changed all references of `snapshot` to `release` for MacVim, to match the new naming convention.

Makes sense :+1:

dylanbr:
I've tested building both Vim and MacVim with:
- `sudo port -vst install +python310 +tcl`
- `sudo port -vst install +python27`

All of these variations build and run without issue.

I have also updated MacVim to use the same mechanism as Vim for applying the Python patches, using `set pythons_suffixes {...}`. This should simplify updates going forward.

dylanbr:
It looks like the problem with `developer_dir` is related to `xcrun`:

```
checking for linking with ncurses library... configure: error: FAILED
Command failed:  cd "/opt/local/var/macports/build/_removed_/vim/work/vim-9.0.0472" && ./configure --prefix=/opt/local --disable-gui --without-x --without-local-dir --disable-gpm --disable-canberra --mandir=/opt/local/share/man --with-tlib=ncurses --enable-multibyte --with-developer-dir=/Applications/Xcode.app/Contents/Developer --enable-fail-if-missing --with-compiledby="MacPorts" --with-features=huge
Exit code: 1
Warning: The following existing files were hidden from the build system by trace mode:
  /Applications/Xcode.app/Contents/Developer/usr/lib/libxcrun.dylib
  /opt/local/bin/strip
  /private/var/select/sh
```

And from `src/auto/config.log`:

```
configure:11873: checking --with-tlib argument
configure:11882: result: ncurses
configure:11885: checking for linking with ncurses library
configure:11898: /usr/bin/clang -o conftest -pipe -Os -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 -I/opt/local/include -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -DMACOS_X -DMACOS_X_DARWIN -L/opt/local/lib -Wl,-headerpad_max_install_names -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 conftest.c  -lncurses >&5
xcrun: error: invalid DEVELOPER_DIR path (/Applications/Xcode.app/Contents/Developer), missing xcrun at: /Applications/Xcode.app/Contents/Developer/usr/bin/xcrun
configure:11898: $? = 1
configure: failed program was:
```

Here are the same logs without `--with-developer-dir`:

```
configure: creating auto/config.status
config.status: creating auto/config.mk
config.status: creating auto/config.h
Warning: The following existing files were hidden from the build system by trace mode:
  /opt/local/bin/strip
  /private/var/select/sh
```

```
configure:11873: checking --with-tlib argument
configure:11882: result: ncurses
configure:11885: checking for linking with ncurses library
configure:11898: /usr/bin/clang -o conftest -pipe -Os -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 -I/opt/local/include -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -DMACOS_X -DMACOS_X_DARWIN -L/opt/local/lib -Wl,-headerpad_max_install_names -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 conftest.c  -lncurses >&5
configure:11898: $? = 0
configure:11899: result: OK
```

Any idea what is causing `libxcrun.dylib` to be hidden when `--with-developer-dir` is set, or how to resolve the issue?

raimue:
> It looks like the problem with `developer_dir` is related to `xcrun`:
> 
> ```
> checking for linking with ncurses library... configure: error: FAILED
> Command failed:  cd "/opt/local/var/macports/build/_removed_/vim/work/vim-9.0.0472" && ./configure --prefix=/opt/local --disable-gui --without-x --without-local-dir --disable-gpm --disable-canberra --mandir=/opt/local/share/man --with-tlib=ncurses --enable-multibyte --with-developer-dir=/Applications/Xcode.app/Contents/Developer --enable-fail-if-missing --with-compiledby="MacPorts" --with-features=huge
> Exit code: 1
> Warning: The following existing files were hidden from the build system by trace mode:
>   /Applications/Xcode.app/Contents/Developer/usr/lib/libxcrun.dylib
>   /opt/local/bin/strip
>   /private/var/select/sh
> ```

Thank you for these logs. The behavior seen in `config.log` is exactly as expected when `xcrun` cannot be executed. Apparently it could not be executed because trace mode denied access to the dynamic library `libxcrun.dylib`.

The implementation in porttrace.tcl is supposed to add `developer_dir` to the list of allowed directories. In this case, it should have allowed all of `/Application/Xcode.app` by stripping away the `Contents/Developer` suffix.

https://github.com/macports/macports-base/blob/887b2fb974c8363413cdbdee65a4962e0cf02dd6/src/port1.0/porttrace.tcl#L235-L241

@neverpanic Maybe you have an idea what could cause the problem with trace mode unexpectedly hiding `/Applications/Xcode.app/Contents/Developer/usr/lib/libxcrun.dylib`? Or any idea how to debug this further?

I am quite sure this is a issue with trace mode in base and not caused by the vim port. But I don't know if there is any workaround we could apply to make trace mode work in this cause.

editors/MacVim/Portfile
```diff
@@ -8,11 +8,11 @@ PortGroup           compiler_blacklist_versions 1.0
 # is because they both share the same set of configuration files and ensures
 # that the user's configuration files will always work with both.
 set vim_version     9.0
-set snapshot        173
-github.setup        macvim-dev macvim ${snapshot} snapshot-
+set release        174
```
raimue, line 11: Nit-pick: indentation not aligned properly.

raimue (changes requested):

raimue:
> I have also updated MacVim to use the same mechanism as Vim for applying the Python patches, using `set pythons_suffixes {...}`. This should simplify updates going forward.

Thank you! That detail is very much appreciated.

neverpanic:
> @neverpanic Maybe you have an idea what could cause the problem with trace mode unexpectedly hiding `/Applications/Xcode.app/Contents/Developer/usr/lib/libxcrun.dylib`? Or any idea how to debug this further?

Does the port set `use_xcode yes`? If it doesn't, trace mode will not allow `/Applications/Xcode.app`, but `/Library/Developer/CommandLineTools`. The debug log also prints the entire sandbox, so as a next step, I'd suggest checking that for `/Applications/Xcode.app`.

dylanbr:
I found that `use_xcode yes` was not present in the `vim` port, but according to the debug log, `/Applications/Xcode.app` was being used.

After replacing `--with-developer-dir=${developer_dir}` with `--with-developer-dir=${configure.developer_dir}`, the build is successful, whether trace mode enabled or not.

From what I can tell, `developer_dir` does not check for `use_xcode yes`, whereas `configure.developer_dir` does.

raimue:
@neverpanic Thanks, `use_xcode` was a good hint to understand what is going on.

So, `configure.developer_dir` will point to `/Library/Developer/CommandLineTools` with `use_xcode no` (the default). I think this is better than requiring a full Xcode installation for vim. This might still break universal builds on legacy macOS versions but I think we just have to wait if it really affects someone.

raimue (approved):
@dylanbr Thank you again for figuring this out as well as for the updated patches!

PR #16724, opened 2022-11-18
qt*-qttranslations: remove qttools dependency

catap:
#### Description

qttranslations is a port which contains only .qm files; this ports is runtime dependency for near of all Qt-based applications like wireshark. qttools depends on clang because qdoc uses it to parce C++ files. That makes wireshark to depends on clang via qttranslations.

Fixed this insanity.

###### Type(s)

- [x] bugfix

###### Tested on

macOS 12.6 21G115 x86_64
Xcode 14.1 14B47b

as removing `qt*-qttools` and running `LC_ALL=ru_RU open /Applications/MacPorts/Wireshark.app` and `LC_ALL=de_DE open /Applications/MacPorts/Wireshark.app` to confirm that different locales are working.

and as `port build qt5-qttranslations` and `port build qt6-qttranslations` to confirm that build dependency to `qttools` is required and working.

###### Verification 
Have you

### DIFF
--- a/aqua/qt5/Portfile
+++ b/aqua/qt5/Portfile
@@ -631,12 +631,12 @@ array set modules {
             1634516
         }
         ""
+        "port:qt5-qttools"
         ""
-        "qttools"
         {"translation files"}
         ""
         "variant overrides: ~examples ~tests ~debug noarch ~docs"
-        "revision 0"
+        "revision 1"
         "License: "
     }
     qtvirtualkeyboard {
--- a/aqua/qt511/Portfile
+++ b/aqua/qt511/Portfile
@@ -547,12 +547,12 @@ array set modules {
             1424160
         }
         ""
+        "port:qt511-qttools"
         ""
-        "qttools"
         {"translation files"}
         ""
         "variant overrides: ~examples ~tests ~debug noarch ~docs"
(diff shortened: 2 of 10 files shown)


MarcusCalhoun-Lopez:
Please forgive me for not commenting on this pull request sooner.
Doesn't the revision of each `qttranslations` need to be increased since a dependency goes from library to build?

catap:
@MarcusCalhoun-Lopez np, anyway you have raised a good question.

This commit helps me to avoid dependency without any revbump. So, I really not sure about that. Base on https://guide.macports.org/#reference.keywords I think that it should be merged without increase dependency because library dependency is useless for that case.

MarcusCalhoun-Lopez:
My understanding is that a revision increase is warranted because the port is changed on the user's machine.
E.g., `port uninstall qt5-qttools` would behave differently for users who installed qt5-qttranslations before and after the change.

catap:
Ok, let’s be on the safe side of the moon.

I’ll bump revisions in couple hours when I’ll be back to the computer

catap:
@MarcusCalhoun-Lopez here it is

MarcusCalhoun-Lopez:
Thank you.

PR #16749, opened 2022-11-21
i2pd: new port, C++ implementation of i2p

barracuda156:
#### Description

`i2pd`: https://i2pd.readthedocs.io/en/latest/

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/security/i2pd/Portfile
+++ b/security/i2pd/Portfile
@@ -1,0 +1,136 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               boost 1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               legacysupport 1.1
+PortGroup               openssl 1.0
+
+github.setup            PurpleI2P i2pd 2.44.0
+revision                0
+categories              security net
+maintainers             {@barracuda156 gmail.com:vital.had} {@catap korins.ky:kirill} openmaintainer
+license                 BSD
+description             End-to-End encrypted and anonymous Internet
+long_description        ${description}
+homepage                https://i2pd.website \
+                        https://i2pd.readthedocs.io
+checksums               rmd160  c7caf1c00ec99208af76f084fe9df523d368cebb \
+                        sha256  4feb50418dc484cf21612a18f7305000e19162af8f75d921e133e85c78afa854 \
+                        size    648654
+
+depends_lib-append      port:zlib
+
+patch.pre_args          -p1
+patchfiles-append       0001-i2pd.conf-adjust-for-MacPorts.patch
+
+set i2pddir             ${prefix}/var/db/${name}
+
+post-patch {
+    reinplace "s|@PREFIX@|${prefix}|g" \
+        ${worksrcpath}/contrib/i2pd.conf \
+        ${worksrcpath}/contrib/i2pd.logrotate
+    reinplace "s|@I2PDDIR@|${i2pddir}|g" ${worksrcpath}/contrib/i2pd.conf
+}
+
+set i2pduser            _i2pd
+add_users               ${i2pduser} group=${i2pduser} realname=I2PD\ user
+
+cmake.source_dir        ${worksrcpath}/build
+
+# Minimum required is C++11, C++17 used if supported.
+compiler.cxx_standard   2011
+
+# https://github.com/PurpleI2P/i2pd/issues/1798
+compiler.blacklist-append \
+                        {clang < 600}
+
+# These are defaults, but better have them explicit:
+configure.args-append   -DWITH_LIBRARY=ON \
+                        -DWITH_BINARY=ON \
+                        -DWITH_STATIC=OFF
+
+if {${build_arch} in [list ppc ppc64]} {
+    configure.args-append \
+                        -Dppc_support=TRUE \
+                        -DWITH_AESNI=OFF
+    # Temporary fix: use std::atomic in place of GCC builtins.
+    # See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107590
+    # https://github.com/PurpleI2P/i2pd/issues/1726
+    # https://github.com/boostorg/smart_ptr/issues/105
+    configure.cxxflags-append \
+                        -DBOOST_SP_USE_STD_ATOMIC
+}
+
+post-destroot {
+    xinstall -d ${destroot}${prefix}/etc/${name}
+    foreach item {i2pd.conf subscriptions.txt tunnels.conf tunnels.d} {
+        move ${worksrcpath}/contrib/${item} ${destroot}${prefix}/etc/${name}/${item}.default
+    }
+
+    xinstall -d ${destroot}${prefix}/share/${name}
+    foreach item {i2pd.logrotate certificates} {
+        move ${worksrcpath}/contrib/${item} ${destroot}${prefix}/share/${name}
+    }
+
+    xinstall -d ${destroot}${prefix}/share/doc/${name}
+    xinstall -W ${worksrcpath} -m 0644 ChangeLog LICENSE README.md ${destroot}${prefix}/share/doc/${name}
+
+    xinstall -d ${destroot}${prefix}/var/log/${name}
+    xinstall -d ${destroot}${prefix}/var/run/${name}
+
+    xinstall -d ${destroot}${i2pddir}
+    ln -sf ${prefix}/share/${name}/certificates ${destroot}${i2pddir}/certificates
+    ln -sf ${prefix}/etc/${name}/tunnels.d ${destroot}${i2pddir}/tunnels.d
+    ln -sf ${prefix}/etc/${name}/i2pd.conf ${destroot}${i2pddir}/i2pd.conf
+    ln -sf ${prefix}/etc/${name}/subscriptions.txt ${destroot}${i2pddir}/subscriptions.txt
+    ln -sf ${prefix}/etc/${name}/tunnels.conf ${destroot}${i2pddir}/tunnels.conf
+}
+
+notes "
+i2pd status can be checked at: http://127.0.0.1:7070
+You may edit i2pd settings in ${prefix}/etc/i2pd
+See also: https://i2pd.readthedocs.io/en/latest/user-guide/configuration
+Re using i2pd with web-browser: https://github.com/PurpleI2P/i2pdbrowser
+"
+
+post-activate {
+    if {![file exists ${prefix}/etc/logrotate.d/i2pd] && [variant_isset logrotate]} {
+        xinstall -d ${destroot}${prefix}/etc/logrotate.d
+        copy ${prefix}/share/${name}/i2pd.logrotate ${prefix}/etc/logrotate.d/i2pd
+    }
+
+    foreach item {i2pd.conf subscriptions.txt tunnels.conf tunnels.d} {
+        if {![file exists ${prefix}/etc/${name}/${item}]} {
+            copy ${prefix}/etc/${name}/${item}.default ${prefix}/etc/${name}/${item}
+        }
+    }
+
+    fs-traverse item "${i2pddir} ${prefix}/var/log/${name} ${prefix}/var/run/${name}" {
+        file attributes ${item} -owner ${i2pduser} -group ${i2pduser}
+    }
+}
+
+startupitem.create      yes
+startupitem.user        ${i2pduser}
+startupitem.group       ${i2pduser}
+startupitem.executable  ${prefix}/bin/${name} --datadir=${i2pddir}
+startupitem.pidfile     clean ${prefix}/var/run/${name}/${name}.pid
+startupitem.logfile     ${prefix}/var/log/${name}/${name}.log
+
+destroot.keepdirs       ${destroot}${i2pddir} \
+                        ${destroot}${prefix}/var/run/${name} \
+                        ${destroot}${prefix}/var/log/${name}
+
+variant upnp description "Support for UPNP" {
+    depends_lib-append  port:miniupnpc
+    configure.args-append \
+                        -DWITH_UPNP=ON
+}
+
+variant logrotate description "Logrotate configuration" {
+    depends_run-append  port:logrotate
+}
+
+default_variants        +upnp +logrotate
--- a/security/i2pd/files/0001-i2pd.conf-adjust-for-MacPorts.patch
+++ b/security/i2pd/files/0001-i2pd.conf-adjust-for-MacPorts.patch
@@ -1,0 +1,47 @@
+From c9e23017bcbc2a35712c3e6739ff27176d51c3c6 Mon Sep 17 00:00:00 2001
+From: barracuda156 <vital.had@gmail.com>
+Date: Fri, 11 Nov 2022 07:48:10 +0800
+Subject: [PATCH] i2pd.conf: adjust for MacPorts
+
+---
+ contrib/i2pd.conf | 8 ++++----
+ 1 file changed, 4 insertions(+), 4 deletions(-)
+
+diff --git a/contrib/i2pd.conf b/contrib/i2pd.conf
+index d0a9e141..06be503e 100644
+--- a/contrib/i2pd.conf
++++ b/contrib/i2pd.conf
+@@ -8,19 +8,19 @@
+ 
+ ## Tunnels config file
+ ## Default: ~/.i2pd/tunnels.conf or /var/lib/i2pd/tunnels.conf
+-# tunconf = /var/lib/i2pd/tunnels.conf
++tunconf = @I2PDDIR@/tunnels.conf
+ 
+ ## Tunnels config files path
+ ## Use that path to store separated tunnels in different config files.
+ ## Default: ~/.i2pd/tunnels.d or /var/lib/i2pd/tunnels.d
+-# tunnelsdir = /var/lib/i2pd/tunnels.d
++tunnelsdir = @I2PDDIR@/tunnels.d
+ 
+ ## Path to certificates used for verifying .su3, families
+ ## Default: ~/.i2pd/certificates or /var/lib/i2pd/certificates
+-# certsdir = /var/lib/i2pd/certificates
++certsdir = @I2PDDIR@/certificates
+ 
+ ## Where to write pidfile (default: i2pd.pid, not used in Windows)
+-# pidfile = /run/i2pd.pid
++pidfile = @PREFIX@/var/run/i2pd/i2pd.pid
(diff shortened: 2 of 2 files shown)


barracuda156:
@i0ntempest Could you please review this?

catap:
I'm fine to be a co-maintaner for this port.

i0ntempest:
Looks fine to me, but that said I'm not a C++ expert

security/i2pd/Portfile
```diff
@@ -0,0 +1,164 @@
...
+    if {[file exists ${worksrcpath}/contrib/certificates]} {
+        xinstall -d ${destroot}${prefix}/share/${name}
+        copy ${worksrcpath}/contrib/certificates ${destroot}${prefix}/share/${name}
+    }
+    xinstall -d ${destroot}${prefix}/share/doc/${name}
+    foreach f {ChangeLog LICENSE Makefile.osx README.md} {
+        if {[file exists ${worksrcpath}/${f}]} {
+            xinstall -m 0644 ${worksrcpath}/${f} ${destroot}${prefix}/share/doc/${name}
+        }
+    }
```
ryandesign, line 89: Why wouldn't these files exist? Same question for the preceding ` foreach` blocks.

Assuming they exist, you can install them all at once with less code:

```suggestion
    xinstall -W ${worksrcpath} -m 0644 ChangeLog LICENSE Makefile.osx README.md \
        ${destroot}${prefix}/share/doc/${name}
```

Do you really want to copy a Makefile into the documentation directory? That's unusual.

security/i2pd/Portfile
```diff
@@ -0,0 +1,164 @@
...
+    configure.ldflags-append \
+                        "-ldl -lminiupnpc"
+}
+
+default_variants        +upnp
+
+depends_test-append     port:gmake
+
+test.run                yes
+test.target             all
```
catap, line 161: are you sure that target `all` is for test? I've expected here `check` or something like that.

security/i2pd/files/0002-util.cpp-do-not-use-pthread_setname_np-on-unsupporte.patch
```diff
@@ -0,0 +1,44 @@
...
+@@ -143,8 +146,15 @@ namespace util
+ 	}
+ 
+ 	void SetThreadName (const char *name) {
+-#if defined(__APPLE__) && !defined(__POWERPC__)
++#if defined(__APPLE__)
++# if (!defined(MAC_OS_X_VERSION_10_6) || \
++		(MAC_OS_X_VERSION_MAX_ALLOWED < 1060) || \
++		defined(__POWERPC__))
++		/* pthread_setname_np is not there on <10.6 and all PPC.
```
catap, line 34: Why not use legacysupport instead?

barracuda156: @catap I think it was tried and didn’t work. Also, that has been merged to upstream already. It is a temporary patch.

security/i2pd/Portfile
barracuda156: Gonna check that. Could be the case that I overlooked it.

security/i2pd/Portfile
```diff
@@ -0,0 +1,164 @@
...
+destroot.keepdirs       ${destroot}${i2pddir} \
+                        ${destroot}${prefix}/var/run/${name} \
+                        ${destroot}${prefix}/var/log/${name}
+
+variant upnp description "Support for UPNP" {
+    depends_lib-append  port:miniupnpc
+    configure.args-append \
+                        -DWITH_UPNP=ON
+    configure.ldflags-append \
+                        "-ldl -lminiupnpc"
```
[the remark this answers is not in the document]
barracuda156: @ryandesign Thank you! I will make corrections today as suggested.

barracuda156:
> Looks fine to me, but that said I'm not a C++ expert

@i0ntempest Let me make a few updates as suggested by @ryandesign above, and then should be good to merge.
I can confirm it is running both directly and via `port load` and status is seen in web console.

security/i2pd/Portfile
barracuda156: @ryandesign `Makefile.osx` is not necessary, of course. I thought to keep it simply because `i2pd` supports alternative build system. Can be dropped.

`file exists` condition is a safe-guard against a case when sources are modified and build gonna fail due to some file not existing or renamed. I have had such issues with other ports.
But also can be dropped, if undesirable.

security/i2pd/Portfile
catap: I've missed that you changed `test.dir`.

barracuda156: > I've missed that you changed `test.dir`.

Let me check if that actually works as supposed though :)

security/i2pd/Portfile
```diff
@@ -0,0 +1,164 @@
...
+    ln -sf ${prefix}/share/${name}/certificates ${i2pddir}/certificates
+    ln -sf ${prefix}/etc/${name}/tunnels.d ${i2pddir}/tunnels.d
+    ln -sf ${prefix}/etc/${name}/i2pd.conf ${i2pddir}/i2pd.conf
+    ln -sf ${prefix}/etc/${name}/subscriptions.txt ${i2pddir}/subscriptions.txt
+    ln -sf ${prefix}/etc/${name}/tunnels.conf ${i2pddir}/tunnels.conf
+}
+
+startupitem.create      yes
+# Option --service has no effect on macOS: https://i2pd.readthedocs.io/en/latest/user-guide/configuration
+startupitem.executable  ${prefix}/bin/i2pd --datadir=${i2pddir} --conf=${i2pddir}/i2pd.conf --certsdir=${i2pddir}/certificates --tunconf=${i2pddir}/tunnels.conf --log=${prefix}/var/log/${name} --daemon
```
catap, line 142:
```suggestion
startupitem.executable  ${prefix}/bin/i2pd \
                        --datadir=${i2pddir} \
                        --conf=${i2pddir}/i2pd.conf \
                        --certsdir=${i2pddir}/certificates \
                        --tunconf=${i2pddir}/tunnels.conf \
                        --log=${prefix}/var/log/${name} \
                        --bob.enabled=true \
                        --i2cp.enabled=true \
                        --daemon
```

security/i2pd/Portfile
```diff
@@ -0,0 +1,164 @@
...
+            copy ${prefix}/etc/${name}/${f} ${destroot}${prefix}/etc/${name}/${f}.mp_backup
+            file attributes ${destroot}${prefix}/etc/${name}/${f}.mp_backup -owner ${i2puser} -group ${i2puser} -permissions 0664
+        }
+    }
+}
+
+post-install {
+    ui_msg "******************************************************"
+    ui_msg "*"
+    ui_msg "* i2pd status can be checked on: https://127.0.0.1:7070"
```
catap, line 103:
```suggestion
    ui_msg "* i2pd status can be checked on: http://127.0.0.1:7070"
```

security/i2pd/Portfile
catap: Future, `port unload` doesn't work. Seems that `--daemon` here is not required.

Also, I suggest to create a dedicated user because running it under `root` isn't good idea to be honest.

security/i2pd/files/0002-util.cpp-do-not-use-pthread_setname_np-on-unsupporte.patch
barracuda156: @catap By the way, we can drop PPC patches now, updating to 2.44.0: https://github.com/PurpleI2P/i2pd/blob/openssl/ChangeLog

security/i2pd/Portfile
barracuda156: @catap Gimme a sec, I will update this and a correction above.

Could you please take a look how to enable tests properly?

security/i2pd/Portfile
barracuda156: > Future, `port unload` doesn't work. Seems that `--daemon` here is not required.
> 
> Also, I suggest to create a dedicated user because running it under `root` isn't good idea to be honest.

User is created above, see the portfile.
What we need then is load it under that user, if it is not done automatically.

Stopping the daemon: `kill -INT $( cat /var/run/i2pd/i2pd.pid )` https://i2pd.readthedocs.io/en/latest/user-guide/run/. I thought MacPorts takes care of executable-type daemons automatically, and `startupitem.stop` is not needed.

barracuda156:
@catap I gonna update to 2.44.0 once the build finishes, drop PPC patches (merged to upstream) and make changes that you are Ryan suggested above.

What is needed still:
– fix unloading of the daemon;
– make tests actually being built and used, given we use CMake system.

security/i2pd/Portfile
```diff
@@ -0,0 +1,164 @@
...
+}
+
+default_variants        +upnp
+
+depends_test-append     port:gmake
+
+test.run                yes
+test.target             all
+test.dir                ${workpath}/test
+test.env-append         CXX="${configure.cxx} [get_canonical_archflags cxx]"
```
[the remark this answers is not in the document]
barracuda156: These has to be fixed properly, I will do it today. Other corrections you have suggested are done.

barracuda156:
@catap Hmm, now after we added `sudo -u ${i2puser}` to `startupitem.executable`, the process loads under `_i2p` user, however webconsole is not reachable :(
Removing `--daemon` from list of flags breaks the daemon: it starts and quits straightaway.

If we cannot fix it with `startupitem.executable`, another option is going back to old `script` way of managing a daemon.

security/i2pd/Portfile
barracuda156: Now (after the rebase) `port unload` works for me (and `--daemon` option _is_ needed), however webconsole access got broken.

security/i2pd/Portfile
```diff
@@ -0,0 +1,153 @@
...
+# Option --service has no effect on macOS: https://i2pd.readthedocs.io/en/latest/user-guide/configuration
+startupitem.executable  sudo -u ${i2puser} ${prefix}/bin/i2pd \
+                        --datadir=${i2pddir} \
+                        --conf=${i2pddir}/i2pd.conf \
+                        --certsdir=${i2pddir}/certificates \
+                        --tunconf=${i2pddir}/tunnels.conf \
+                        --log=${prefix}/var/log/${name} \
+                        --bob.enabled=true \
+                        --i2cp.enabled=true \
+                        --daemon
```
catap, line 139:
```suggestion
startupitem.create      yes
startupitem.user        ${i2puser}
startupitem.group       ${i2puser}
# Option --service has no effect on macOS: https://i2pd.readthedocs.io/en/latest/user-guide/configuration
startupitem.executable  ${prefix}/bin/i2pd \
                        --datadir=${i2pddir} \
                        --conf=${i2pddir}/i2pd.conf \
                        --certsdir=${i2pddir}/certificates \
                        --tunconf=${i2pddir}/tunnels.conf \
                        --log=${prefix}/var/log/${name} \
                        --bob.enabled=true \
                        --i2cp.enabled=true \
                        --daemon
startupitem.pidfile     auto ${prefix}/var/run/${name}/i2pd.pid
startupitem.logfile     ${destroot}${prefix}/var/log/${name}/daemon.log
startupitem.logevents   yes
```

I suggest to read "docs" for startupitem: https://github.com/macports/macports-base/blob/master/src/port1.0/portstartupitem.tcl :)

I not sure that I've pointed the right log file (it should be different one), and not sure about pid file and its behaviour, but you should get an idea :)

barracuda156: I actually had those earlier, but decided to minimize the code. `pid` and `log` are specified in `conf` file. But let me try to add these back.

catap: @barracuda156 `pid` and `log` are specified in `i2pd` config, here I've specified them for `startupitem`. Log should allow to debug it (right now all logs from startupitem is going to `/dev/null`) and pid should fix unload issue (I hope)

barracuda156:
@catap Rebased with latest changes. Loading and unloading looks fine. Processes do run under `_i2p` user.
Two questions remaining:

1. Webconsole is not accessible now, even though the daemon runs. Something got broken in the process of fixing (I could access it before) or it is a consequence of setting a dedicated user? _We probably want webconsole to be accessible._
2. If we want tests, I guess target has to be written from scratch?

security/i2pd/Portfile
```diff
@@ -0,0 +1,159 @@
...
+
+# These are defaults, but better have them explicit:
+configure.args-append   -DWITH_LIBRARY=ON \
+                        -DWITH_BINARY=ON \
+                        -DWITH_STATIC=OFF
+
+# See Makefile.osx
+configure.ldflags-append \
+                        -lz -lcrypto -lpthread \
+                        -dead_strip -dead_strip_dylibs -bind_at_load
```
catap, line 55: You don't need it since you'd migrated to cmake.

```suggestion
```

security/i2pd/Portfile
```diff
@@ -0,0 +1,159 @@
...
+destroot.keepdirs       ${destroot}${i2pddir} \
+                        ${destroot}${prefix}/var/run/${name} \
+                        ${destroot}${prefix}/var/log/${name}
+
+variant upnp description "Support for UPNP" {
+    depends_lib-append  port:miniupnpc
+    configure.args-append \
+                        -DWITH_UPNP=ON
+    configure.ldflags-append \
+                        -ldl -lminiupnpc
```
catap, line 156: You don't need also it since you'd migrated to cmake.

```suggestion
```

catap (changes requested):
BTW `sudo port load i2pd` doesn't work anymore :( I can't find the process.

catap:
@barracuda156 I've made a push that fixes this port: https://github.com/catap/macports-ports/tree/i2pd ; feel free to squash my commit into your.

I suggest to remove `db`, `run` and `log` before you try the updated version.

On my machine it creates a user and `port load i2pd` runs a services, and `port unload i2pd` kills it.

barracuda156:
> @barracuda156 I've made a push that fixes this port: https://github.com/catap/macports-ports/tree/i2pd ; feel free to squash my commit into your.
> 
> 
> 
> I suggest to remove `db`, `run` and `log` before you try the updated version.
> 
> 
> 
> On my machine it creates a user and `port load i2pd` runs a services, and `port unload i2pd` kills it.

@catap I suspect it did not work for you because conf was backed up and re-used. On my machine load and unload were working.
But I will try the updated version.

Upstream confirmed there is no CMake support for tests. We have to make it. Anyway, I will try today, if it works, gonna add here and submit PR to upstream along.

barracuda156:
@catap Provisionally I have fixed tests. Right now all pass but one (`test-blinding` fails with SIGTRAP).

security/i2pd/files/0001-i2pd.conf-adjust-for-MacPorts.patch
```diff
@@ -0,0 +1,52 @@
...
+ ##  * file - log entries to a file
+ ##  * syslog - use syslog, see man 3 syslog
+-# log = file
++log = file
+ ## Path to logfile (default - autodetect)
+-# logfile = /var/log/i2pd/i2pd.log
++logfile = @PREFIX@/var/log/i2pd/i2pd.log
+ ## Log messages above this level (debug, info, *warn, error, none)
+ ## If you set it to none, logging will be disabled
+ # loglevel = warn
```
catap, line 49: Let remove remove this chunk?

Right now startupitem has log, and redirects all output to the log file. Why not use it instead?

security/i2pd/Portfile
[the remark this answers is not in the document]
barracuda156: Irrelevant now, this did not work, I have re-written test implementation from scratch.

barracuda156:
@catap Could you try tests on your end?

catap:
@barracuda156 tested, it works. I've pushed small polish around startupscript into https://github.com/catap/macports-ports/tree/i2pd — feel free to include it.

The only things that seems wired for me is specific variant for testing.

Maybe keep tests disabled until upstream doesn't merge your PR with cmake-based tests? It makes port file cleaner, and keep used patch much simpler :)

barracuda156:
> @barracuda156 tested, it works. I've pushed small polish around startupscript into https://github.com/catap/macports-ports/tree/i2pd — feel free to include it.

@catap Great!

> The only things that seems wired for me is specific variant for testing.
> Maybe keep tests disabled until upstream doesn't merge your PR with cmake-based tests? It makes port file cleaner, and keep used patch much simpler :)

I do not really insist on adding tests right now, but regardless of upstream merging it or not, we would not want to build tests unless a user ask for that, would we?

catap:
Anyway, it produces a lot of logs and I feel that installing logrotate is mandatory :(

barracuda156:
> Anyway, it produces a lot of warn-level logs and I feel that installing logrotate is mandatory :( Like this one:

@catap If you could commit a needed change to your tree, I gonna pull that over and rebase.

catap:
> @catap If you could commit a needed change to your tree, I gonna pull that over and rebase.

It's done!

security/i2pd/Portfile
```diff
@@ -0,0 +1,166 @@
...
+                copy ${prefix}/etc/${name}/${f}.mp_backup ${prefix}/etc/${name}/${f}
+            } else {
+                copy ${prefix}/etc/${name}/${f}.new ${prefix}/etc/${name}/${f}
+            }
+            file attributes ${prefix}/etc/${name}/${f} -owner ${i2pduser} -group ${i2pduser} -permissions 0644
+        }
+    }
+
+    if {![file exists ${i2pddir}]} {
+        xinstall -m 0644 -d ${i2pddir}
```
catap, line 132:
```suggestion
        xinstall -m 0755 -d ${i2pddir}
```

security/i2pd/Portfile
```diff
@@ -0,0 +1,166 @@
...
+    }
```
catap, line 133:
```suggestion
    # restore conf files
    foreach f {i2pd.conf subscriptions.txt tunnels.conf tunnels.d} {
        set item ${prefix}/etc/${name}/${f}
        if {![file exists ${item}]} {
            if {[file exists ${item}.mp_backup]} {
                copy ${item}.mp_backup ${item}
            } else {
                copy ${item}.new ${item}
            }
            if {[file isdirectory ${item}]} {
                file attributes ${item} -owner ${i2pduser} -group ${i2pduser} -permissions 0755
            } elseif {[file isfile ${item}]} {
                file attributes ${item} -owner ${i2pduser} -group ${i2pduser} -permissions 0644
            }
        }
    }

    xinstall -d ${i2pddir}

```

barracuda156: Changed this!

barracuda156:
@catap @i0ntempest @ryandesign I guess we are ready with this port.

PR #16758, opened 2022-11-21
soplex: requires c++14

catap:
#### Description

Fix a build https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/201871

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/math/soplex/Portfile
+++ b/math/soplex/Portfile
@@ -24,6 +24,8 @@ checksums           rmd160  30745ef666d7493864fb8e54815c829b9e8bcb4d \
                     sha256  47150ddd6eab9e51003fd2c303ebe8ec225e8ee9742f48548b9c74617d449962 \
                     size    984865
 
+compiler.cxx_standard 2014
+
 depends_lib-append  port:gmp \
                     port:zlib
 


math/soplex/Portfile
```diff
@@ -24,6 +24,8 @@ checksums           rmd160  30745ef666d7493864fb8e54815c829b9e8bcb4d \
                     sha256  47150ddd6eab9e51003fd2c303ebe8ec225e8ee9742f48548b9c74617d449962 \
                     size    984865
 
+compiler.cxx_standard 2011
```
ryandesign, line 27:
```suggestion
compiler.cxx_standard 2014
```

ryandesign (commented):
> Fix a build https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/201871

That log shows the flag `-std=gnu++14` is being used, so should we specify that the port needs the 2014 standard instead?

math/soplex/Portfile
catap:
```
/opt/local/libexec/boost/1.76/include/boost/multiprecision/detail/number_base.hpp:36:2: error: "This library now requires a C++11 or later compiler - this message was generated as a result of BOOST_NO_CXX11_HDR_TYPE_TRAITS being set"
```

That lead to thinking that c++2011 is required. But you're right => c++-2014 is required.

PR #16773, opened 2022-11-22
PULL_REQUEST_TEMPLATE.md: add line about variants

cooljeanius:
#### Description

One of the things I like about MacPorts is how many ports offer variants to enable extra features. I often try to install my ports with maximum possible variants enabled when possible. Unfortunately, it seems that some of the non-default variants don't get tested as much, which often leads to breakage when I try to update. This PR simply adds an extra checkbox to the PR template reminding PR authors to check their variants. The header for that section includes a `<!-- (delete not applicable items) -->` comment, so if it isn't relevant to certain ports, it can always be skipped. This is just to give a bit of an extra reminder for the cases where variants *are* relevant.

###### Type(s)
- [x] enhancement

###### Tested on
macOS 11.7.1 20G918 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

- [x] checked that any [variants](https://trac.macports.org/wiki/Variants) the Portfile may have haven't been broken?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -27,5 +27,6 @@ Have you
 - [ ] tried existing tests with `sudo port test`?
 - [ ] tried a full install with `sudo port -vst install`?
 - [ ] tested basic functionality of all binary files?
+- [ ] checked that any [variants](https://trac.macports.org/wiki/Variants) the Portfile may have haven't been broken?
 
 <!-- Use "skip notification" (surrounded with []) to avoid notifying maintainers -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -27,6 +27,6 @@ Have you
 - [ ] tried existing tests with `sudo port test`?
 - [ ] tried a full install with `sudo port -vst install`?
 - [ ] tested basic functionality of all binary files?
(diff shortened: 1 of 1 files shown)


catap:
Not sure that it is good idea.

For example `mpi` PG introduces a low of heavy variants: https://ports.macports.org/port/mumps/details/ which conflicts and simple enable all of them is not possible.

barracuda156:
There are, I believe, arch-specific variants, which cannot be built on some other archs.

cooljeanius:
I mean yeah, I get that testing every single combination of variants is often impossible, but I still think it ought to be possible for maintainers to at least test more variants than they're currently testing.

neverpanic (approved):
I like the idea. This gets a +1 from me.

catap:
To be clear: I also like this idea, but I afraid that it can't be required for a lot of ports. But, still, as reminder to test it... it is very nice idea.

cooljeanius:
> To be clear: I also like this idea, but I afraid that it can't be required for a lot of ports. But, still, as reminder to test it... it is very nice idea.

Is the change in wording in [c0d7f5a](https://github.com/macports/macports-ports/pull/16773/commits/c0d7f5aa3e52a8da4bb65547b0015be9e9f57890) an improvement? I tried to make it sound like less of a requirement to do all of them, and instead more of a reminder just to check the most important ones.

catap:
@cooljeanius cleaner, thanks! I like that!

cooljeanius:
> @cooljeanius cleaner, thanks! I like that!

OK, someone else will need to merge it, then, as I don't have the permissions necessary for that...

PR #16808, opened 2022-11-25
libptytty: set cxx_standard 11, fixes build on old systems

barracuda156:
#### Description

The port fails to build with `gcc-4.2`, and sure enough, source code explicitly requires C++ 11.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/devel/libptytty/Portfile
+++ b/devel/libptytty/Portfile
@@ -17,3 +17,5 @@ master_sites        http://dist.schmorp.de/libptytty/ \
 checksums           rmd160  402bd9793974ccba5c2a1319bc533c5ee275ccf4 \
                     sha256  8033ed3aadf28759660d4f11f2d7b030acf2a6890cb0f7926fb0cfa6739d31f7 \
                     size    45673
+
+compiler.cxx_standard 2011


devel/libptytty/Portfile
```diff
@@ -17,3 +17,5 @@ master_sites        http://dist.schmorp.de/libptytty/ \
 checksums           rmd160  402bd9793974ccba5c2a1319bc533c5ee275ccf4 \
                     sha256  8033ed3aadf28759660d4f11f2d7b030acf2a6890cb0f7926fb0cfa6739d31f7 \
                     size    45673
+
+compiler.cxx_standard 2011
```
exg, line 21: Does this alone work? The cmake-1.1 port group propagates the preference to cmake only if you also set
```
cmake.set_cxx_standard yes
```

barracuda156: @exg Maybe that is the default? But yes, I have only added `compiler.cxx_standard 2011`, it solved the issue.

kencu: This ```compiler.cxx_standard 2011``` makes sure that a compiler supporting c++11 is chosen by MacPorts, correct.

What c++ standard the build will actually use depends on the build settings and the compiler defaults. 

A build should set the c++ standard it needs directly, and you will see it specified on every build line.

Compilers have a 'built-in' fallback setting that they use if none is set, however. This changes as time goes by, and is less reliable. Most, but not all, of the compilers macports will choose if ```compiler.cxx_standard 2011``` is set will default to at least c++11, so in such situations, the build might be noted to work on some systems but not on others. This causes confusion.

barracuda156: @kencu In this case, are we good with compiler.cxx_standard, or rather add something? (I think the source code sets the standard to 2011, so perhaps what we have is sufficient.)

kencu: I have never looked at this build, just explaining what macports does, and why it doesn't always work out the way you think it should...

barracuda156: @exg I think setting the standard is sufficient, but if you would like to add that extra line, I will add it.

exg (approved):

PR #16843, opened 2022-11-29
Move Leopard to libgcc12, fix GCC targets for PPC

barracuda156:
#### Description

This PR aims to finally move 10.5 to using the latest `libgcc` and `gcc`, like it is the case with 10.6. Aside of changing relevant settings, it uses `gcc10-bootstrap` to build `gcc8` and up on PPC.
It also fixes wrong target names for PPC: it should be `powerpc*-` and not `ppc*-`.

At the moment is **does not address** building GCC as `+universal`. It is certainly desirable to have universal GCC on Leopard, but here I need help of @catap and @kencu to get it done. This is a bigger problem, perhaps, since currently “universal” GCCs in Macports are fake universal, AFAIU.

A question: do we need to use `gcc10-bootstrap` on 10.5 Intel? I have no idea and cannot test it. @kencu I think you know this, should I use `if {${os.major} < 10 || ({os.major} == 10 && ${build_arch} eq "ppc")}` instead of PPC-only condition?

@catap I followed your clang-11-bootstrap here in setting `depends_build` instead of `depends_lib` for `gcc10-bootstrap`. I have used `depends_lib` earlier and that worked fine, but I guess we only need build dependency?

Relevant Trac ticket: https://trac.macports.org/ticket/64694
Not closing it, since Tiger remains on `libgcc7`.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.5.8
Xcode 3.1.4

macOS 10A190
Xcode 3.2

macOS 10.6.8
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/lang/gcc9/Portfile
+++ b/lang/gcc9/Portfile
@@ -12,7 +12,7 @@ name                gcc9
 version             9.5.0
 revision            0
 subport             libgcc9 { revision 2 }
-platforms           {darwin >= 10 < 15}
+platforms           {darwin >= 9 < 15}
 categories          lang
 maintainers         nomaintainer
 # an exception in the license allows dependents to not be GPL
--- a/lang/gcc9/Portfile
+++ b/lang/gcc9/Portfile
@@ -81,6 +81,8 @@ depends_run         port:gcc_select \
 
 platform darwin {
     if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        configure.compiler.add_deps \
+                        no
         depends_build-append \
                         port:gcc10-bootstrap
         configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
--- a/lang/gcc9/Portfile
+++ b/lang/gcc9/Portfile
@@ -79,6 +79,15 @@ depends_lib         port:cctools \
 depends_run         port:gcc_select \
                     port:libgcc9
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
+    }
+}
+
 depends_skip_archcheck-append gcc_select ld64 cctools
 license_noconflict  gmp mpfr ppl libmpc zlib
 
--- a/lang/gcc9/Portfile
+++ b/lang/gcc9/Portfile
@@ -85,7 +85,13 @@ license_noconflict  gmp mpfr ppl libmpc zlib
 set major           [lindex [split ${version} .-] 0]
 
 platform darwin {
-    configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    if {${configure.build_arch} eq "ppc"} {
+        configure.pre_args-append --build=powerpc-apple-darwin${os.major}
+    } elseif {${configure.build_arch} eq "ppc64"} {
+        configure.pre_args-append --build=powerpc64-apple-darwin${os.major}
+    } else {
+        configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    }
 }
 
 set gcc_configure_langs {c c++ objc obj-c++ lto fortran}
--- a/_resources/port1.0/compilers/gcc_compilers.tcl
+++ b/_resources/port1.0/compilers/gcc_compilers.tcl
@@ -6,19 +6,19 @@
 global os.major os.arch os.platform
 
 #3 GCC 10 and above on OSX10.6+
-if {${os.major} >= 10 || ${os.platform} ne "darwin"} {
+if {${os.major} >= 9 || ${os.platform} ne "darwin"} {
     lappend compilers macports-gcc-13 macports-gcc-12 macports-gcc-11 macports-gcc-10
 }
 
 # GCC 9 and older only on OSX10.10 and older
 # https://trac.macports.org/ticket/65472
 if { ${os.major} < 15 } {
-    if { ${os.major} >= 10 } {
+    if { ${os.major} >= 9 } {
         lappend compilers macports-gcc-9 macports-gcc-8
     }
     lappend compilers macports-gcc-7 macports-gcc-6 macports-gcc-5
 }
 
-if { ${os.major} >= 10 } {
+if { ${os.major} >= 9 } {
     lappend compilers macports-gcc-devel
 }
--- a/_resources/port1.0/compilers/gcc_dependencies.tcl
+++ b/_resources/port1.0/compilers/gcc_dependencies.tcl
@@ -4,7 +4,7 @@ global os.major os.platform
 
 # GCC version providing the primary runtime
 # Note settings here *must* match those in the lang/libgcc port and compilers PG
-if {${os.platform} eq "darwin" && ${os.major} < 10} {
+if {${os.platform} eq "darwin" && ${os.major} < 9} {
     set gcc_main_version 7
 } else {
     set gcc_main_version 13
--- a/_resources/port1.0/group/compilers-1.0.tcl
+++ b/_resources/port1.0/group/compilers-1.0.tcl
@@ -77,7 +77,7 @@ options compilers.allow_arguments_mismatch
 default compilers.allow_arguments_mismatch no
 
 # Set a default gcc version
-if {${os.major} < 10 && ${os.platform} eq "darwin" } {
+if {${os.major} < 9 && ${os.platform} eq "darwin" } {
     # see https://trac.macports.org/ticket/57135
     set compilers.gcc_default gcc7
 } else {
@@ -95,13 +95,13 @@ if { ${os.arch} eq "arm" || ${os.platform} ne "darwin" } {
     if { ${os.major} < 15 } {
         lappend gcc_versions 5 6 7 8 9
     }
-    if { ${os.major} >= 10 } {
+    if { ${os.major} >= 9 } {
         lappend gcc_versions 10 11 12 13 devel
     }
 }
 # GCC version providing the primary runtime
 # Note settings here *must* match those in the lang/libgcc port.
-if { ${os.major} < 10 && ${os.platform} eq "darwin" } {
+if { ${os.major} < 9 && ${os.platform} eq "darwin" } {
     set gcc_main_version 7
 } else {
     set gcc_main_version 13
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -73,6 +73,8 @@ depends_run-append  port:gcc_select \
 
 platform darwin {
     if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        configure.compiler.add_deps \
+                        no
         depends_build-append \
                         port:gcc10-bootstrap
         configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -11,7 +11,7 @@ name                gcc-devel
 
 homepage            https://gcc.gnu.org/
 
-platforms           {darwin >= 10}
+platforms           {darwin >= 9}
 categories          lang
 maintainers         nomaintainer
 # an exception in the license allows dependents to not be GPL
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -71,6 +71,15 @@ depends_lib-append  port:cctools \
 depends_run-append  port:gcc_select \
                     port:libgcc-devel
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
+    }
+}
+
 depends_skip_archcheck-append gcc_select ld64 cctools
 license_noconflict  gmp mpfr ppl libmpc zlib
 
--- a/lang/gcc-devel/Portfile
+++ b/lang/gcc-devel/Portfile
@@ -80,7 +80,13 @@ license_noconflict  gmp mpfr ppl libmpc zlib
 set major           devel
 
 platform darwin {
-    configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    if {${configure.build_arch} eq "ppc"} {
+        configure.pre_args-append --build=powerpc-apple-darwin${os.major}
+    } elseif {${configure.build_arch} eq "ppc64"} {
+        configure.pre_args-append --build=powerpc64-apple-darwin${os.major}
+    } else {
+        configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    }
 }
 
 set gcc_configure_langs {c c++ objc obj-c++ lto fortran}
--- a/lang/gcc10/Portfile
+++ b/lang/gcc10/Portfile
@@ -89,6 +89,8 @@ depends_run-append  port:gcc_select \
 
 platform darwin {
     if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        configure.compiler.add_deps \
+                        no
         depends_build-append \
                         port:gcc10-bootstrap
         configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
--- a/lang/gcc10/Portfile
+++ b/lang/gcc10/Portfile
@@ -12,7 +12,7 @@ name                gcc10
 version             10.4.0
 revision            4
 
-platforms           {darwin >= 10 < 22}
+platforms           {darwin >= 9 < 22}
 categories          lang
 maintainers         nomaintainer
 # an exception in the license allows dependents to not be GPL
--- a/lang/gcc10/Portfile
+++ b/lang/gcc10/Portfile
@@ -93,7 +93,13 @@ license_noconflict  gmp mpfr ppl libmpc zlib
 set major           [lindex [split ${version} .-] 0]
 
 platform darwin {
-    configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    if {${configure.build_arch} eq "ppc"} {
+        configure.pre_args-append --build=powerpc-apple-darwin${os.major}
+    } elseif {${configure.build_arch} eq "ppc64"} {
+        configure.pre_args-append --build=powerpc64-apple-darwin${os.major}
+    } else {
+        configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    }
 }
 
 set gcc_configure_langs {c c++ objc obj-c++ lto fortran}
--- a/lang/gcc10/Portfile
+++ b/lang/gcc10/Portfile
@@ -87,6 +87,15 @@ depends_lib-append  port:cctools \
 depends_run-append  port:gcc_select \
                     port:libgcc10
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
+    }
+}
+
 depends_skip_archcheck-append gcc_select ld64 cctools
 license_noconflict  gmp mpfr ppl libmpc zlib
 
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -13,7 +13,7 @@ name                gcc11
 version             11.4.0
 revision            0
 
-platforms           {darwin >= 10 < 22}
+platforms           {darwin >= 9 < 22}
 categories          lang
 maintainers         nomaintainer
 # an exception in the license allows dependents to not be GPL
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -100,7 +100,13 @@ license_noconflict  gmp mpfr ppl libmpc zlib
 set major           [lindex [split ${version} .-] 0]
 
 platform darwin {
-    configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    if {${configure.build_arch} eq "ppc"} {
+        configure.pre_args-append --build=powerpc-apple-darwin${os.major}
+    } elseif {${configure.build_arch} eq "ppc64"} {
+        configure.pre_args-append --build=powerpc64-apple-darwin${os.major}
+    } else {
+        configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    }
 }
 
 set gcc_configure_langs {c c++ objc obj-c++ lto fortran}
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -94,6 +94,15 @@ depends_lib-append  port:cctools \
 depends_run-append  port:gcc_select \
                     port:libgcc11
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
+    }
+}
+
 depends_skip_archcheck-append gcc_select ld64 cctools
 license_noconflict  gmp mpfr ppl libmpc zlib
 
--- a/lang/gcc11/Portfile
+++ b/lang/gcc11/Portfile
@@ -96,6 +96,8 @@ depends_run-append  port:gcc_select \
 
 platform darwin {
     if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        configure.compiler.add_deps \
+                        no
         depends_build-append \
                         port:gcc10-bootstrap
         configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
--- a/lang/gcc12/Portfile
+++ b/lang/gcc12/Portfile
@@ -66,6 +66,8 @@ depends_run-append  port:gcc_select \
 
 platform darwin {
     if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        configure.compiler.add_deps \
+                        no
         depends_build-append \
                         port:gcc10-bootstrap
         configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
--- a/lang/gcc12/Portfile
+++ b/lang/gcc12/Portfile
@@ -11,7 +11,7 @@ name                gcc12
 
 homepage            https://gcc.gnu.org/
 
-platforms           {darwin >= 10}
+platforms           {darwin >= 9}
 categories          lang
 maintainers         nomaintainer
 # an exception in the license allows dependents to not be GPL
--- a/lang/gcc12/Portfile
+++ b/lang/gcc12/Portfile
@@ -70,7 +70,13 @@ license_noconflict  gmp mpfr ppl libmpc zlib
 set major           [lindex [split ${version} .-] 0]
 
 platform darwin {
-    configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    if {${configure.build_arch} eq "ppc"} {
+        configure.pre_args-append --build=powerpc-apple-darwin${os.major}
+    } elseif {${configure.build_arch} eq "ppc64"} {
+        configure.pre_args-append --build=powerpc64-apple-darwin${os.major}
+    } else {
+        configure.pre_args-append --build=${build_arch}-apple-darwin${os.major}
+    }
 }
 
 set gcc_configure_langs {c c++ objc obj-c++ lto fortran}
--- a/lang/gcc12/Portfile
+++ b/lang/gcc12/Portfile
@@ -64,6 +64,15 @@ depends_lib-append  port:cctools \
 depends_run-append  port:gcc_select \
                     port:libgcc12
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
+    }
+}
+
 depends_skip_archcheck-append gcc_select ld64 cctools
 license_noconflict  gmp mpfr ppl libmpc zlib
 
--- a/lang/gcc13/Portfile
+++ b/lang/gcc13/Portfile
@@ -66,6 +66,8 @@ depends_run-append  port:gcc_select \
 
 platform darwin {
     if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        configure.compiler.add_deps \
+                        no
         depends_build-append \
                         port:gcc10-bootstrap
         configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
--- a/lang/gcc13/Portfile
+++ b/lang/gcc13/Portfile
@@ -64,6 +64,15 @@ depends_lib-append  port:cctools \
 depends_run-append  port:gcc_select \
                     path:share/doc/libgcc/README:libgcc
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
+    }
+}
+
 depends_skip_archcheck-append gcc_select ld64 cctools
 license_noconflict  gmp mpfr ppl libmpc zlib
 
--- a/lang/gcc13/Portfile
+++ b/lang/gcc13/Portfile
@@ -11,7 +11,7 @@ name                gcc13
 
 homepage            https://gcc.gnu.org/
 
-platforms           {darwin >= 10}
+platforms           {darwin >= 9}
 categories          lang
 maintainers         nomaintainer
 # an exception in the license allows dependents to not be GPL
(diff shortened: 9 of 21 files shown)


barracuda156:
@kencu @catap Please take a look, let us finally sort this out.

`gcc11` and `gcc12` work nicely across all my systems (10.5.8, 10.6 PPC, 10.6.8 Rosetta).

kencu:
see https://trac.macports.org/ticket/66307#comment:6 for some comments and other steps to handle.

IMHO whatever is done for ppc also has to be done for Intel, as we can't have 10.5 PPC default to libgcc12 and 10.5 Intel default to libgcc7, for example. (Well, of course someone could try to do that, but it would be a mess to support.).

I have not seen gcc8/libgcc8 through gcc11/libgcc11 building on MacPorts older systems as yet, so those all have to be demonstrated to build.

kencu:
JFTR -- it is time to get this done, for sure. Just have to do it all, and do it right.

barracuda156:
> see https://trac.macports.org/ticket/66307#comment:6 for some comments and other steps to handle.

@kencu Thank you, I will look into that today.

> IMHO whatever is done for ppc also has to be done for Intel, as we can't have 10.5 PPC default to libgcc12 and 10.5 Intel default to libgcc7, for example. (Well, of course someone could try to do that, but it would be a mess to support.).

Sure. My only question is whether we should use `gcc10-bootstrap` on 10.5 Intel or Clang. Could you or @catap clarify this? Of course, using `gcc10-bootstrap` will be easier, just modify the condition in a way I mentioned above.

> I have not seen gcc8/libgcc8 through gcc11/libgcc11 building on MacPorts older systems as yet, so those all have to be demonstrated to build.

I have `libgcc8` through `libgcc12` built on PPC (as we know, `libgcc8` and `libgcc9` as useless and install nothing, but they do build).
I have `gcc10` through `gcc12` built and working perfectly fine on PPC.

I did not try building and using `gcc8` and `gcc9` and see no point in that, TBH.

barracuda156:
> see https://trac.macports.org/ticket/66307#comment:6 for some comments and other steps to handle.

Some additional comments re steps:

1. I need someone to handle 10.5 Intel case. This is not something I am able to do. @catap Could you help us with 10.5 Intel here?
2. I suggest we move Leopard to `libgcc12` but not Tiger. No testing has been done on Tiger at all. While I can try building a new GCC on Tiger, given that `gcc10-bootstrap` was fixed for Tiger some days ago, that is not enough, and we cannot be sure it works on Tiger smoothly. Leopard is the latest official OS for PPC. Leopard is very important. Tiger is a super-niche OS. (I do not object against Tiger in principle, but there is no one to test it, and it is by far less important that Leopard.)

> I have not seen gcc8/libgcc8 through gcc11/libgcc11 building on MacPorts older systems as yet, so those all have to be demonstrated to build.

While all versions of `libgcc` build fine, there are three totally parasitic which we perhaps want to avoid building – not for PPC, but in general.

Here is what gets installed on 10.6:

`libgcc6`:
```
libgfortran.3.dylib
```

`libgcc7`:
```
libgfortran.4.dylib
```

`libgcc8`: NOTHING

`libgcc9`:
```
libgcc_ext.10.4.dylib
libgcc_ext.10.5.dylib
```

`libgcc10`: NOTHING

`libgcc11`: NOTHING

`libgcc12`: full runtime, of which into `libgcc`:
```
libatomic.1.dylib
libatomic.dylib
libgcc_ehs.1.1.dylib
libgcc_ehs.dylib
libgcc_s.1.1.dylib
libgcc_s.1.dylib
libgcc_s.dylib
libgfortran.5.dylib
libgfortran.dylib
libgomp.1.dylib
libgomp.dylib
libitm.1.dylib
libitm.dylib
libobjc-gnu.4.dylib
libobjc-gnu.dylib
libssp.0.dylib
libssp.dylib
libstdc++.6.dylib
libstdc++.dylib
```

That is, we got three parasitic dependencies which install strictly nothing but are required to be built. On the fastest G5 – I have G5 Quad with 16 GB RAM and SSD – it still takes about 3 hours per arch, so if we consider Leopard and universal builds, that translated into 18 hours of useless compilation.

kencu:
Move Tiger too. It's too much of a pain in the ass to come back and try to do all this again, we handle those systems together all the time, and don't want to have to special-case the crap out of every single portfile.

IMHO -- forget about gcc 8,9,10,11 on 10.4 - 10.6. 

Just have gcc7 and then gcc12+, and done.

kencu:
Chris set up the builds of libgccN that way for a reason, as per the scalapack ticket, so he won't want to change it now.

We're talking about the same things in 10 different places again, BTW.

kencu:
The problem with Tiger was always the endless "bootstrap comparison failures" -- but somewhere along the way everyone decided we didn't care about that any more, and we've turned it off almost everywhere :>

That is one way to solve the problem ... heh.

barracuda156:
> Chris set up the builds of libgccN that way for a reason, as per the scalapack ticket, so he won't want to change it now.

Yes, I agree. Let us not touch multiple libgccs issue in this PR. I do not want unnecessary resistance :)
After all, that problem is relevant only if someone wants to install old GCC in addition to the latest ones.

Here is rather want to fix PPC (and 10.5 Intel, if you or @catap help) and avoid making controversial changes affecting other systems.

barracuda156:
So to sum up the current state of things, this is what to be added _in this PR:_

1. Add Tiger.
2. Take care of 10.4–10.5 Intel – **I need someone to help here**.
3. Add force deactivation for `libgcc7` in a manner it was done earlier for 10.6.
4. Either remove blacklisting of gccs in gcc12 portfile (old gccs won’t be used anyway since we explicitly set the compiler to `gcc10-bootstrap`), or sort the problem of dependency cycle otherwise. I described it earlier, need to find a ref.

barracuda156:
@kencu Okay, so replying here to comment from: https://trac.macports.org/ticket/66307

1. No `libgcc` version used in Macports, to my knowledge, is broken on PPC (the only one I did not build is `libgcc-devel`), older ones just require a ridiculous amount of compilation time, mostly for nothing, due to parasitic ports which install nothing. This _is_ a problem, at least until we have a build bot for PPC. It is not a major problem, however, since aside of few specific cases, old GCCs are unneeded, and benefits of new GCCs arguably outweigh inconvenience of building few extra ports once in a blue moon.
2. I would definitely prefer not having to be required to build empty `libgccX` ports just to make some `gcc48` work, but I do not have authority here, frankly speaking, to introduce such changes which gonna affect Macports in general and not just PPC users. So I leave this to you, @mascguy or @catap – or anyone from members or experienced contributors. This introduces considerable inconvenience on Intel systems too, so there should be an incentive to sort this out eventually.
3. Regarding a “proof” that libgcc8+ build on older systems (meaning Leopard and SL, I never made any claims about Tiger), while it does feel a bit strange that you don’t take my word here, it is trivial to verify – just add a dependency on `gcc10-bootstrap`, as specified in the notes to that port, and you are good to go. Only 10A190 will require minor patching akin to what you did for `gcc7` and actually less than that, since `darwin.h` in RS6000 takes care of half of needed changes in newer GCCs. Leopard or conventional 10.6, whether Intel or Rosetta, should need nothing of extra patches.

kencu:
as I said above “ IMHO -- forget about gcc 8,9,10,11 on 10.4 - 10.6.”…. you said you wanted to keep gcc11, so OK. But the others are not needed, which solves lots of your heartburn.

but it has to then be sorted out how that can work/be done. Look forward to seeing that.

kencu:
there is a way, of course. libgcc7 can look at libgcc12 instead of libgcc8 on such systems, and use libgcc12 for anything that is a newer version of one of it’s libs. Then libgcc7 installs the rest of it’s libs for gcc7 and older.

libgcc6 and earlier don’t know the difference, they still depend on libgcc7 just like always.

kencu:
let’s take the part about changing the build of every gcc compiler out of this PR. 

It’s not relevant to this, and TBH it was set up this way by an Apple engineer who we all trust and I’d have to look at why you want to change them all, and see if that is reasonable, given they’ve been working well for 20+ years like this.

kencu:
TBH, it’s looking like it would take less time for me to just do this, rather than try to explain to you how to do it, but change things so that libgcc12 is default on all older systems, so that gcc10-bootstrap is build compiler for gcc (at least 10+, probably 8+) on all older systems, and there is an auto-upgrade of libgcc7 -> new libgcc7 using the epoch force-deactivate mechanism.

Then we decide if gcc 8,9,10 will or won’t be supported on these systems. It is simpler to leave them supported, as that is what newer systems support and so all various PGs and ports expect that. But it is some needless building that might be avoided if there is nothing to install for a given libgcc with a simple tweak to that libgcc to install a stub.

Then, once that is all in place, run it yourself to see if it works on your ppc system. If it does, and you get gcc7,8,9,10,11, and 12 all in a working state, we can move on to 10.5 Intel (will be issues to fix), and then the Tigers.

Last time we did this, I did all this building, fixing, and verifying myself, on all these systems. 

But then the cctools upgrade broke gcc8 on Intel, so things bogged down.

barracuda156:
Okay, I am returning to this and hope to finalize and get this merged, finally. It is a pain to stay on archaic GCC on Leopard, and it is one of two major reasons I prefer not to touch it.

I will review this and rebase to master now.

@kencu If we get this merged, I could test stuff on 10.5.8 more easily. I also lack any interest to maintain anything for `gcc7` specifically; I believe, by now we have evidence that is it suboptimal – and in essence a waste of time, IMO. Move everything to the same GCC, and life is easier.
@mascguy @catap Please let me know if we have any issues with this.

mascguy:
While Ken and others need to verify that there aren't any issues with such a switch, there is one issue that needs to be fixed:

Use of `build_arch`, which should be changed to `configure.build_arch`. (Assuming this isn't one of those rare cases where `build_arch` is more appropriate...)

barracuda156:
@mascguy @kencu @catap I guess, that’s it.

CI perhaps will time-out. In fact, all changes are irrelevant to new systems.

barracuda156:
UPD. I guess one more hack needed, to go around a bug somewhere in Macports, which creates circular dependency for gcc which is set to be the main one. Let me verify it is the case on 10.5.8.

barracuda156:
@kencu @catap Could someone please tell me if we should use `gcc10-bootstrap` to build `gcc8+` on 10.5.x **Intel**? I have no idea about that and cannot test it. If it is not needed, then I will drop the condition for it, and use `gcc10-bootstrap` exclusively for PowerPC builds.

barracuda156:
@catap Other thing, my PR does not address universal builds of new gccs on PowerPC (i.e., ppc + ppc64). I believe, they are broken with the current `gcc10-bootstrap` version (since the time it has changed the way it is built as universal itself). Also, AFAIR, with the exception of `gcc10-bootstrap` now, all other universal builds of gccs are “fake” in Macports: they allow building for ppc and ppc64, but do not have universal binaries. (Sorry if I am wrong here.)

This is out of my reach, unfortunately, to fix this. It is desirable to have it fixed, of course, but we may need your help here. Testing I can do on my side.

barracuda156:
@mascguy @kencu @catap This presumably should be good. Please review.

barracuda156:
Well, I am getting a silly issue on Leopard: https://trac.macports.org/ticket/68544
@catap Could you please take a look? Maybe I am just missing something.

barracuda156:
@kencu Here is what I have been referring to, for Leopard. While I am sure it is fixable, and possibly trivially for someone who knows what actually causes the problem, it is not obvious for me at all. Or maybe my brain is dead at the moment, and I miss something obvious. But I tried whatever came to my mind, and nothing works.
And these things totally demotivate dealing with this system. How can I work on any ports, if some very basic stuff fails on every corner? :)
We still have universal pythons broken, all of them, on powerpc. Now, this – I cannot force a non-universal build of `libgcc13` once `gcc10-bootstrap` is built as `+universal` (which is the default variant for it). Presumably I can rebuild `gcc10-bootstrap` as non-universal, and then it should not force universal builds, or maybe uninstall everything and start from scratch. And then have `m4` broken (I believe, your fix is not merged), then something else not building for `ppc64` which I have fixed locally and forgot. And waste another 5–10 hrs to find out something does not work again. This is on the Quad – Macports gcc still takes several hours, since it does not use stage-1 flags which make it faster; imagine someone trying to handle this on a single G4.

kencu:
Yes, it's a project to get it right (which is why it still sits here on the to do list, I guess...)

gcc11+ needs to bootstrap with a c++11-capable compiler. 

gcc10-bootstrap should be used on 10.4 to 10.4, both Intel and PPC, as this bootstrap compiler as it has no deps, AND is vetted by Iain to build gcc11+.

Each system we plan to support in the 10.4 to 10.6 range needs to be shown to be able to build gcc10-bootstrap and use that to build gcc13 (currently gcc13).

Once that can be demonstrated, all those systems can move to libgcc13/gcc13 as their primary gcc compiler. They will need to use the force-deactivate hack for the older libgcc.

========

I am strongly tempted to forget about all the gcc versions between gcc13 and gcc7. There is no requirement for those versions of gcc, gcc8-12, to be supported. We have many missing clang versions that have been retired, for example. If we all agree to retire gcc8-12, then we're done, once we use libgcc7 to fill in the needed libraries that libgcc13 doesn't have.

If someone really sees a need for gcc8-12 on older systems (and please think carefully about why they would be needed, as supporting them takes time we don't have) then each would have to be worked through one-by-one. Probably bootstrapping with the system compilers, with plenty of headaches along the way to work out - yuk.

kencu:
Someday, there will come along some new gcc version, say gcc22, that will not build with gcc10-bootstrap. When that day comes, if we're all still alive, we would use gcc10-bootstrap, to build, say, gcc21-bootstrap, and then that becomes the new bootstrap compiler for all newer gccs.

kencu:
Final point:

cross-arch universal builds are full of headaches. This is not properly sorted out even for the newest systems, as our current "universal" gcc13 is hopelessly broken and is not universal at all, much less able to build things universal.

Restricting the older systems < 10.6 to be one-family universal (either Intel universal OR ppc/ppc64 universal but not cross-arch universal) would go a long way to making things simpler for everyone.

barracuda156:
@kencu

1. While I have personally built _every_ gcc version from gcc8 through gcc13 with gcc10-bootstrap (due to a weird arrangement in Macports which required sequential builds of every libgcc* – thankfully, it is gone now, AFAIU), I see no specific reason to have all those in Macports.
However there is no reason to do it one by one either (nothing will be ever completed this way). _Just do not require them to be built,_ that’s it. If anyone wishes to build any of those, there are no reason to prevent that. As long as those ports are not removed.

I do not want to throw away gcc versions here though. Let’s do everything in a reasonable order. If someone wishes to get rid of old gccs first, please, and I rebase this. Otherwise go my suggested way, and throw away later on, if desired.

2. At the moment all gcc versions are allowed on 10.6. How are they built on Intel? I am fine switching to `gcc10-bootstrap` for all 10.6, of course, but let us make sure it is a preferred way for Intel. _Somebody, please confirm this point._

3. I only implied “one-family” universal builds. Those are broken for multiple ports. In fact, almost for everything which needs modern gcc. But also for some basic ports.

4. The current issue I am facing with `ppc64` on Leopard has to be resolved, that _is_ a stopper.

kencu:
> @kencu
> 
> 1. ...due to a weird arrangement in Macports which required sequential builds of every libgcc* – thankfully, it is gone now, AFAIU)

I know understanding how the gcc ports work can be confusing, but I don't think you quite get the picture yet. The "weird arrangement" is not gone:

```
% /opt/local/bin/port rdeps libgcc6
The following ports are dependencies of libgcc6 @6.5.0_4:
...
  libgcc7
    libgcc8
      libgcc9
        libgcc10
          libgcc11
            libgcc12
              libgcc
                libgcc13
```
This is madness, but newer systems 10.6+ are stuck with it for the moment, until somebody gets the initiative together to dump those useless mid-versions.

Such madness doesn't yet exist on < 10.6. 

There is no libgcc8-12 on these systems, so:

libgcc7 depends on libgcc13.

DONE. 

much easier, and that is my suggested approach, because it is just stupid to build 10 gcc versions just to get gcc6 working.

barracuda156:
@kencu What I mean is whether those useless libgcc* are built as opposed to making a dummy registry entry. Earlier each of them was built, even when literally nothing got installed. A waste of multiple hours of compilation time for nothing. Madness. When I upgraded to gcc13, however, for libgcc12 it installed a dummy without building it. gcc12 surely works, so what it did was correct. I did not check what goes on with earlier versions.

barracuda156:
@kencu I am not even sure we need gcc7 for anything. I know of zero examples where it is needed. (Of course, this does not prove it is not needed, but FWIW.)
We may need gcc6 for Java (if it is restored) and we do need gcc48 for TenFourFox.

lang/gcc9/Portfile
```diff
@@ -79,13 +79,28 @@ depends_lib         port:cctools \
 depends_run         port:gcc_select \
                     port:libgcc9
 
+platform darwin {
+    if {(${configure.build_arch} in [list ppc ppc64]) || ${os.major} < 10} {
+        depends_build-append \
+                        port:gcc10-bootstrap
+        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
+        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
```
catap, line 87:
```suggestion
        configure.compiler.add_deps \
                        no
        depends_build-append \
                        port:gcc10-bootstrap
        configure.cc    ${prefix}/libexec/gcc10-bootstrap/bin/gcc
        configure.cxx   ${prefix}/libexec/gcc10-bootstrap/bin/g++
```
and so on, I guess

catap:
@barracuda156 `gcc10-boostrap` was build universal for a reason, and you may enforce `-universal` globally if you wish (probably you need to use `port upgrade --enforce-variant...`), or add `depends_skip_archcheck` into Portfile to skip checking architecture at all.

BTW see how I've enforced build via `arch` inside `clang-11-bootstrap` by using this compiler, I guess you need that hack as well.

barracuda156:
@catap I believe, we should just fix this for everyone. It is a wrong behavior to force a universal build when non-universal is asked for. `--enforce-variants` is irrelevant, aside of upgrading, it is not used for `port build`. I even uninstalled `libgcc` altogether, it made no change.
(I am ready to accept it is my error somewhere, but so far it does not look like that.)

lang/gcc9/Portfile
barracuda156: I will try this.

kencu:
> @kencu I am not even sure we need gcc7 for anything. I know of zero examples where it is needed. (Of course, this does not prove it is not needed, but FWIW.) We may need gcc6 for Java (if it is restored) and we do need gcc48 for TenFourFox.

Yeah, I would say dump gcc6 too (and forget about gcc6-java, which is useless).

I am tempted to suggest gcc7 stay as an option, just in case, as it's been used for so many years now and we know it's great for everything...

barracuda156:
@kencu gcj6 has been supported as an option to build jdk. Unless we manage to build jdk8 with jdk7, gcj6 remains relevant, even if in need of a few fix-ups.

gcc7 – well, aside of being unable to build ICU, and then R, and then not having multiple fixes backported, yeah, it still works for something. But do we have an example where gcc13 cannot build a port but gcc7 can, any arch?
(To be clear, I am not talking about pre-Leopard systems, which are hopeless anyway.)

kencu:
> @kencu gcj6 has been supported as an option to build jdk. 

Yes, I agree, gcj6 is completely useless. gcj5 works fine, for the systems who need it. gcj6 is a total waste of time.


> gcc7 – well...

dump it too then -- I don't care -- but's it's been the defacto gcc for macports older systems for five+ years, so -- no harm in keeping it until we see what gcc13 breaks  (likely lots of stuff).

kencu:
> > @kencu gcj6 has been supported as an option to build jdk.
> 
> Yes, I agree, gcj6 is completely useless. gcj5 works fine, for the systems who need it. gcj6 is a total waste of time.


Oh, did I misread you? You feel there is a special reason to keep gcc6 and gcj6 around? I have no idea what that would be -- pdftk, which is the only thing I know of that requires gcj to build, builds fine with the current gcj5.

No need to complicate things for absolutely no point at all. And then, hopefully, gcc6 can someday be just dumped as useless too.

barracuda156:
@kencu Aren’t gcc4* using libgcc6 for runtime?

P. S. Well, there is no harm in having all gcc (or clang) versions, the only important thing is to make them non-required.

kencu:
> @kencu Aren’t gcc4* using libgcc6 for runtime?
> 
> For gcj6 and jdk – theoretically, this is the reason to keep gcc6. In practice, we are not at the point to test it yet.

I don’t quite follow this, but pdftk build great with gcj5 and nothing needs a gcj6.

> P. S. Well, there is no harm in having all gcc (or clang) versions, the only important thing is to make them non-required.

No, there is huge harm in having all these gcc versions, as explained (extensively) above.

barracuda156:
@mascguy @catap @kencu A fix suggested by Kirill worked, universal builds are not forced anymore on 10.5.

This should be good to go.

I suggest doing drastic changes to libgcc setups separately. This PR will just move 10.5 to an identical gccs with 10.6.
And gcc on 10.6 is very well tested.

kencu:
There are a bunch to 10.6-for-PPC specific hacks in here which I can't support.

Tiger and Leopard should be done together.

The libgcc fixes need to be done at the same time.

gcc8-12 should not be brought to the older systems due to the previously-described issues that will require extensive deps of libcc8,9,10,11,and 12 needlessly.

There needs to be the standard "deactivate hack" done to bring these systems up to the current libgcc version, like I did when I updated 10.6 to use the current gcc/libgcc two years ago.

Somebody needs to actually try this process out on 10.4 PPC, 10.5 PPC, 10.4 Intel, and 10.5 Intel, and put their names down on the list to say they tested it and it properly works.

barracuda156:
> bunch to 10.6-for-PPC specific hacks

@kencu No, there are no 10A190-specific hacks here. I know very well what 10A190 needs, it is a 4 lines patch in `gcc/config/darwin.h` (significantly less than what you did for gcc7 earlier), but it is _not_ included in this PR.

> Tiger and Leopard should be done together.

_I am not going to deal with Tiger_. I have no resources for that technically, no motivation personally, I also see absolutely no point in that in principle. Too many things are broken on Tiger, and no one is fixing them presently, and no one will, I believe. Capable G4 machines support Leopard. G3 or early G4 – why would someone use those for development?

Leopard is the last official OS for PowerPC and _the last_ system to support ppc64. Snow Leopard is the last development version to run on PowerPC and _the last official system to support running ppc binaries_ in Rosetta.
Tiger should be retired like Jaguar or Panther. There is no need to proactively break support for it, of course, but no point wasting time to work on its support.

If you intend to veto a cardinal improvement for 10.5, because 10.4 is not included – do not wonder next time why I only care about allegedly inferior 10.6 PPC and not amazing and glorious 10.5.8 :)

I will not keep pushing this move in a dire hope someone maybe merges it in another 6 months. Anyone can test this, suggest improvements and fixes or propose a better implementation altogether, and we close this PR.

If no one has interest to have it merged (or merge a different, better alternative), fine, I will close this PR, but I will not do anything for Leopard further on.

> Somebody needs to actually try this process out on 10.4 PPC, 10.5 PPC, 10.4 Intel, and 10.5 Intel, and put their names down on the list to say they tested it and it properly works.

_This has been hanging for almost a year already_. Aside of minor revision of macros, patches are essentially the same.
Nothing prevents anyone to test it. However, neither I cannot force anyone to test it. So what do you want from me?
If you doubt that gcc works on 10.5 (which is supported by gcc upstream, which you are well-aware of), please try using it and put your name here :) If something does not work on 10.5, then it should be reported to gcc upstream.
On our side it is sufficient to ensure 10.5 has same defaults as 10.6 with libstdc++ right now. No need to invent a bicycle.

> There needs to be the standard "deactivate hack" done to bring these systems up to the current libgcc version, like I did when I updated 10.6 to use the current gcc/libgcc two years ago.

This is a good point, I can borrow that.

Overall, it is not helpful to extend clearly unrealistic demands and apply standards which are never used otherwise (we have breaking commits on a daily basis, and no one dies, we just keep fixing and improving whatever feasible). Instead, if this is not gonna be merged, save your, my and everyone time and please express that directly. If this is a decision of Macports members, so be it.

P. S. On a side note, we have this unfixed since a move to `gcc12` and then to `gcc13`:
https://github.com/macports/macports-ports/blob/6f2fed8cf25a6512166733b565b80cf62ae0c69a/_resources/port1.0/compilers/gcc_dependencies.tcl#L10
_Must match,_ but it does not, and nobody is concerned for a year; at the same time, God forbid we commit something which affects only < 10.7 and which has not been tested by the whole community on every possible combo of OS, arch and cpu, with a list of authorized signatures. A procedure followed, I believe, about zero times in recent history, if ever.

kencu:
1. there are plenty of hacks for 10.6 ppc
2. as the person who maintained the compilers on older systems for years, the ones that you now enjoy, and as the person who taught you how to build and use gcc in the first place…


I’m just describing what would have to happen for _me_ to commit this. 

I’m fed up with half-baked fixes and complaints that “nobody cares…”

Of course, what others do is up to them.

kencu:
The fact is that if I'm not going to do this compiler project myself, and it appears that I'm not, I should just leave this to others to sort out as best they can manage.

I left my thoughts, for what they are worth, but I'll leave the rest of this to you folks.

barracuda156:
@kencu Once again, you use phrasing which is misleading.
There are zero 10A190-specific hacks here.
There are minimal settings allowing build on Rosetta on released 10.6.8.
There are no clangs on Rosetta. We have to use gcc10-bootstrap.

The only difference between released 10.6.8 and pre-released 10A190 for gcc is a need for one patch to the header I pointed. (Ironically, even that patch in its unwinder-related part is compatible with 10.6.8, it can be applied and nothing will change in practice for Rosetta.)

That patch is NOT included in this PR.

Rosetta must be supported as long as PowerPC is supported. It is a part of standard released macOS.

kencu:
macports supports pcc just fine, mostly thanks to work I did to support it :)

your endless specific spaghetti hacks to make ppc work on 10.6 are what I won’t commit, for all the reasons outlined previously.

anyway, even leaving out this ongoing difference of opinion that has not been decided by the admins (if Josh commits your python hack for 10.6-ppc we’ll call it decided) this PR still has lots of unresolved issues as I outlined above that would need to be resolved for this to be committable.

no shortcuts in the toolchain business…

kencu:
let me give you an example.

This PR https://github.com/macports/macports-ports/pull/17102 is trying to sort out a good, general plan for when to use certain triplets when cross-compiling on macports.

That is great. A goid, general fix, once sorted out and tested, would be excellent. Andvperhaps in the process, would make ppc builds work on 10.6 when building for ppc too.

Almost all your suggestions fix, sorta, your own specific issue of the moment with a spaghetti hack that is fragile, hard to test and vet, and completely inappropriate for inclusion in a general package manager.

barracuda156:
> your endless specific spaghetti hacks to make ppc work on 10.6

@kencu There have been like 2–3 fixes out of thousands which were needed for 10A190, harmless for Rosetta and having strictly no effect on anything else.
There have been a few fixes needed only for Rosetta, which are easy to drop once a general solution is implemented. I never opposed that and in fact encourage.

You get a wrong impression because every time, including totally unrelated cases, you keep bringing up this topic. Even 10A190 would have needed about 15 minimal fixes overall, affecting strictly nothing else. It is not done for the sole reason: two years back someone at Macrumors wrote something inappropriate to you (I do think that individual was wrong, inconsiderate and behaved not smart), and while I was in no way involved, that somehow keeps being extrapolated on me. And since then anything remotely related to ppc on 10.6 is a taboo, which you oppose on every occasion.
I do not know what you gain or expect to gain, honestly. The only result is that I waste time to carry fixes locally, and have less time to contribute to Macports.

If you think that my contributing to Macports is restricted to PowerPC or 10.6, please, take a pain to actually check, history is public.

I respect and appreciate what you did earlier to support PowerPC systems. I am aware of your contribution. But you chose not to support it further at some point, or at least that was what I have understood. Someone else has to do that.
Whenever you can suggest a better solution, please do. (And yes, I remember cases when you actually did propose a superior fix.) I do not care whose fixes get merged as long as the desired result is there.

Cooperation is and will be greatly appreciated. But if you are not willing to cooperate on something (which you do not have to, of course), perhaps not creating uncalled for obstacles is the option to benefit Macports and everyone.

> Almost all your suggestions fix, sorta, your own specific issue of the moment with a spaghetti hack that is fragile, hard to test and vet, and completely inappropriate for inclusion in a general package manager.

No, it is in fact trivial to test. While those hacks, if you wish, are in reality minimal and a small percentage of commits, there is absolutely no difficulty to verify they do not break anything. And I am yet to see a single example of 10.6 ppc-related stuff breaking anything at all.

>  this PR still has lots of unresolved issues as I outlined above

Well, I do not ask anyone merge it right now as is.

I am not dealing with Tiger though. If someone is interested, please; if not, there are three options: go with 10.5, leaving 10.4 where it is; update 10.4 blindly; do not fix 10.5 because of 10.4 is untested.
I am not the one to make a decision here, but my proposal is to go with 10.5 and leave 10.4 alone where it is now. Unless someone is eager to test 10.4.

kencu:
> > your endless specific spaghetti hacks to make ppc work on 10.6
> 
> @kencu There have been like 2–3 fixes out of thousands which were needed for 10A190, harmless for Rosetta and having strictly no effect on anything else. 

Nobody builds PPC on 10.6... These "rosetta" hacks are just a coy back-door to get your 10.6-PPC nonsense in.

Anyway, if Josh commits your 10.6-PPC-specific hack into his carefully-cultivated python ports, we'll all know that the admins support that and move on. I won't support it, but macports might.

dbsoft:
I have not been following the details of what is holding up this getting merged, and I hope I am not sticking my nose into something where it doesn't belong.  But I'll just say I had been waiting for this gcc upgrade for a long time now.  I had been using the gcc 7 port, but it isn't really sufficient for my needs.  I got tired of waiting and just compiled Iain's tree directly since this pull has been in limbo for a year. So I just want to put my 2 cents in that this is a desired feature for MacPPC. I don't want either of you to give up on it.

barracuda156:
> Nobody builds PPC on 10.6... These "rosetta" hacks are just a coy back-door to get your 10.6-PPC nonsense in.

@kencu Repeating a statement does not amount to an argument, and repeating it again still does not :)

1. Nobody may be using ports built in Rosetta for _end-user purposes,_ but it does make sense for development purposes: it is the easiest way to test something for `ppc` without having a PowerPC hardware.

2. If you do not do something, it does not prove no one does. @catap at some point wanted to try testing something in Rosetta, just to find out that some basic ports were broken.

3. Not liking something does not make it nonsense in a general case, for others.

4. Motivation is absolutely irrelevant as long as the result is correct (in discussed context). Maybe fixes you did were to sneak in some hacks for Debian and never intended to be used on macOS? Who knows. But it does not matter.

5. When there are hacks in Macports like building ports against foreign SDKs, recommendations to replace Apple-provided SDK with some extracted from another Xcode version one, which totally breaks one of architectures for that system, hacks to use libc++ on a system where libstdc++ should be used, by Apple design, – give me a break. There is plenty of stuff which is potentially or actually breaking; there is plenty of stuff altering system behavior and modifying what has been installed and how it was supposed to be used. This is not to say these hacks did not improve things; but they make your core argument out of line with reality.

6. If in order to add a fix we need a certain minimum number of users to vote for it, please add this to Macports policies explicitly (and sure enough, nothing gets fixed for old Intel just as well then). Until then it does not matter how many users benefit from a fix. Even if it is a single user – even if that user is one of us – that should not prevent fixing something.

kencu:
clean fixes…yes

spaghetti code hacks ==> no. (and this is pretty much all your 10.6-PPC hacks).

I realize you don’t know how to do clean fixes most of the time, so you spaghetti code in hacks for your needs and move on.

All I’m saying is: I won’t commit them :)

kencu:
> I have not been following the details of what is holding up this getting merged, and I hope I am not sticking my nose into something where it doesn't belong. But I'll just say I had been waiting for this gcc upgrade for a long time now. I had been using the gcc 7 port, but it isn't really sufficient for my needs. I got tired of waiting and just compiled Iain's tree directly since this pull has been in limbo for a year. So I just want to put my 2 cents in that this is a desired feature for MacPPC. I don't want either of you to give up on it.

When I last brought 10.6 up to parity, I did this:

https://github.com/macports/macports-ports/commit/9aa2182afcd93d2baeffc480c0b19067c1717bc6

We need something like that, including Tiger and Leopard, actually tested by someone.

And to avoid a lot of pain, dropping gcc8-12 on older systems.

It’s not hard. I pointed it out many times before. But it seems nobody wants to actually do the work on that, so here we sit. This PR is not that.

kencu:
> > Nobody builds PPC on 10.6... These "rosetta" hacks are just a coy back-door to get your 10.6-PPC nonsense in.
> 
> @kencu Repeating a statement does not amount to an argument, and repeating it again still does not :)
> 
> 1. Nobody may be using ports built in Rosetta for _end-user purposes,_ but it does make sense for development purposes: it is the easiest way to test something for `ppc` without having a PowerPC hardware.
> 2. If you do not do something, it does not prove no one does. @catap at some point wanted to try testing something in Rosetta, just to find out that some basic ports were broken.
> 3. Not liking something does not make it nonsense in a general case, for others.
> 4. Motivation is absolutely irrelevant as long as the result is correct (in discussed context). Maybe fixes you did were to sneak in some hacks for Debian and never intended to be used on macOS? Who knows. But it does not matter.
> 5. When there are hacks in Macports like building ports against foreign SDKs, recommendations to replace Apple-provided SDK with some extracted from another Xcode version one, which totally breaks one of architectures for that system, hacks to use libc++ on a system where libstdc++ should be used, by Apple design, – give me a break. There is plenty of stuff which is potentially or actually breaking; there is plenty of stuff altering system behavior and modifying what has been installed and how it was supposed to be used. This is not to say these hacks did not improve things; but they make your core argument out of line with reality.
> 6. If in order to add a fix we need a certain minimum number of users to vote for it, please add this to Macports policies explicitly (and sure enough, nothing gets fixed for old Intel just as well then). Until then it does not matter how many users benefit from a fix. Even if it is a single user – even if that user is one of us – that should not prevent fixing something.

If Josh commits your 10.6-PPC-specific hack into his carefully-cultivated python ports, we'll all know that the admins support that and move on. If he doesn’t, I’d vote we get started pulling it all out before this mess gets any worse.

acjones8:
I can offer an old iBook G4 for verifying 10.4 and 10.5 PPC compatibility. It's not very fast, so it'd take a couple of days... but it's available, if that would help.

chf2000:
GCC13 successfully built and run on MacOS X 10.4.11 PPC:
gcc13

kencu:
newer gcc versions are known to build on PPC 10.4 and up (although Iain doesn't test 10.4 PPC any more, he said).

It's just doing the proper integration into MacPorts that nobody has got around to doing.

See https://github.com/macports/macports-ports/commit/9aa2182afcd93d2baeffc480c0b19067c1717bc6 for the last time gcc was updated for an OS (10.6 in that case) for what needs to be done.

Lots of what needs to be done is still not in this PR -- after more than a year.

barracuda156:
@kencu Well, I do not see anything “lots of” there, tbh, but okay, let me go through, add the missing hack and rebase again.

AJenbo:
> Macports gcc still takes several hours, since it does not use stage-1 flags which make it faster; imagine someone trying to handle this on a single G4.

@barracuda156 are you saying there is a way to speed this up ... it took 42 hours to build gcc10-bootstrap on my G4 (800Mhz)

barracuda156:
@AJenbo Oh wow, that’s a lot.
Check flags here: https://github.com/iains/gcc-6-branch/issues/1#issuecomment-1001223448

PR #16896, opened 2022-12-05
Scan manfix

eborisch:
Fix manpage location.


artkiver:
Good catch! Seems like a reasonable change to me.

artkiver:
@eborisch Thank you! 🙏

ryandesign:
Note that this did not in fact put the manpages in the correct location. #18212 fixes it properly.

PR #16922, opened 2022-12-08
icu-devel: update PowerPC patch

catap:
#### Description

This is updated version of patch by request from upstream.

I've skip ci because it can't work on `icu-devel` port because of conflict with `icu` port.

See: https://github.com/unicode-org/icu/pull/2257

###### Type(s)

- [x] enhancement

###### Tested on

macOS 12.6 21G115 x86_64
Xcode 14.1 14B47b

###### Verification 
Have you

### DIFF
--- a/devel/icu-devel/files/patch-balign-to-align.diff
+++ b/devel/icu-devel/files/patch-balign-to-align.diff
@@ -1,15 +1,122 @@
+--- acinclude.m4
++++ acinclude.m4
+@@ -77,6 +77,7 @@ x86_64-*-cygwin)
+ *-*ibm-openedition*|*-*-os390*)	icu_cv_host_frag=mh-os390 ;;
+ *-*-os400*)	icu_cv_host_frag=mh-os400 ;;
+ *-apple-rhapsody*)	icu_cv_host_frag=mh-darwin ;;
++powerpc*-apple-darwin*)	icu_cv_host_frag=mh-darwin-ppc ;;
+ *-apple-darwin*)	icu_cv_host_frag=mh-darwin ;;
+ *-*-beos)       icu_cv_host_frag=mh-beos ;; 
+ *-*-haiku)      icu_cv_host_frag=mh-haiku ;; 
+--- /dev/null
++++ config/mh-darwin-ppc
+@@ -0,0 +1,85 @@
++## -*-makefile-*-
++## Copyright (C) 2016 and later: Unicode, Inc. and others.
++## License & terms of use: http://www.unicode.org/copyright.html
++## Darwin-specific setup (Darwin is the Mac OS X developer preview, successor
++## to Rhapsody, aka Mac OS X Server)
++## Copyright (c) 1999-2012 International Business Machines Corporation and
++## others. All Rights Reserved.
++
++## Flags for position independent code
++SHAREDLIBCFLAGS = -dynamic
++SHAREDLIBCXXFLAGS = -dynamic
++SHAREDLIBCPPFLAGS =
++
++# Do not export internal methods by default
++LIBCFLAGS += -fvisibility=hidden
++LIBCXXFLAGS += -fvisibility=hidden
++
++# Pad out the paths to the maximum possible length
++LD_FLAGS += -headerpad_max_install_names
++
++# We do not need to see the stderr message that the archive was made.
++ARFLAGS += -c
++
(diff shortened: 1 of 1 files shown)


kencu:
the balign to align thing is not needed -- that was just a misconfiguration on that user's system (PATH was set wrong, so wrong assembler was being used).

catap:
@kencu do you mean that clean PowerPC (without MacPorts) can build it?

kencu:
> @kencu do you mean that clean PowerPC (without MacPorts) can build it?

not many builds are tested like that — almost nothing will build on a clean powerpc without macports, esp Tiger. We only use the system toolchain to bootstrap, as you know. 

What I mean is the whole balign issue was a mistake, based on that user’s misconfigured system. You can ignore it and revert the balign patch.

catap:
@kencu upstream would like to include it :) and asked me to adjust my patch. Here I'd like simple to test it before push to upstream.

kencu:
> @kencu upstream would like to include it :) and asked me to adjust my patch. Here I'd like simple to test it before push to upstream.

Yeah, I always hate it when we recommend things to upstream that really don't belong there.

This was a nothingburger that never should have gone anywhere -- one of the reasons I set up my PPC system again was to start nipping these in the budy.

kencu:
Yeah, OK. I guess we are where we are then :>

kencu:
Attention: it turns out that it is quite possible that "balign" is not the same as "align" in PowerPC assembly.

"balign" refers to "align-to-bytes", ie .balign 8 aligns on an 8-byte boundary.
"p2align" refers to "align-to-power-of-two" ie .p2align 3 aligns on an 8-byte boundary.

"align" was not very specific, and meant different things on different platforms, so compilers moved away from using it.

eg: <https://reviews.llvm.org/D16549>

but it looks like "align" in PowerPC assembly means "p2align" and not "balign", according to that LLVM review. If so, substituting "balign" for "align" might do the wrong thing.

I don't have the 100% definitive PowerPC assembler manual right at hand -- I do have it somewhere. Looking.

But NB regarding your upstream PR.

As before, the current cctools assembler we use on Tiger appears to support ".balign" just fine, although the ancient ancient ancient system one at /usr/bin/as appears to not support that.

catap:
@kencu also, switch from `.align` to `.baling` was made a couple years ago: https://unicode-org.atlassian.net/browse/ICU-10305

:)

kencu:
Yeah, that was the right move, and so it should probably not to "align" at all, but p2align instead.
 
".balign 16 == .align 4 == .p2align 4"

Or just scrap the plan ;> -- that would be my vote. After all, this was all based on one user's misconfigured system :>

kencu:
let me try building icu-devel on my "pure" Tiger PPC system without the balign patch. I presume it will build just fine, as ICU built just fine there.

kencu:
Just to confirm that icu-devel builds just fine with TIger PPC with no balign patch at all (as did icu).

catap:
@kencu out of my curiosity: may I ask you to try to build ICU without macports?

kencu:
the system tools on tiger are too old to build anything beyond bootstrapping to something newer.

as a first step, we build apple-gcc42, then cctools and ld64-97, then rebuild apple-gcc42 against the new tools. Then, after that, we can start building newer stuff, and that is macports starting position for all Tiger support.

icu already requires a c++11 compiler to build. 

So trying to fix icu to build against the system “as” on Tiger is a waste of time, can easily cause mistakes as we have seen when the wrong directives get used, and the only reason you ever heard about this is Pete’s system was misconfigured, very unfortunately, which sent us down this rabbit hole.

Now you can keep going on this if you want, as you’ve got the PR open and it’s kinda embarassing to just say “never mind” I guess, but if so, upstream will have to know about the proper figure to use with the “.align” assembler directive, which appears to be different that the one used with “.balign”. Currently you have it patched to use the same figure, which appears to be wrong, which would be very unfortunate, which is why I’m commenting again.

So then we’d be patching something incorrectly and for no good reason :)

kencu:
actually, never mind that last post… I’m too much of a purist sometimes. Will leave you to it.

catap:
I'll switch it to draft until it is merged

catap:
It is irrelevant since the patch was included into upstream and update of ICU delivers it

PR #16926, opened 2022-12-09
dokku-client: new port

bashu:
#### Description

Command-line client for the Dokku PaaS

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6.1 21G217 x86_64
Xcode 14.1 14B47b

###### Verification 
Have you

### DIFF
--- a/sysutils/dokku-client/Portfile
+++ b/sysutils/dokku-client/Portfile
@@ -1,0 +1,51 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+
+github.setup            dokku dokku 0.29.0 v
+github.tarball_from     archive
+
+name                    dokku-client
+
+categories              sysutils
+
+maintainers             {@bashu gmail.com:bashu.was.here} openmaintainer
+
+description             Command-line client for the Dokku PaaS
+
+long_description        {*}${description}
+
+license                 MIT
+
+supported_archs         noarch
+
+checksums               rmd160  c1e871648a84a4341272b7f2f9f9e70a332e2673 \
+                        sha256  b7949af5a16981896dc01ddf0b372d9681304a94bd260f057928d69009c8cbc5 \
+                        size    2137220
+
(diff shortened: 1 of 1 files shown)


bashu:
I skipped: `port test ...` cuz there are no tests for this port

sysutils/dokku/Portfile
```diff
@@ -0,0 +1,49 @@
...
+github.setup            dokku dokku 0.29.0 v
+github.tarball_from     archive
+
+categories              sysutils
+
+maintainers             {@bashu gmail.com:bashu.was.here} openmaintainer
+
+description             Command-line client for the Dokku PaaS
+
+long_description        ${description}
```
herbygillot, line 15: Reusing the `${description}` variable like this requires the use of Tcl's expansion operator (`{*}`), so this should be written like so:

```suggestion
long_description        {*}${description}
```

bashu: fixed

herbygillot:
I see `dokku` has Go components - is the single shell script really all that needs to be installed?

bashu:
> I see dokku has Go components - is the single shell script really all that needs to be installed?

Yes, we only use one standalone script from contrib subdirectory. On second thought should I rename the package to something like, dokku-cli ? What do you think ?

herbygillot:
Yeah, perhaps it might make more sense to name this port `dokku-client`

bashu:
> Yeah, perhaps it might make more sense to name this port dokku-client

Done

sysutils/dokku-client/Portfile
```diff
@@ -0,0 +1,51 @@
...
+                        size    2137220
+
+depends_run             path:bin/bash:bash
+
+use_configure           no
+
+build                   {}
+
+destroot {
+    xinstall -m 755 ${worksrcpath}/contrib/dokku_client.sh ${destroot}${prefix}/bin/dokku
```
herbygillot, line 34: Perhaps the installed binary should also be renamed to `dokku_client`?

herbygillot: Also make the permission `0755`

bashu: > Also make the permission 0755

Fixed

herbygillot (approved):

herbygillot:
Merged. Thanks both for this submission, and for working through these changes.

PR #16974, opened 2022-12-15
clojure: update to 1.11.1.1208

jtrtik:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.1 22C65 arm64
Xcode 14.1 14B47b

###### Verification 
Have you

### DIFF
--- a/lang/clojure/Portfile
+++ b/lang/clojure/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 
 name                clojure
 epoch               20110929
-version             1.11.1.1200
+version             1.11.1.1208
 revision            0
 license             EPL-1
 categories          lang java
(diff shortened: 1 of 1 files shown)


lang/clojure/Portfile
```diff
@@ -16,14 +16,14 @@ long_description    Clojure is a dynamic programming language for the JVM. \
                     dialect of Lisp, with a rich set of immutable, \
                     persistent data structures and a macro system.
 homepage            https://clojure.org
-platforms           any
+platforms           darwin
```
herbygillot, line 19: `platforms darwin` is now implicit by default - you can simply delete the line instead of setting `platforms darwin`:

```suggestion
```

jtrtik: thanks for the hint. I removed the 'platforms' definition

jtrtik: should I squash the commits?

herbygillot: Please do

jtrtik: ok. I force-pushed the squashed commits...

herbygillot:
Thank you @jtrtik

lang/clojure/Portfile
jmroot: Why change `platforms` from `any` to `darwin` at all?

jtrtik: because portlint was failing with "any"... 
https://github.com/macports/macports-base/commit/84d8683465b79829d64269ae624fe0f323006a06

jmroot: Right, that was a bug (only) in lint; the value worked correctly in all other ways. It should be restored to `any` so that we aren't redundantly building a separate archive per macOS version.

jtrtik: ok. I'll make the PR...

jmroot: I can just push it directly if you'd prefer.

jtrtik: the direct update will be much quicker, thanks...

jtrtik: should I prepare the patch or will you do it yourself?

jmroot: Thanks. Done in d906bac.

jtrtik: thanks...

PR #16977, opened 2022-12-15
xar: use path:lib/libssl.dylib:openssl for LibreSSL compatibility.

artkiver:
* tested with libressl-devel and openssl3

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 13.1 22C65 arm64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/archivers/xar/Portfile
+++ b/archivers/xar/Portfile
@@ -3,7 +3,6 @@
 PortSystem          1.0
 PortGroup           clang_dependency 1.0
 PortGroup           github 1.0
-PortGroup           openssl 1.0
 
 set apple_version   494
 github.setup        apple-oss-distributions xar ${apple_version} xar-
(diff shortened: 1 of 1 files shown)


Schamschula:
It really should link against openssl3. openssl is ambiguous.

artkiver:
> It really should link against openssl3. openssl is ambiguous.

Ah, so you think the line should read:

depends_lib path:lib/libssl.dylib:openssl3 \

?

I have not seen that nomenclature used by other Portfiles which use the libssl.dylib.

However, I modified the Portfile locally as suggested above and it does still seem to function OK with libressl-devel, so if that is your preference it seems as if a minor 1 byte modification to the Pull Request is not an unreasonable one.

My guess is, the ambiguity is handled elsewhere?

e.g. if I run % sudo port install openssl:

--->  Computing dependencies for openssl
The following dependencies will be installed:  openssl3
Continue? [Y/n]: 

So OpenSSL3 seems to be implied, regardless of whether or not that stanza in the Portfile is so specific?

I guess either way, it seems to function for me.

Schamschula:
I guess it's OK. I just ran `port info openssl` and found out that it is a

> Shim port installing symbolic links in primary prefix pointing to the openssl3 installation.

All good!

artkiver:
Thank you @Schamschula !🙏

PR #16997, opened 2022-12-17
devel/pnpm: trivial version update

adfernandes:
* trivial version update 7.13.2 -> 7.18.2


adfernandes:
Merging as per `openmaintainer` policy

PR #17005, opened 2022-12-18
ruby/rbenv: trivial update from 1.1.2 to 1.2.0

adfernandes:
ruby/rbenv: trivial update from 1.1.2 to 1.2.0


adfernandes:
@mojca, can you please test?

I believe that the autocomplete patch is no longer required for `rbenv 1.2.0`, and it seems to work for me.

PR #17071, opened 2022-12-27
qt6: remove useless & confusing cmake symlinks

tenzap:
These symlinks are not useful IMO.

- qmake shall find the qt6 installation it belongs to correctly
- when building a Qt application with cmake, one should actually use
  qt-cmake which will also find the qt6 installation it belongs to.

The important thing is to call the qmake & qt-cmake executable from their
installation location. Eg /opt/local/libexec/qt6/bin

especially, qt-cmake shouldn't be called through a symlink

If these symlinks are present and configuration is made with cmake (and not
qt-cmake) one may have errors reported at configure time saying that some Qt
components/tools are not installed. Actually, this is due to heavy use by Qt
of CMAKE_CURRENT_LIST_DIR in the cmake files. Changing this will mess the Qt6
installation. It doesn't resolve to the realpath but to the path to the
symlink.

Configuration should be made with qt-cmake (or if one wants to use cmake, one
has to add additional parameters to the cmake call) or qmake.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

CI

###### Verification 
Have you

### DIFF
--- a/aqua/qt6/Portfile
+++ b/aqua/qt6/Portfile
@@ -113,7 +113,7 @@ array set modules {
         {"Qt 3D"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qt5compat {
@@ -128,7 +128,7 @@ array set modules {
         {"Qt5 Compatibility"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtbase {
@@ -149,7 +149,7 @@ array set modules {
          "Qt OpenGL" "Qt Platform Headers" "Qt Print Support" "Qt XML"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtcharts {
@@ -164,7 +164,7 @@ array set modules {
         {"Qt Charts"}
         "GPLv3 license only"
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: GPL-3"
     }
     qtconnectivity {
@@ -179,7 +179,7 @@ array set modules {
         {"Qt Bluetooth" "Qt NFC"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtdeclarative {
@@ -194,7 +194,7 @@ array set modules {
         {"Qt QML" "Qt Quick" "Qt Quick Layouts" "Qt Quick Widgets"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtimageformats {
@@ -209,7 +209,7 @@ array set modules {
         {"Qt Image Formats"}
         ""
         "variant overrides: "
-        "revision 2"
+        "revision 3"
         "License: "
     }
     qtmultimedia {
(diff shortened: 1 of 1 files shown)


TellowKrinkle:
IMO the port should be able to be found by cmake without needing to add a custom search path.  This is how the qt6 packages work on pretty much every other package manager (including Homebrew).

How about adding the following to qtbase, which will make the standard method of importing qt6, `find_package(Qt6 COMPONENTS ...)` work properly with no modifications to the qt6 files:
```tcl
post-destroot {
    if { ${qt_libs_dir} ne "${mp_prefix}/lib" } {
        # Add cmake find modules that forward to the real ones in the prefix
        set cmakepath ${destroot}${qt_libs_dir}/cmake
        file mkdir "${destroot}${mp_prefix}/lib/cmake/Qt6"
        foreach configfile [list "Qt6Config.cmake" "Qt6ConfigVersion.cmake"] {
            set fd [open "${destroot}${mp_prefix}/lib/cmake/Qt6/${configfile}" "w"]
            puts $fd "set(_PREFIX_PATH_BAK \${CMAKE_PREFIX_PATH})"
            puts $fd "list(APPEND CMAKE_PREFIX_PATH \"${qt_dir}\")"
            puts $fd "include(\"${qt_libs_dir}/cmake/Qt6/${configfile}\")"
            puts $fd "set(CMAKE_PREFIX_PATH \${_PREFIX_PATH_BAK})"
            close $fd
        }
    }
}
```
The above will generate the following cmake find scripts that temporarily add the real qt directory to the prefix path, then load the real find module:
```cmake
set(_PREFIX_PATH_BAK ${CMAKE_PREFIX_PATH})
list(APPEND CMAKE_PREFIX_PATH "/opt/local/libexec/qt6")
include("/opt/local/libexec/qt6/lib/cmake/Qt6/Qt6Config.cmake")
set(CMAKE_PREFIX_PATH ${_PREFIX_PATH_BAK})
```

tenzap:
Will this work with all dirs that hold qt6 cmake files? IIRC  there are many such dirs starting with qt6 (qt6qml, qt6*...). Will all qt modules build correctly and be found? There used to be such issues with the old way.

tenzap:
These are the recommended ways to build qt6 projects according to qt
https://doc.qt.io/qt-6/cmake-build-on-cmdline.html

jhoyt4:
To carry a conversation over from [PR 16735](https://github.com/macports/macports-ports/pull/16735) - IMHO this portfile is required to allow applications to be build outside the macports structure. I do this myself (mythtv) and was my main purpose when the portfile was created.

I have not had a chance to check @TellowKrinkle's proposed solution, but believe it to both be better than my original hack and the current solution.

TellowKrinkle:
Side note, the cmake forwarding script should work with the other cmake files if you wanted to do that

(If you forwarded everything, you wouldn't even need to mess with the cmake prefix path and could just do `include("target-script.cmake")`.)

I just think it'll make it really hard to do a `port select` in the future when we have multiple qt6s, since the select would need to select files from multiple subports at once, and I have the feeling it's not set up to support that well (though I haven't actually checked).  With just doing those two files, you could have the `port select` only target qtbase and still make most cmake projects work fine.

MarcusCalhoun-Lopez:

PR #17087, opened 2022-12-28
python311: fix build on G3 processors

evanmiller:
#### Description

There seems to be a bug in the GCC7 linker on 10.4 that causes `-all_load` to bundle up regular `ppc` (non-Altivec) functions and report the collection as `ppc7400` (Altivec), resulting in a failed build on non-Altivec machines. Adding `-Wl,-force_cpusubtype_ALL` works around this behavior. I don't *think* this change will have untoward consequences for the Altivec crowd, but it does let me build Python 3.11 on a G3.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.4.11

###### Verification 
Have you

### DIFF
--- a/lang/python311/Portfile
+++ b/lang/python311/Portfile
@@ -42,6 +42,9 @@ if {${os.platform} eq "darwin" && ${os.major} <= 10} {
 if {${configure.build_arch} in "ppc ppc64"} {
     # https://trac.macports.org/ticket/66483
     configure.ldflags-append -Wl,-read_only_relocs,suppress
+    # G3 builds fail without this flag, as converting a static library
+    # to dynamic using -all_load inadvertently sets the arch to ppc7400
+    configure.ldflags-append -Wl,-force_cpusubtype_ALL
 }
 
 depends_build       port:pkgconfig


barracuda156:
Hmm, maybe make it a variant at least? I do not think it is a great idea to degrade performance on G4/G5 for the sake of supporting G3. And why is this applied to `ppc64`? G3 does not support it.

kencu:
having more CPU subtypes has no degrading effect on any current builds, and variants are really a bad idea for things like this as people can have no idea if they should have them or not.

it should really be the default for cctools to do this -- I think it doesn't because somewhere in cctools or ld64-97 something is defaulting to 10.6 as a base OS and that only supports ppc7400, so the system tools just use that. 

LGTM

kencu:
This is not an open maintainer port, so I can't push it for you.

barracuda156:
> having more CPU subtypes has no degrading effect on any current builds

@kencu Ah, so it won’t prevent AltiVec from being used? (Not referring to Python here, but generally.)

evanmiller:
> > having more CPU subtypes has no degrading effect on any current builds
> 
> @kencu Ah, so it won’t prevent AltiVec from being used? (Not referring to Python here, but generally.)

This change affects the linking stage, i.e. how the code self-reports and not how it is actually compiled.

barracuda156:
@evanmiller @kencu Then awesome!

On a side note, anyone checked if this is fixed in Iain’s `xtools`?

kencu:
finding clear documentation in 2023 about exactly what force_cpusubtype_ALL will do to code optimization on darwin ppc is actually not that easy. I know force_cpusubtype_ALL is used broadly. I may have to run some test experiments to prove exactly what it changes. As is so often the case these days, Iain may be the only one who 100% can say exactly what it does to links.

barracuda156:
> finding clear documentation in 2023 about exactly what force_cpusubtype_ALL will do to code optimization on darwin ppc is actually not that easy. I know force_cpusubtype_ALL is used broadly. I may have to run some test experiments to prove exactly what it changes. As is so often the case these days, Iain may be the only one who 100% can say exactly what it does to links.

Fair enough.
If there is some transparent way to test this, I can do testing later in Jan, once back to native PPC hardware.

kencu:
this is useful…

https://books.google.ca/books?id=K8vUkpOXhN4C&pg=PA1012&lpg=PA1012&dq=force_cpusubtype_ALL&source=bl&ots=ONohWXZyXy&sig=ACfU3U37Ag3l_ybxgjFeN4ydvSUpBQLNDA&hl=en&sa=X&ved=2ahUKEwia9bDQ2Kv8AhWzKH0KHUtDC444HhDoAXoECAQQAw#v=onepage&q=force_cpusubtype_ALL&f=false

evanmiller:
As a linker option, `-force_cpusubtype_ALL` can't rewrite compiled object code. I believe it only controls the linked archive's self-reporting. If Altivec code is present but linked with `-force_cpusubtype_ALL`, a G3 will crash when it attempts to execute an unrecognized instruction. This happened to me in my (unsuccessful) attempts to get OpenBLAS working with G3s.

kencu:
Yes, in the end I guess the way to think of it is "don't check the CPU subtype, just run it..."

PR #17100, opened 2022-12-28
gdb: Fix x86_64 builds on Apple silicon a different way

Gcenx:
#### Description

@kencu heres a much cleaner way to handle this.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/gdb/Portfile
+++ b/devel/gdb/Portfile
@@ -1,7 +1,7 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem      1.0
-PortGroup       muniversal 1.0
+PortGroup       muniversal 1.1
 PortGroup       compiler_blacklist_versions 1.0
 
 name            gdb
@@ -65,6 +65,8 @@ compiler.thread_local_storage \
 # compiler supports the standard - getting rid of old Apple GCC versions and the like?
 compiler.cxx_standard   2011
 
+triplet.add_build       cross
+
 configure.dir ${workpath}/build
 configure.cmd ${worksrcpath}/configure
 configure.args \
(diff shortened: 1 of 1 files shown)


barracuda156:
Typo in the title, should be `gdb`.

kencu:
magical mystical port group magic. -- hah hah, gotta love it.

Now why would the universal PortGroup be doing anything at all on a build that is not universal?

Would anyone (including me) ever think it would be?

Gcenx:
> magical mystical port group magic. -- hah hah, gotta love it.

Yep I’ve moved everything I’m maintaining in my overlay to muniversal-1.1 for all these special little things.

> Now why would the muniversal PortGroup be doing anything at all on a build that is not universal?

When cross-compiling it’s common to set --build & --host, these are not always needed when you have a cross-toolchain installed but we don’t in the case of macOS so we simply telling configure we’re xyz host and building for xyz since we match no need to look for prefixed tools.

Apples entire toolchain is a cross-toolchain but configure based projects can get confused so we’re simply helping them understand what’s todo.

> Would anyone (including me) ever think it would be?

If you don’t commonly cross-compile you might not look for something like this.

kencu:
Yeah, but what I mean is the muniversal PortGroup should do **nothing** when you are not building universal.

Gcenx:
> Yeah, but what I mean is the muniversal PortGroup should do **nothing** when you are not building universal.

That’s not how muniversal-1.1 works though with legitimate reason.

You can have separate configure.args etc for specific architectures without special blocks, this makes makes Portfiles way less complicated.

To build a project for x86_64 from arm64 we required setting --build & --host, this being an example of a port that requires this.

Gcenx:
> Typo in the title, should be `gdb`.

That’s due to the commit being wrong that’s what I get for doing that on mobile…., Ken fixed the Pr title and I’ve just pushed a fixed commit message thanks for noticing.

kencu:
I think we'll stick with the one I tested, but I will look at Marcus' group in future, thanks, perhaps for the next update.

Gcenx:
> I think we'll stick with the one I tested, but I will look at Marcus' group in future, thanks, perhaps for the next update.

You really should muniversal-1.1 has really helped shrink down the usual B.S especially when specific arguments are required per arch.

PR #17102, opened 2022-12-28
muniversal-1.1: Set --build when cross-compiling

Gcenx:
This is almost always required when building for x86_64 from Apple Silicon

Commonly configure based projects will be looking for ${triplet.cpu.${arch}}-${triplet.vendor}-${triplet.os}-${tool}

e.g. 
An x86_64 build from Apple silicon will be looking for x86_64-apple-darwin22-ar when --build is not set

###### Type(s)

- [x] enhancement

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/muniversal-1.1.tcl
+++ b/_resources/port1.0/group/muniversal-1.1.tcl
@@ -166,7 +166,7 @@ default triplet.add_host    {cross}
 
 # possible values: none, all, cross, or a list of architectures
 options triplet.add_build
-default triplet.add_build   {none}
+default triplet.add_build   {cross}
 
 foreach arch {arm64 x86_64 i386 ppc ppc64} {
     options triplet.cpu.${arch}


kencu:
I thought about this too -- but have to see what happens with meson, cmake, etc, etc

Gcenx:
> I thought about this too -- but have to see what happens with meson, cmake, etc, etc

cmake and meson wouldn’t be affected as these zero these out already as it breaks these build systems.

kencu:
how do they zero them out?

how does anyone ever know how these portgroups interact, anyway? Must be hell on Earth for new users to sort out, as it is nearly hell on Earth for me to sort out.

Gcenx:
> how do they zero them out?

https://github.com/macports/macports-ports/blob/be3c4c3524c3853e32991f25a26893ecc62acf3f/_resources/port1.0/group/muniversal-1.1.tcl#L211-L241

> how does anyone ever know how these portgroups interact, anyway? Must be hell on Earth for new users to sort out, as it is nearly hell on Earth for me to sort out.

I’ve only dug into the ones I make use of but yeah sometimes it’s very confusing….

kencu:
btw

> An x86_64 build from Apple silicon will be looking for x86_64-apple-darwin22-ar when --build is not set

that is not what I saw fixing gdb recently. I had to override AR=ar to make it work…even though build was set to aarch-etc.

I personally don’t like hiding these fixes in Portgroups. I prefer it out in the open where I can see it, understand it, and debug it.

Have asked Marcus to review your suggestion, as he’s back abit now.

Gcenx:
That’s because you’d actually set `configure.args-append --build=arm64-apple-${os.platform}${os.major}` when it should have been `configure.args-append --build=x86_64-apple-${os.platform}${os.major}`

--build & --host need to match to force none prefixed tools.

kencu:
Yeah, but -- I set it right -- build is the build system, host is the host system, target is the target system.

<https://wiki.osdev.org/Target_Triplet>

I don't want to tell my arm64 system it is an x86_64 system when it is an arm64 system, even if I'm cross compiling...

It won't be long until this arm64 system doesn't have Rosetta2 any longer, and then all this cross compiling that is set up like that will be broken.

And this is what I mean, by hiding all this in PortGroups that nobody understands, and nobody can debug when it is broken.

Gcenx:
> Yeah, but -- I set it right -- build is the build system, host is the host system, target is the target system.
> 
> https://wiki.osdev.org/Target_Triplet

That’s only true when you already have the target binaries hence _why_ configure wants prefixed tools over the none prefixed tools. 

There’s no actual point in setting --build in this scenario unless configure didn’t correctly detect it’s been ran on an arm64 system. That’s why you were required to manually set AR.

The entire macOS toolchain is a cross-compilation toolchain that doesn’t require prefixed tools when compiling for supported targets aka darwin/iOS/etc.

> I don't want to tell my arm64 system it is an x86_64 system when it is an arm64 system, even if I'm cross compiling...

As explained above you set --build & --host match so prefix tools are not expected. If you want to do more standard cross-compiling you really should be installing prefixed tools for example the mingw toolchain.

> It won't be long until this arm64 system doesn't have Rosetta2 any longer, and then all this cross compiling that is set up like that will be broken (and we will have to rethink some things we are assuming at present, but that is for another day :> )

We lightly have a couple more years of rosetta2 since this time it’s Apples own tech and not third-party.

> And this is what I mean, by hiding all this in PortGroups that nobody understands, and nobody can debug when it is broken.

This could be said for every single PortGroup.

Gcenx:
I’ll need to tweak this as this only really applies when it’s darwin building for other darwin targets.

MarcusCalhoun-Lopez (commented):
Please forgive me if I am missing something, but if `--build=` sets [the platform on which the compilation tools are executed](https://wiki.osdev.org/Target_Triplet) and [defaults to the result of running `config.guess`](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Specifying-Target-Triplets.html), don't we want, for most ports, to trust the output of `config.guess`?

Gcenx:
> Please forgive me if I am missing something, but if `--build=` sets [the platform on which the compilation tools are executed](https://wiki.osdev.org/Target_Triplet) and [defaults to the result of running `config.guess`](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Specifying-Target-Triplets.html), don't we want, for most ports, to trust the output of `config.guess`?

When building on Apple Silicon (not running under Rosetta2) and want to build for x86_64 using `--host=x86_64-apple-darwin22` configure will then require prefix tools e.g. `x86_64-apple-darwin22-ar`. The reason for that is usually the system building can’t use it’s own tools for another system/architecture (unlike the macOS toolchain)

<br>

However if we also set `--build=x86_64-apple-darwin22` configure now looks for none prefix tools, e.g. `ar` instead of `x86_64-apple-darwin22-ar`

<br>

The changes to `meson` a while back caused this exact change in behavior on macOS causing the crossfiles to require almost everything to be manually configured.

Gcenx:
As far as configure (and most other build systems) are concerned

aarch64-apple-darwin22 & x86_64-apple-darwin22 are two different platforms.

Gcenx:
As Macports mostly using llvm/clang a more relevant reference would be https://clang.llvm.org/docs/CrossCompilation.html

Most other documentation focuses on using a targeted GNU toolchain.

barracuda156:
@Gcenx We do have gcc on PowerPC systems and Rosetta (original one). Just in case, this should not be forgotten and should be tested.
On Rosetta the working and pretty well-tested setting is `--build=powerpc-apple-darwin=`.

kencu:
So this has been sitting a year because we don't treasure the idea of setting "build" to other than the real "build" machine, as Marcus said above. It seems fragile. Although I can see it can work with Rosetta working the right direction (x86_64 builds work on Apple silicon), it is likely to be broken in the other direction, for Apple silicon on x86_64, for example, and perishable.

So, without fibbing to the build system, if the problem boils down to missing tools named ```${triplet.cpu.${arch}}-${triplet.vendor}-${triplet.os}-${tool}``` then would it be an idea to generate such tool names that are symlinks to said tools that we know work? That seems to be using the mechanisms as they are supposed to be used, which just feels like the right thing to do.

We already do plenty of symlinks in the cctools port.

eg 

we know "ar" on darwin works on both arm64 and x86_64 binaries through our deep knowledge of the OS. So we do this for the supported universal archs:

x86_64-apple-darwin22-ar -> ar
arm64-apple-darwin22-ar -> ar

and -- we are done? No further shenanigans? 

In fact, it's kinda too bad Apple didn't do this by default in the first place, I suppose.

pmetzger:
What do we need to do in order to resolve this one way or another? It's been open for over a year.

kencu:
> @kencu as I'd explained already heres what happens when you only set `build` and skip setting `host` [wine-devel-main.log](https://github.com/macports/macports-ports/files/13694321/main.log)
> 
> As you can see that build was started from aarch64 Terminal session and cross-compiled for x86_64, it doesn't look for prefixed tools, if you however set both 'build' & `host` it will look for host prefixed tools.

I don't think anyone is against setting both build and host.

pmetzger:
I'm sincerely sorry about how long this took. Mostly we get PRs handled very quickly, but infrastructure ones often result in long discussions that don't terminate. I apologize about that.

PR #17124, opened 2022-12-29
hstr: update to 2.6.0

herbygillot:
###### Verification 
Have you

### DIFF
--- a/shells/hstr/Portfile
+++ b/shells/hstr/Portfile
@@ -3,19 +3,20 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        dvorka hstr 2.5
+github.setup        dvorka hstr 2.6
 while {[llength [split ${version} .]] < 3} {
     version         ${version}.0
 }
 revision            0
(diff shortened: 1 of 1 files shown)


ryandesign:
I hadn't updated the port because I noticed they hadn't posted a release tarball. I assumed they had forgotten to do so and would do so later, at which time I would update the port. But now it's been a few weeks and they still haven't. Do you know whether they have intentionally stopped publishing release downloads? (If so, can you point to where that's stated?) Otherwise, I would prefer that we file an issue with them asking them to publish the release download for 2.6.0 and then update to that and not switch to archive downloads.

herbygillot:
Well there is a release, but it's pointing to the archive tarball ( https://github.com/dvorka/hstr/releases/tag/2.6 ). So I was thinking to just use archives for this version, and switch to release tarballs later when they become available.

But that said, I opened an issue: https://github.com/dvorka/hstr/issues/474

herbygillot:
Updated to use the newly added release tarball for 2.6.0

ryandesign:
`use_autoreconf yes` shouldn't be needed with the release tarball, right?

herbygillot:
> 

Removed.

PR #17125, opened 2022-12-30
libressl: revision 1 to Portfile, improve livecheck.

artkiver:
* Closes: https://trac.macports.org/ticket/66568

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 13.1 22C65 arm64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/security/libressl/Portfile
+++ b/security/libressl/Portfile
@@ -49,5 +49,5 @@ platform darwin {
 }
 
 livecheck.type      regex
-livecheck.url       https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/
-livecheck.regex     "(\\d+\\.\\d+\\.\\d+)"
+livecheck.url       https://www.libressl.org/
+livecheck.regex     "The latest stable release is (\\d+\\.\\d+\\.\\d+)"


herbygillot:
If you're only making changes to the livecheck, you don't have to increment the revision.

This is because the livecheck is only used by MacPorts infra, and does not result in any changes to the files or binaries produced by this port.

`revision` should only be incremented when a change will modify the resulting binary, libraries, bundled assets, or any other file produced by the port's build process that gets packaged, installed and distributed.

artkiver:
> If you're only making changes to the livecheck, you don't have to increment the revision.
> 
> This is because the livecheck is only used by MacPorts infra, and does not result in any changes to the files or binaries produced by this port.
> 
> `revision` should only be incremented when a change will modify the resulting binary, libraries, bundled assets, or any other file produced by the port's build process that gets packaged, installed and distributed.

Ah, good to know! My bad.

I also submitted a PR for libressl-devel with a similar improvement, though in that instance I noticed the Portfile did not even have a revision field, so I added one.

In both instances, it is just for the livecheck bits, should I amend the commits? If so, guidance on what to modify would be appreciated.

herbygillot:
> Ah, good to know! My bad.
> 
> I also submitted a PR for libressl-devel with a similar improvement, though in that instance I noticed the Portfile did not even have a revision field, so I added one.
> 
> In both instances, it is just for the livecheck bits, should I amend the commits? If so, guidance on what to modify would be appreciated.

Yes, you can amend to just keep the revision fields the same as they were. In the instance of `libressl-devel`, since it was missing the field to begin with, just add a `revision` field set to zero.

herbygillot:
Great, the commit for this PR looks good. Will merge. Thank you @artkiver.

artkiver:
Thanks @herbygillot ! 🙏 The instructions you provided on that snac update from a while back were somewhat helpful, though ultimately it seemed as if git push -f was required?

I ran into some different issues with the libressl-devel attempt to amend that, I'll see if I can improve it hopefully.

herbygillot:
Yes, if you amend a commit, you will need to use `push -f` to override the remote branch with your now different local branch.

PR #17224, opened 2023-01-05
tk: add patch to allow building on OSX 10.6

michaelld:
#### Description

add patch to tk to allow building on OSX 10.6, by removing 10.7+ API call

I think this is relevant to just `+quartz`, but it doesn't hurt to apply it regardless

NOTE: the functionality is relevant for retina displays only, which were introduced in 2012 & 10.8 -- maybe late 10.7 ; retina was not part of 10.6 hardware from what I can determine, so this change should be safe

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.7.3 20G1102 x86_64
Xcode 12.3 12C33

macOS 10.6.8 10K549 i386
Xcode 4.2 4C199

###### Verification 
Have you

### DIFF
--- a/x11/tk/Portfile
+++ b/x11/tk/Portfile
@@ -38,6 +38,12 @@ patch.dir           ${workpath}/${name}${version}
 # see https://trac.macports.org/ticket/57594
 patchfiles-append   patch-dyld_fallback_library_path.diff
 
+# https://github.com/tcltk/tk/commit/b0cb1a48cb0c4a0118d45e8804476a6b4ab502c8
+# this is the only code that fails on OSX 10.6, so enable it for 10.7 or newer only; see also
+# https://developer.apple.com/documentation/quartzcore/calayer/1410746-contentsscale
+# retina displays were introduced in 2012 with OSX 10.8, so this removal should be safe for 10.6
+patchfiles-append   patch-enable-viewDidChangeBackingProperties-on-10.7-or-newer.diff
+
 configure.checks.implicit_function_declaration.whitelist-append \
                     closedir64 \
                     readdir64 \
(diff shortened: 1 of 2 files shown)


barracuda156:
@michaelld
> I think this is relevant to just `+quartz`

Likely yes, since it builds on Rosetta as-is.

michaelld:
> @michaelld
> 
> > I think this is relevant to just `+quartz`
> 
> Likely yes, since it builds on Rosetta as-is.

@barracuda156 Rosetta is used to run PPC code on Intel, yes? If that's the case, then this makes sense as `+quartz` is disabled for < 10.6, which includes all PPC

barracuda156:
> @barracuda156 Rosetta is used to run PPC code on Intel, yes? If that's the case, then this makes sense as `+quartz` is disabled for < 10.6, which includes all PPC

@michaelld I would believe that `+quartz` is not invoked on any PPC, yes.

MarcusCalhoun-Lopez:
Tk has code to support older systems.
Might it make sense to have the patch wrap the function in `#if !(MAC_OS_X_VERSION_MAX_ALLOWED < 1070)` as [other parts of the tk codebase do](https://github.com/tcltk/tk/blob/5e7bcea28317ca7b25fe5cad6c4878ae6d048148/macosx/tkMacOSXWindowEvent.c#L312) and then apply the patch unconditionally.
We might then be able to get the patch merged upstream.

michaelld:
@MarcusCalhoun-Lopez great idea! I just pushed that change

michaelld:
I wonder why they used `#if !(MAC_OS_X_VERSION_MAX_ALLOWED < 1070)` instead of `#if MAC_OS_X_VERSION_MAX_ALLOWED >= 1070` like they do elsewhere ... I know the version numbering went to 6 digits -- for example `101500` is for OSX 10.15 -- but I'd think those are all greater or equal to the older 4-digit numbers ... hmmmm well whatever this change does work and can be submitted upstream since it's nicely generic

MarcusCalhoun-Lopez:
Tcl/Tk does not have the nicest bug reporting system, but I tried to [submit the patch upstream](https://core.tcl-lang.org/tcl/tktview?name=d84d334ae6).
Thank you for working on this.

PR #17230, opened 2023-01-06
various: set new extract.rename to yes if it exists

michaelld:
#### Description

set new extract.rename to yes if it exists, in various ports

This feature is in current master and will be part of the next MacPorts release. When set to yes, it tries to rename the extracted tarball top-level directory to what the Portfile specifies, if different. Some ports require this while others do not. Here are 4 that I've come across in my recent work; I'm sure there are more.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.7.3 20G1102 x86_64
Xcode 12.3 12C33

as well as 10.5, 10.6, 10.7, and 10.11 ... same need in each of these with the current MacPorts base master HEAD

###### Verification 
Have you

### DIFF
--- a/databases/lmdb/Portfile
+++ b/databases/lmdb/Portfile
@@ -42,6 +42,15 @@ makefile.override-append \
                     CFLAGS \
                     PREFIX
 
+# have to do the extract rename for this port to work
+# this will not do anything in the current release 3.8.0 or prior
+# it will be part of the next release: 3.8.1 or 3.9.0
+# https://trac.macports.org/ticket/66415
+# remove the 'if' clause with the next 'port' release, leaving just the middle line
+if {[exists extract.rename]} {
+    extract.rename  yes
+}
+
 post-destroot {
     # Install pkg-config .pc file because Linux distributions have done
     # so and other software now depends on this but the developer of
--- a/devel/xxhash/Portfile
+++ b/devel/xxhash/Portfile
@@ -25,6 +25,15 @@ if {[string match "*gcc-4.*" ${configure.compiler}]} {
                     patch-Makefile-gcc.diff
 }
 
+# have to do the extract rename for this port to work
+# this will not do anything in the current release 3.8.0 or prior
+# it will be part of the next release: 3.8.1 or 3.9.0
+# https://trac.macports.org/ticket/66415
+# remove the 'if' clause with the next 'port' release, leaving just the middle line
+if {[exists extract.rename]} {
+    extract.rename  yes
+}
+
 subport ${name} {
     revision        2
 
--- a/multimedia/dav1d/Portfile
+++ b/multimedia/dav1d/Portfile
@@ -33,6 +33,15 @@ checksums           rmd160  7a4a3f9ffc2c1d40b8da7077648b8e6c0fd41f99 \
 
 use_bzip2           yes
 
+# have to do the extract rename for this port to work
+# this will not do anything in the current release 3.8.0 or prior
+# it will be part of the next release: 3.8.1 or 3.9.0
+# https://trac.macports.org/ticket/66415
+# remove the 'if' clause with the next 'port' release, leaving just the middle line
+if {[exists extract.rename]} {
+    extract.rename  yes
+}
+
 # see https://trac.macports.org/ticket/62618
 # patch can be applied always, but limit to < darwin 10 so no chance of troubles
 if {${os.platform} eq "darwin" && ${os.major} < 10} {
--- a/textproc/asciidoc/Portfile
+++ b/textproc/asciidoc/Portfile
@@ -31,6 +31,15 @@ installs_libs       no
 
 homepage            https://asciidoc-py.github.io/
 
+# have to do the extract rename for this port to work
+# this will not do anything in the current release 3.8.0 or prior
+# it will be part of the next release: 3.8.1 or 3.9.0
+# https://trac.macports.org/ticket/66415
+# remove the 'if' clause with the next 'port' release, leaving just the middle line
+if {[exists extract.rename]} {
+    extract.rename  yes
+}
+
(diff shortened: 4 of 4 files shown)


ctreleaven:
Hi Michael:  I'm sure you're doing the right thing but I'm not grokking why https://trac.macports.org/ticket/66415 affects dav1d??  When the new version of MacPorts base is released, dav1d would fail at the extract stage?

Is this because the dav1d port needs to set the distname (line 28)?  With the change in base, would we still need to do that?

Thanks for any help you can offer to make me a little less clueless than before!

jmroot (commented):
Looks fine as far as it goes. But it looks like the github portgroup should be defaulting this to `yes` in the `github.tarball_from tarball` case (this is one good reason to switch to `github.tarball_from archive` BTW), and taking the tag prefix and suffix into account when setting up `distname` and/or `worksrcdir` in the other cases.

jmroot:
> When the new version of MacPorts base is released, dav1d would fail at the extract stage?
> 
> Is this because the dav1d port needs to set the distname (line 28)? With the change in base, would we still need to do that?

Setting distname isn't the issue, it's that the directory the archive extracts to doesn't match the archive name. In this case, the archive is `dav1d_1.0.0.tar.bz2` which when extracted gives you a directory named `dav1d-1.0.0-99172b11470776177939c3d2bc366fe8d904eab7`.

Normally you would handle this by just setting `worksrcdir` to the right value (instead of its default of `$distname`). But the hash in the directory name can't be derived automatically, which makes this a manual process for each release. So it's easier to have base automatically rename whatever directory does exist. (Ideally of course upstream would publish tarballs with predictably-named contents.)

ctreleaven:
@jmroot Thanks for the explanation.  If I understand correctly, the port zim (that I happened to be working on yesterday) will not be affected since we can make changes to the github portgroup to automagically workaround the issue.  With current base, zim's work directory is now:

```
$ cd /opt/local/var/macports/build/_Users_craigtreleaven_mp_macports-ports_editors_zim/zim/work
CT-MBP11:work craigtreleaven$ ls -l
total 0
lrwxr-xr-x  1 macports   41 Jan  6 13:06 zim-desktop-wiki-0.75.1 -> zim-desktop-wiki-zim-desktop-wiki-eea32ce
drwxr-xr-x 32 macports 1024 Oct 25 15:16 zim-desktop-wiki-zim-desktop-wiki-eea32ce
```

neverpanic (approved):
OK for asciidoc.

jmroot:
> If I understand correctly, the port zim (that I happened to be working on yesterday) will not be affected since we can make changes to the github portgroup to automagically workaround the issue.

Sure, the workaround can be added to the portgroup. But pretty much every port that uses the github portgroup should still be setting `github.tarball_from archive` (if using `releases` isn't possible). Unfortunately the default value for that option is the worst choice, but the default can't be changed because it would break the checksums for all ports using it.

michaelld:
@ctreleaven `zim` works for me without tweaks. As noted, `dav1d` doesn't use GitHub & the tarball contains a hash that would need to be manually set. Doing this `extract.rename yes` is quick, easy, and concise. I do think it's the way to go for `dav1d` and some other similar ports that don't use GitHub.

michaelld:
@neverpanic turns out `asciidoc` can use `releases` ... here's the change:
```
checksums           rmd160  e0dcc5db1a0cfa038cfb56fb259ee55412d5378e \
                    sha256  237b2ba5c35c0ae7ccd4cd44ebf1d87c20b2695dae01798954416d492ef7fa0e \
                    size    1276134
name                asciidoc

distname            ${name}-${version}
github.tarball_from releases
dist_subdir         ${name}/${version}_1
```

This is a more robust way to go than what's in this PR. I will remove that port from this PR and create a new one with this change.

michaelld:
@Schamschula `xxhash` only does archives, so it will need an `extract.rename yes` to work (whether in the port or in GitHub PG)

michaelld:
`lmdb` also only does archives, so it will need an `extract.rename yes` to work (whether in the port or in GitHub PG)

michaelld:
> `lmdb` also only does archives, so it will need an `extract.rename yes` to work (whether in the port or in GitHub PG)

Actually this port can be done by setting `worksrcdir          ${name}-${github.author}_${version}` ... this seems consistent between recent versions.

jmroot:
> Actually this port can be done by setting `worksrcdir ${name}-${github.author}_${version}` ... this seems consistent between recent versions. I will remove it from this PR and do a separate PR with this change

The updated portgroup should handle this automatically by setting `worksrcdir` to the full tag name.

michaelld:
OK so with that change to the GH PG, I now see that `lmdb` can be tweaked as follows:
```
github.setup        LMDB lmdb 0.9.29 LMDB_
revision            1
checksums           rmd160  17d59b57e99865cc75af229cc6204e931d8934c5 \
                    sha256  1f73a97da602346568f40a03030169d6f7dcd491ad31e59c7b1cb2670c4efaa4 \
                    size    144669

dist_subdir         ${name}/${version}_1
```
This is effectively a stealth update ... need to also test on `port` before the `extract.rename` changes

michaelld:
`xxhash` works as-is now with the GH PG changes ... no changes required!  ... need to also test on port before the extract.rename changes

michaelld:
Given the 3 out of 4 no-longer-necessary changes to this PR, I'm going to close it & start individual ones for those changes instead

jmroot:
It's not clear to me why lmdb needs any changes, can you elaborate?

michaelld:
> It's not clear to me why lmdb needs any changes, can you elaborate?

Ah ... you are correct! I had removed `github.tarball_from archive` during testing and not put it back. With that back & the GH PG changes, `lmdb` works as-is now. Yay!

PR #17234, opened 2023-01-07
fpm: new port, Fortran package manager, and new PG for it

barracuda156:
#### Description

New port: https://fortran-lang.github.io/fpm/index.html
New PG to use it to build Fortran packages.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/fortran-1.0.tcl
+++ b/_resources/port1.0/group/fortran-1.0.tcl
@@ -1,0 +1,84 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+#
+# This PortGroup provides support for building Fortran packages.
+#
+# Usage:
+# PortGroup         fortran 1.0
+
+PortGroup           compiler_blacklist_versions 1.0
+PortGroup           compilers 1.0
+PortGroup           github 1.0
+
+default categories  "fortran"
+
+depends_build-append \
+                    port:fpm
+
+compiler.blacklist-append \
+                    *gcc-4.*
+
+global os.platform os.arch
+platform darwin powerpc {
+    compiler.blacklist-append *clang*
+}
+
+compilers.choose    fc f90
+compilers.setup     require_fortran -g95
+
+# For now, avoid universal:
+default universal_variant no
+
+use_configure       no
+
+pre-build {
+    # Make gfortran symbolic link
+    xinstall -m 0755 -d ${workpath}/bin
+    ln -s ${configure.fc} ${workpath}/bin/gfortran
+
+    build.env-append \
+        PATH=$env(PATH):${workpath}/bin \
+        FC=${configure.fc}
+}
+
+build.cmd           ${prefix}/bin/fpm install
+build.target
+
+options             fortran.profile
+global              prefix
+default profile     release
+build.cmd-append    --prefix="${workpath}${prefix}" --profile="${profile}"
+
+global name
+destroot {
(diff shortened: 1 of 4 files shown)


barracuda156:
@herbygillot Could you review this, please, when possible?

devel/fpm/Portfile
```diff
@@ -0,0 +1,55 @@
...
+compilers.setup     require_fortran
+
+use_configure       no
+
+build.cmd           ${worksrcpath}/install.sh
+build.target
+build.cmd-prepend   FPM_CFLAGS="${configure.cflags} [get_canonical_archflags cc]" \
+                    FPM_LDFLAGS="${configure.ldflags} [get_canonical_archflags ld]"
+
+destroot            {}
```
herbygillot, line 55: So it looks like this Portfile is installing straight into the destroot path during the `build` phase, and so doesn't need to do anything during the `destroot` phase.  Is there any way that this can be changed?

Does `install.sh` only copy files? If that's the case, then it will make more sense to have an empty 'build' phase and copy files in the `destroot` block.

barracuda156: The thing is that `$BOOTSTRAP_DIR/fpm install --compiler "$FC" --flag "$FFLAGS" --prefix "$PREFIX"` both builds a full version of `fpm` and installs it into specified prefix. Perhaps this can be moved into destroot stage, but that gonna be a part of building. In either case stages are not neat – but that is the case with quite a number of ports, in fact.

Let me try modifying the script to see if doing `fpm install` in destroot works. Otherwise perhaps prefix can be redefined and destroot implemented manually.

herbygillot: Well if it's actually building the software, then it should be left as-is, in the `build` phase.

The problem is that things are being installed into the `destroot` before the `destroot` phase gets run, where accessory files are installed before the software is to be staged into the `destroot`.  I guess if you see this in other ports, we can leave it, but it's not optimal.

barracuda156: @herbygillot Once CI pass, should be good, I think.

barracuda156: @herbygillot Okay, now hopefully everything works as supposed. Took a bit to figure out how to make it correctly in the PG.

_resources/port1.0/group/fortran-1.0.tcl
```diff
@@ -0,0 +1,50 @@
...
+build.cmd           ${prefix}/bin/fpm install
+build.target
+
+options             fortran.profile
+global              prefix
+default profile     release
+build.cmd-append    --prefix="${workpath}${prefix}" --profile="${profile}"
+
+destroot {
+    move ${workpath}${prefix} ${destroot}${prefix}
```
herbygillot, line 49: One lingering concern is whether this `move` will overwrite files already in place.  Pre-destroot, MacPorts installs certain core files into the destroot prefix before running the port-specific `destroot` block or command. So it could be that all gets blown away by this `move` command.

barracuda156: Ok, so what gonna be the correct way? We do not know exactly which files gonna be installed, since different ports may install different components. Use foreach with copy and if fileexists?

herbygillot: `fs-traverse` could be an option here

barracuda156: > `fs-traverse` could be an option here

Possibly I have implemented it :)

barracuda156:
@herbygillot Thank you!

herbygillot:
Actually thank you as well for your efforts on this

PR #17235, opened 2023-01-07
perlN: add supported platforms

kencu:
this is kind of a new process, and the perl ports are more complicated than usual, but this seems to do the right thing.

As far as I can see, you have to actually run a portindex on the modified port file for it to actually pick up the platform changes in the subports.

### DIFF
--- a/lang/perl5/Portfile
+++ b/lang/perl5/Portfile
@@ -216,6 +216,11 @@ foreach {perl5.v perl5.subversion perl5.revision perl5.rmd160 perl5.sha256 perl5
             }
         }
 
+        # https://trac.macports.org/ticket/66478
+        if {${perl5.major} < 5.28} {
+            platforms           {darwin < 22} freebsd linux
+        }
+
         post-configure {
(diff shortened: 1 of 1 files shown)


jmroot:
Interesting that it works on macOS 11 and 12 but not 13. I guess maybe Ventura removed support for `SYSTEM_VERSION_COMPAT`?

lang/perl5/Portfile
```diff
@@ -55,6 +55,11 @@ foreach {perl5.v perl5.subversion perl5.revision perl5.rmd160 perl5.sha256 perl5
         depends_lib-append  port:db48 \
                             port:gdbm
 
+        # https://trac.macports.org/ticket/66478
+        if {${perl5.major} < 5.28} {
+            platforms           {darwin < 22} freebsd linux
```
jmroot, line 60: It might make sense to put this with the related code around line 214/219. Up to Mojca though.

kencu: done

PR #17289, opened 2023-01-12
Qt6: Clean up unused dependencies

TellowKrinkle:
#### Description

Follow-up to #16735, extracting the controversial part so the non-controversial (and more important) part can get merged faster.  See the comments on that for the beginning of the discussion on whether these dependencies are needed.

(Built on #16735, so will have its changes until that gets merged.  Just look at the last commit.)

- Remove dependencies that are only needed by other subpackages
- Remove hunspell which is only needed by qtvirtualkeyboard, which isn't currently included in the supported qt6 subpackages

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 13.0 22A380 arm64
Xcode 14.1 14B47b

###### Verification
Have you

### DIFF
--- a/aqua/qt6/Portfile
+++ b/aqua/qt6/Portfile
@@ -601,13 +601,6 @@ foreach {module module_info} [array get modules] {
                 }
             }
 
-            if { [variant_isset universal] } {
-                pre-fetch {
-                    ui_warn "Multiple architectures is not a Reference Configuration for Qt."
-                    ui_warn "See https://doc.qt.io/qt-6/supported-platforms.html#reference-configurations"
-                }
-            }
-
             minimum_developerversions 18 9
 
             # see https://trac.macports.org/ticket/63805#comment:13
@@ -750,9 +743,9 @@ foreach {module module_info} [array get modules] {
                     move -force ${worksrcpath}/mkspecs-save/common/${conf} ${worksrcpath}/mkspecs/common/${conf}
                 }
                 foreach spec {macx-clang macx-g++ macx-icc} {
-                        move -force ${worksrcpath}/mkspecs-save/${spec}/qmake.conf ${worksrcpath}/mkspecs/${spec}/qmake.conf
-                    }
+                    move -force ${worksrcpath}/mkspecs-save/${spec}/qmake.conf ${worksrcpath}/mkspecs/${spec}/qmake.conf
                 }
+            }
 
             # --prefix is not recognized.
             configure.pre_args-delete       --prefix=${prefix}
@@ -821,6 +814,10 @@ foreach {module module_info} [array get modules] {
                     -DCMAKE_CXX_COMPILER_LAUNCHER=${prefix}/bin/ccache \
                     -DCMAKE_OBJC_COMPILER_LAUNCHER=${prefix}/bin/ccache \
                     -DCMAKE_OBJCXX_COMPILER_LAUNCHER=${prefix}/bin/ccache
+            }
+            if {[variant_isset universal]} {
+                configure.post_args-append \
+                    "\"-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64\""
             }
 
             configure.args-append \
@@ -907,6 +904,7 @@ foreach {module module_info} [array get modules] {
             configure.universal_cflags
             configure.universal_cxxflags
             configure.universal_cppflags
+            configure.universal_args
 
             # configure script uses gawk if it can find it,
             #    so require it for consistency
@@ -960,6 +958,12 @@ foreach {module module_info} [array get modules] {
             configure.pre_args-delete   --prefix=${prefix}
             configure.pre_args-append   ${worksrcpath}
 
+            configure.universal_ldflags
+            configure.universal_cflags
+            configure.universal_cxxflags
+            configure.universal_cppflags
+            configure.universal_args
+
             # build using using cmake
             build.dir       ${workpath}/build
             build.cmd       cmake --build ${build.dir}
@@ -989,6 +993,20 @@ foreach {module module_info} [array get modules] {
                 -DCMAKE_OBJCXX_COMPILER=[option configure.objcxx]
             if {[option configure.ccache]} {
                 configure.post_args-append -DQT_USE_CCACHE=1
+            }
+            if {[variant_isset universal]} {
+                configure.post_args-append \
+                    "\"-DCMAKE_OSX_ARCHITECTURES=x86_64;arm64\""
+            } else {
+                # If we don't specify anything, the qtbase universal flags will be used
+                platform i386 {
+                    configure.post_args-append \
+                        "-DCMAKE_OSX_ARCHITECTURES=x86_64"
+                }
+                platform arm {
+                    configure.post_args-append \
+                        "-DCMAKE_OSX_ARCHITECTURES=arm64"
+                }
             }
 
             # determine which variants are to be turned off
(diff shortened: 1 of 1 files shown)


TellowKrinkle:
> From @jhoyt4 on #16735: Referencing how [homebrew builds qt6](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt.rb), they also install assimp, hunspell, jasper, nodejs, tiff, and webp.

Homebrew imports them because it squishes all qt subpackages into one big qt6 package
So if any of the subpackages have the dependency, qt6 depends on it

But here, we split them up into qtbase, qtimageformats, etc
And while e.g. qtimageformats needs jasper, qtbase does not

jhoyt4:
> Homebrew imports them because it squishes all qt subpackages into one big qt6 package So if any of the subpackages have the dependency, qt6 depends on it

Thank you for checking and verifying. 

Once last sanity check shows hunspell, jasper, and webp to be unused and able to be moved to the correct subports
```
qt6 % port contents qt6-qtbase|xargs grep -i -s hunspell
qt6 % port contents qt6-qtbase|xargs grep -i -s jasper  
qt6 % port contents qt6-qtbase|xargs grep -i -s wepb
```
assimp and tiff do get flagged (as you noted before).  

assimp - may be benign (nothing of note in otool as you've previsouly stated - and I'll trust you on looking into the hex.  

```
qt6 % port contents qt6-qtbase|xargs grep -i -s assimp  
Binary file /opt/local/libexec/qt6/lib/QtNetwork.framework/QtNetwork matches
Binary file /opt/local/libexec/qt6/lib/QtNetwork.framework/Versions/A/QtNetwork matches
```

tif is a bit weirder.  It's all over the place, but this appears to be a referencing "tiffany" in QtNetwork
```
qt6 % port contents qt6-qtbase|xargs grep -i -s tiff  
Binary file /opt/local/libexec/qt6/lib/QtGui.framework/QtGui matches
Binary file /opt/local/libexec/qt6/lib/QtGui.framework/Versions/A/QtGui matches
Binary file /opt/local/libexec/qt6/lib/QtNetwork.framework/QtNetwork matches
/opt/local/libexec/qt6/lib/QtNetwork.framework/Versions/A/Headers/6.2.4/QtNetwork/private/qurltlds_p.h:"\xd0\xb8\xd0\xba\xd0\xbe\xd0\xbc.museum\0\xd8\xa7\xd9\x84\xd8\xb3\xd8\xb9\xd9\x88\xd8\xaf\xd9\x8a\xd8\xa9\0tiffany\0\xd0\xb1\xd0\xb8\xd0\xb7.\xd1\x80\xd1\x83\xd1\x81\0"
Binary file /opt/local/libexec/qt6/lib/QtNetwork.framework/Versions/A/QtNetwork matches
Binary file /opt/local/libexec/qt6/plugins/platforms/libqcocoa.dylib matches
```

artkiver:
I haven't tested this yet, but given I love fewer dependencies, the PR title caught my eye and I am looking into doing so momentarily.

However, running:

% port lint --nitpick 

yields:
```

--->  Verifying Portfile for qt6
Warning: Line 249 has trailing whitespace before newline
Error: Line 944 repeats inclusion of PortGroup qt6
Warning: Line 982 should be a newline (after PortGroup)
--->  1 errors and 2 warnings found.

```
249 is easy enough to fix.

Less sure what to do about 944.

Added a newline after PortGroup to 982 and got:


`Error: Line 984 uses compiler.blacklist entry {clang < 1100} which requires the compiler_blacklist_versions portgroup which has not been included`

So less sure about those. I will test it anyway as is, but I was already previously running into some issues with this port.

artkiver:
Also noticing this eyebrow raiser:

 ```
port -v install
--->  Computing dependencies for qt6....
 openssl3
 qt6-qtbase
 qt6-sqlite-plugin
Continue? [Y/n]: 

```

I see that the Portfile does have            

`                depends_lib-append path:lib/libssl.dylib:openssl
`
Which, typically facilitates usage of other TLS libraries (e.g. LibreSSL, not merely OpenSSL) yet that seems to be invoked via

`            default_variants-append +openssl
`

Meanwhile, I am on a system with libressl-devel installed

```
% openssl version
LibreSSL 3.7.0
```

So, it presumably *shouldn't* be asking to install openssl3 as a dependency?

**Comment by artkiver** [comment]
port -v install progressed but ended here: 


```

Error: Failed to build qt6-qtbase: command execution failed
Error: See /opt/local/var/macports/logs/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_aqua_qt6/qt6-qtbase/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets if you believe there
is a bug.
Error: Processing of port qt6 failed
```
[main.log](https://github.com/macports/macports-ports/files/10406514/main.log)

Attaching main.log for reference. My apologies if that is a rabbit hole, but thank you for your efforts to attempt to reduce the dependencies of this port!

**Comment by TellowKrinkle** [comment]
> `port lint` things

I think 944 is a false positive, it's because we do the `PortGroup` in an if/else, so while each path sees one copy of `PortGroup`, the linter (which doesn't follow if/else's) sees two

Not sure what's up with 982 either, since there's clearly a `PortGroup compiler_blacklist_versions 1.0` right before it...

> `default_variants-append +openssl`

That's just making it so `+openssl` is the default (and you have to opt out with `-openssl` if you don't want it)

> Failed compile
> `/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_aqua_qt6/qt6-qtbase/work/qtbase-everywhere-src-6.2.4/src/plugins/tls/openssl/qsslsocket_openssl_symbols_p.h:241:22: error: unknown type name 'OPENSSL_STACK'`

Looks like all the errors are openssl related.  Did you let it install openssl3 or are you trying to compile against libressl?

It seems to be finding its openssl headers in `/opt/local/include/openssl`, if you want to check which package those came from

**Comment by artkiver** [comment]
> > `port lint` things
> 
> I think 944 is a false positive, it's because we do the `PortGroup` in an if/else, so while each path sees one copy of `PortGroup`, the linter (which doesn't follow if/else's) sees two
> 
> Not sure what's up with 982 either, since there's clearly a `PortGroup compiler_blacklist_versions 1.0` right before it...
> 
> > `default_variants-append +openssl`
> 
> That's just making it so `+openssl` is the default (and you have to opt out with `-openssl` if you don't want it)

Ah, indeed: 

` % sudo port -v install qt6 -openssl
`

(bunch of stuff omitted)

```
--->  Installing qt6 @6.2.4_0
--->  Activating qt6 @6.2.4_0
x ./
x ./+PORTFILE
x ./+CONTENTS
x ./+DESC
x ./+STATE
x ./opt/
x ./+COMMENT
x ./opt/local/
x ./opt/local/share/
x ./opt/local/share/doc/
x ./opt/local/share/doc/qt6/
x ./opt/local/share/doc/qt6/README.txt
--->  Cleaning qt6
--->  Removing work directory for qt6
--->  Updating database of binaries
--->  No broken files found.
--->  No broken ports found.
```
Yay!

So, thanks for that suggestion. Honestly, I think I may have also tried it with openssl3 installed, but I don't remember now. I didn't spend a lot of time looking at it as I got lost down _this_ rabbit hole: https://github.com/macports/macports-ports/pull/17298#pullrequestreview-1246798349 and I should probably go back to focusing on that and seeing about implementing the improvements @neverpanic mentioned.

Regardless, relative to my earlier efforts to attempt to build qt6: your Portfile functioned! Which is more than my previous attempts.

MarcusCalhoun-Lopez:

PR #17296, opened 2023-01-12
cargo: do not use extract.rename if available

michaelld:
#### Description

cargo: do not use extract.rename if available

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.7.3 20G1102 x86_64
Xcode 12.3 12C33

###### Verification 
Have you

### DIFF
--- a/devel/cargo/Portfile
+++ b/devel/cargo/Portfile
@@ -22,6 +22,9 @@ long_description    Cargo downloads your Rust project’s dependencies and \
 
 homepage            https://crates.io
 github.tarball_from archive
+if {[exists extract.rename]} {
+    extract.rename  no
+}
 
 installs_libs       no
 


MarcusCalhoun-Lopez:
I have no strong feelings about this.
However, just out of curiosity, if `extract.rename yes` is the default, is there a compelling reason to change it?

If we are going to change `extract.rename`, aren't we well past the two weeks where we have to check for its existence?

michaelld:
`extract.rename` is part of `base` but not a release, so we have to check for its existence until it is part of a release & then the existence check can be removed (though it doesn't hurt to keep it around)

michaelld:
This PR might be unnecessary now with this commit on base since the PR was opened: https://github.com/macports/macports-base/commit/d43dd953c6b0ff9e9b0bb0c18c8e6c1da17f9617 ... I will check & see and if not then close the PR

michaelld:
Indeed that tweak to base does away with the need for this specific PR ... closing

PR #17298, opened 2023-01-13
rpki-client: update to version 8.2.

artkiver:
* closes: https://trac.macports.org/ticket/66677
 * add variants for libressl, libressldevel, openssl (requires libretls)
 * add maintainer
 * added default_variants-append +libressldevel for the CI buildbots

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 13.1 22C65 arm64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/net/rpki-client/Portfile
+++ b/net/rpki-client/Portfile
@@ -7,13 +7,13 @@ PortGroup           legacysupport 1.0
 legacysupport.newest_darwin_requires_legacy 15
 
 name                rpki-client
-version             7.4
-revision            2
+version             8.2
+revision            0
 
 categories          net
 platforms           darwin
 license             ISC
-maintainers         nomaintainer
+maintainers         {@artkiver gmail.com:artkiver} openmaintainer
 
 description         OpenBSD RPKI Validator
 long_description    rpki-client is a FREE, easy-to-use implementation \
@@ -26,10 +26,27 @@ homepage            https://www.rpki-client.org/index.html
 master_sites        https://cdn.openbsd.org/pub/OpenBSD/rpki-client/ \
                     https://ftp.openbsd.org/pub/OpenBSD/rpki-client/
 
-checksums           rmd160  74c8efaba732d3b5619c8c3a56fbc77c2e33c624\
-                    sha256  efd336137707ad73fa51f8480cff686c21fea7086619432069c1a02553b68d57\
-                    size    439919
+checksums           rmd160 a8c0d3783e765a5bea2d7f6652d015da4690a5f6 \
+                    sha256 dc0d19679b57ae657b92d21730b1678823974d43300faa8c24ee155c1e2b1d64 \
+                    size   475141
 
 depends_lib-append  port:expat \
-                    port:libretls \
+
+variant libressl description "use LibreSSL Stable" {
+depends_lib         port:libressl
+conflicts           openssl libretls libressl-devel
+}
+
+variant libressldevel description "use LibreSSL Development" {
+depends_lib         port:libressl-devel
+conflicts           openssl libretls libressl
+
+}
+
+variant openssl description "use OpenSSL" {
+depends_lib         port:libretls \
                     port:openssl
+conflicts           libressl libressl-devel
+}
+
+default_variants-append +libressldevel
--- a/net/rpki-client/Portfile
+++ b/net/rpki-client/Portfile
@@ -32,9 +32,8 @@ checksums           rmd160 a8c0d3783e765a5bea2d7f6652d015da4690a5f6 \
 
 depends_lib-append  port:expat \
 
-variant libressl description "use LibreSSL Stable" {
-depends_lib         port:libressl
-conflicts           openssl libretls libressl-devel
+variant libressl conflicts libressldevel openssl description {use LibreSSL Stable} {
+    depends_lib-append port:libressl
 }
 
 variant libressldevel description "use LibreSSL Development" {


net/rpki-client/Portfile
```diff
@@ -26,10 +26,27 @@ homepage            https://www.rpki-client.org/index.html
...
+                    sha256 dc0d19679b57ae657b92d21730b1678823974d43300faa8c24ee155c1e2b1d64 \
+                    size   475141
 
 depends_lib-append  port:expat \
-                    port:libretls \
+
+variant libressl description "use LibreSSL Stable" {
+depends_lib         port:libressl
+conflicts           openssl libretls libressl-devel
+}
```
neverpanic, line 38: Your variants also need to conflict with each other (since only one of them can be active). You can do that with `conflicts` on the variant line, e.g.

```suggestion
variant libressl conflicts libressldevel openssl description {use LibreSSL Stable} {
    depends_lib-append port:libressl
}
```

Using `depends_lib` on its own will overwrite the expat dependency, so this must be `depends_lib-append`.

Usually, we would add a configure flag to make sure that the port will only use the variant's TLS implementation, and not rely on dependency auto-detection. In this particular case, though, it seems as if all of openssl, libressl and libressl-devel install an `openssl.pc` pkg-config file, so that argument would always be `--with-openssl=openssl`, so we might as well skip it. We should, however, specify `configure.args-append --without-libretls` for those variants that will not be using libretls.

If this port compiled against libressl will run with libressl-devel (i.e., if libressl and libressl-devel are sufficiently binary compatible), you may be able to remove the libressldevel variant and instead use a `path:lib/libtls.dylib:libressl` dependency, which could then be transparently be fulfilled by libressl-devel.

In any case, you should probably not set the default variant to libressldevel – we'd expect users to commonly have either openssl or libressl installed, but probably not the -devel port.

`default_variants-append` is a special case and does the same thing as `default_variants`, iirc, so you might as well use the shorter version. Note that you must add code to only set a default variant if no conflicting variant is active, though, i.e.

```tcl
if {![variant_isset openssl]} {
  default_variants +libressl
}
```

And, finally, you can provide a nicer user experience by pre-selecting the default variant that matches what the user has already installed on their system. The gnupg2 port does this with pinentry and pinentry-mac for example. Try

```tcl
if {![catch {registry_active openssl}] && ![variant_isset libressl]} {
  # openssl is installed and the libressl variant isn't manually requested, default to openssl
  default_variant +openssl
} elseif {(![catch {registry_active libressl}] || ![catch {registry_active libressl-devel}]) && ![variant_isset openssl]} {
  # either libressl or libressl-devel are already installed, and the openssl variant is not explicitly requested, default to libressl
  default_variant +libressl
} elseif {![variant_isset openssl]} {
  # neither openssl nor libressl-devel are installed, and the openssl variant isn't selected, so libressl should be the default
  default_variant +libressl
}
```

neverpanic (changes requested):

net/rpki-client/Portfile
artkiver: Hah, I am a little less sure how GitHub's "commit suggestion" functions, but reading this was very informative. Thank you. I am guessing I may need to amend the PR, but will re-read this a few more times and am focused on some other things for a bit first.  Regardless, what a wonderful learning experience. Thank you for the wisdom @neverpanic !

herbygillot: As per what @neverpanic is saying, you need to define a `conflicts` for each variant. So this isn't complete simply adding the `conflicts` to the `libressl` variant, but you also need to do so to the `libressldevel` and `openssl` variant definitions as well.

`variant libressldevel description "use LibreSSL Development" {` -> `variant libressldevel conflicts libressl openssl description "use LibreSSL Development" {` 

`variant openssl description "use OpenSSL" {` -> `variant openssl conflicts libressl libressldevel description "use OpenSSL" {`

artkiver:
OK, I think I have @neverpanic's modifications incorporated and tested locally and looking pretty good!

port install will default to installing libressl
port install +openssl functions OK and will add libretls
port install +libressl functions OK too
Moreover, if libressl-devel is installed, it seems to acknowledge that too.

I'm not sure what other tests may be worthwhile (boringssl maybe?) but for the time being it seems as if it is definitely better than my earlier attempts for facilitating compatibility with different TLS libraries and hopefully won't pester the user if they already have one previously installed, and if they don't have any installed it will provide an acceptable stable default.

Unfortunately, whatever git wizardry (Herby has been helpful in this in the past but I am still running into issues it seems) is required to get my local branch happy with my GitHub branches still seems as if it's not going smoothly, so I am, regrettably, going to close this PR and re-submit with a cleaner branch and hope that goes well.

*fingers crossed*

(After all, I haven't seen how this new attempt runs through the CI buildbots yet)

net/rpki-client/Portfile
artkiver: > As per what @neverpanic is saying, you need to define a `conflicts` for each variant. So this isn't complete simply adding the `conflicts` to the `libressl` variant, but you also need to do so to the `libressldevel` and `openssl` variant definitions as well.
> 
> `variant libressldevel description "use LibreSSL Development" {` -> `variant libressldevel conflicts libressl openssl description "use LibreSSL Development" {`
> 
> `variant openssl description "use OpenSSL" {` -> `variant openssl conflicts libressl libressldevel description "use OpenSSL" {`

Ah, understood.

I ended up removing the libressldevel variants since @neverpanic suggested the path:lib/libtls.dylib:libressl approach which seems as if it successfully condensed down a +libressl variant to function with either libressl or libressl-devel installed (though this will default to stable libressl, which seems like a saner choice as suggested).

I hope I added the conflicts correctly for the two remaining variants?

I tried to test it pretty exhaustively with both openssl and libressl (and libressl-devel for good measure) and it seemed OK locally and I am relieved that the new PR's CI buildbot checks appear to have gone OK as well:

https://github.com/macports/macports-ports/pull/17311/files

Of course, there may still be room for improvement, but that is a lot better than my initial PR efforts and hopefully covers a lot more bases than 7.4 did as well?

I'm still somewhat consternated with the git/GitHub workflow; I tried your branching suggestions again but while some of the efforts would see that I had pushed again, it didn't actually reflect the changes I had --amended in the commits and I am not sure where I am fumbling that still.

Regardless, thank you for both taking a gander at this and providing useful perspective!🙏

PR #17386, opened 2023-01-19
ruby30: Fix build on 10.4, 10.5

fhgwright:
Also enables universal builds, where possible.

This incorporates the new patches from ruby27, all of which are relevant to ruby30.

All source patches are now in a single patchfile derived from the ruby fork at github.com/fhgwright/ruby.  The changes are much more readable in the git repo.  A second patchfile (also derived from the GitHub fork) patches 'configure' to reflect the source changes without rerunning autoconf.

Although this update mostly only affects previously failing builds, it also changes the i386 context handling in successful builds as well, therefore requiring a revbump.

In theory, the installed content should be unchanged on successful non-i386 builds.  In practice, there appear to be nonreproducible build issues affecting libruby.3.0-static.a and many .gemspec files, though there are presumably no effective differences.

Narrows compiler blacklisting to the one offending case.

Universal builds are now enabled using ruby's built-in universal mechanism (as in ruby27).  Compiler issues currently prevent building universal on 10.4 and 10.5, though it's believed that they would work otherwise.

Fixes a couple of typos in a Portfile comment.

Cleans up long lines in Portfile comments.

Also adds self as co-maintainer.

TESTED:
Built successfully -universal on OSX 10.4-10.5 ppc, 10.4-10.6 i386, 10.6-10.15 x86_64, and 11.x-13.x arm64.  Also successfully built +universal in all 10.6+ cases except 10.14-10.15 (SDK restriction).

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 11.7.2 20G1020, arm64, Xcode 13.2.1 13C100
macOS 12.6.2 21G320, arm64, Xcode 14.2 14C18
macOS 13.1 22C65, arm64, Xcode 14.2 14C18

###### Verification 
Have you

- [KNOWN FAIL] tried existing tests with `sudo port test`?


barracuda156:
@fhgwright Thank you, I will try that on 10.5.8 and 10.6 for ppc. If universal works on 10.5.8 for ppc+ppc64, gonna be great.
Some of your changes may be worth upstreaming perhaps?

BTW, why do you use `ppc*-darwin8*` instead of `ppc*-darwin*`? We can build for PPC at least on 3 major versions of Darwin.

fhgwright:
@barracuda156 
> @fhgwright Thank you, I will try that on 10.5.8 and 10.6 for ppc. If universal works on 10.5.8 for ppc+ppc64, gonna be great.

I wouldn't bother with universal just for ppc+ppc64, since it's very rare that building for ppc64 is useful.  On most architectures, 64-bit builds are slower due to the poorer cache locality caused by the larger pointers, and should only be used if either a really large address space is needed, or _lots_ of 64-bit integer arithmetic is being used.  There's even a "ppc970" target (albeit not well supported) that's a 32-bit build optimized for the G5.

The one exception to this rule is x86, since i386 is so register-starved that a typical program tends to run about 15% faster on x86_64 than i386, using the same source code and the same CPU.

Because of the above, I don't have ppc64 in universal_archs anywhere, since it's fairly unimportant as a test case.  Of course if anyone were ever to add the missing register dump code for ppc, it would need to be tested for both ppc and ppc64, possibly by temporarily removing the realpath() fix. :-)

> Some of your changes may be worth upstreaming perhaps?

In principle, yes, though I wouldn't try to bug upstream until there are suitable changes for the latest code, and only then trying to get them to accept changes for older but still updated versions.

> BTW, why do you use `ppc*-darwin8*` instead of `ppc*-darwin*`? We can build for PPC at least on 3 major versions of Darwin.

That's explained in the commit messages in my fork, as referenced by the comments in the Portfile.  Patchfiles make a terrible primary representation of changes.

The relevant code relates to some context stuff, which can be handled in one of three ways, in order of preference:

1. CPU-specific code provided by Ruby
2. The ucontext mechanism provided by the OS.
3. The generic C "copy" mechanism provided by Ruby.

The default is option 2, which is CPU-independent (at least from the application perspective), but it doesn't work prior to 10.5.  Option 1 doesn't exist at all for ppc, so on 10.4 it needs to select option 3 (hence the `ppc*-darwin8*`).  Option 1 exists for x86, but they neglected to include a case for i386 (as opposed to x86_64), so I added that one, for all OS versions.  This affects all i386 builds (including ones that previously worked), hence the revbump.

Since that logic is in the configure phase, it can't be CPU-specific for universal builds, and the CPU type it uses is "universal".  That has to use "copy" on 10.4 as the only fully compatible choice, but the default "ucontext" is acceptable on 10.5+.

I think they did a major overhaul of some of that stuff in 3.1, which might be good in principle but also presents new challenges.

As far as 10.6 ppc goes, AFAIK that's never been a supported combination, and IIRC, Xcode 3.2 (the last version for 10.6) has a mixture of universal and non-universal binaries such that some things work on ppc and some don't.  I expect that anyone running 10.6 on ppc is in for a world of hurt, and it's not worth my trying to do the same just to provide testing for that oddball case.

barracuda156:
@fhgwright 
> Of course if anyone were ever to add the missing register dump code for ppc, it would need to be tested for both ppc and ppc64, possibly by temporarily removing the realpath() fix. :-)

Could you elaborate a bit? I may be interested to work on it, but I need to understand more.

> The default is option 2, which is CPU-independent (at least from the application perspective), but it doesn't work prior to 10.5.

Then why not use `ucontext` for 10.5 ppc/ppc64, leaving  `copy` to 10.4? Or do I miss something?
P. S. You reminded me about `copy`. Now it is replaced (?) by `pthread`: https://trac.macports.org/ticket/64906#comment:13

> I think they did a major overhaul of some of that stuff in 3.1, which might be good in principle but also presents new challenges.

_If you refer to PPC, it was me!_ Ha-ha
But with great help of three Ruby developers and Iain.

> As far as 10.6 ppc goes, AFAIK that's never been a supported combination, and IIRC, Xcode 3.2 (the last version for 10.6) has a mixture of universal and non-universal binaries such that some things work on ppc and some don't. I expect that anyone running 10.6 on ppc is in for a world of hurt, and it's not worth my trying to do the same just to provide testing for that oddball case.

Well, 10.6.8 is standard and supports ppc32 fully.
(I can explain quite a bit on Xcode versions of 3.2, since I actually tested multiple, but perhaps you are not interested/it is irrelevant. Suffice to say that Xcode tools from 10A190 are fully compatible with PPC and work nearly identically to 10.6.8.)

fhgwright:
@barracuda156 
> @fhgwright
> 
> > Of course if anyone were ever to add the missing register dump code for ppc, it would need to be tested for both ppc and ppc64, possibly by temporarily removing the realpath() fix. :-)
> 
> Could you elaborate a bit? I may be interested to work on it, but I need to understand more.

The crash code includes CPU-specific register dumping, but simply omits that part where it doesn't have the CPU-specific code.  It never had it for ppc, hence no registers in the crash dump.  It shouldn't be difficult to add, but it would need to be tested on both ppc and ppc64.  I believe the general ppc concept is that most of the registers have the same names on ppc and ppc64, but with different sizes.  Though I think there are a few that are always 32-bit, and some variations in which are present.  I have the necessary docs, but wouldn't implement it without being able to test both cases, and my Quad G5 is inoperable until I get ambitious enough to do an LCS overhaul on it.  And there's also the question of how far one wants to depart from upstream for things that aren't bugs in the usual sense.

Interestingly enough, although arm64 support began with 2.6, they didn't get around to adding the arm64 register dump until 3.1.

> > The default is option 2, which is CPU-independent (at least from the application perspective), but it doesn't work prior to 10.5.
> 
> Then why not use `ucontext` for 10.5 ppc/ppc64, leaving `copy` to 10.4? Or do I miss something?

You missed the part about `ucontext` being the default.  What you're saying above is exactly what happens.

 P. S. You reminded me about `copy`. Now it is replaced (?) by `pthread`: https://trac.macports.org/ticket/64906#comment:13

Well, `copy` vs. `pthread` seems like apples 'n oranges, but I haven't dug into what they actually mean by that.  I did run afoul of that thread id issue when getting llvm-7.0 to build on 10.5, but it looked like it was only for debugging and therefore OK to disable when it doesn't work.

> > I think they did a major overhaul of some of that stuff in 3.1, which might be good in principle but also presents new challenges.
> 
> _If you refer to PPC, it was me!_ Ha-ha But with great help of three Ruby developers and Iain.

I was mainly referring to the universal handling, which is what I noticed when looking briefly at 3.1 at one point.  The most important thing in universal builds is to avoid doing anything CPU-specific in the configure phase.  That includes indirect dependencies like sizeof(time_t) (which they should have fixed at the same time as ino_t, but didn't).

barracuda156:
@fhgwright This is off-topic, but just FYI: some people on MacRumors made air-cooled mode to the Quad: https://forums.macrumors.com/threads/quad-g5-air-cooled-dxf-included.2229488 (and a couple of other threads there).

barracuda156:
@fhgwright `ruby30` from your branch builds fine on 10.6 PPC. All good.
The only config we have not tested is ppc64, but I guess we can return to it later. (I will not be able to deal with 10.5.8 ppc64 any soon.)

P. S. There are some warnings (which are not specific to your branch or `ruby30`, I get the same with the latest `ruby-devel`). They should be fixable, however solutions that I found did not work. Like `lgamma_r` is defined in system headers, requiring `_REENTRANT` – however that does not work, it is still implicitly defined. Do you know how we could fix these?
```
dir.c:157:15: warning: implicit declaration of function 'fgetattrlist'; did you mean 'getattrlist'? [-Wimplicit-function-declaration]
  157 |     int ret = fgetattrlist(dirfd(dirp), &al, attrbuf, sizeof(attrbuf), 0);
      |               ^~~~~~~~~~~~
      |               getattrlist

math.c: In function 'math_lgamma':
math.c:927:17: warning: implicit declaration of function 'lgamma_r'; did you mean 'lgammal'? [-Wimplicit-function-declaration]
  927 |     v = DBL2NUM(lgamma_r(d, &sign));
      |                 ^~~~~~~~
      |                 lgammal
```

fhgwright:
> @fhgwright `ruby30` from your branch builds fine on 10.6 PPC. All good. The only config we have not tested is ppc64, but I guess we can return to it later. (I will not be able to deal with 10.5.8 ppc64 any soon.)

No surprise there.  In general, the fixes either related to the CPU type or the OS version, not the combination thereof, and I'd already tested all OS versions and all CPU types (except ppc64).

As I mentioned previously, there's not much motivation to build for ppc64 even on a G5, so I wouldn't worry about that case too much.  Though if you look closely at the fixes, you'll notice that one was needed only for _32-bit_ PPC. :-)

> P. S. There are some warnings (which are not specific to your branch or `ruby30`, I get the same with the latest `ruby-devel`). They should be fixable, however solutions that I found did not work. Like `lgamma_r` is defined in system headers, requiring `_REENTRANT` – however that does not work, it is still implicitly defined. Do you know how we could fix these?

In general, it's rare to find any open-source software that doesn't get a ton of compiler warnings (and Ruby is no exception), so I don't usually worry about those when just fixing broken builds on particular OS versions or CPU types.  Even in cases where they represent real bugs, the bugs are unlikely to be specific to the local fixes.  One usually tends to see more warnings with newer compilers, but there are exceptions to that.

barracuda156:
> No surprise there. In general, the fixes either related to the CPU type or the OS version, not the combination thereof, and I'd already tested all OS versions and all CPU types (except ppc64).

Yes, I agree, and I didn’t expect anything but successful build on native PPC, but wanted to test just for the record.
(By the way, I was not able to build any Ruby in Rosetta for some reason. Not recent ones at least. At some point the build just freezes and nothing happens then.)

> not much motivation to build for ppc64 even on a G5

I certainly am interested in ppc64; however, reality is that Macports is largely broken for ppc64 or ppc+ppc64 universal. And having just _some_ ports built for ppc64 becomes an annoyance to maintain.
(Plus the technical difficulty – Rosetta does not support ppc64, so I can only work on it on a PowerMac, which means if I am out of station, no ppc64 anymore.)

> Even in cases where they represent real bugs, the bugs are unlikely to be specific to the local fixes.

I have encountered that `lgamma_r` issue already, with `Stan`. There developers just advised to use Boost implementation instead. But I wonder why native does not work, eventhough it should.

fhgwright:
> > No surprise there. In general, the fixes either related to the CPU type or the OS version, not the combination thereof, and I'd already tested all OS versions and all CPU types (except ppc64).
> 
> Yes, I agree, and I didn’t expect anything but successful build on native PPC, but wanted to test just for the record.

If you're suggesting that my tests weren't on native PPC, that's incorrect.  I have a G4 Mini set up to dual-boot 10.4 and 10.5.  I don't do the unofficial 10.6 case, and in fact, the Snow Leopard install disc (10.6.3) won't even boot on that machine.

> (By the way, I was not able to build any Ruby in Rosetta for some reason. Not recent ones at least. At some point the build just freezes and nothing happens then.)

I haven't tried building anything with Rosetta or Rosetta2.  I don't currently have coverage for 11.x-13.x x86_64, but I can build for those OS versions on arm64, and earlier versions on x86 (or PPC, where applicable).

Targeting x86_64 via Rosetta2 might be useful for Ruby <2.6 on arm64, if a native fix is too hard.

> > not much motivation to build for ppc64 even on a G5
> 
> I certainly am interested in ppc64; however, reality is that Macports is largely broken for ppc64 or ppc+ppc64 universal. And having just _some_ ports built for ppc64 becomes an annoyance to maintain. (Plus the technical difficulty – Rosetta does not support ppc64, so I can only work on it on a PowerMac, which means if I am out of station, no ppc64 anymore.)

The need for ppc64 is sufficiently rare that it might make sense to make it a variant on the few ports where it's useful.

> > Even in cases where they represent real bugs, the bugs are unlikely to be specific to the local fixes.
> 
> I have encountered that `lgamma_r` issue already, with `Stan`. There developers just advised to use Boost implementation instead. But I wonder why native does not work, eventhough it should.

Chasing those sorts of things often involves a lot of digging through the config flag mess, and looking at lots of -E output.

barracuda156:
> If you're suggesting that my tests weren't on native PPC, that's incorrect.

No, I meant that a failure on _emulated_ PPC is expected (given my experience). Since that is one of the environments I normally use and test, I made this clarification.

lang/ruby30/Portfile
```diff
@@ -50,18 +50,59 @@ depends_skip_archcheck pkgconfig
...
+#-------------------------------------------------------------------------------
 compiler.blacklist-append \
-                    { clang < 901 }
+                    { clang >= 900 < 901 }
 
 compiler.thread_local_storage yes
 
+# At least one version of clang doesn't offer cstdbool (introduced in C++11)
+# with -stdlib=libstc++ but still reports as C++14.  This causes an include
+# failure in include/ruby/internal/stdbool.h whel built as C++.  Setting the
```
michaelld, line 94: typo: `whel` -> `when`

michaelld (changes requested):
Thanks for all of your work here @fhgwright ! This is a complicated set of changes, and so nontrivial to determine functionality / viability. The QA/CI checks pass on macOS 11 and 12, and the vast majority of the changes are clearly for 10.4 and 10.5 PPC -- and thus not the core MacPorts user group. Hence, I think it's safe to merge this and see what happens, once the typo is fixed.

lang/ruby30/Portfile
fhgwright: I guess it's two steps forward and one back in the comment typo department. :-)

The changes are easier to read in my fork than in the patchfiles.

I'll force-push twice - once for the rebase and once for the real change, since this PR is currently 578 commits behind.

fhgwright: OK, the second forced push is the typo correction.

michaelld (commented):
LGTM ; thanks for the quick change!

michaelld:
Given that the port in this PR is openmaintainer and the listed maintainer has not responded, and the changes here are complicated but broadly target just OSX 10.4 and 10.5 for PPC, I'm going to merge this with maintainer timeout & we'll see what happens

barracuda156:
@michaelld It certainly does not hurt merging this. If any additional fixes will be needed (like, for ppc64), we can take care of that in another PR after testing.

@fhgwright Thank you for working on this!
Besides, does `ruby-devel` work for you on PPC?

PR #17426, opened 2023-01-22
misc noarch ports: update platforms

jmroot:
#### Description

These ports all set `supported_archs noarch`, and appear (from my inspection of the Portfiles and lists of installed files) to install the same files regardless of OS version. This PR updates the `platforms` option so that we don't build a separate archive per macOS version.

If you know that this is not correct for your port or even have significant doubts, please say so and I will remove it from this PR. This change is an optimisation which has the potential to cause incorrect behaviour if applied inappropriately. In case of doubt, it's better for the maintainer to investigate thoroughly before changing this.

###### Type(s)
- [x] enhancement

### DIFF
--- a/databases/lmdbxx-cxx17/Portfile
+++ b/databases/lmdbxx-cxx17/Portfile
@@ -6,10 +6,10 @@ PortGroup           github 1.0
 github.setup        hoytech lmdbxx 1.0.0
 name                lmdbxx-cxx17
 categories          databases devel
-platforms           darwin
+platforms           any
 supported_archs     noarch
 license             public-domain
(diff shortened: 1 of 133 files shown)


aalmiray:
👍 on jreleaser

casey:
Fine for edmv 👍

brucemiller:
latexml should be fine; there are no explicit OS dependences.

michaelld (approved):
LGTM for my 2 ports here

maherg (approved):

pmoura:
Fine for `logtalk`.

PR #17435, opened 2023-01-22
smlnj: fix the build on PPC

barracuda156:
#### Description

Fixes: https://trac.macports.org/ticket/66745

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

macOS 10A190
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/lang/smlnj/Portfile
+++ b/lang/smlnj/Portfile
@@ -92,6 +92,9 @@ configure {
         ${worksrcpath}/config/targets
 }
 
+# error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘ml_val_t’
+configure.cflags-append -std=c99
+
 ### build ###
 build.env           URLGETTER=curl
 build.cmd           ${worksrcpath}/config/install.sh
--- a/lang/smlnj/Portfile
+++ b/lang/smlnj/Portfile
@@ -64,7 +64,8 @@ extract.mkdir       yes
 extract.only        config.tgz
 
 ### patch ###
-patchfiles          patch-config__install.sh-correct-compiler.diff
+patchfiles          patch-config__install.sh-correct-compiler.diff \
+                    patch-_arch.diff
 
 post-patch {
     reinplace "s|@CC@|${configure.cc}|" ${worksrcpath}/config/install.sh
--- a/lang/smlnj/files/patch-_arch.diff
+++ b/lang/smlnj/files/patch-_arch.diff
@@ -1,0 +1,12 @@
+--- config/_arch-n-opsys.orig	2022-07-28 12:55:45.000000000 +0800
++++ config/_arch-n-opsys	2023-01-23 03:30:12.000000000 +0800
+@@ -66,7 +66,9 @@
+       powerpc)
+ 	ARCH=$(pick_arch ppc ppc64)
+ 	case `uname -r` in
++	  8*) OPSYS=darwin;  HEAP_OPSYS=darwin ;; # MacOS X 10.4 Leopard
+ 	  9*) OPSYS=darwin;  HEAP_OPSYS=darwin ;; # MacOS X 10.5 Leopard
++	  10*) OPSYS=darwin;  HEAP_OPSYS=darwin ;; # MacOS X 10.6 Snow Leopard
+ 	  *) exit 1;;
+ 	esac;;
+       i386)


barracuda156:
Works nicely on PPC, I compiled `hamlet` with it: https://github.com/macports/macports-ports/pull/17438

tobypeterson:
Would it be possible to make these patches be a single patch that will apply and work cleanly regardless of platform? I don't love the conditional patches in general. In this case it's even worse; the larger patch explicitly removes a section that is often modified by other patches (e.g, when we have to support darwin 23 before upstream does).

tobypeterson:
I suspect that the small patch would apply cleanly to everything, and you should be able to control the size by passing specific arguments to the build program (maybe `-default 32`).

barracuda156:
@tobypeterson Honestly I do not know how to make it nicely with makefiles which call for physical cpu info: that inevitably gets wrong on Rosetta and breaks the build.
If you could advise a neat fix, I will happily follow through.

Otherwise I am also okay with dropping Rosetta patch and just keeping native PPC.

barracuda156:
Main issue is that `uname` reports x86_64, so even if 32-bit is passed via a flag, arch will be wrong (for Rosetta case).

tobypeterson:
patch-rosetta.diff does three things:
- fiddles SIZE
- same changes as patch-_arch.diff
- deletes i386/arm stuff

SIZE should be controllable via arguments to install.sh (build.args).
Shouldn't need to delete i386/arm stuff, as you should go down the powerpc path regardless.

So my recommendation is to make patch-_arch.diff apply unconditionally, because it shouldn't affect other platforms.

For controlling SIZE... I think I understand what's happening. You're building for powerpc, but on an x86_64 machine, and it's checking uname instead of your build target. For this, I'd do something like:
```
if {${build_arch} eq "x86_64"} {
    build.args-append   -default 64
} elseif {${build_arch} eq "ppc"} {
    build.args-append    -default 32
}
```

tobypeterson:
> Main issue is that `uname` reports x86_64, so even if 32-bit is passed via a flag, arch will be wrong (for Rosetta case).

I think `-default 32` will override, based on the code in `install.sh`:

```
DEFAULT_SIZE=32
case `uname -m` in
  x86_64) DEFAULT_SIZE=64 ;;
esac

# process options
SIZE_OPT=
nolib=false
while [ "$#" != "0" ] ; do
    arg=$1; shift
    case $arg in
      -32) SIZE_OPT=$arg ;;
      -64) SIZE_OPT=$arg ;;
      -default)
	case x"$1" in
	  x32) DEFAULT_SIZE="32"; shift ;;
	  x64) DEFAULT_SIZE="64"; shift ;;
	  x) complain "missing size argument for '-default'" ;;
	  *) complain "invalid size argument for '-default'; should be 32 or 64" ;;
```

barracuda156:
@tobypeterson Thank you! I will test and update the PR.

tobypeterson:
I think you may need to use `${configure.build_arch}` - on my Apple silicon machine, it seems like `${build_arch}` is always arm64, even though this port is trying to build for x86_64.

tobypeterson:
I pushed a change (in the wrong commit, oops) that changes the x86_64 check to use configure.build_arch.

barracuda156:
Thank you, I will deal with this today and drop a clumsy Rosetta patch :)
Sorry for a delay, been very busy with fixing gfortran.

barracuda156:
@tobypeterson Dropping Rosetta patch in favor of passing a flag to build for 32 bit, it fails because of x86 is detected:
```
/usr/bin/gcc-4.2 -m32 -std=c99 -c -g -O2 -D_DARWIN_C_SOURCE -mmacosx-version-min=10.6 -Os -std=c99 -arch ppc -DARCH_X86 -DDSIZE_32 -DOPSYS_UNIX -DOPSYS_DARWIN -DGNU_ASSEMBLER -DDLOPEN -DINDIRECT_CFUNC -I../objs -I../include ../mach-dep/unix-fault.c
../mach-dep/unix-fault.c: In function ‘FaultHandler’:
../mach-dep/unix-fault.c:75: error: ‘struct __darwin_ppc_thread_state’ has no member named ‘__eip’
../mach-dep/unix-fault.c:102: error: ‘struct __darwin_ppc_thread_state’ has no member named ‘__eip’
make[1]: *** [unix-fault.o] Error 1
```

tobypeterson:
Is that building for ppc on an intel machine? In many places, it seems to look at the host instead of the target

barracuda156:
> Is that building for ppc on an intel machine? In many places, it seems to look at the host instead of the target

@tobypeterson I was trying to fix the build on Rosetta (that is, 10.6.8, where physical cpu is Intel, but OS emulates PPC). It needs hacking the script, because `uname` reports Intel.

But let us just drop Rosetta case, I guess. Anyone trying to build on it perhaps can figure out what to change himself.

barracuda156:
@tobypeterson Let us merge it without explicit Rosetta support, I am totally okay with that. Thus changes are minimal on our side, and whomever needs this on Rosetta, it will be one-line fix.

tobypeterson:
Sorry for the delay!

barracuda156:
@tobypeterson Thank you for merging!

PR #17482, opened 2023-01-26
meson-1.0: Add setup using configure.pre_args

Gcenx:
Adding via configure.pre_args to avoid https://trac.macports.org/ticket/66761

###### Type(s)

- [x] bugfix

###### Tested on

macOS x.y

###### Verification 

### DIFF
--- a/_resources/port1.0/group/meson-1.0.tcl
+++ b/_resources/port1.0/group/meson-1.0.tcl
@@ -17,7 +17,7 @@ depends_skip_archcheck-append \
 
 # TODO: --buildtype=plain tells Meson not to add its own flags to the command line. This gives the packager total control on used flags.
 default configure.cmd       {${prefix}/bin/meson}
-default configure.pre_args  {setup}
+default configure.pre_args  {setup --prefix=${prefix}}
 default configure.post_args {[meson::get_post_args]}
 configure.universal_args-delete \
                             --disable-dependency-tracking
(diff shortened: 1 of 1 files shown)


kencu:
But I believe the default configure.pre_args already contain some things…like prefix , set by base. 

So if they are overridden, the prefix will be unset…

Did this change work for you when you tested it?

kencu:
I tried a couple of simple fixes, but realized it was going to take a bit of a deeper test.

IDKY, but maybe the meson PG might have to use

configure.pre_args-prepend setup

to get this robustly right? I can’t imagine any Portfiles using meson using that, so it should be robust (although if a meson-using Portfile did use configure.pre_args-prepend, things would break).

Or tack setup back onto the configure.cmd, but rejig muniversal 1.0 and 1.1 to work properly with that (I tried that too, it works, bit less elegant, may be best…)

Gcenx:
I’ll dig into this more after work today

Gcenx:
@kencu maybe we simply do:
```
default configure.pre_args  {setup --prefix=${prefix}}
```

kencu:
> @kencu maybe we simply do:
> 
> ```
> default configure.pre_args  {setup --prefix=${prefix}}
> ```

That might well work. Is there any other part of base, or these myriad of Portgroups we have, that might monkey with the configure.pre_args, I wonder? 

It's just almost impossible to know the answer to that. Might have to just go for it and see what fallout might occur.

Gcenx:
> That might well work. Is there any other part of base, or these myriad of Portgroups we have, that might monkey with the configure.pre_args, I wonder?

Only one that comes to mind is the cmake Portgroups and those wouldn't affect correctly configured meson projects _unless someone forgotten to use a "port" and meson falls back to a wrap for a cmake project, aka the reason for #16810_
 
> It's just almost impossible to know the answer to that. Might have to just go for it and see what fallout might occur.

Agreed

Gcenx:
Simply put there's no legitimate reason to have a cmake portgroup added to a meson project.

There are some weird and wonderful meson wraps that handle many different deps but these are not _real_ meson projects.

PR #17545, opened 2023-01-31
yabai: remove xcode requirement as xcode cli should be sufficient

Ci7rix:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0.1 14A400

###### Verification 
Have you

### DIFF
--- a/sysutils/yabai/Portfile
+++ b/sysutils/yabai/Portfile
@@ -57,13 +57,15 @@ destroot {
     }
 }
 
-set minxcodever 13.3
-if {([vercmp $xcodeversion ${minxcodever}] < 0)} {
+if {${os.major} < 20} {
     known_fail yes
     pre-fetch {
-        ui_error "${name} ${version} requires Xcode ${minxcodever} or greater to build."
-        return -code error "incompatible Xcode version"
+        # See https://github.com/koekeishiya/yabai/issues/1287
+        ui_error "${name} ${version} cannot build on macOS 10.15 and earlier."
+        return -code error "incompatible macOS version"
     }
+} elseif {${os.major} == 20} {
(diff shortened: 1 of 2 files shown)


Ci7rix:
Is there a way to check the command line tool version that is installed ? 
See here : https://github.com/koekeishiya/yabai/issues/1200#issuecomment-1184981914

sysutils/yabai/Portfile
```diff
@@ -57,12 +57,12 @@ destroot {
...
 
-set minxcodever 13.3
-if {([vercmp $xcodeversion ${minxcodever}] < 0)} {
+if {${os.major} < 20} {
     known_fail yes
     pre-fetch {
-        ui_error "${name} ${version} requires Xcode ${minxcodever} or greater to build."
-        return -code error "incompatible Xcode version"
+        # See https://github.com/koekeishiya/yabai/issues/1287
+        ui_error "${name} ${version} cannot build on macOS 11.0 and earlier."
```
chrstphrchvz, line 64: Based on the condition, shouldn’t this instead be the following?

```suggestion
        ui_error "${name} ${version} cannot build on macOS 10.15 and earlier."
```

chrstphrchvz:
The patch approach may also be preferable due to MacPorts builders only having SDKs for the OS version they are running, and not more recent SDKs.

Ci7rix:
> The patch approach may also be preferable due to MacPorts builders only having SDKs for the OS version they are running, and not more recent SDKs.

Thank you very much for all your explanations, I think it's fine now !

TheKevJames (approved):
Makes sense to me! Thanks :)

PR #17552, opened 2023-02-01
skhd: remove xcode requirement as xcode cli should be sufficient

Ci7rix:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6 21G115 arm64
Xcode 14.0.1 14A400

###### Verification 
Have you

### DIFF
--- a/sysutils/skhd/Portfile
+++ b/sysutils/skhd/Portfile
@@ -2,7 +2,6 @@
 PortSystem              1.0
 PortGroup               github 1.0
 PortGroup               makefile 1.0
-PortGroup               xcodeversion 1.0
 
 github.setup            koekeishiya skhd 0.3.5 v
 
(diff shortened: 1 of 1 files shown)


TheKevJames (approved):
I'm not sure what the normal macports policies are on this sort of thing -- looks like this would drop support for some older OSX versions "unnecessarily" -- but overall I'm fine with it.

Ci7rix:
> I'm not sure what the normal macports policies are on this sort of thing -- looks like this would drop support for some older OSX versions "unnecessarily" -- but overall I'm fine with it.

From what I could find, the port won't build on macOS El Capitan, see here : https://ports.macports.org/port/skhd/details/

https://build.macports.org/ seems to be down for me right now, but yesterday I found out it wouldn't build because of this error : ```unknown type name 'CFNotificationName'```. 
CFNotificationName is available starting with Sierra (https://developer.apple.com/documentation/corefoundation/cfnotificationname)

TheKevJames:
Well that definitely makes sense! If the current approach doesn't work on the older versions anyway, it's better to explicitly show that in the portfile. :+1:

Ci7rix:
Anything else to do so it can be merged ?

PR #17678, opened 2023-02-12
OpenBLAS: fix regression on ppc/i386 Rosetta

MarcusCalhoun-Lopez:
Fixes https://trac.macports.org/ticket/64986

No revbump since port either builds correctly or not at all.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/math/OpenBLAS/Portfile
+++ b/math/OpenBLAS/Portfile
@@ -129,10 +129,6 @@ pre-build {
             }
             ppc {
                 puts $makeINC "TARGET = PPCG4"
-                if { ${os.arch} eq "i386" } {
-                    # This fixes Rosetta build: https://trac.macports.org/ticket/64986
-                    puts $makeINC "HOSTARCH = PPCG4"
-                }
             }
             ppc64 {
                 puts $makeINC "TARGET = PPC970"
@@ -141,6 +137,11 @@ pre-build {
 
             }
         }
+    }
+
+    # This fixes Rosetta build: https://trac.macports.org/ticket/64986
+    if {${os.platform} eq "darwin" && ${os.major} == 10 && ${muniversal.build_arch} eq "ppc"} {
+        puts $makeINC "HOSTARCH = PPCG4"
     }
 
     if {${muniversal.build_arch} ni [list i386 x86_64]} {


MarcusCalhoun-Lopez:
Unfortunately, I have no way of testing this.
If someone can verify that it works, we can remove the WIP label.

barracuda156:
> Unfortunately, I have no way of testing this.
> If someone can verify that it works, we can remove the WIP label.

@MarcusCalhoun-Lopez Thank you for dealing with the issue.

It does not work in present form, because `${os.arch}` is `powerpc` for Rosetta, so condition will not apply. It should rather use something like I made earlier (which worked fine): `${os.major} == 10 && ${build_arch} eq "ppc"`.
I think it is acceptable to disable universal on this combination, if handling universal is problematic (building ppc+x86 is largely broken anyway, and there is no ppc64 on either Rosetta or 10.6 PPC, so nothing is lost).

MarcusCalhoun-Lopez:
Please forgive me for not understanding.
On an **Intel** machine, MacPorts was installed as a **PowerPC** binary.
That is the only way I could think of that `${os.arch}` would be `powerpc` on an Intel machine. 
Is that correct?

The reason I ask is that I tried that once and had no end of troubles.

barracuda156:
@MarcusCalhoun-Lopez 
I should say sorry in fact, I did not analyze _why_ it is detected as powerpc. (And then not only here `${os.arch}` cannot be reliably used, but elsewhere as well for Rosetta case – I will look where I used it before to add Rosetta fixes.)

> On an **Intel** machine, MacPorts was installed as a **PowerPC** binary. That is the only way I could think of that `${os.arch}` would be `powerpc` on an Intel machine. Is that correct?
> 
> The reason I ask is that I tried that once and had no end of troubles.

Yes, I build everything for `ppc` on 10.6.8 – to keep it as close to native as possible. So Macports is configured explicitly for `ppc`:
```
./configure --build=powerpc-apple-darwin10 --with-curlprefix=/opt/bootstrap CC='gcc-4.2 -m32 -arch ppc' CXX='g++-4.2 -m32 -arch ppc'
```
(I also have x86 installation of Macports on 10.6.8, but in a separate prefix.)

MarcusCalhoun-Lopez:
Thank you for the clarification.
These are the types of problems I ran into when I tried the same strategy as you.
I could not find a reliable way to distinguish between  using Rosetta  and a native PowerPC system in the Portfile.
May I suggest a different approach?

Please correct me if I am wrong, but the essential problem with OpenBLAS is that on your machine, `uname -m` returns x86_64 or i386 when you would like it to return "Power Macintosh" (or something similar).
I have not tried it, but one solution might be to
- Create a directory ${prefix}/bin/rosetta
- Edit the value of "binpath" in ${prefix}/etc/macports/macports.conf so that it begins with "${prefix}/bin/rosetta:"
- Create a shell script called ${prefix}/bin/rosetta/uname that runs `exec arch /usr/bin/uname "$@"`
- Since the muniversal PG already sets [ARCHPREFERENCE](https://github.com/macports/macports-ports/blob/3b2af0ed541206430cf4c6f42d12a40366170be3/_resources/port1.0/group/muniversal-1.1.tcl#L727), your `uname -m` should return the desirable value.

Might this work?
It might help reduce the number of changes in Portfiles, especially since the configuration you are running is probably not common.

barracuda156:
@MarcusCalhoun-Lopez Makefiles that use `uname` and then force arch-specific choices based on that are a pain – it’s a bad code without an obvious solution to fix it (besides forcing a correct choice via conditional patch). If only we could switch to CMake build system – that would have been easier to fix.

IMO the desirable outcome is to have Rosetta fixed for everyone and not just for me (locally I do not need to bother about a neat fix, I can just write a needed value into Makefile), since it is the only generally accessible way to test and develop anything for PPC. Admittedly, few will use it and perhaps not often – but if it is left broken, we will get even less testing for already poorly supported platform.

By the way, what is the problem with the earlier solution that used os.major with build.arch? We do not need to distinguish between Rosetta and [unsupported] native 10.6 PPC here: both are ppc32 for all relevant to existing implementation practical purposes. Identical settings will work.

MarcusCalhoun-Lopez:
Thank you for taking the time to explain.
I did not understand the use of `${os.major}` before.
I have updated the pull request.

barracuda156:
> Thank you for taking the time to explain.
> I did not understand the use of `${os.major}` before.
> I have updated the pull request.

Thank you! Now it works correctly. Build is running.
(Just for readability, maybe place `${os.platform}` before ${os.major}`?)

MarcusCalhoun-Lopez:
> Thank you! Now it works correctly. Build is running. (Just for readability, maybe place `${os.platform}` before `${os.major}`?)

As suggested, I tweaked the pull request for readability.

barracuda156:
> As suggested, I tweaked the pull request for readability.

LGTM, thank you.

michaelld (approved):
LGTM

michaelld:
Passes current checks for non-PPC (which is always good) and has been tested with PPC & Rosetta(1) ... merging

PR #17715, opened 2023-02-15
mm-common: correct license

szhorvat:
#### Description

 - Correct the license from GPL-2 to GPL-2+. 

All copyright notices I could find at https://gitlab.gnome.org/GNOME/mm-common/ state something like:

>mm-common is free software: you can redistribute it and/or modify
> it under the terms of the GNU General Public License as published
> by the Free Software Foundation, either version 2 of the License,
> **or (at your option) any later version.**

For example, see https://gitlab.gnome.org/GNOME/mm-common/-/blob/master/macros/mm-common.m4.in

Debian states that the licenses relevant for this software are GPL-2+ and LGPL-2.1+, see https://metadata.ftp-master.debian.org/changelogs//main/m/mm-common/mm-common_1.0.5-1_copyright Presumably Debian did a thorough survey of all files in this software and their licenses.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/gnome/mm-common/Portfile
+++ b/gnome/mm-common/Portfile
@@ -8,7 +8,7 @@ version             1.0.4
 revision            0
 
 categories          gnome devel
-license             GPL-2
+license             GPL-2+
 maintainers         {devans @dbevans} {mascguy @mascguy} openmaintainer
 supported_archs     noarch
 


szhorvat:
CC @neverpanic

The motivation for this PR is to allow binaries for `py-graph-tool` to be distributed. I believe this PR will fix that:

```
$ ./port_binary_distributable.tcl -v py310-graph-tool
"py310-graph-tool" is not distributable because its license "LGPL-3" conflicts with license "GPL-2" of dependency "mm-common"
```

`py310-graph-tool` depends on `cairomm`, which in turn depends on `mm-common`. `py-graph-tool` has license LGPL-3 which is not compatible with GPL-2, but GPL-2+ is fine.

Furthermore, my understanding is that `mm-common` should only be a build dependency in most cases. Therefore, I would like to ask @essandess to review whether it is really necessary to keep it as a library dependency for `cairomm`, or a build dependency is sufficient.

szhorvat:
Question: Once this PR is merged, is it necessary to revbump `py310-graph-tool` to make sure that its binaries will be distributed?

neverpanic (approved):
This looks good to me, but let's wait for the maintainers to chime in for a change like that, maybe they are more familiar with the component's licensing than I am.

This will need a re-trigger of a build, which can either be done manually on the buildbot web interface, or by making a no-op change that changes the Portfile's checksum.

It should not require a revbump, but maybe our build caching setup has changed since the last time I looked at it and the relevant step is no longer being run, in which case this would need a revbump.

Try pinging me, I can trigger a build on the web interface of build.macports.org.

essandess:
> I would like to ask @essandess to review whether it is really necessary to keep it as a library dependency for `cairomm`, or a build dependency is sufficient.

@count0 Would you please weigh in on this question?

essandess:
> I would like to ask @essandess to review whether it is really necessary to keep it as a library dependency for `cairomm`, or a build dependency is sufficient.

It is necessary to specify `cairomm` as a `depends_lib` of` py*-graph-tool`:
```sh
sudo port install py310-graph-tool
port contents py310-graph-tool | egrep -e '\.so$' | xargs otool -L | grep cairo | awk '{print $1}' | sort -u | xargs port provides

/opt/local/lib/libcairo.2.dylib is provided by: cairo
/opt/local/lib/libcairomm-1.0.1.dylib is provided by: cairomm
```

szhorvat:
Yes, `cairomm` is most definitely needed by `py*-graph-tool`. 

My question was whether `mm-common` should be a `depends_lib` of `cairomm`, or most of the other ports where it is currently a dependency. `mm-common` does not contain any executable binaries, nor any C/C++ headers, only some scripts. Its [description](https://gitlab.gnome.org/GNOME/mm-common/) appears to suggest that it is part of a build system, so perhaps it's only needed at build time? But then I'm not sure. 

> The mm-common module provides the build infrastructure and utilities
shared among the GNOME C++ binding libraries.  It is only a required
dependency for building the C++ bindings from the gnome.org version
control repository.  An installation of mm-common is not required for
building tarball releases, unless configured to use maintainer-mode.

essandess:
> My question was whether `mm-common` should be a `depends_lib` of `cairomm`

Yes, confirm, `mm-common` should be a `depends_lib` of `cairomm`.

jmroot:
Really? How does cairomm use mm-common at runtime?

It's not just that a lot of things that depend on cairomm also need mm-common at build time?

essandess:
> Really? How does cairomm use mm-common at runtime?
> 
> It's not just that a lot of things that depend on cairomm also need mm-common at build time?

My mistake. I meant `depends_run`.

jmroot:
> I'm still miswriting this! I meant `depends_build`.

OK. Glad we got that cleared up. 😄

szhorvat:
@essandess Thanks for confirming! I just realized that I pinged you about this by accident, as I thought you were a maintainer of `cairomm` (I was looking at `py-graph-tool`, sorry).

The move from `depends_lib` to `depends_build` can probably be done for most other ports that depend on `mm-common` as well.

ryandesign:
> It should not require a revbump, but maybe our build caching setup has changed since the last time I looked at it and the relevant step is no longer being run, in which case this would need a revbump.

There has been no change to the buildbot setup. It remains the case that any change to a Portfile, even a no-op whitespace change, will cause buildbot to reevaluate whether building or distributing is necessary.

szhorvat:
@mascguy @dbevans Can this be merged as is?

mascguy (approved):

mascguy:
Sorry for the delay in reviewing this. And thank you so much for the fix!

PR #17721, opened 2023-02-16
legacy-support: use makefile & muniversal 1.1 PGs

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/legacy-support/Portfile
+++ b/devel/legacy-support/Portfile
@@ -1,8 +1,8 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 PortSystem          1.0
-PortGroup           github 1.0
 PortGroup           clang_dependency 1.0
+PortGroup           github 1.0
 PortGroup           makefile 1.0
 PortGroup           muniversal 1.1
 
--- a/devel/legacy-support/Portfile
+++ b/devel/legacy-support/Portfile
@@ -3,7 +3,8 @@
 PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           clang_dependency 1.0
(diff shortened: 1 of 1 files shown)


kencu:
Marcus, Chris doesn’t actually have or use any older OS versions. 

This looks OK, and I know if it’s notbOK in some way, you’ll surely fix it. 

So go ahead, I think, and please consider whether you want to be an official maintainer here too, so you can just push fixes…

devel/legacy-support/Portfile
```diff
@@ -3,7 +3,8 @@
 PortSystem          1.0
 PortGroup           github 1.0
 PortGroup           clang_dependency 1.0
-PortGroup           muniversal 1.0
+PortGroup           muniversal 1.1
+PortGroup           makefile 1.0
```
ryandesign, line 7: Don't we usually have portgroups in alphabetical order? Or is there an extenuating circumstance why that doesn't work here?

MarcusCalhoun-Lopez: I am afraid I was unaware of the convention of alphabetizing PortGroup names.
Thank you for the information.
I have updated the pull request.

ryandesign: Complex portgroups have a tendency to conflict with one another in subtle ways. `pre-` and `post-` blocks run in the order in which they were declared (i.e. the order in which the portgroups were included), and the ultimate value of options can vary depending on portgroup order if both portgroups set option defaults. For example the legacysupport 1.1 portgroup must be included before the makefile 1.0 portgroup or it will not work properly; once we discovered this, a test was added to the legacysupport 1.1 portgroup to ensure the correct inclusion order, but that doesn't mean the other portgroups are immune from this type of problem. To minimize the likelihood of these problems, it's recommended to adopt a standardized order for including portgroups and alphabetical order is what we seem to have standardized on.

PR #17729, opened 2023-02-17
apr: use the right compiler

MarcusCalhoun-Lopez:
See https://trac.macports.org/wiki/UsingTheRightCompiler

No revbump since installed files do not change.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/apr/Portfile
+++ b/devel/apr/Portfile
@@ -33,7 +33,8 @@ if {[vercmp ${xcodeversion} 3.0] >= 0} {
 use_parallel_build yes
 configure.ccache   no
 #configure's tests don't work right for sed/gsed if we just set SED here
-configure.env      ac_cv_prog_AWK=awk ac_cv_path_SED=sed lt_ECHO=/bin/echo lt_cv_path_SED=sed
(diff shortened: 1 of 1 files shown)


danielluke:
This looks ok, but where is CC_FOR_BUILD being used (instead of CC?)

danielluke:
This looks like an upstream bug - it appears to only be used for building tools/gen_test_char - can you file a bug upstream about it? The change here looks good for us for now.

MarcusCalhoun-Lopez:
I am not sure this is an upstream bug.
The use of `CC_FOR_BUILD` seems to have been a [conscious choice](https://svn.apache.org/viewvc?view=revision&revision=1871998) in response to a cross-compiling issue.

danielluke:
I think it's still worthy of a bug. If you set CC you don't expect to also have to set CC_FOR_BUILD in order to get the compiler you want used. If they decide not to fix it at least we'll know that we have to carry the patch forward forever...

PR #17739, opened 2023-02-18
tenfourfox: new port, web-browser for PPC

barracuda156:
#### Description

Perhaps a well-known browser: https://github.com/classilla/tenfourfox
This port is only for PPC.

P. S. If anyone interested to support Intel version, you are welcome. I asked earlier on Trac, but no one expressed interest: https://trac.macports.org/ticket/66185

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/devel/strip7/Portfile
+++ b/devel/strip7/Portfile
@@ -1,0 +1,239 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+
+name                    strip7
+
+# Xcode 8.1 - matching cctools
+version                 895
+revision                0
+set ld64_version        274.1
+categories              devel
+maintainers             nomaintainer
+license                 {APSL-2 GPL-2+}
+installs_libs           no
+description             A modified version of strip that ignores variant n_sects
+long_description        ${name}. Useful only for building TenFourFox with modern gcc versions.
+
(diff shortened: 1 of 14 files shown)


devel/strip7/Portfile
```diff
@@ -0,0 +1,239 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+
+name                    strip7
+
+# Xcode 8.1 - matching cctools
+version                 895
+revision                7
```
herbygillot, line 9: `revision` should be starting at zero for a brand new port.

barracuda156: @herbygillot I can do that, but I suspect this corresponds to actually used cctools time ago. Doesn’t matter?

herbygillot: If the `name` and `version` of this new port is exactly the same as a port that _used_ to exist, then yes, it would actually make sense to have an incremented `revision`.  But if the port name was `cctools` and not `strip7`, then this would be considered the first instance of the `strip7` port ever and should have a rev of zero.

barracuda156: Fair enough, I will set it to zero and rebase in a few min.

barracuda156: UPD. Done!

barracuda156:
Sorry, rebased wrongly. Gimme a few min.

PR #17849, opened 2023-03-03
R-igraph: add support for tests

barracuda156:
#### Description

Wrapped in a `tests` variant due to dependency on `tcltk`, which is not a port, but a non-default variant for `R`.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/R/R-igraph/Portfile
+++ b/R/R-igraph/Portfile
@@ -25,3 +25,20 @@ depends_lib-append  port:glpk \
                     port:R-rlang
 
 compilers.setup     require_fortran
+
+variant tests description "Enable tests" {
+    require_active_variants R tcltk
+    depends_test-append \
+                    port:R-ape \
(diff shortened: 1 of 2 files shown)


szhorvat:
Does it really require a separate variant for testing?

barracuda156:
> Does it really require a separate variant for testing?

No, if we have R built with `+tcltk`. But it is not a default variant, so we run into trouble: https://trac.macports.org/ticket/67016

barracuda156:
@szhorvat P. S. I dislike these ugly variants for tests, but to get rid of these hacks we need `R` to be built with `+tcltk` by default. That would solve other problems as well, like we can bring `Rcmdr` and some other ports.

PR #17913, opened 2023-03-08
cairomm: Update to version 1.16.2

essandess:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.2.1 22D68 x86_64
Xcode 14.2 14C18

###### Verification 
Have you


ryandesign:
Checks fail:
```
Run-time dependency sigc++-3.0 found: NO (tried pkgconfig, framework and cmake)
Looking for a fallback subproject for the dependency sigc++-3.0

Executing subproject sigc++-3.0 

sigc++-3.0| Project name: libsigc++
sigc++-3.0| Project version: 3.4.0
sigc++-3.0| C++ compiler for the host machine: /usr/bin/clang++ (clang 13.0.0 "Apple clang version 13.0.0 (clang-1300.0.29.30)")
sigc++-3.0| C++ linker for the host machine: /usr/bin/clang++ ld64 711
sigc++-3.0| Program python3 found: YES 3.8.9 (/usr/bin/python3)
sigc++-3.0| Run-time dependency Boost found: NO (tried system)
sigc++-3.0| Program mm-common-get found: YES (/opt/local/bin/mm-common-get)
sigc++-3.0| Program doxygen found: NO

subprojects/libsigcplusplus/meson.build:124:0: ERROR: Program 'doxygen' not found or not executable
```

graphics/cairomm/Portfile
```diff
@@ -1,35 +1,41 @@
...
+
+use_xz              yes
 
 depends_build       port:pkgconfig \
                     port:mm-common
 
 depends_lib         path:lib/pkgconfig/cairo.pc:cairo \
+                    port:doxygen \
+                    port:graphviz \
                     port:libsigcxx2
```
ryandesign, line 37: doxygen would be a build dependency, not a lib dependency. doxygen sometimes uses graphviz, so graphviz would be a build dependency too. When you depend on graphviz, make sure to do it in a way that allows graphviz-devel to satisfy it; usually that's `path:bin/dot:graphviz`.

However doxygen and graphviz are only being called for because the port is trying to build its own copy of libsigc++ 3 because it didn't find one already installed. Switching the libsigcxx2 dependency to libsigcxx3 works for me:

```suggestion
                    port:libsigcxx3
```

graphics/cairomm/Portfile
```diff
@@ -1,35 +1,41 @@
...
 
 PortSystem          1.0
+
 PortGroup           active_variants 1.1
+PortGroup           boost 1.0
+PortGroup           meson 1.0
 
 name                cairomm
-version             1.12.2
+version             1.16.2
```
ryandesign, line 10: This version of cairomm has a new major version of the libcairomm library. All ports that link with the library will be broken after this update unless you revbump them.

graphics/cairomm/Portfile
```diff
@@ -1,35 +1,41 @@
...
 categories          graphics
 license             LGPL-2+
 maintainers         {devans @dbevans} openmaintainer
+platforms           darwin
```
ryandesign, line 14: That's the default so it can be removed.
```suggestion
```

graphics/cairomm/Portfile
essandess: Thank you! Done.

graphics/cairomm/Portfile
essandess: I'll add these commits after I see that the CI builds this version.

graphics/cairomm/Portfile
essandess: Thanks done.

graphics/cairomm/Portfile
essandess: Added. Note that `py*-graph-tool` must be updated in #17914.

ryandesign:
pangomm failed with a curious error:

```
checking for glibmm-2.4 >= 2.48.0 cairomm-1.0 >= 1.2.2 pangocairo >= 1.41.0... no
configure: error: Package requirements (glibmm-2.4 >= 2.48.0 cairomm-1.0 >= 1.2.2 pangocairo >= 1.41.0) were not met:

No package 'cairomm-1.0' found
```

I see the same on my system. The config.log says the reason is:

```
configure:17886: checking for glibmm-2.4 >= 2.48.0 cairomm-1.0 >= 1.2.2 pangocairo >= 1.41.0
configure:17893: $PKG_CONFIG --exists --print-errors "$PANGOMM_MODULES"
Package cairomm-1.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `cairomm-1.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'cairomm-1.0' found
configure:17896: $? = 1
configure:17910: $PKG_CONFIG --exists --print-errors "$PANGOMM_MODULES"
Package cairomm-1.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `cairomm-1.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'cairomm-1.0' found
configure:17913: $? = 1
configure:17927: result: no
```

Indeed, cairomm 1.16.2 doesn't install cairomm-1.0.pc; it installs cairomm-1.16.pc.

essandess:
It appears that the latest version of `pangomm` uses `cairomm-1.16`, so I updated this too.

https://github.com/GNOME/pangomm/blob/7083b4deaa6401bf1f5b771752a46632ecc8d303/configure.ac#L52

essandess:
> Are you testing these on your system first? 

Yes, I confirm a successful build on my system. I forgot to sync my local branch with my local ports.

mascguy:
Rebased against master, fixing the merge conflict with `cherrytree`.

mascguy:
Since I'm working on Cairo-related items at the moment, I'll add this to the list.

Next step is to test locally with these changes, to see whether the CI failures are legitimate. Stay tuned...

mascguy:
Okay, so after digging into this, here are the key takeaways:

For `cairomm`:
* This update _might_ be fine as-is, but need to do more testing and validation.
* So more to follow on this port.

For `pangomm`:
* The project depends on a newer version of `glibmm`, so it tries to download and install it publicly. That's a big problem, and simply won't work.
* So essentially this depends on me updating `glibmm` first, which is still on my to-do list.
* I'm also seeing `autoconf`-related failures, which I'm not easily able to work around with patches. But building with Meson appears to be fine, and we're overdue to switchover anyway, so I'll take care of that.

Still pondering whether we can/should update `cairomm` by itself, and defer `pangomm` until `glibmm` has been updated.

Converting to draft for now. Stay tuned...

mascguy:
Okay, so after digging into the details, it became clear that upstream has accommodated backward compatibility.

In short, older bindings for the various C++ lib - including, but not limited to, `glibmm`, `atkmm`, `cairomm`, `pangomm`, etc - are provided by our outdated versions. Meanwhile, the newest upstream releases provide non-conflicting functionality, by installing into different subdirs. (Along with non-conflicting pkgconfigs and such.)

So if we follow a different approach - providing the older/legacy bindings via new ports - we can also update to the latest versions. All without having to rebuild dependents, as the new legacy ports, will provide exact ABI (and API) compatibility.

The first phase of that was done last night, for `atkmm` and `glibmm`. Now we simply need to follow the same model for `cairomm` and `pangomm`.

I'm currently testing all of that locally, across the full swath of macOS releases - 10.6 thru Monterey - to ensure there aren't any issues. (I'm not yet able to create a working Ventura VM, but that's the only blind-spot.) Once that's done, I'll commit to master.

Net-Net, all of the various rev-bumps and such here, aren't necessary: The new legacy binding ports will provide drop-in ABI compatibility, and will be pulled in automatically during an update.

PR #17924, opened 2023-03-10
gmtl: Fix build issues

ryandesign:
#### Description

Add patches to fix build failures when using clang, when using the latest version of scons, when scons is using python 3, and to prevent the creation of the unwanted file gmtl-0.6.1.tar.gz and to install the file gmtl.pc.

Add runtime dependencies on flagpoll and python which are used by gmtl-config. flagpoll is currently incompatible with python 3 and uses python27 so use the same python here to reduce the number of pythons a user needs to install.

Specify absolute path to scons.

Specify prefix at build time so that build output does not give the erroneous impression that files will be installed in /usr/local.

Specify empty directories for Boost and CppUnit at build time so that any Boost or CppUnit that might be installed in /usr/local is not inadvertently used.

No need to specify destroot.cmd or destroot.target; the defaults work.

Add modeline.

Modernize checksums.

Set platforms to any and set supported_archs to noarch because this port doesn't install anything architecture- or OS-specific.

Add GitHub handle to maintainers line.

Install documentation files.

Indicate license. The software claims to be licensed LGPL-2.1+ but there is also a license addendum that imposes additional restrictions and the LGPL does not permit the imposition of additional restrictions.

Rewrite master_sites to avoid redirects.

Closes: https://trac.macports.org/ticket/57421

-----

@mccdo as with #17916 I've added your GitHub handle to the maintainers line. If you don't want to maintain this port anymore let us know and we'll remove you. If the email address needs updating let us know the new address. If you allow others to make changes to this port without consulting you let us know and we'll add `openmaintainer`.

I did already make one change to this port without consulting you — d4a979bd0a5ca18d5d9104394eabe30cc94c7c44 — apologies. This was based on the fact that you had not responded to [the ticket that this PR fixes](https://trac.macports.org/ticket/57421) for over four years and the port had not been installable at all since Xcode 5 was released almost ten years ago.

###### Type(s)

- [X] bugfix
- [X] enhancement
- [ ] security fix

###### Tested on
macOS 12.3.1 21E258 x86_64
Xcode 13.2.1 13C100

###### Verification 
Have you

### DIFF
--- a/math/gmtl/Portfile
+++ b/math/gmtl/Portfile
@@ -1,27 +1,67 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
 PortSystem          1.0
 
 name                gmtl
 version             0.6.1
+revision            1
+checksums           rmd160  19ca8c25cf54b1650272f7a387f5f8b8c5c3140c \
+                    sha256  f7d8e6958d96a326cb732a9d3692a3ff3fd7df240eb1d0921a7c5c77e37fc434 \
+                    size    339777
+
 categories          math
-platforms           darwin
-maintainers         iastate.edu:mccdo
+platforms           any
+supported_archs     noarch
+maintainers         {iastate.edu:mccdo @mccdo}
+# LGPL-2.1+ modified by an addendum
+license             Copyleft GPLConflict
+
 description         a lightweight math library
 long_description    The math library used by vrjuggler. It is \
                     implemented with c++ templates.
 homepage            http://ggt.sourceforge.net/
-master_sites        sourceforge:ggt
-checksums           md5 1391af2c5ea050dda7735855ea5bb4c1 \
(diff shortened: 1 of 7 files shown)


math/gmtl/Portfile
```diff
@@ -1,27 +1,66 @@
...
+                    sha256  f7d8e6958d96a326cb732a9d3692a3ff3fd7df240eb1d0921a7c5c77e37fc434 \
+                    size    339777
+
 categories          math
-platforms           darwin
-maintainers         iastate.edu:mccdo
+platforms           any
+supported_archs     noarch
+maintainers         {iastate.edu:mccdo @mccdo}
+license             LGPL-2.1+ Restrictive/Distributable
```
ryandesign, line 16: @jmroot Is this the correct license line? This software claims to be under LGPL-2.1+ but also includes a license addendum that imposes additional restrictions.

jmroot: Most of the license addendum is actually granting additional permissions, which is fine, but section 4 definitely appears to be an additional restriction, which is disallowed by the LGPL. I think this has to be treated as a distinct GPL-incompatible copyleft license, which would be written `Copyleft GPLConflict`.

PR #17966, opened 2023-03-14
Updated inline comment about VideoToolbox

tux-o-matic:
FFmpeg can use the HW accelerated H265 encoder exposed by VideoToolbox via hevc_videotoolbox

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix


ryandesign:
The same comment is in the ffmpeg-devel, ffmpeg-upstream, and mpv ports; should update all ports at once.

tux-o-matic:
@ryandesign, thanks for the pointer. It's aligned with all these Ports as well now.

PR #17973, opened 2023-03-14
goimapnotify: New port

neverpanic:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 12.6.3 21G419 arm64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/mail/goimapnotify/Portfile
+++ b/mail/goimapnotify/Portfile
@@ -1,0 +1,96 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           golang 1.0
+
+go.setup            gitlab.com/shackra/goimapnotify 2.4-rc3
+categories          mail
+maintainers         {cal @neverpanic} openmaintainer
+license             GPL-3 MIT BSD
+
+description         \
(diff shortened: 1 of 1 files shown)


mail/goimapnotify/Portfile
```diff
@@ -0,0 +1,96 @@
...
+go.setup            gitlab.com/shackra/goimapnotify 2.4-rc3
+categories          mail
+maintainers         {cal @neverpanic} openmaintainer
+license             GPL-3 MIT BSD
+
+description         \
+    Execute scripts on IMAP mailbox changes (new/deleted/updated messages) \
+    using IDLE
+
+long_description    ${description}
```
herbygillot, line 15:
```suggestion
long_description    {*}${description}
```

neverpanic: Does that make a difference? I thought both of those were strings, so `{*}` should do exactly nothing here, but maybe I'm missing something.

kencu: This has been a thing for the past year or so, and virtually every single Portfile has had this change made. It is thought to be important.

neverpanic: 03b9df6ce4e75b14bc373ab89c09457bd32f36ff

jmroot: Both options are lists. It often doesn't make a difference if the description doesn't contain any characters that need to be quoted in lists. Using `{*}` always gives a correct result, so that's the better example to set.

PR #17975, opened 2023-03-14
qemu: update to 7.2.0

vallon:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6.3 21G419 arm64
Xcode 14.2 14C18

###### Verification 
Have you

- [x] tested basic functionality of all binary files? *ran arm qemu virtual machine*
- [ ] checked that the Portfile's most important [variants](https://trac.macports.org/wiki/Variants) haven't been broken? *lots of variants, checked default builds and my custom flavor runs my VM*

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -8,8 +8,8 @@ PortGroup legacysupport 1.1
 legacysupport.newest_darwin_requires_legacy 16
 
 name                    qemu
-version                 7.1.0
-revision                1
+version                 7.2.0
+revision                0
 categories              emulators
 license                 GPL-2+
 maintainers             {raimue @raimue} \
@@ -25,11 +25,12 @@ homepage                https://www.qemu.org
 master_sites            https://download.qemu.org/
 use_xz                  yes
 
-checksums               rmd160  01c82bed8b0219935c00981e16b573004b3802d0 \
-                        sha256  a0634e536bded57cf38ec8a751adb124b89c776fe0846f21ab6c6728f1cbbbe6 \
-                        size    121833004
+checksums               rmd160  50661dd45f91d900241e0e6b919140a6b8a2b2cf \
+                        sha256  5b49ce2687744dad494ae90a898c52204a3406e84d072482a1e1be854eeb2157 \
+                        size    122408576
 
 depends_build           port:texinfo \
+                        port:libslirp \
                         port:libtool \
                         port:meson \
                         port:ninja \


raimue:
Thank you for the port update!

However, I have some notes based on the changelog provided by QEMU upstream.

https://wiki.qemu.org/ChangeLog/7.2

The changelog mentions that libslirp is no longer bundled. I think we might need to use that from a port now to get feature parity with the previous version.

Additionally, it is now possible to build combinations of Cocoa, GTK and SDL UI interfaces into the same binary. Therefore the variants should no longer be conflicting.

vallon:
Yes, I just realized networking is not working.  I see there is libslirp in ports.  I'll try to update.

I don't know enough about the rest to know whether/how to test the variants.  Maybe another PR?

emulators/qemu/Portfile
```diff
@@ -8,7 +8,7 @@ PortGroup legacysupport 1.1
 legacysupport.newest_darwin_requires_legacy 16
 
 name                    qemu
-version                 7.1.0
+version                 7.2.0
 revision                1
```
herbygillot, line 12: Also `revision` resets to zero whenever the `version` gets updated:

```suggestion
revision                0
```

vallon: fixed

vallon:
Adding libslirp dependency restored default networking behavior (fixed -nic user: network backend 'user' is not compiled into this binary)

emulators/qemu/Portfile
```diff
@@ -25,11 +25,12 @@ homepage                https://www.qemu.org
...
 
-checksums               rmd160  01c82bed8b0219935c00981e16b573004b3802d0 \
-                        sha256  a0634e536bded57cf38ec8a751adb124b89c776fe0846f21ab6c6728f1cbbbe6 \
-                        size    121833004
+checksums               rmd160  50661dd45f91d900241e0e6b919140a6b8a2b2cf \
+                        sha256  5b49ce2687744dad494ae90a898c52204a3406e84d072482a1e1be854eeb2157 \
+                        size    122408576
 
 depends_build           port:texinfo \
+                        port:libslirp \
```
raimue, line 33: This is good, but we should also add an explicit `--enable-slirp` to the `configure.args` to ensure we always link with libslirp and do not rely on the auto-detection of the configure script.

This is small enough that we do not need it as a variant (user networking is so common that I doubt anyone really needs to deselect it for minimal gains).

raimue:
> I don't know enough about the rest to know whether/how to test the variants. Maybe another PR?

Fine with me. Let's focus on getting the new QEMU 7.2.0 version in with this PR. The no longer conflicting variants can still be resolved later. I did not have a chance to play with building a single binary for multiple GUIs.

catap:
FYI: 8.0.0 was released a few weeks ago

raimue:
Thank you for the pull request. However, this has now been superseded by #19283, which also makes the correct changes to handle the switch to libslirp.

PR #18002, opened 2023-03-18
submission: resource_dasm and phosg

fracai:
#### Description

Portfiles for `resource_dasm` and dependency `phosg`

`phosg` doesn't have any releases and uses the latest commit (as of submission).
`resource_dasm` does have some release tags, but there have been many commits since the last tag, so also uses the latest commit (as of submission).

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.2.1 22D68 arm64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/devel/phosg/Portfile
+++ b/devel/phosg/Portfile
@@ -1,0 +1,34 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               compiler_blacklist_versions 1.0
+
+github.setup            fuzziqersoftware phosg e1bd000
+github.tarball_from     archive
+
+# blacklisting to select C++20 capable compilers
+compiler.blacklist-append {clang < 1300}
+
+version                 2023.03.04
+revision                0
+categories              devel
+maintainers             {alum.wpi.edu:arno+macports @fracai} openmaintainer
+license                 MIT
+
+description             Phosg is a basic C++ wrapper library around some common C routines.
+long_description        {*}${description}
+
+checksums               rmd160  526fb52a0eadbb3e33da6e2711c4c43afca04eda \
+                        sha256  802170acf4b8bae5383183c9361812d5309cb3e79d703e365b96feffe5d776cd \
+                        size    147998
+
+set pyver               3.11
+set python_version      [string index ${pyver} 0][string range ${pyver} 2 end]
+
+patchfiles              CMakeLists-txt.diff
+
+depends_build-append    port:python${python_version}
+
+configure.args-append   -DPYTHON_EXECUTABLE=${prefix}/bin/python${pyver}
--- a/devel/resource_dasm/Portfile
+++ b/devel/resource_dasm/Portfile
@@ -1,0 +1,38 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               compiler_blacklist_versions 1.0
+
+github.setup            fuzziqersoftware resource_dasm 46ad32e
+github.tarball_from     archive
+
+# blacklisting to select C++20 capable compilers
+compiler.blacklist-append {clang < 1300}
+
+version                 2023.03.15
+revision                0
+categories              devel
+maintainers             {alum.wpi.edu:arno+macports @fracai} openmaintainer
+license                 MIT
+
+description             Tools for reverse-engineering classic Mac OS applications and games.
+long_description        {*}${description}
+
+checksums               rmd160  190a5eb9a174d8eafdc17ca736f730e2d12862ec \
+                        sha256  db5e42d9ef9df9d4d40f42fd7c1b67cc2f5108a06e30ae983fd6e40046ad22e0 \
+                        size    457269
+
+set pyver               3.11
+set python_version      [string index ${pyver} 0][string range ${pyver} 2 end]
+
+patchfiles              CMakeLists-txt.diff
+
+depends_build-append    port:python${python_version}
+depends_lib-append      port:phosg \
+                        port:zlib
+depends_run-append      port:netpbm
+
+configure.args-append   -DPYTHON_EXECUTABLE=${prefix}/bin/python${pyver} \
+                        -Dphosg_DIR=${prefix}/lib/cmake/phosg
--- a/devel/phosg/files/CMakeLists-txt.diff
+++ b/devel/phosg/files/CMakeLists-txt.diff
@@ -1,0 +1,10 @@
+--- CMakeLists.txt.orig
++++ CMakeLists.txt
+@@ -71,5 +71,5 @@
+ install(TARGETS jsonformat DESTINATION bin)
+ install(TARGETS parse-data DESTINATION bin)
+ install(FILES ${Headers} DESTINATION include/phosg)
+-install(FILES phosg-config.cmake DESTINATION lib)
+-install(EXPORT phosg DESTINATION lib)
++install(FILES phosg-config.cmake DESTINATION lib/cmake/phosg)
++install(EXPORT phosg DESTINATION lib/cmake/phosg)
--- a/devel/resource_dasm/files/CMakeLists-txt.diff
+++ b/devel/resource_dasm/files/CMakeLists-txt.diff
@@ -1,0 +1,10 @@
+--- CMakeLists.txt.orig
++++ CMakeLists.txt
+@@ -130,5 +130,5 @@
+ file(GLOB EmulatorsHeaders ${CMAKE_SOURCE_DIR}/src/Emulators/*.hh)
+ install(FILES ${EmulatorsHeaders} DESTINATION include/resource_file/Emulators)
+ 
+-install(FILES resource_file-config.cmake DESTINATION lib)
+-install(EXPORT resource_file DESTINATION lib)
++install(FILES resource_file-config.cmake DESTINATION lib/cmake/resource_dasm)
++install(EXPORT resource_file DESTINATION lib/cmake/resource_dasm)


devel/phosg/Portfile
```diff
@@ -0,0 +1,28 @@
...
+revision                1
+categories              devel
+maintainers             {alum.wpi.edu:arno+macports @fracai} openmaintainer
+platforms               darwin
+license                 MIT
+
+description             Phosg is a basic C++ wrapper library around some common C routines.
+long_description        {*}${description}
+
+homepage                https://github.com/fuzziqersoftware/phosg
```
harens, line 17: This is set by the github portgroup

devel/phosg/Portfile
```diff
@@ -0,0 +1,28 @@
...
+
+checksums               rmd160  18a3d315b3668f1089a7396fe25a918bf78055b4 \
+                        sha256  2280afd7a677a7f2c570c86469fc43a827b04fdf5e06bb682b04e6fcf333458b \
+                        size    147748
+
+set pyver               3.10
```
harens, line 23: Could we use Python 3.11?

harens (commented):
Hi there @fracai 👋 Thank you for your first contribution to MacPorts! You've done a great job here. I've left a few minor comments on phosg, which also apply to resource_dasm.

If you could separate the two portfiles into different commits in the form `NAME: new port`, that would be great. Thank you for contributing @fracai 👍

fracai:
> If you could separate the two portfiles into different commits in the form `NAME: new port`, that would be great.

Do you mean that I should create two PRs? Or do you want me to force push (or open a new replacement PR) with one commit for each Portfile, with `NAME: new port` as the commit message?

Thanks for the review, I'll get the changes up once I understand the commit format you're looking for.

devel/phosg/Portfile
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+
+github.setup            fuzziqersoftware phosg e1bd000
+revision                1
```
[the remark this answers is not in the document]
fracai: I originally set a revision because I was only setting a commit hash to checkout. If I now set the version number as the commit date, should I hold off on setting a revision until it's needed?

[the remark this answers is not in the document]
harens: > If I now set the version number as the commit date, should I hold off on setting a revision until it's needed?

Setting the revision number explicitly as zero should be good.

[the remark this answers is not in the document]
fracai: `revision 0` in both

devel/phosg/Portfile
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+
+github.setup            fuzziqersoftware phosg e1bd000
```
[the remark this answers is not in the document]
fracai: `version` dated in both

devel/phosg/Portfile
fracai: removed

devel/phosg/Portfile
```diff
@@ -0,0 +1,28 @@
...
+revision                1
+categories              devel
+maintainers             {alum.wpi.edu:arno+macports @fracai} openmaintainer
+platforms               darwin
```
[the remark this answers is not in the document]
fracai: removed

devel/phosg/Portfile
fracai: `pyver` set to 3.11

fracai:
comments resolved, split to two commits, and force pushed

harens:
```
   CMake Error at CMakeLists.txt:30 (find_package):
    By not providing "Findphosg.cmake" in CMAKE_MODULE_PATH this project has
    asked CMake to find a package configuration file provided by "phosg", but
    CMake did not find one.
  
    Could not find a package configuration file provided by "phosg" with any of
    the following names:
  
      phosgConfig.cmake
      phosg-config.cmake
  
    Add the installation prefix of "phosg" to CMAKE_PREFIX_PATH or set
    "phosg_DIR" to a directory containing one of the above files.  If "phosg"
    provides a separate development package or SDK, be sure it has been
    installed.
```

fracai:
Gah, I forgot to include cmake as a dependency.
I actually figured it would have been included as a dependency as it's part of the cmake portgroup.

harens:
> I actually figured it would have been included as a dependency as it's part of the cmake portgroup.

It is a dependency, so that's not the problem:

https://github.com/macports/macports-ports/blob/fd259e6117af72e16caa5c8e85b4aa5ad3d532ca/_resources/port1.0/group/cmake-1.1.tcl#L91-L93

The issue is that `phosg-config.cmake` (as well as `phosg-debug.cmake` and `phosg.cmake`) are installed into `${prefix}/lib`, when they should be in their own subdirectory. Maybe `${prefix}/lib/cmake/phosg/`?

fracai:
The latest workflow failed `resource_dasm` because `cmake` couldn't find `phosg`. It suggests adding the `phosg` installation prefix, but the `Portfile` already does that with `configure.args-append … -DCMAKE_PREFIX_PATH=${prefix}/lib`. And when building the Portfile, adding this arg did get the build working. Does the workflow use trace mode ([currently broken on Ventura](https://trac.macports.org/ticket/66358)) and that's causing an issue?
I note that line [1304 of the log](https://github.com/macports/macports-ports/actions/runs/4495960166/jobs/7919345614#step:8:1310) has two occurrences of `DCMAKE_PREFIX_PATH`. I wonder if only the first is taking effect.
The build also suggests setting `phosg_DIR`, so I'll give that a try instead of `DCMAKE_PREFIX_PATH`.

fracai:
Hah, didn't see your comment while posting mine.
So I need to change the installation of just those `*.cmake` files to a phosg subdir? And then update the `resource_dasm` argument accordingly?
And then remove `cmake` as a dependency as well?

fracai:
The latest push installs the cmake files to `lib/cmake/<port>` and everything seems to compile without the extra cmake arguments. I'll be interested to see if it works with the workflow builds.

I just saw your additional notes, so I haven't addressed those yet. I'll take a look.

fracai:
So, my initial pass sums up to 🤷 

It definitely compiles and is useable by resource_dasm as I've compiled both and successfully tested resource_dasm while phosg was installed under /opt/local. So I'm tempted to say that issue might be down to how that individual built, installed, and linked to phosg. And the C++20 request is perhaps not actually using features introduced in 20 as, again, it did build for me?

I don't know. If the workflow build succeeds I'm tempted to keep an eye on those issues and make updates accordingly, but maybe not let them become blockers… ?

fracai:
This latest sets `phosg_DIR`
🤞

fracai:
I'm at a loss. The arg `DCMAKE_PREFIX_PATH` already exists with a different path, so setting it in the Portfile adds a second copy, but doesn't build on the workflow. Setting `Dphosg_DIR` also doesn't build on the workflow. Both of these build for me locally.

I'm assuming the workflow uses tracemode, which I can't test with as it's broken on Ventura.

Is it possible that there's some other difference how the workflow builder operates? Or something I can do to test in a different way?

devel/phosg/Portfile
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
```
harens, line 5: Would you mind adding the following (or something similar) to both this and the other port? (see [wiki](https://guide.macports.org/#reference.portgroup.github.distfilestrategy))

```tcl
github.tarball_from archive
```

By default, the tar is fetched from `https://codeload.github.com/fuzziqersoftware/phosg/legacy.tar.gz/e1bd000?dummy=`, which is a legacy method.

Note that this would then require the checksums to be changed. Run `sudo port clean --all NAME` to remove the old downloaded files.

harens (commented):
Ok I don't know how I didn't spot this before, but phosg hasn't been set as a dependency of resource_dasm. That's likely the issue.

The C++20 thing affects both ports, and might prevent them from building on older hardware ([wiki](https://trac.macports.org/wiki/CompilerSelection)). Maybe take a look at https://github.com/macports/macports-ports/pull/9932#issuecomment-774422453 or https://github.com/macports/macports-ports/pull/14963#issuecomment-1140330450.

I think after that we can merge this and see what happens with the buildbots. Thanks for all your work on this @fracai :)

fracai:
How did I manage to not specify `phosg` as a dependency. Well, I know exactly how. I'm testing locally without trace so I installed `phosg` and then `resource_dasm` and the dependency was met. But still, how did I forget to specify that?

Oh well.

This latest adds the missing dependency and the compiler blacklist.

Thanks for your help in figuring this out.

devel/resource_dasm/Portfile
```diff
@@ -0,0 +1,37 @@
...
+checksums               rmd160  190a5eb9a174d8eafdc17ca736f730e2d12862ec \
+                        sha256  db5e42d9ef9df9d4d40f42fd7c1b67cc2f5108a06e30ae983fd6e40046ad22e0 \
+                        size    457269
+
+set pyver               3.11
+set python_version      [string index ${pyver} 0][string range ${pyver} 2 end]
+
+patchfiles              CMakeLists-txt.diff
+
+depends_build-append    port:phosg \
```
harens, line 31: phosg should probably be a library dependency (i.e. required at both build time and runtime). I'm guessing the same is true for zlib as well.

devel/resource_dasm/Portfile
```diff
@@ -0,0 +1,37 @@
...
+                        port:zlib \
+                        port:netpbm \
```
harens, line 33: [Upstream docs](https://github.com/fuzziqersoftware/resource_dasm#building) note that "Also, this [Netpbm] is a runtime dependency only; you can install it later if you find that you need it, and you won't have to rebuild resource_dasm."

That can then be a runtime rather than build time dependency.

devel/resource_dasm/Portfile
```diff
@@ -0,0 +1,37 @@
...
+
+PortSystem              1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               compiler_blacklist_versions 1.0
+
+github.setup            fuzziqersoftware resource_dasm 46ad32e
+github.tarball_from     archive
+
+compiler.blacklist-append {clang < 1300}
```
harens, line 11: This might need some tweaking depending on what C++20 features (if any) are used, but we'll see what happens on the buildbots.

Would you mind adding a comment above the blacklist in both ports mentioning that this is for C++20 support? Thanks.

fracai: Thanks, I had meant to add such a comment, but neglected to add it before pushing.
Latest push addresses the dependency categories as well.

devel/resource_dasm/Portfile
fracai: Thanks. I noted that when building the Portfile and intended to add them under different categories, but simply forgot to get them sorted.

harens (approved):
Thank you for your contribution @fracai! 🎉

PR #18009, opened 2023-03-18
py-pytorch: Update to version 2.0.0

essandess:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.2.1 22D68 x86_64
Xcode 14.2 14C18

###### Verification 
Have you


essandess:
@reneeotten Like all versions of PyTorch, this PR build and runs fine locally, but the CI fails. Do you see why? I do not.

I don’t mind a couple rounds of CI debugging, but it would be good this major version improvement merged ASAP.

reneeotten:
> @reneeotten Like all versions of PyTorch, this PR build and runs fine locally, but the CI fails. Do you see why? I do not.
> 
> I don’t mind a couple rounds of CI debugging, but it would be good this major version improvement merged ASAP.

You can download the logs from the CI run and look at them. The error for all subports is:
```
2023-03-18T18:24:32.7624930Z CMake Error at third_party/fbgemm/CMakeLists.txt:76 (message):
2023-03-18T18:24:32.7638400Z   A compiler with AVX512 support is required.
```

essandess:
> ```
> 2023-03-18T18:24:32.7624930Z CMake Error at third_party/fbgemm/CMakeLists.txt:76 (message):
> 2023-03-18T18:24:32.7638400Z   A compiler with AVX512 support is required.
> ```

Is this an architecture issue with the CI box? My successful build uses:
```sh
:debug:build CXX='/usr/bin/clang++'
…
:info:build -- Performing Test C_HAS_AVX512_2 - Success
…
:info:build -- Performing Test CXX_HAS_AVX512_2 - Success
```

I believe the I've used the [right compiler](https://trac.macports.org/wiki/UsingTheRightCompiler). Any suggestions, or does this complicated build fail with the CI? Here's a successful build log:

[main.log.gz](https://github.com/macports/macports-ports/files/11012205/main.log.gz)

Dave-Allured:
From the CI build:
```
CLANG_VERSION_STRING:  Apple clang version 13.0.0 (clang-1300.0.29.30)
Performing Test C_HAS_AVX512_2 - Failed
```

Your private build:
```
CLANG_VERSION_STRING:  Apple clang version 14.0.0 (clang-1400.0.29.202)
Performing Test C_HAS_AVX512_2 - Success
```
You might try selecting a newer Clang version for the CI.  A quick check of Clang release notes shows that some extra AVX512 capability was added in Clang 14.0.  However, Clang 13 should have included core AVX512 support, so I am mystified.

I also see that the SDK versions differ between your builds.  I am not sure about Xcode versions.

Dave-Allured:
Pytorch release notes say that minimum supported python version is now 3.8.  So drop 3.7 from portfile.  This is probably unrelated to AVX512.

essandess:
@Dave-Allured, thanks for taking a look.

> A quick check of Clang release notes shows that some extra AVX512 capability was added in Clang 14.0. However, Clang 13 should have included core AVX512 support, so I am mystified.
> 
> I also see that the SDK versions differ between your builds. I am not sure about Xcode versions.

There is a long-standing problem with the CI building PT that predates these clang versions: https://github.com/macports/macports-ports/commits/54e660f99dd7681d05252a2e40478b6e66e5eb67/python/py-pytorch/Portfile. I don't believe that this is a version issue.

> Pytorch release notes say that minimum supported python version is now 3.8. So drop 3.7 from portfile. This is probably unrelated to AVX512.

Agreed, but that also means dropping the dependencies below, some of which do not currently build. That opens a dependency hell that will block an update of this import update, so I think we should defer until the dependencies all build.
```sh
port echo depends:py37-pytorch
py37-torchaudio                 
py37-torchvision                
py37-fairseq                    
py37-pytorch-lightning
```

reneeotten:
> > A quick check of Clang release notes shows that some extra AVX512 capability was added in Clang 14.0. However, Clang 13 should have included core AVX512 support, so I am mystified.
> > I also see that the SDK versions differ between your builds. I am not sure about Xcode versions.
> 
> There is a long-standing problem with the CI building PT that predates these clang versions: https://github.com/macports/macports-ports/commits/54e660f99dd7681d05252a2e40478b6e66e5eb67/python/py-pytorch/Portfile. I don't believe that this is a version issue.

there are no build logs anymore from these builds so it's hard to figure out whether or not this is the same issue. As @Dave-Allured suggests it could be that it is actually an issue with the Clang version. If it works for you with Clang 14.0 and doesn't work on the CI or buildbot with an earlier versions, please try to blacklist the older clangs and see if the CI works then.

> > Pytorch release notes say that minimum supported python version is now 3.8. So drop 3.7 from portfile. This is probably unrelated to AVX512.
> 
> Agreed, but that also means dropping the dependencies below, some of which do not currently build. That opens a dependency hell that will block an update of this import update, so I think we should defer until the dependencies all build.
> 
> ```shell
> port echo depends:py37-pytorch
> py37-torchaudio                 
> py37-torchvision                
> py37-fairseq                    
> py37-pytorch-lightning
> ```

If the current version doesn't work anymore with Python 3.7 it should be removed from the Portfile. There is no "dependency hell", you just need to remove the py37 subport from (recursive) dependents as well. Of course you can do locally whatever you want, but if you want this committed/merged there are certain "rules" to be followed.

Dave-Allured:
I was mistaken, it is not Clang 13.  Don't spend time on compiler version.

On closer inspection. the AVX512 error is coming from a config test in submodule [FBGEMM](https://github.com/pytorch/FBGEMM).  This module dynamically generates CPU-optimized runtime code.  More importantly, it runs CPU tests.  I suspect that the failing test is either a runtime CPU test done in the wrong context, or else the test is simply buggy or  misconfigured in current context.

I will keep looking, as time permits.  FBGEMM code is not contained within the pytorch main distro.  Rather, it is fetched as a third party submodule.  Also it depends on other third parties, such as "cpuinfo", that are also fetched independently.  This keeps me busy chasing down chunks of code to puzzle over.

essandess:
@reneeotten Would you please merge this upgrade?

reneeotten:
> @reneeotten Would you please merge this upgrade?

thank you for dropping the PY37 subport that is  no longer supported from the `py-pytorch` port and dependents. However, I don't see a solution for the build failures so to me it doesn't make sense to merge this at this point.

essandess:
We cannot block updates to this important port because the CI machine doesn’t build it. As I point out on the [dev](https://lists.macports.org/pipermail/macports-dev/2023-March/044964.html) list:

> PyTorch build and runs fine locally, but the CI consistently fails with error below. This issue has created unnecessary delays in merging new PT versions.

The failure is with the CI, not the port. The CI has **always** failed to build this port, but it builds and works fine locally.

Would someone please merge this major update to this important port?

> > @reneeotten Would you please merge this upgrade?
> 
> thank you for dropping the PY37 subport that is no longer supported from the `py-pytorch` port and dependents. However, I don't see a solution for the build failures so to me it doesn't make sense to merge this at this point.

reneeotten:
both @cjones051073 and @jmroot responded to that thread saying what should be fixed. @Dave-Allured pointed out in which package the issue occurs, which you can see from the logs as well. So this is not a problem with the CI - it's a problem with the port; it just happens to work on the latest macOS version. 

So why should I/we keep merging PRs that do not actually fix the problem that indeed has existed for a long time? I understand that it works for you but it will not work for anyone running an older OS.

essandess:
> both @cjones051073 and @jmroot responded to that thread saying what should be fixed. @Dave-Allured pointed out in which package the issue occurs, which you can see from the logs as well. So this is not a problem with the CI - it's a problem with the port; it just happens to work on the latest macOS version.
> 
> So why should I/we keep merging PRs that do not actually fix the problem that indeed has existed for a long time? I understand that it works for you but it will not work for anyone running an older OS.

This is not correct. No specific fix to the CI issue has been suggested.

Please do not cause unnecessary delays merging updates to important ports because of an issue with the CI. If you believe there is a fix to then CI build, please link to the fix. To the best of my knowledge none exists.

Dave-Allured:
Unfortunately I do not have the free time right now to further investigate the config problem within [FBGEMM](https://github.com/pytorch/FBGEMM).  Anyone else?

essandess:
> The port needs to be configured so that, by default, it does not attempt to use AVX512 (or AVX2) instructions.

The vast majority of PT users need these instructions for performance. Turning this off, if it’s possible, is a very bad idea.

jmroot:
The bare minimum change for this to be merged, if no one is able to find and fix the issue, would be to mark it as only supporting Ventura. That's still very bad, but it's not a regression and users would then have their expectations set appropriately.

cjones051073:
> > The port needs to be configured so that, by default, it does not attempt to use AVX512 (or AVX2) instructions.
> 
> The vast majority of PT users need these instructions for performance. Turning this off, if it’s possible, is a very bad idea.

The *default* build, i.e. the one provided by the build bots, needs to be one that assumes a low minimum hardware so will run on the largest cross section of machines. 

This though does not mean users that want to enable the more advanced cannot do it. The standard practise we have for this sort of thing in MacPorts is to add a `native` variant to the port, which is not enabled by default, but users are free do use it and doing so configures the build to then use more advanced instructions.

jmroot:
> The vast majority of PT users need these instructions for performance. Turning this off, if it’s possible, is a very bad idea.

It may be that all that is needed is a recent enough compiler that passes the AVX512 test used here. Maybe pytorch checks CPU capabilities at runtime. I don't know, but that needs to be investigated since it potentially affects the viability of binary distribution and the hardware requirements for the port.

cjones051073:
> > The vast majority of PT users need these instructions for performance. Turning this off, if it’s possible, is a very bad idea.
> 
> It may be that all that is needed is a recent enough compiler that passes the AVX512 test used here. Maybe pytorch checks CPU capabilities at runtime. I don't know, but that needs to be investigated since it potentially affects the viability of binary distribution and the hardware requirements for the port.

Indeed, it is an un-answered question as to how PyTorch (or the sub-modules it uses) enables support for these instructions. If done `properly` then it will be using runtime detection of the machine capabilities and enabling them on the fly. In which case its fine to allow the use by default, and it might just be a case of blacklisting compilers accordingly to force the use of more than supports these instructions.

If runtime detection is not used, then the only option is to do what I describe above, and disable the use of the instructions  at configuration time for the default builds.

jmroot:
> The CI build issue predates Ventura, so that’s not the issue

And yet on the buildbot, pytorch succeeds on Ventura and fails on every older OS version: https://ports.macports.org/port/py310-pytorch/builds/

cjones051073:
> The CI build issue predates Ventura, so that’s not the issue.
> 
> None of us know what causes this issue with the CI builds.
> 
> There is no constructive purpose to delay merging this update indefinitely for an unknown solution to an unknown problem with the CI. The existing and previous PT versions all have this issue, so let’s just merge the PR and have the issue with an up-to-date PR.

It is true this MR has not introduced the issue, but the fact it was either ignored, or not known about, until now is not a good reason to carry on ignoring it. Now it is known a fix should be found.

its also not true that its an unknown problem. The problem is clear

```
2023-03-26T12:47:27.8058950Z CMake Error at third_party/fbgemm/CMakeLists.txt:76 (message):
2023-03-26T12:47:27.8059440Z   A compiler with AVX512 support is required.
```

A simple test you could try right now is to see if blacklisting the Xcode compilers on macOS12 and older, and thus force the use of a recent macports-clang, help or not.

essandess:
> > > The port needs to be configured so that, by default, it does not attempt to use AVX512 (or AVX2) instructions.
> > 
> > 
> > The vast majority of PT users need these instructions for performance. Turning this off, if it’s possible, is a very bad idea.
> 
> The _default_ build, i.e. the one provided by the build bots, needs to be one that assumes a low minimum hardware so will run on the largest cross section of machines.
> 
> This though does not mean users that want to enable the more advanced instruction sets cannot do it. The standard practise we have for this sort of thing in MacPorts is to add a `native` variant to the port, which is not enabled by default, but users are free to use it and doing so configures the build to then use more advanced instructions.

Okay thank you. I’ll look into how to turn off `AVX512` outside of a new `native` variant.

cjones051073:
> > > > The port needs to be configured so that, by default, it does not attempt to use AVX512 (or AVX2) instructions.
> > > 
> > > 
> > > The vast majority of PT users need these instructions for performance. Turning this off, if it’s possible, is a very bad idea.
> > 
> > 
> > The _default_ build, i.e. the one provided by the build bots, needs to be one that assumes a low minimum hardware so will run on the largest cross section of machines.
> > This though does not mean users that want to enable the more advanced instruction sets cannot do it. The standard practise we have for this sort of thing in MacPorts is to add a `native` variant to the port, which is not enabled by default, but users are free to use it and doing so configures the build to then use more advanced instructions.
> 
> Okay thank you. I’ll look into how to turn off `AVX512` outside of a new `native` variant.

You should I think first look to answering @jmroot 's question, s if the build does things 'correctly' and uses runtime capabilities detection, then this native variant is not needed, and instead if might hopefully just be a matter of compiler blacklisting.

cjones051073 (changes requested):

jmroot:
What's really needed of course is the cmake log from a failed build.

Dave-Allured:
Up top, the original poster said:

> Is this an architecture issue with the CI box??

Ironically, you might be right in one sense.  But it shouldn't matter!  The compiler should be able to cross compile regardless of the local architecture.

@essandess, what is your local machine CPU type, and does it have AVX512 support?  Likewise, does anyone know the CPU type and AVX512 support on the MacOS-11 CI?  I don't find this in the build log.

cjones051073:
> Up top, the original poster said:
> 
> > Is this an architecture issue with the CI box??
> 
> Ironically, you might be right in one sense. But it shouldn't matter! The compiler should be able to cross compile regardless of the local architecture.
> 
> @essandess, what is your local machine CPU type, and does it have AVX512 support? Likewise, does anyone know the CPU type and AVX512 support on the MacOS-11 CI? I don't find this in the build log.

Remember the issue is not just in the GitHub CI builds. Its likely the buildbots are seeing the same (and quite possibly for the same reasons as they are also I believe VMs). For these though you can ask @ryandesign for more details.

Dave-Allured:
@cjones051073, agreed, buildbots too.  I confused the two, not being very familiar with Macports systems.

Dave-Allured:
> The file should be patched to remove that line

**"-Werror"** -- That's debatable.  This supports a "clean build" policy, where the authors wish to deal immediately with all possible new warnings as they arise.  If you keep this flag, you are more dependent on quick turnaround from upstream, but you also catch potential problems sooner.

Personally, I use this one all the time for my local code development.  But there I am able to respond immediately to anything new.

Dave-Allured:
How to check your AVX512 hardware support on Mac:

`sysctl -a | & grep -i avx`

Look for **hw.optional.avx512f**.  This is the modern standard AVX512 core capability.  1 = supported, 0 = not supported.  I get 0 on my 2019 Macbook Pro.

jmroot:
Building with `-Werror` when you are the developer is a great idea. Building with no warnings on your dev machine and CI is something to aspire to. Building with no warnings on every machine with every version of every compiler, on the other hand, is practically impossible. In the context of a distro like MacPorts that builds from source on end-user systems, `-Werror` is empirically observed to cause new build failures with new compilers pretty much constantly, whereas I can't think of a single time that it has alerted us to an actual bug that had been hitherto unnoticed.

jmroot:
@Dave-Allured My advice to patch it out was only intended to refer to the port. I certainly wouldn't want upstream to remove it entirely (though a configure option to control it would be convenient.)

essandess:
> > @reneeotten Like all versions of PyTorch, this PR build and runs fine locally, but the CI fails. Do you see why? I do not.
> > I don’t mind a couple rounds of CI debugging, but it would be good this major version improvement merged ASAP.
> 
> You can download the logs from the CI run and look at them. The error for all subports is:
> 
> ```
> 2023-03-18T18:24:32.7624930Z CMake Error at third_party/fbgemm/CMakeLists.txt:76 (message):
> 2023-03-18T18:24:32.7638400Z   A compiler with AVX512 support is required.
> ```

Dave-Allured:
> **"-Werror"** --

> Personally, I use this one all the time for my local code development.

Ouch.  I lied.  I gave up on **"-Werror"** a couple years ago on larger projects, after being overwhelmed by "function parameter not used" and unwarranted "might be used uninitialized" and such out of recent GCC.  It is still good practice in controlled situations.

python/py-pytorch/Portfile
```diff
@@ -172,6 +177,14 @@ if {${name} ne ${subport}} {
                     USE_ZMQ=ON \
                     USE_ZSTD=OFF
 
+    # https://github.com/macports/macports-ports/pull/18009#issuecomment-1490276882
+    set fbgemm_minimum_supported_major_version 22
+    if {${os.platform} eq {darwin}
+        && ${os.major} < ${fbgemm_minimum_supported_major_version}} {
+        build.env-append \
+                    USE_FBGEMM=OFF
```
jmroot, line 185: Shouldn't we check if the removal of `-Werror` fixes the fbgemm configure check, before resorting to disabling entirely?

essandess:
I made two changes in the latest force-push:
* Commented out the `-Werror` line in https://github.com/pytorch/FBGEMM/blob/03b2046676707da64504e898490ab46104d4682a/CMakeLists.txt#L147
* Added `USE_FBGEMM=OFF` to `build.env` for OS's prior to Ventura.

I do not have access to an architecture that doesn't have `AVX512 ` support, so I can only test the ideas above by submitting amended commits to the CI.

python/py-pytorch/Portfile
essandess: `AVX512` support is required for `FBGEMM`, and @jmroot's [comment](https://github.com/macports/macports-ports/pull/18009#issuecomment-1490316527) above suggests that such support only exists on MacPorts's Ventura machines.

https://github.com/pytorch/FBGEMM/blob/177ba08dbc3bae56d11e4c89bdf19b04974374dc/CMakeLists.txt#L91

jmroot: The check that is failing is just testing whether compiling something with `-mavx512f` succeeds. My hypothesis is that there may be a warning generated in the test compile which is being turned into an error and making the test fail.

essandess: Okay, thanks for the explanation. I'll let this CI complete then try a version with this part removed.

essandess: > My hypothesis is that there may be a warning generated in the test compile which is being turned into an error and making the test fail.

Also, if your hypothesis is correct, wouldn't that imply a bug in upstream's `cmake` code?

jmroot: Yes, it would.

python/py-fairseq/Portfile
```diff
@@ -23,7 +23,7 @@ checksums           rmd160  05969920d445518f940eb3bc0bb4fefe74e64d05 \
                     sha256  a34ccef8f779da9dd878fd6d21633de8af8573075e6ef087fd945598cf170c92 \
                     size    12042596
 
-python.versions     37 38 39 310
```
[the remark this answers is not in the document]
essandess: These `py311-pytorch` dependencies don't exist:
* `py311-gmpy`
* `py311-onnx`
* `py311-protobuf3`

And a simple test of trying to install  `py311-gmpy` shows that it fails to build:
```sh
:info:build src/gmpy.c:243:10: fatal error: 'longintrepr.h' file not found
:info:build #include "longintrepr.h"
:info:build          ^~~~~~~~~~~~~~~
:info:build error: command '/usr/bin/clang' failed with exit code 1
```

The issues in this PR are complicated enough without addressing Python 311 dependency issues.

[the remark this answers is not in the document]
jmroot: It's actually not clear why the gmpy dependency is in the port. I don't see it documented as needed anywhere obvious upstream.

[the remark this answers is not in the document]
essandess: `py311-mkl` doesn't build. Perhaps this issue.

https://trac.macports.org/ticket/66778

Too much going on already to address Python 311 in this PR.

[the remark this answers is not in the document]
jmroot: Sure, I agree that py311 can be handled in a later PR.

jmroot:
> What's really needed of course is the cmake log from a failed build.

I've made a change to the CI config that should allow us to get at the `CMakeError.log`.

essandess:
Are these CI failures timeouts or errors? I don't see any build error messages in the logs to which I have access.

jmroot:
> Are these CI failures timeouts or errors? I don't see any build error messages in the logs to which I have access.

Errors. The macOS 12 build was cancelled early though because the macOS 11 one failed.

jmroot:
The error for `py310-pytorch` now is:
```
2023-03-30T23:36:40.0000490Z [ 11%] Building CXX object third_party/googletest/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o
2023-03-30T23:36:40.0344950Z error: unknown warning option '-Werror=bitwise-instead-of-logical' [-Werror,-Wunknown-warning-option]
2023-03-30T23:36:40.0347400Z make[2]: *** [third_party/googletest/googletest/CMakeFiles/gtest.dir/src/gtest-all.cc.o] Error 1
2023-03-30T23:36:40.0364060Z make[1]: *** [third_party/googletest/googletest/CMakeFiles/gtest.dir/all] Error 2
```

Dave-Allured:
> -Werror=bitwise-instead-of-logical

Ouch.  This is a new warning added in Clang 14.0.  I was wrong when I said it's not the compiler.

[https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#major-new-features](https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#major-new-features)

So, something easy to try would be to blacklist Clang 13.  Any second opinion?

Now I wonder if that was also the problem with FBGEMM.

Dave-Allured:
Forensics.
Clang 13.0.0 was released on 2021 Sep 30.
Clang 14.0.0 was released on 2022 Mar 23.
Clang 15.0.0 was released on 2022 Sep 6.
PyTorch 2.0 was released on 2023 Mar 15.
Clang 16.0.0 was released on 2023 Mar 20.

cjones051073:
Please try adding this

```
compiler.blacklist-append {clang < 1400} {macports-clang-1[0-4]}
```

That will blacklist all but the clang version in the latest Xcode, and also allow only macports-clang-15/16 as the fallbacks (15 is the default so is what will be picked unless otherwise blacklisted).

jmroot:
Let's see how this CI run goes first.

jmroot:
Now that looks like a failure due to CI limitations. I assume that means it got a lot further at the very least, so might as well merge and see how it goes on the buildbot.

ryandesign:
> And yet on the buildbot, pytorch succeeds on Ventura and fails on every older OS version: https://ports.macports.org/port/py310-pytorch/builds/

I doubt the OS version is relevant; it's probably the CPU model. Our macOS 12 and earlier x86_64 buildbot workers run on 2009 Xserve3,1 machines with Intel Xeon Nehalem processors which predate the existence of AVX-512. The macOS 13 x86_64 buildbot worker is temporarily running on a 2016 MacBookPro13,3 with an Intel Core i7 Skylake processor which supports AVX-512.

jmroot:
The OS version was indirectly relevant because the newer compiler supports the warning option that was causing problems due to -Werror. Those build issues are fixed now in any case (though there's a different one on < 10.15). Runtime CPU requirements are still an open question.

PR #18075, opened 2023-03-24
repo: update to 2.32

armatys:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 13.2.1 22D68 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

- [x] ~~tried a full install with `sudo port -vst install`~~ tried a full install with `sudo port -vs install`?

### DIFF
--- a/python/repo/Portfile
+++ b/python/repo/Portfile
@@ -5,7 +5,7 @@ PortGroup           github 1.0
 
 name                repo
 epoch               20160202
-github.setup        GerritCodeReview git-repo 2.26 v
+github.setup        GerritCodeReview git-repo 2.32 v
 revision            0
 license             Apache-2
 categories          python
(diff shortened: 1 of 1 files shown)


reneeotten (commented):
it seems we don't explicitly depend on a Python port here - is that what we want to do? It assumes that for this to work the system should have Python 3.5+ installed, I don't think that's always the case on newer systems? I *think* newer macOS versions only have "python3" which wouldn't be used here - or?

In other words, should we just depend on MacPorts' default Python versions?

armatys:
Yes, when I installed the `repo` on its own, and then tried to run it, it would complain: `env: python: No such file or directory`.
So it probably makes sense to depend on default Python version as well.

Would adding `python311` to `depends_run` be enough (or is there a way to use a default version, whatever that version might be)?

ryandesign:
> Would adding `python311` to `depends_run` be enough

No. You would need to tell the build system (if there is one, or write a patch and/or reinplace if there is not) to use that specific version of python as well.

> (or is there a way to use a default version, whatever that version might be)?

There is not.

herbygillot:
I've updated this Portfile to modify `repo` to use MacPort's `python311`. Should be good now.

PR #18147, opened 2023-03-31
pooler: compile with OpenMP so multi-core works

ssb22:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 13.2.1 22D68 x86_64
Xcode 14.2 14C18

###### Verification 
Have you


kencu:
thanks,

I see you’re the author of the software, thanks for taking the time for MacPorts.

There are some MacPorts idiosyncracies we use for openmp and compiler selection that someone will help with. The way you’ve done it works, but we have a somewhat more general approach.

science/pooler/Portfile
```diff
@@ -27,6 +27,9 @@ maintainers         {@ssb22} \
                     openmaintainer
 
 build.dir           ${worksrcpath}/pooler
+depends_lib         port:libomp
+depends_build       port:clang-16
+configure.cc        clang-mp-16
```
chrstphrchvz, line 32: Please see https://trac.macports.org/wiki/CompilerSelection#Parallelism and https://trac.macports.org/wiki/CompilerSelection#GettingtheRightCompilerinthePortfile

`compiler.openmp_version …` should be set instead; that should automatically take care of the compiler and libomp dependencies. Depending on what OpenMP version is needed, you may also need to specify `configure.cppflags-append -fopenmp-version=…`.

ssb22: Thanks, now updated to use `compiler.openmp_version` instead

PR #18192, opened 2023-04-05
clang-*: fix use of -stdlib=macports-libstdc++ on newer machines

MarcusCalhoun-Lopez:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/lang/llvm-16/Portfile
+++ b/lang/llvm-16/Portfile
@@ -28,7 +28,7 @@ version                 ${llvm_version}.0.1
 name                    llvm-${llvm_version}
 revision                0
 subport                 mlir-${llvm_version}  { revision [ expr ${revision} + 0 ] }
-subport                 clang-${llvm_version} { revision [ expr ${revision} + 0 ] }
+subport                 clang-${llvm_version} { revision [ expr ${revision} + 1 ] }
 subport                 lldb-${llvm_version}  { revision [ expr ${revision} + 0 ] }
 subport                 flang-${llvm_version} { revision [ expr ${revision} + 0 ] }
 
@@ -508,17 +508,10 @@ if {${subport} eq "clang-${llvm_version}"} {
         patchfiles-append 0005-clang-support-macports-libstdcxx.patch
 
         post-patch {
-            if { ${configure.build_arch} eq "arm64" } {
-                # see https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483
-                set gcc_arch    aarch64
-            } else {
-                set gcc_arch    ${configure.build_arch}
-            }
-
             reinplace "s|@@MACPORTS_GCC_INCLUDE_DIR@@|${prefix}/include/gcc/c++|g" \
                 ${patch.dir}/clang/lib/Lex/InitHeaderSearch.cpp \
                 ${patch.dir}/clang/lib/Driver/ToolChains/Darwin.cpp
-            reinplace "s|@@MACPORTS_HOST_NAME@@|${gcc_arch}-apple-darwin${os.major}|g" \
+            reinplace "s|@@MACPORTS_HOST_NAME@@|${configure.build_arch}-apple-darwin${os.major}|g" \
                 ${patch.dir}/clang/lib/Driver/ToolChains/Darwin.cpp
             reinplace "s|@@MACPORTS_libstdc++@@|${prefix}/lib/libgcc/libstdc++.6.dylib|g" \
                 ${patch.dir}/clang/lib/Driver/ToolChains/Darwin.cpp
--- a/lang/clang-11-bootstrap/Portfile
+++ b/lang/clang-11-bootstrap/Portfile
@@ -8,7 +8,7 @@ PortGroup           muniversal 1.0
 name                clang-11-bootstrap
 
 version             11.1.0
-revision            3
+revision            4
 epoch               0
 
 platforms           darwin
@@ -300,17 +300,10 @@ if {${os.platform} eq "darwin" && ${os.major} < 18} {
 }
 
 post-patch {
-    if { ${configure.build_arch} eq "arm64" } {
-        # see https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483
-        set gcc_arch    aarch64
-    } else {
-        set gcc_arch    ${configure.build_arch}
-    }
-
     reinplace "s|@@MACPORTS_GCC_INCLUDE_DIR@@|${prefix}/include/gcc/c++|g" \
         ${worksrcpath}/tools/clang/lib/Frontend/InitHeaderSearch.cpp \
         ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
-    reinplace "s|@@MACPORTS_HOST_NAME@@|${gcc_arch}-apple-darwin${os.major}|g" \
+    reinplace "s|@@MACPORTS_HOST_NAME@@|${configure.build_arch}-apple-darwin${os.major}|g" \
         ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
     reinplace "s|@@MACPORTS_libstdc++@@|${prefix}/lib/libgcc/libstdc++.6.dylib|g" \
         ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
--- a/lang/llvm-10/Portfile
+++ b/lang/llvm-10/Portfile
@@ -12,7 +12,7 @@ set clang_executable_version 10
 set lldb_executable_version 10
 name                    llvm-${llvm_version}
 revision                3
-subport                 clang-${llvm_version} { revision 7 }
+subport                 clang-${llvm_version} { revision 8 }
 subport                 lldb-${llvm_version} { revision 3 }
 set suffix              mp-${llvm_version}
 set sub_prefix          ${prefix}/libexec/llvm-${llvm_version}
@@ -562,17 +562,10 @@ if {${subport} eq "llvm-${llvm_version}"} {
         patchfiles-append 9001-macports-libstdcxx.diff
 
         post-patch {
-            if { ${configure.build_arch} eq "arm64" } {
-                # see https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483
-                set gcc_arch    aarch64
-            } else {
-                set gcc_arch    ${configure.build_arch}
-            }
-
             reinplace "s|@@MACPORTS_GCC_INCLUDE_DIR@@|${prefix}/include/gcc/c++|g" \
                 ${worksrcpath}/tools/clang/lib/Frontend/InitHeaderSearch.cpp \
                 ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
-            reinplace "s|@@MACPORTS_HOST_NAME@@|${gcc_arch}-apple-darwin${os.major}|g" \
+            reinplace "s|@@MACPORTS_HOST_NAME@@|${configure.build_arch}-apple-darwin${os.major}|g" \
                 ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
             reinplace "s|@@MACPORTS_libstdc++@@|${prefix}/lib/libgcc/libstdc++.6.dylib|g" \
                 ${worksrcpath}/tools/clang/lib/Driver/ToolChains/Darwin.cpp
--- a/lang/llvm-11/Portfile
+++ b/lang/llvm-11/Portfile
@@ -26,7 +26,7 @@ checksums               rmd160  f566b4b75c8f30418f19069a9a84864ead766401 \
 
 name                    llvm-${llvm_version}
 revision                4
-subport                 clang-${llvm_version} { revision 6 }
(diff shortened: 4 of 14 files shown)


cjones051073:
Not sure I see the need for the new 'port' `libstdcxx_clang_fix` ? why cannot you use legacy-support for this ? Seems like it would fit there just perfectly ?

lang/llvm-16/Portfile
```diff
@@ -506,19 +506,13 @@ if {${subport} eq "clang-${llvm_version}"} {
...
+        depends_run-append  port:libstdcxx_clang_fix
 
         post-patch {
-            if { ${configure.build_arch} eq "arm64" } {
-                # see https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483
-                set gcc_arch    aarch64
-            } else {
-                set gcc_arch    ${configure.build_arch}
-            }
-
```
cjones051073, line 517: Not sure I see why you can just remove this ? If the comment in

https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483

that aarch64 should be used on ARM no longer valid for some reason ?

cjones051073: As @michaelld committed the above he should comment on if it is now valid to just remove this...

cjones051073:
So.... The more I look at the changes here the less I like them. I really am not a fan of this new port that installs a fix up header. As per the discussion in

https://trac.macports.org/ticket/61778

I would like to understand why the math.h fixes in  legacysupport, which on the fact of it should do exactly what is needed, are not working.

https://github.com/macports/macports-legacy-support/blob/master/include/math.h

cjones051073:
b.t.w. as llvm-15 and newer (so currently 16 and devel) no longer supports being built on OSX10.6 and older, and as your proposed changes here can only have an effect on those old platforms, its completely pointless to make any fixes to these newer clang versions, so any changes to these newer compilers should, at bear minimum, be removed.

That said, I am still quite dubious as to if the approach here is the right why at all, as I think first more effort is needed to understand why what legacysupport does is not good enough to address the issue, as I think the intention is it should.

kencu:
you can also see 

https://github.com/macports/macports-legacy-support/pull/51

if you like.

kencu:
@catap was working on this, but ran out of steam

MarcusCalhoun-Lopez:
If the. compiler is not working properly, the fix should be made there.
LegacySupport may need work, but this PR is a separate issue.
`echo "#include <cmath>" | clang++-mp-7.0 -c -stdlib=macports-libstdc++ -std=c++11 -xc++ -`
does not run on older systems.
The compiler should work for such a trivial program as `#include <cmath>` *without* LegacySupport.

lang/llvm-16/Portfile
MarcusCalhoun-Lopez: https://github.com/macports/macports-ports/commit/be1e11a368f672d927a7bdb381f2fa71a79ba483
was reverted in
https://github.com/macports/macports-ports/commit/ed352751afc52a036f329e7d34d418d367cf6cb5
I am not sure why the change was reverted, but currently `-stdlib=macports-libstdc++` does not work on Apple Silicon.

MarcusCalhoun-Lopez:
> b.t.w. as llvm-15 and newer (so currently 16 and devel) no longer supports being built on OSX10.6 and older, and as your proposed changes here can only have an effect on those old platforms, its completely pointless to make any fixes to these newer clang versions, so any changes to these newer compilers should, at bear minimum, be removed.

The changes for Apple Silicon still need to be included regardless of the situation with older systems.
The only reason both sets of changes are included in one PR is to avoid multiple revision increases.

Is there any hope for newer Clang versions to someday working on older systems?
If so, libstdcxx_clang_fix will still be needed.
On newer systems, libstdcxx_clang_fix is just a stub port.

lang/llvm-16/Portfile
cjones051073: TBH the whole variant that enables `macports-libstdc++` should only be active on the system that actually needs it. Currently its on by default across the board, which I also don't think is correct.

What systems actually need to be able to use `macports-libstdc++` ?

cjones051073:
> > b.t.w. as llvm-15 and newer (so currently 16 and devel) no longer supports being built on OSX10.6 and older, and as your proposed changes here can only have an effect on those old platforms, its completely pointless to make any fixes to these newer clang versions, so any changes to these newer compilers should, at bear minimum, be removed.
> 
> The changes for Apple Silicon still need to be included regardless of the situation with older systems. The only reason both sets of changes are included in one PR is to avoid multiple revision increases.

I would very much prefer they where in different MRs, particularly as I strongly system there shouldn't be any overlap in the two. i.e. the systems that actually needs macports-libstdc++ (old) and those that support apple silicon (new) probably do not overlap, but some margin...

> 
> Is there any hope for newer Clang versions to someday working on older systems? If so, libstdcxx_clang_fix will still be needed. On newer systems, libstdcxx_clang_fix is just a stub port.

Unlikely ptr this time, but who can say in the future.

Even if it is a stub port, I would still prefer it wasn't a dep at all on systems that do not need it.

lang/llvm-16/Portfile
MarcusCalhoun-Lopez: > TBH the whole variant that enables `macports-libstdc++` should only be active on the system that actually needs it. Currently its on by default across the board, which I also don't think is correct.
> 
> What systems actually need to be able to use `macports-libstdc++` ?

Much has changed with the libc++ vs libstdc++ issue since `variant libstdcxx` was first created.
So it is entirely possible that the use of `macports-libstdc++` can be significantly reduced.
Default or not however, the variant is broken on Apple Silicon machines.

cjones051073:
> > I would very much prefer they where in different PRs, particularly as I strongly suspect there shouldn't be any overlap in the two. i.e. the systems that actually needs macports-libstdc++ (old) and those that support apple silicon (new) probably do not overlap, but some margin...
> 
> I would be happy to put the changes in two pull requests. We just need to make sure that the PR that increases the revision number is merged after the one that does not.
> 
> > Even if it is a stub port, I would still prefer it wasn't a dep at all on systems that do not need it.
> 
> I will make the change.

To my eye the variant enabling macports-libstdc++ should be turned off (and frankly not even defined) on newer systems that do not need it. What I do not know is quite where that cut off is. Does this variant have any use ion anything newer than OSX10.6 ?

MarcusCalhoun-Lopez:
> To my eye the variant enabling macports-libstdc++ should be turned off (and frankly not even defined) on newer systems that do not need it. What I do not know is quite where that cut off is. Does this variant have any use ion anything newer than OSX10.6 ?

I am looking through the history to see where the cutoff is.
Removing the variant entirely gets into the whole issue of support for older system, which is important but should include the larger community and is beyond the scope of this pull request.

cjones051073:
> > To my eye the variant enabling macports-libstdc++ should be turned off (and frankly not even defined) on newer systems that do not need it. What I do not know is quite where that cut off is. Does this variant have any use ion anything newer than OSX10.6 ?
> 
> I am looking through the history to see where the cutoff is. Removing the variant entirely gets into the whole issue of support for older system, which is important but should include the larger community and is beyond the scope of this pull request.

I am not suggesting removing it entirely. Just 

```
if {${os.major} < N} {
     variant libstdcxx { .... }
    default_variants-append +libstdcxx
}
```

As long as N is chosen at the point where there is never any need to use macports-libstdc++, its then IMHO it is better to do the above as opposed to

```
variant libstdcxx { .... }
if {${os.major} < N} {
    default_variants-append +libstdcxx
}
```

As it removes the ability to accidentally enable it when it does no good. But if others prefer the second is fine. Way better than what we do now which is enable it across the board, which is really not required.

cjones051073:
> If the. compiler is not working properly, the fix should be made there.
> LegacySupport may need work, but this PR is a separate issue.
> `echo "#include <cmath>" | clang++-mp-7.0 -c -stdlib=macports-libstdc++ -std=c++11 -xc++ -`
> does not run on older systems.
> The compiler should work for such a trivial program as `#include <cmath>` _without_ LegacySupport.

The above command runs just fine. It just fails to compile the code, which is exactly what I would expect *without* legacy support, as the system math.h does not have those prototypes defined.

This is *exactly* the sort of thing legacy support is there to fix, so I completely disagree that this is something to fix in the compilers. The fix here should be in the legacy port library, no where else.

MarcusCalhoun-Lopez:
> > If the. compiler is not working properly, the fix should be made there.
> > LegacySupport may need work, but this PR is a separate issue.
> > `echo "#include <cmath>" | clang++-mp-7.0 -c -stdlib=macports-libstdc++ -std=c++11 -xc++ -`
> > does not run on older systems.
> > The compiler should work for such a trivial program as `#include <cmath>` _without_ LegacySupport.
> 
> The above command runs just fine. It just fails to compile the code, which is exactly what I would expect _without_ legacy support, as the system math.h does not have those prototypes defined.
> 
> This is _exactly_ the sort of thing legacy support is there to fix, so I completely disagree that this is something to fix in the compilers. The fix here should be in the legacy port library, no where else.

LegacySupports is used only when a program program **needs** a missing function.
A program that makes no use of `llrint` in any way should not have an error related to `llrint`.
This is an error with the compiler.

Also, fixing the compiler in this case is the precedent the compilers have already established.
We already have [a patch for Clang](https://github.com/macports/macports-ports/blob/master/lang/llvm-devel/files/0011-Fix-missing-long-long-math-prototypes-when-using-the.patch) that essentially does what is proposed in this PR.
GCC also provides the missing prototypes in, e.g., `/opt/local/lib/gcc7/gcc/i386-apple-darwin9/7.5.0/include-fixed/architecture/i386/math.h`.

MarcusCalhoun-Lopez:
As requested, I have broken the issues discussed here into three pull requests.
This one fixes issues on newer machines.
https://github.com/macports/macports-ports/pull/18203/ fixes issues on older machines.
https://github.com/macports/macports-ports/pull/18204 limits these issues entirely.

cjones051073:
> As requested, I have broken the issues discussed here into three pull requests.
> This one fixes issues on newer machines.
> #18203 fixes issues on older machines.
> #18204 limits these issues entirely.

Thanks, but looking now at the changes and taking https://github.com/macports/macports-ports/pull/18204 into account I do not think the changes here actually have any material effect. The only really change is to remove the build_arch workaround, inside the libstdcxx variant. This change only has any effect for arm machines, on intel it makes no difference. However, with https://github.com/macports/macports-ports/pull/18204 I suspect this variant will in practise now never be used on arm machines, as I doubt anyone is going too set ${cxx_stdlib} to "libstdc++" on these machines. So in practise its likely the changes here are a no op.

cjones051073:
> #18203

https://github.com/macports/macports-ports/pull/18203 is in fact the PR that fixes macports-libstdc++ on older machines.

Could you update the PR title and commit messages here to properly say what this PR now does please.

MarcusCalhoun-Lopez:
> #18203 is in fact the PR that fixes macports-libstdc++ on older machines.
> 
> Could you update the PR title and commit messages here to properly say what this PR now does please.

I appreciate you taking the time to evaluate these changes.
I would humbly suggest that this PR is more than cleanup.

On an Apple Silicon machine,
`echo "#include <cmath>" | clang++-mp-15 -c -stdlib=macports-libstdc++ -std=c++11 -xc++ - -v`
gives the following error:
```
...
clang -cc1 version 15.0.7 based upon LLVM 15.0.7 default target arm64-apple-darwin22.3.0
ignoring nonexistent directory "/usr/local/include"
ignoring nonexistent directory "/opt/local/include/gcc/c++/aarch64-apple-darwin22"
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/local/include"
ignoring nonexistent directory "/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/Library/Frameworks"
#include "..." search starts here:
#include <...> search starts here:
 /opt/local/include/gcc/c++
 /opt/local/include/gcc/c++/backward
 /opt/local/libexec/llvm-15/lib/clang/15.0.7/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include
 /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks (framework directory)
End of search list.
In file included from <stdin>:1:
/opt/local/include/gcc/c++/cmath:41:10: fatal error: 'bits/c++config.h' file not found
#include <bits/c++config.h>
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```

The reason for the error is that the search path:
`/opt/local/include/gcc/c++/aarch64-apple-darwin22`
should be
`/opt/local/include/gcc/c++/arm64-apple-darwin22`
because of how GCC [now installs headers](https://github.com/macports/macports-ports/commit/ed352751afc52a036f329e7d34d418d367cf6cb5#diff-d5df50b875863519e7b254a6c43b789d34a7deb4838ce332f8f1f2c49bb6bb22L99).

This PR attempts to fix the problem.
If the commit message and title are unclear, I would be happy to change them, but the PR is a fix on newer machines.

Of course, you are right that the number of people who actually need this fix is vanishingly small.

PR #18204, opened 2023-04-06
clang-*: default to +libstdcxx only when needed

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/lang/llvm-10/Portfile
+++ b/lang/llvm-10/Portfile
@@ -606,7 +606,9 @@ if {${subport} eq "llvm-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 
     post-patch {
         reinplace "s|@@PREFIX@@|${prefix}|" \
--- a/lang/llvm-11/Portfile
+++ b/lang/llvm-11/Portfile
@@ -604,7 +604,9 @@ if {${subport} eq "llvm-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 
     post-patch {
         reinplace "s|@@PREFIX@@|${prefix}|" \
--- a/lang/llvm-12/Portfile
+++ b/lang/llvm-12/Portfile
@@ -462,7 +462,9 @@ if {${subport} eq "clang-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 }
 
 livecheck.type      regex
--- a/lang/llvm-13/Portfile
+++ b/lang/llvm-13/Portfile
@@ -533,7 +533,9 @@ if {${subport} eq "clang-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 }
 
 livecheck.type      regex
--- a/lang/llvm-14/Portfile
+++ b/lang/llvm-14/Portfile
@@ -546,7 +546,9 @@ if {${subport} eq "clang-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 }
 
 livecheck.type      regex
--- a/lang/llvm-15/Portfile
+++ b/lang/llvm-15/Portfile
@@ -552,7 +552,9 @@ if {${subport} eq "clang-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 }
 
 livecheck.type      regex
--- a/lang/llvm-16/Portfile
+++ b/lang/llvm-16/Portfile
@@ -552,7 +552,9 @@ if {${subport} eq "clang-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 }
 
 livecheck.type      regex
--- a/lang/llvm-5.0/Portfile
+++ b/lang/llvm-5.0/Portfile
@@ -606,7 +606,9 @@ if {${subport} eq "llvm-${llvm_version}"} {
             }
         }
     }
-    default_variants-append +libstdcxx
+    if { ${cxx_stdlib} eq "libstdc++" } {
+        default_variants-append +libstdcxx
+    }
 
     post-patch {
         reinplace "s|@@PREFIX@@|${prefix}|" \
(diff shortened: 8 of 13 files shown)


cjones051073:
Can you remind me what the relationship is between `${configure.cxx_stdlib}` and `${cxx_stdlib` ?

MarcusCalhoun-Lopez:
> Can you remind me what the relationship is between `${configure.cxx_stdlib}` and `${cxx_stdlib` ?

`${cxx_stdlib}` can be [set by the user ](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/macports1.0/macports.tcl#L858) but [defaults](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/macports1.0/macports.tcl#L1247) to either `libc++` or `libstdc++`, depending on `${os.major}`.

`${configure.cxx_stdlib}` is an [option](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/port1.0/portutil.tcl#L194) that can be set in the Portfile but [defaults](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/port1.0/portconfigure.tcl#L137) to `libc++`, `libstdc++`, or `macports-libstdc++`, depending on `${cxx_stdlib}` and `${compiler.cxx_standard}`.

cjones051073:
> > Can you remind me what the relationship is between `${configure.cxx_stdlib}` and `${cxx_stdlib` ?
> 
> `${cxx_stdlib}` can be [set by the user ](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/macports1.0/macports.tcl#L858) but [defaults](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/macports1.0/macports.tcl#L1247) to either `libc++` or `libstdc++`, depending on `${os.major}`.
> 
> `${configure.cxx_stdlib}` is an [option](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/port1.0/portutil.tcl#L194) that can be set in the Portfile but [defaults](https://github.com/macports/macports-base/blob/564391982b671dfc2a30ddd001149cd995bdde25/src/port1.0/portconfigure.tcl#L137) to `libc++`, `libstdc++`, or `macports-libstdc++`, depending on `${cxx_stdlib}` and `${compiler.cxx_standard}`.

Sorry, Was not clear. The above I know, what I am wondering is if there is a (valid) scenario where `${cxx_stdlib}` is set to `libc++` (which of course is the vast majority of cases now) but a port could decide to set `${configure.cxx_stdlib}` to `libstdc++`, or `macports-libstdc++` ? if thats never a valid combination, then your changes look look just fine to me.

MarcusCalhoun-Lopez:
> Sorry, Was not clear. The above I know, what I am wondering is if there is a (valid) scenario where `${cxx_stdlib}` is set to `libc++` (which of course is the vast majority of cases now) but a port could decide to set `${configure.cxx_stdlib}` to `libstdc++`, or `macports-libstdc++` ? if thats never a valid combination, then your changes look look just fine to me.

Sorry for my confusion.
I cannot think of a scenario as you describe.
I would be happy to wait, but is this PR ready to merge?

cjones051073:
> > Sorry, Was not clear. The above I know, what I am wondering is if there is a (valid) scenario where `${cxx_stdlib}` is set to `libc++` (which of course is the vast majority of cases now) but a port could decide to set `${configure.cxx_stdlib}` to `libstdc++`, or `macports-libstdc++` ? if thats never a valid combination, then your changes look look just fine to me.
> 
> Sorry for my confusion. I cannot think of a scenario as you describe. I would be happy to wait, but is this PR ready to merge?

I think so but I would like @kencu thoughts here.

One thing to note, I think even without a revision or version change merging this on its own will trigger a number of port rebuilds in the buildbots, because ports using these compilers to build will start to request a different combination of variants, the binaries for which will thus need to be made. So I think it makes sense to even if this is ready merge this simultaneously with https://github.com/macports/macports-ports/pull/18192 and https://github.com/macports/macports-ports/pull/18203 to minimise the number of rebuilds.

MarcusCalhoun-Lopez:
> I think so but I would like @kencu thoughts here.
> 
> One thing to note, I think even without a revision or version change merging this on its own will trigger a number of port rebuilds in the buildbots, because ports using these compilers to build will start to request a different combination of variants, the binaries for which will thus need to be made. So I think it makes sense to even if this is ready merge this simultaneously with #18192 and #18203 to minimise the number of rebuilds.

Thanks, I had forgotten about how the buildbots would be affected.

cjones051073:
> If you commit this, then the default macports’ clangs will no longer support using -stdlib=macports-libstdc++.
> 
> Is that what you’re intending?

It will not be supported by default on platforms where libc++ is the default.

The main question for me is where do we *need* to support this option. My feeling is currently we enable it in way more places where it is actually needed.

kencu:
If that is the intention, then this PR would accomplish that.

macports-clang-N would then only sometimes support -stdlib=macports-libstdc++, depending on the users setting when it is built.

This will probably get messy quickly... if you're going down this route, perhaps it is time to consider having MacPorts drop all support for systems < 10.9 then?

cjones051073:
> If that is the intention, then this PR would accomplish that.
> 
> macports-clang-N would then only sometimes support -stdlib=macports-libstdc++, depending on the users setting when it is built.
> 
> This will probably get messy quickly... if you're going down this route, perhaps it is time to consider having MacPorts drop all support for systems < 10.9 then?

Isn't the *default* now libc++ on 10.6 ? On these systems the changes shouldn't have much effect, as libc++ is used in the vast majority of ports. 

One even older systems where libstdc++ is still the default, support is retained.

So I don't see the need for this change to have such a drastic impact.

kencu:
Yes, you are probably right.

I can't seem to recall any ports that are forced to build with -stdlib=macports-libc++ on systems that are defaulting to libc++.

jmroot:
> I am wondering is if there is a (valid) scenario where `${cxx_stdlib}` is set to `libc++` (which of course is the vast majority of cases now) but a port could decide to set `${configure.cxx_stdlib}` to `libstdc++`, or `macports-libstdc++` ? if thats never a valid combination, then your changes look look just fine to me.

Yes, that is a valid thing for a Portfile to do. There would be no reason to have an option for it otherwise.

Such ports of course usually only work on older OS versions, and tend to build with gcc or the system clang, not macports-clang.

jmroot:
> I can't seem to recall any ports that are forced to build with -stdlib=macports-libstdc++ on systems that are defaulting to libc++.

Affected ports would be those that specify `configure.cxx_stdlib libstdc++` and also a newer C or C++ standard than is supported by the old system compilers. Probably rare, but presumably also more than zero ports given that someone bothered adding specific compiler selection logic to base that covers that case (and which is of course rendered incorrect by this change).

kencu:
OK @cjones051073 -- you're sunk then.

Time to rever this.

kencu:
It was going to be a mess anyway, I thought. But if you're not up to revert it, I guess you can debate that with Josh.

cjones051073:
> It was going to be a mess anyway, I thought. But if you're not up to revert it, I guess you can debate that with Josh.

I am nit going to revert this, as it is still my opinion we where enabling support for macpoets-libstdcxx in way more places than really needed, and these changes are a step to addressing that. 

Honestly I am not yet sure I see a need for using this in anything post 10.5, but if there are a handful of ports affected by this, on some older systems, i would prefer to then address these on a case by case basis.

PR #18206, opened 2023-04-06
coreutils-uutils: update to 0.0.18

dgilman:
#### Description

This enables ACL support in the cp command. I have also started installing the manpages.

One question: i have it installing the manpages in `uu-xyz` form to the standard manpath under the MP prefix. This is working fine. I'd like to symlink unprefixed manpages under the `libexec/uutils` path. I tried several attempts at this but was unable to make it work. If you do it during the destroot phase you get symlinks pointing to the worksrcdir which is unusable. I tried adding a post-destroot (as I notice the GNU packages set up their symlinks in a post-destroot) but couldn't get that to work for whatever reason. Is there a standard way to do this, or some other package that's figured it out?

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 10.15.7 19H2026 x86_64
Xcode 12.4 12D4e

###### Verification 
Have you

### DIFF
--- a/sysutils/coreutils-uutils/Portfile
+++ b/sysutils/coreutils-uutils/Portfile
@@ -6,7 +6,7 @@ PortGroup           cargo 1.0
 
 name                coreutils-uutils
 revision            0
-github.setup        uutils coreutils 0.0.17
+github.setup        uutils coreutils 0.0.18
 github.tarball_from archive
 categories          sysutils
 platforms           darwin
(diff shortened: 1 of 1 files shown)


ryandesign:
You can create manpage symlinks at any time during the destroot phase. Since this port already overrides the destroot phase with a destroot block that would seem to be the most logical place to do it but it makes no difference if you do it in a pre-destroot or post-destroot block instead.

There's nothing to it really, just `ln -s source_file target_file` or `ln -s source_file ... target_dir`, as the ln(1) manpage says. `source_file` is the existing file or files, `target_file` or `target_dir` is where you want the symlink(s) to be created. For the source file(s), you'll need to specify a path that makes sense after the port has been installed, such as an absolute path that begins with `${prefix}`. For the target file or directory, you'll need to specify a path within `${destroot}` so that the symlink are included in the set of files registered to the port and properly removed when the port is uninstalled.

PR #18209, opened 2023-04-06
hptt: fix compiler choice on <10.7, fix build on >10.6 Intel after recent change to gcc settings

barracuda156:
#### Description

On 10.6 gcc-4.2 gets picked; fix that.

Macports, unfortunately, moved GCC to use libc++, which is untested. To no surprise, it does not work. Set libstdc++ manually, fix the build.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/math/hptt/Portfile
+++ b/math/hptt/Portfile
@@ -22,13 +22,14 @@ patchfiles          0001-CMakeLists-add-PPC-support-unbreak-build-on-Darwin.patc
                     0002-CMakeLists-fix-install-path-for-headers.patch \
                     0003-Add-hptt.pc-config-file.patch
 
+# Keep this above, otherwise a wrong compiler can be picked.
+compiler.cxx_standard   2011
 # It does not build with new clang: https://github.com/springer13/hptt/issues/21
 # See also: https://bugs.llvm.org/show_bug.cgi?id=36915
 compiler.blacklist-append \
                     *clang*
 compiler.fallback-append \
                     macports-gcc-12 macports-gcc-11
-compiler.cxx_standard   2011
 
 configure.args-append \
                     -DENABLE_ARM=OFF \
--- a/math/hptt/Portfile
+++ b/math/hptt/Portfile
@@ -31,6 +31,8 @@ compiler.blacklist-append \
 compiler.fallback-append \
                     macports-gcc-12 macports-gcc-11
 
+configure.cxx_stdlib    macports-libstdc++
+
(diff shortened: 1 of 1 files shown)


barracuda156:
Hmm, we get an error which was not there earlier (`hiop` has not been updated, so it is either `gcc` or something else).

cjones051073:
I do not think the fix is is the right way to go. See discussion in https://github.com/macports/macports-ports/commit/30a50c174346b502af883ce3d9ca829cf8d256b5#commitcomment-108191462

We should first see if the issues with gcc and libc++ can be addressed, as this goes beyond just this port.

barracuda156:
> I do not think the fix is is the right way to go. See discussion in [30a50c1#commitcomment-108191462](https://github.com/macports/macports-ports/commit/30a50c174346b502af883ce3d9ca829cf8d256b5#commitcomment-108191462)
> 
> We should first see if the issues with gcc and libc++ can be addressed, as this goes beyond just this port.

We can always switch it back for Intel to use `libc++`. What benefit we get by having it broken until someone fixes Clang headers?

cjones051073:
The 'quick_exit' errors above where introduced I suspect by

https://github.com/macports/macports-ports/commit/30a50c174346b502af883ce3d9ca829cf8d256b5#commitcomment-108191462

which updated the LLVM headers gcc uses from clang 14 to clang 16. This improved support for newer c++ standards, but if it also caused this error I will just revert that change and go back to clang-14. At which point you should remove  the line forcing `macports-libstdc++`

barracuda156:
> The 'quick_exit' errors above where introduced I suspect by
> 
> [30a50c1#commitcomment-108191462](https://github.com/macports/macports-ports/commit/30a50c174346b502af883ce3d9ca829cf8d256b5#commitcomment-108191462)
> 
> which updated the LLVM headers gcc uses from clang 14 to clang 16. This improved support for newer c++ standards, but if it also caused this error I will just revert that change and go back to clang-14. At which point you should remove the line forcing `macports-libstdc++`

@cjones051073 Converted this to draft for now – if you could fix Clang header, that gonna solve the problem in a better way indeed.

barracuda156:
Fixed in https://github.com/macports/macports-ports/commit/a13f740688545c194bbac873cb6af818015cc425

PR #18258, opened 2023-04-13
llvm-cx: Seems "darwin any" isn't helpful here

Gcenx:
the prebuilt binary on https://packages.macports.org/llvm-cx has an internal deployment target of 13, so macports builds on the latest OS not the minimum OS

###### Type(s)

- [x] bugfix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/lang/llvm-cx/Portfile
+++ b/lang/llvm-cx/Portfile
@@ -6,9 +6,10 @@ PortGroup                   cmake 1.1
 # This custom compiler toolchain is based on llvm|clang-8 but each release adds fixes for newer SDKs.
 name                        llvm-cx
 version                     22.1.1
+revision                    1
 dist_subdir                 wine
 categories                  lang
-platforms                   {darwin any >= 18}
+platforms                   {darwin >= 18}
 license                     NCSA
 maintainers                 nomaintainer
 description                 CodeWeavers custom compiler for -mwine32 targets ${version}.


Gcenx:
@kencu seems "darwin any" doesn't work the way I'd expected it to, it's building on macOS Ventura with a deployment target of 13 where I'd have expected this to be the reverse so guessing this requires a rev-bump.

kencu:
Setting "darwin any" doesn't do anything to the deployment target, and I think neither will the change in this PR...

what are you trying to do?

Gcenx:
> Setting "darwin any" doesn't do anything to the deployment target, and I think neither will the change in this PR...

Yeah seems to be the case wish that were explained somewhere.

> what are you trying to do?

I was attempting to make use of macports new single prebuilt package system for multiple versions of macOS using {darwin any}, but as the package was built in macOS Ventura with a deployment target of 13.0 is useless for lower versions of macOS.

So I’d removed the “any” so the build bots instead build llvm-cx for each supported version.

kencu:
OK, I see... it would be useful if things worked the way you thought they did. Let's try this then.

jmroot:
`platforms {darwin any}` means the port installs the same files regardless of which darwin version it is built on, so if it's applicable, it doesn't matter which OS version the archive is built on (and in practice it will be built on whichever builder is available first.)

Gcenx:
> `platforms {darwin any}` means the port installs the same files regardless of which darwin version it is built on, so if it's applicable, it doesn't matter which OS version the archive is built on (and in practice it will be built on whichever builder is available first.)

So the current way this works isn’t too useful then, as if this were built on the macOS Mojave buildbot it would have worked just fine in all later versions without issue but as it was built o blacks Ventura is become useless to all lower versions.

jmroot:
It's very useful if we would otherwise build a dozen identical archives. The option is not there to make the port build identically on different OS versions (and that's often quite difficult to achieve and not necessarily desirable anyway), it's there so you can signal that it does in fact do that and thus separate archives per OS version are not needed.

Gcenx:
> It's very useful if we would otherwise build a dozen identical archives. The option is not there to make the port build identically on different OS versions (and that's often quite difficult to achieve and not necessarily desirable anyway), it's there so you can signal that it does in fact do that and thus separate archives per OS version are not needed.

With the way it works it might be useful for say `winetricks` since it’s just a shell script on all systems.

For `llvm-cx` if it would built on macOS Mojave it could have been usable for all later versions without issue, I know that’s not always the case.

jmroot:
Right, I'm sure there would be uses for such a feature, it's just not a feature that we have currently. Roughly half of the ports in the tree currently have `any` in their `platforms`, so it seems that that also has its uses. :)

kencu:
the needed bit would be to have a way to spec which system would be the builder…

but I can think of issues right away, eg non-default prefixes…

Gcenx:
> the needed bit would be to have a way to spec which system would be the builder…

Here’s what I’d assumed was the case so maybe a good example

`{darwin any > 15` I’d assume the archive would be built on darwin15 but used on anything later I’d arch matched.

Obviously darwin15 doesn’t support arm64 so that would then fallback to being built on darwin20 etc.

> but I can think of issues right away, eg non-default prefixes…

Yeah that’s where it could cause issues for something like `MoltenVK` as that requires at minimum of Xcode11.7 to build but the resulting dylib could run on El Capitan.

That’s why id handled it in a more _fun_ way aka extracting the installers contents but that’s no longer working for newer versions and they’d denied changing back to a more universal archive format….

PR #18269, opened 2023-04-15
cairo-devel: update to 1.17.8; fixes major memory leak

lukaso:
#### Description

Fixes major memory leak

https://trac.macports.org/ticket/66852

This is causing issues for GIMP https://gitlab.gnome.org/GNOME/gimp/-/issues/9278

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 13.3.1 22E261 arm64
Xcode 14.3 14E222b

###### Verification 
Have you

- [x] tried existing tests with `sudo port test`? - error but nothing in logs
- [x] checked that the Portfile's most important [variants](https://trac.macports.org/wiki/Variants) haven't been broken? couldn't really check


lukaso:
This builds and functions on my local mac. Not sure what is going wrong in the build phase. Is it possible that extra checks have been enabled on the build servers?

ryandesign:
From the CI logs, it fails because it cannot fetch the distfile.

I would prefer that you not change the `homepage` or `master_sites`. That may fix the fetch failure.

As I mentioned in https://trac.macports.org/ticket/66852, I recommend we do cairo-devel first and make sure there are no problems with it before doing the update on cairo.

Will it work if the meson portgroup is included in alphabetical order? If so, that would be preferable.

I haven't looked at the rest of it or tested it; I'll leave that to @mascguy.

lukaso:
> From the CI logs, it fails because it cannot fetch the distfile.
> 
> I would prefer that you not change the `homepage` or `master_sites`. That may fix the fetch failure.

Couldn't be helped because the tar file isn't at the original sites. For whatever reason, they've not pushed the file there.

> As I mentioned in https://trac.macports.org/ticket/66852, I recommend we do cairo-devel first and make sure there are no problems with it before doing the update on cairo.

Given it has a lot of dependencies, will those use `cairo-devel` instead of `cairo`? I couldn't see how to get that to work locally. I'm not really sure how the `-devel` ports are supposed to work.

> Will it work if the meson portgroup is included in alphabetical order? If so, that would be preferable.

I'll try this.

mascguy:
> > As I mentioned in https://trac.macports.org/ticket/66852, I recommend we do cairo-devel first and make sure there are no problems with it before doing the update on cairo.
> 
> Given it has a lot of dependencies, will those use `cairo-devel` instead of `cairo`? I couldn't see how to get that to work locally. I'm not really sure how the `-devel` ports are supposed to work.

For cases like this - where we also maintain a newer `-devel` port - dependents are setup to use either. This is done via a special `path:` syntax like the following, when declaring a dependency.

```
# Example path dependencies, copy-pasted from port 'gegl'
depends_lib-append \
                    path:lib/pkgconfig/babl.pc:babl \
                    path:lib/pkgconfig/cairo.pc:cairo \
                    path:lib/pkgconfig/pango.pc:pango \
# ...etc...
```

Basically the dependency is fulfilled by the presence of the file after `path:`, such as the pkgconfig file in these cases. (The last part of the dependency is the default port - `:<default_port_name>` - which is used if the dependency isn't installed yet.)

So for such a case, you'll need to forcibly deactivate the non-Devel port, and then install the Devel variation:

```
$ sudo port -f deactivate cairo
$ sudo port -N install cairo-devel
```

At that point, `cairo-devel` will fulfill the dependency for all ports, and you're good-to-go.

Let us know if that makes sense.

lukaso:
Thanks for the guidance on testing. In the meantime, `adwaita-icon-theme` no longer can be installed on my mac and actually hangs it https://trac.macports.org/ticket/67243, so I'm not able to test this any further. Before, when I could compile, it was working fine.

ryandesign:
> Fixes major memory leak

I merged your other PR #18282 that backported the fix for the memory leak, so now this PR needs that conflict to be resolved.

> > I would prefer that you not change the `homepage` or `master_sites`. That may fix the fetch failure.
> 
> Couldn't be helped because the tar file isn't at the original sites. For whatever reason, they've not pushed the file there.

I don't follow. As far as I can see, cairo-1.17.8.tar.xz is published at https://www.cairographics.org/snapshots/

ryandesign:
And also from the ticket, before merging I would want confirmation that for each library, the library name and major version (the one in the library filenames and the install name) hasn't changed, that the library minor current version and minor compatibility version have not decreased, and that at least a couple ports built against cairo 1.17.6 or earlier are confirmed to continue working after upgrading to 1.17.8 without needing to rebuild those other ports.

graphics/cairo-devel/Portfile
```diff
@@ -51,12 +52,6 @@ depends_lib                 path:lib/pkgconfig/pixman-1.pc:libpixman \
 
 minimum_xcodeversions       {8 2.4.1}
 
-# Prevent cairo from using librsvg, libspectre, poppler.
-patchfiles-append           patch-configure.diff
```
ryandesign, line 55: You've removed the Portfile code here that ensured that cairo would not use librsvg, libspectre, or poppler, and below for ghostscript or lzo2, even if they were installed. What's the status of each of these now? Have you verified that, if those are installed, they aren't used? (Does the meson script disable these features by default and require you to opt into them?)

graphics/cairo-devel/Portfile
```diff
@@ -68,44 +63,18 @@ if {${configure.build_arch} eq "x86_64" || (${universal_possible} && [variant_is
...
-
-# Prevent cairo from using lzo2 because its GPL license makes cairo's effective license GPL too.
-configure.args-append       ac_cv_lib_lzo2_lzo2a_decompress=no
-
-variant opengl requires x11 description {Add OpenGL graphics interface} {
-    depends_lib-append      port:mesa
-
-    configure.args-replace  --disable-gl --enable-gl
-    configure.args-append   --enable-glx
-}
```
ryandesign, line 102: You've removed the opengl variant... Is this functionality no longer available in cairo?

graphics/cairo-devel/Portfile
lukaso: I didn't validate that this is covered. It needs more work.

graphics/cairo-devel/Portfile
lukaso: I couldn't see it anywhere in the meson_options.txt file. Same is true for all the options I deleted.

graphics/cairo-devel/Portfile
```diff
@@ -51,12 +52,6 @@ depends_lib                 path:lib/pkgconfig/pixman-1.pc:libpixman \
 
 minimum_xcodeversions       {8 2.4.1}
 
-# Prevent cairo from using librsvg, libspectre, poppler.
-patchfiles-append           patch-configure.diff
-
-# Don't check stderr in configure tests.
-patchfiles-append           dont-check-stderr.patch
```
lukaso, line 58: I've left a question regarding this patch here: https://gitlab.freedesktop.org/cairo/cairo/-/issues/121

graphics/cairo-devel/Portfile
lukaso: BTW: I don't see where I've removed the ghostscript or lzo2 checks.

ryandesign:
I would also want to pay special attention to the universal variant: not just whether it builds, but whether the result is correct for both architectures. For example, build universal on an Intel Mac (running the tests, if possible, and verifying they pass for both architectures), transfer the archive to an Apple Silicon Mac running the same OS version, install it, and verify that cairo works correctly. (The MacPorts buildbot system may build ports universal, and if it does, the archive will be published to the packages server, and users of both machine architectures might receive that same archive, and it is undefined which architecture of build machine will perform the build.)

I see code in the meson build script that sets `WORDS_BIGENDIAN` on big-endian systems like PowerPC, so testing a Mac OS X 10.4 or 10.5 ppc+i386 universal build in the same way would be good too. I don't expect you to have such old machines lying around. I do, as do Ken and some other contributors, and one of us could possibly test it, though it does take awhile to build on those old machines. Such testing needn't hold up merging the PR to update the cairo-devel port, but I would want to know that universal works correctly before doing the update on the cairo port.

graphics/cairo-devel/Portfile
ryandesign: Ok, the NEWS file says for 1.17.8:

> In a continuing effort to reduce the amount of legacy code, and increase
> the long-term maintainability of cairo, the following backends have been
> removed:
> 
> - GL and GLES drawing

mascguy:
> I haven't looked at the rest of it or tested it; I'll leave that to @mascguy.

I'll try to take a closer look over the next few days, if possible.

graphics/cairo-devel/Portfile
mascguy: The key output I see during the configure phase -- without trace mode, but with the various dependencies installed -- is:

```
Requested 'libspectre >= 99999' but version of libspectre is 0.2.12
configure:25672: WARNING: PDF backend will not be tested since poppler >= 99999 is not available
configure:26038: WARNING: SVG backend will not be tested since librsvg >= 99999 is not available
```

And the configure script defines the following variables, which appear to be utilized/imported by the Meson build (?), are defined as (matching the previous output):

```
configure:SPECTRE_VERSION_REQUIRED=99999
configure:POPPLER_VERSION_REQUIRED=99999
configure:LIBRSVG_VERSION_REQUIRED=99999
```

I haven't had a chance to review upstream's commit history yet. But is the current release setup to explicitly avoid these three dependencies? (Superficially that appears to be the case, but curious whether you folks have more insight into the details...)

mascguy: Sorry for the confusion, my previous message (RE: various versions required for dependencies) was due to working with the wrong branch (1.17.6, rather than Lukas' 1.17.8).

Never Mind...

mascguy:
I'm seeing build errors, both locally as well as via CI, like the following:

```
  ../cairo-1.17.8/src/cairo-tee-surface.c:223:69: error: too few arguments to function call, expected 5, have 4
          status = _cairo_surface_wrapper_paint (&slaves[n], op, source, clip);
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~                              ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:102:1: note: '_cairo_surface_wrapper_paint' declared here
  _cairo_surface_wrapper_paint (cairo_surface_wrapper_t *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:228:76: error: too few arguments to function call, expected 5, have 4
      return _cairo_surface_wrapper_paint (&surface->master, op, source, clip);
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                    ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:102:1: note: '_cairo_surface_wrapper_paint' declared here
  _cairo_surface_wrapper_paint (cairo_surface_wrapper_t *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:247:34: error: too few arguments to function call, expected 7, have 5
                                                op, source, mask, clip);
                                                                      ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:109:1: note: '_cairo_surface_wrapper_mask' declared here
  _cairo_surface_wrapper_mask (cairo_surface_wrapper_t *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:253:28: error: too few arguments to function call, expected 7, have 5
                                          op, source, mask, clip);
                                                                ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:109:1: note: '_cairo_surface_wrapper_mask' declared here
  _cairo_surface_wrapper_mask (cairo_surface_wrapper_t *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:281:11: error: too few arguments to function call, expected 11, have 10
                                                  clip);
                                                      ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:118:1: note: '_cairo_surface_wrapper_stroke' declared here
  _cairo_surface_wrapper_stroke (cairo_surface_wrapper_t    *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:291:12: error: too few arguments to function call, expected 11, have 10
                                            clip);
                                                ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:118:1: note: '_cairo_surface_wrapper_stroke' declared here
  _cairo_surface_wrapper_stroke (cairo_surface_wrapper_t    *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:316:16: error: too few arguments to function call, expected 9, have 8
                                                clip);
                                                    ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:150:1: note: '_cairo_surface_wrapper_fill' declared here
  _cairo_surface_wrapper_fill (cairo_surface_wrapper_t  *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:325:10: error: too few arguments to function call, expected 9, have 8
                                          clip);
                                              ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:150:1: note: '_cairo_surface_wrapper_fill' declared here
  _cairo_surface_wrapper_fill (cairo_surface_wrapper_t  *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:370:14: error: too few arguments to function call, expected 13, have 12
                                                            clip);
                                                                ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:161:1: note: '_cairo_surface_wrapper_show_text_glyphs' declared here
  _cairo_surface_wrapper_show_text_glyphs (cairo_surface_wrapper_t     *wrapper,
  ^
  ../cairo-1.17.8/src/cairo-tee-surface.c:383:17: error: too few arguments to function call, expected 13, have 12
                                                        clip);
                                                            ^
  ../cairo-1.17.8/src/cairo-surface-wrapper-private.h:161:1: note: '_cairo_surface_wrapper_show_text_glyphs' declared here
  _cairo_surface_wrapper_show_text_glyphs (cairo_surface_wrapper_t     *wrapper,
  ^
  10 errors generated.
```

lukaso:
I'm very much struggling to have time to spend on this. Looks like I did the easy part so far (unfortunately).

mascguy:
> I'm very much struggling to have time to spend on this. Looks like I did the easy part so far (unfortunately).

No worries Lukas, I'm planning to take a closer look this weekend. More to follow.

mascguy:
Haven't forgotten about this, but still working through a five-month backlog of work.

Hoping to get back to this in the coming days.

mascguy:
Relative to the build errors I saw previously, it turns out that upstream's 1.17.8 snapshot was missing some minor changes. They provided those via a subsequent merge request, and I added a patch with those updates.

So the build issues are fixed, which is a big step forward!

We're still not done yet though, as I need to address the concerns which Ryan mentioned previously. More to follow, but making progress!

mascguy:
There's one major issue I noticed, which is that the various libs are being versioned at 2.0.0. That's confirmed by what I'm seeing in the build logs during linking, with the following arguments being passed:

```
-compatibility_version 2 -current_version 2
```

This is a show-stopper, as the versions should be greater than those for 1.17.6 (11709.0.0). But since the libs are being built with version 2.0.0, every dependent binary and dylib is broken.

Based on a quick review of `meson.build`, the versioning appears to be correct there. But somehow what's ultimately used at lower levels is v2.

My Meson knowledge is a bit light, particularly when it comes to low-level things like how these versions are passed down. Is there something fundamental I'm missing here...?

lukaso:
The library versioning stuff is a dark art. I had to figure it out about a year ago, but I can't remember anything other than it was difficult. There's some adding and subtracting that goes on as well. There are numbers that go into the naming of the file, and other numbers that go into the version inside. The one thing I remember for sure is that it is the same on linux so it's not a mac specific thing.

mascguy:
Hey Lukas, I think you lost the updated I pushed last time. One of the most important being, an upstream patch to fix an issue with the 1.17.8 release.

I'll add a separate commit, to reconcile against your latest changes.

mascguy:
> Hey Lukas, I think you lost the updates I pushed last time. One of the most important being, an upstream patch to fix an issue with the 1.17.8 release.
> 
> I'll add a separate commit, to reconcile against your latest changes.

Done.

Unfortunately, I'm still not able to determine why the dylib versioning is wrong. The only thing different about this project's `meson.build`, is the use of a Python script to extract the version:

```
project('cairo', 'c', 'cpp',
  meson_version: '>= 0.56.0',
  version: run_command(find_program('version.py'), check: true).stdout().strip(),
  default_options: ['warning_level=2'],
)
```

But I also tested patching that, to include a hard-coded version string... and still no dice. And the previous versions worked fine, so I'm stumped.

Any thoughts/ideas?

lukaso:
> Hey Lukas, I think you lost the updates I pushed last time. One of the most important being, an upstream patch to fix an issue with the 1.17.8 release.

> I'll add a separate commit, to reconcile against your latest changes.

Sorry, I was force pushing the master branch to my fork and it clearly pushed this as well. There's unfortunately no chance that I'll be able to bring this PR too a conclusion, just too busy. Making sure my GIMP build keeps working is soaking up a lot of time (e.g. the broken libgcc12 build). And currently it's not working.

mascguy:
> > Hey Lukas, I think you lost the updates I pushed last time. One of the most important being, an upstream patch to fix an issue with the 1.17.8 release.
> > I'll add a separate commit, to reconcile against your latest changes.
> 
> Sorry, I was force pushing the master branch to my fork and it clearly pushed this as well. There's unfortunately no chance that I'll be able to bring this PR too a conclusion, just too busy. Making sure my GIMP build keeps working is soaking up a lot of time (e.g. the broken libgcc12 build). And currently it's not working.

No worries, this work is still on my to-do list. The only remaining issue is working out the dylib versioning problem, which is a show-stopper until we figure it out.

mascguy:
We'll be updating `cairo/cairo-devel` to 1.18 soon, so this PR is no longer relevant.

PR #18291, opened 2023-04-16
nrsc5: new port

trodemaster:
#### Description
Adding new port nrsc5

###### Type(s)
- [x] enhancement

###### Tested on
macOS 13.3.1 22E261 arm64
Xcode 14.1 14B47b

###### Verification 

### DIFF
--- a/audio/nrsc5/Portfile
+++ b/audio/nrsc5/Portfile
@@ -1,0 +1,42 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           cmake 1.1
+
+github.setup        theori-io nrsc5 eb76474193f228fd8fa697fc267b123c6c845a22
+
+revision            0
+version             1.0-20230418
+categories          audio
+maintainers         {netjibbing.com:blake @trodemaster} \
+                    openmaintainer
+license             GPL-3+
+
+description         This program receives NRSC-5 digital radio stations using an RTL-SDR dongle
+long_description    \
+This program receives NRSC-5 digital radio stations using an RTL-SDR dongle. It offers a command-line interface as well as an API upon which other applications can be built.
+
+checksums           rmd160  c0bade3db2333ebe2a8bc7a2ce27b88b1f229a59 \
+                    sha256  793a70b25c585babe5041267ab7bccf7d8f549bebe53d9228c4bf1e00c20ab41 \
+                    size    23168318
+
+patchfiles         CMakeLists.txt.diff
+
+configure.args-append \
+    -DUSE_NEON=ON \
+    -DUSE_SSE=ON \
+    -DUSE_FAAD2=ON
+
+depends_build-append \
+                    port:libtool \
+                    port:cmake \
+                    port:autoconf \
+                    port:automake
+
+depends_run         port:fftw \
+                    port:faad2
+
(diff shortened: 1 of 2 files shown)


herbygillot:
Thanks for this contribution @trodemaster.  Some comments below.  Please make all changes within the same commit.

audio/nrsc5/Portfile
```diff
@@ -0,0 +1,40 @@
...
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           cmake 1.1
+
+github.setup        theori-io nrsc5 f1a32a551471dc5045193027e4c828746c5790fd
+
+revision            0
+
+categories          audio
+maintainers         {netjibbing.com:blake @trodemaster}
```
herbygillot, line 12: Ports should also be `openmaintainer` unless there is a very specific reason why you and only you can make modifications to this port:

```suggestion
maintainers         {netjibbing.com:blake @trodemaster} \
                    openmaintainer
```

audio/nrsc5/Portfile
```diff
@@ -0,0 +1,40 @@
...
+                    size    23168025
+
+patchfiles         CMakeLists.txt.diff
+
+configure.args-append \
+    -DUSE_NEON=ON \
+    -DUSE_SSE=ON \
+    -DUSE_FAAD2=ON
+
+depends_build       port:libtool \
```
herbygillot, line 30: This would be best as `depends_build-append`.  Especially when you're using portgroups, implicit dependencies may be added by portgroups, and using the `-append` suffix is safer for all `depends_*` specifications.

audio/nrsc5/Portfile
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           cmake 1.1
+
+github.setup        theori-io nrsc5 f1a32a551471dc5045193027e4c828746c5790fd
```
jmroot, line 7: A monotonically increasing `version` needs to be set. If upstream doesn't have a version scheme in place, something like `0.0-20230411` will do, with the date stamp changing to the date of future commits that are used.

audio/nrsc5/Portfile
```diff
@@ -0,0 +1,40 @@
...
+
+revision            0
+
+categories          audio
+maintainers         {netjibbing.com:blake @trodemaster}
+license             GPL
```
jmroot, line 13: `GPL` means any version of the GPL can be used. If this is not the case, specify the allowed version range.

jmroot (changes requested):
If the include order issues discussed [on the mailing list](https://lists.macports.org/pipermail/macports-dev/2023-April/044987.html) have not been resolved, they need to be.

herbygillot:
Thanks for making these changes @trodemaster. In addition to @jmroot's comments, please squash all changes back into the original commit. MacPorts submission guidelines want to see a single commit per new submission or port update.

audio/nrsc5/Portfile
trodemaster: I have added a version based on your suggestion

audio/nrsc5/Portfile
trodemaster: Updated to GPL-3+

trodemaster:
I was able to squash the commits and included all feedback so far, including reordering of include dirs as discussed on the email thread.

herbygillot:
Looks good @trodemaster, merged. Thank you for your submission.

PR #18310, opened 2023-04-18
libgit2: update to 1.6.4

ioquatix:
Previous version bump: https://github.com/macports/macports-ports/pull/17800

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/cargo/Portfile
+++ b/devel/cargo/Portfile
@@ -8,7 +8,7 @@ PortGroup           cargo   1.0
 name                cargo
 epoch               1
 github.setup        rust-lang ${name} 0.69.1
-revision            0
+revision            1
 
 categories          devel
 license             {MIT Apache-2}
--- a/devel/libgit2/Portfile
+++ b/devel/libgit2/Portfile
@@ -7,7 +7,7 @@ PortGroup           legacysupport 1.0
 PortGroup           muniversal 1.0
 
 # don't forget to update py-pygit2 and libgit2-glib as well
-github.setup        libgit2 libgit2 1.5.2 v
+github.setup        libgit2 libgit2 1.6.4 v
 github.tarball_from archive
 revision            0
 epoch               1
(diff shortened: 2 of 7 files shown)


ioquatix:
cc @herbygillot I have tested this with the Ruby rugged gem and it looks good to me.

herbygillot:
1.5.2 to 1.6.4 is a major version change, which has the potential to be disruptive to many of the ports which depend on `libgit2`.

The PR you linked to only updates the minor version; it updated `libgit2` from 1.5.1 to 1.5.2. A minor version update does not change the API in a major enough way to be a risky update.

So we need to be more careful with upgrading to 1.6.4 and test more than just Ruby's rugged.  For example, we'd also need to test that Rust's `cargo` can build against this version, amongst many other dependent ports.

herbygillot:
We can add commits incrementing the revision of dependent ports to the branch after this update commit to rebuild them against the updated `libgit2`.  The one limitation with this approach is that the build process here on Github is time constrained, so the CI job will cancel if it's taking too long to rebuild everything we revision bump. I'll add some revision bump commits in a few.  We should probably start with `cargo`.

mascguy:
> We can add commits incrementing the revision of dependent ports to the branch after this update commit to rebuild them against the updated `libgit2`. The one limitation with this approach is that the build process here on Github is time constrained, so the CI job will cancel if it's taking too long to rebuild everything we revision bump. I'll add some revision bump commits in a few. We should probably start with `cargo`.

Sounds good to me Herby!

ioquatix:
Thanks, that sounds like a good plan!

herbygillot:
Everything seems to have built successfully.  I am merging these.  There are additional ports that need to be bumped, but they are downstream of `rust` and `cargo`, and so should build fine since `rust` & `cargo` built fine.

ioquatix:
Thanks!

PR #18363, opened 2023-04-21
mariadb-10.11: fix the build for 32-bit

barracuda156:
#### Description

A very simple fix to allow it build.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/databases/mariadb-10.11/Portfile
+++ b/databases/mariadb-10.11/Portfile
@@ -214,6 +214,11 @@ if {$subport eq $name} {
     # temporary fix for C/C++ flag discovery
     patchfiles-append patch-fix-flag-discovery.diff
 
+    # Disable a static assert that breaks the build, see: https://trac.macports.org/ticket/67278
+    if {${build_arch} in [list arm i386 ppc]} {
+        patchfiles-append patch-fix-32-bit.diff
+    }
+
     post-patch {
         reinplace "s|@NAME@|${name_mysql}|g" \
             ${worksrcpath}/cmake/install_layout.cmake
--- a/databases/mariadb-10.11/Portfile
+++ b/databases/mariadb-10.11/Portfile
@@ -2,6 +2,7 @@
 
 PortSystem          1.0
 PortGroup           legacysupport 1.0
+
 # CLOCK_REALTIME
 legacysupport.newest_darwin_requires_legacy 15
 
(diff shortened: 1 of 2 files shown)


evanmiller:
Disabling a static_assert is generally a bad idea as the program usually depends on the assertion in order for the code to work. Have you run the test suite?

barracuda156:
> Disabling a static_assert is generally a bad idea as the program usually depends on the assertion in order for the code to work. Have you run the test suite?

@evanmiller Unfortunately, the port does not have any tests enabled.
Judging from comments to the code, should not be a big deal IMO. (Isn’t it verified at compile time only?)

kencu:
I presumed that ```sizeof(void*)*2+8``` should probably be ```sizeof(void*)*2+4``` for 32 bit, but I didn't have enough insight to know for certain.

This YYSTYPE business is apparently related to yacc, of which I know next to nothing:

https://silcnitc.github.io/ywl.html

barracuda156:
> I presumed that `sizeof(void*)*2+8` should probably be `sizeof(void*)*2+4` for 32 bit, but I didn't have enough insight to know for certain.

That explains the error `note: the comparison reduces to '(20 == 16)`.
https://trac.macports.org/ticket/67278

evanmiller:
> > I presumed that `sizeof(void*)*2+8` should probably be `sizeof(void*)*2+4` for 32 bit, but I didn't have enough insight to know for certain.
> 
> That explains the error `note: the comparison reduces to '(20 == 16)`. https://trac.macports.org/ticket/67278
> 
> So I can try using `sizeof(void*)*2+4` instead and if it works, we do not need to disable it.

You are missing the point. The comment indicates that the code relies on `YYSTYPE` being a certain size. Please don't modify this without either running a test suite or fully understanding what the code is doing.

barracuda156:
> You are missing the point. The comment indicates that the code relies on `YYSTYPE` being a certain size. Please don't modify this without either running a test suite or fully understanding what the code is doing.

Sorry, missed that it is reverse: reducing right side by 4 won’t make it equal to `YYSTYPE`. Need more caffeine.

kencu:
perhaps the yystype for 32 bit is supposed to be 16. 

you’d have to look at what the union is thinking it is made up of. Why does this check even exist? Somebody thought it important…

perhaps this is a good upstream question.

I did look to see this code seems old, and hasn’t changed recently afaict.

So something odd is happening.

PR #18392, opened 2023-04-24
icu-devel: enable rpath via config, not patch

catap:
#### Description

Modern `icu4c` allows to control rpath via configure script, that allows to replace a patch by configure option that makes a support of it much easy.

It also fixed a regression with rpath which was introduced by last version of `icu-devel` update.

As usual I've skip CI because it can't build that port due to conflict of dependency with non devel version of icu.

###### Type(s)

- [x] bugfix
- [x] enhancement

###### Tested on

macOS 12.6.3 21G419 x86_64
Xcode 14.2 14C18

###### Verification 
Have you


catap:
@barracuda156 may you check it on your setup?

kencu:
this configure arg works -- but it still seems to me that "--enable-rpath" is completely broken here, as it does not actually enable "@rpath" references to the libraries nor does it add rpath search paths to the libraries or the binaries.

Instead "--enable-rpath" instead disables rpaths and puts in the full pathname to the library.

But for our purposes it at least does the thing we want, even if it is ass-backwards and accidental!

catap:
@kencu may I ask you to merge it when?

catap:
Thanks!

PR #18401, opened 2023-04-24
retdec-devel: update to 20230404

catap:
###### Type(s)

- [x] enhancement

###### Tested on

macOS 12.6.3 21G419 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/devel/retdec/Portfile
+++ b/devel/retdec/Portfile
@@ -98,16 +98,16 @@ if {${name} eq ${subport}} {
 subport retdec-devel {
     conflicts       $name
 
-    github.setup    avast retdec 53e55b4b26e9b843787f0e06d867441e32b1604e
-    version         20221208
+    github.setup    avast retdec 5821bbe735262187bd5d48930d0eda8b7b48bba8
+    version         20230404
     revision        0
     epoch           1
 
     checksums-append \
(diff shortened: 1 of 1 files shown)


kencu:
@jmroot 

any better idea for the extract.rename issue here?

catap:
@kencu I suggest to use https://github.com/macports/macports-base/pull/289 as discussion to find the better solution to this use-case.

I don't say that suggest approach is good, but it fixes that port.

kencu:
I just want @jmroot or @ryandesign to sign off on this ahead of the commit rather than have issues with it after.

jmroot:
Better to handle it in the github portgroup for everyone, I would think. (See Trac comments.)

catap:
@jmroot anyway it shouldn't be a part of this PR, do you agree?

catap:
@jmroot , @kencu maybe merge this PR? It seems quite clear that an issue isn't related to it, and should be fixed inside Github PG or MacPorts itself.

herbygillot:
Merging this.  We can open a separate PR to enhance the `github` portgroup if needed.

catap:
thanks, Herby!

PR #18416, opened 2023-04-25
R: fix for 10.6 and add +tcltk to default variants to support ports that require it

barracuda156:
#### Description

Fixes for:
https://trac.macports.org/ticket/67298
https://trac.macports.org/ticket/67016

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

macOS 10A190
Xcode 3.2

###### Verification 
Have you


i0ntempest:
Can we have this patch only apply for <10.7? Also compilation fixes do not need a revbump.

barracuda156:
> Can we have this patch only apply for <10.7? Also compilation fixes do not need a revbump.

1. Technically yes, but the patch does not affect anything 10.7+ as you can see from the code. (A reason why I thought unconditional patch is better because system-specific patches are too often being forgotten to be updated, and we are left with a broken software.) But if you favor a conditional patch – I will definitely prefer that to having `R` defunct :)
2. Revbump is for `+tcltk`, not for the fix. Did I do it in a wrong commit? I will swipe it then.

PR #18428, opened 2023-04-26
coreutils{,-devel}, m4: avoid a patch to overstep max path bug; ncurses, gettext: avoid max path bug on macOS 10.5

catap:
#### Description

This is is much more stright forward way to overstep that bug that makes supporting it much easy.

Half year ago I've tried to sent it to upstream, but hasn't been successful, but still we may enjoy it as an easy workaround

See: https://trac.macports.org/ticket/62994
See: https://lists.gnu.org/r/bug-gnulib/2022-09/msg00045.html

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

###### Verification 
Have you


mascguy (approved):
LGTM, much cleaner and simpler than patches. Glad this is an option!

@ryandesign ?

catap:
@mascguy and much easy to move to another portfiles if it required ;)

catap:
I've just catch one more port which requires such hack: ncurses

kencu (approved):

ryandesign:
I'm sad to hear that the developers of gnulib have been uninterested in fixing this problem. It sounds like they disagree that this test result should be overridden in this way. However, if you are sure that it should always have this value on macOS, then overriding it seems reasonable to me.

mascguy:
@jmroot Josh, any concerns relative to `ncurses`?

kencu:
the developers of gnulib asked that the test program be fixed, and gave some examples how to go about doing that.

But nobody responded to them.

PR #18459, opened 2023-04-28
Fix building polymake on 10.10 and 10.7

catap:
#### Description

Here I've fixed building on 10.10 `scip` and on 10.7 `mumps` and `OpenBLAS`.

I've also revbumped `scalapack`.

And make explicit requirements of compiler for `polymake`.

All of it allows to build `polymake` on both target platforms.

Closes: https://trac.macports.org/ticket/67322
Closes: https://trac.macports.org/ticket/67175
Closes: https://trac.macports.org/ticket/67350

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.10.5 14F2511 x86_64
Xcode 7.2.1 7C1002

macOS 10.7.5 11G63 x86_64
Xcode 4.6.3 4H1503

###### Verification 
Have you

### DIFF
--- a/math/OpenBLAS/Portfile
+++ b/math/OpenBLAS/Portfile
@@ -10,7 +10,7 @@ PortGroup           makefile 1.0
 name                OpenBLAS
 categories          math science
 license             BSD
-maintainers         {nicos @NicosPavlov} {michaelld @michaelld} openmaintainer
+maintainers         {nicos @NicosPavlov} {michaelld @michaelld} {@catap korins.ky:kirill} openmaintainer
 description         OpenBLAS is an optimized BLAS library based on GotoBLAS2
 long_description    {*}${description}
 
--- a/math/OpenBLAS/files/patch-MacOSX-mktemp.diff
+++ b/math/OpenBLAS/files/patch-MacOSX-mktemp.diff
@@ -1,11 +1,51 @@
---- c_check~	2023-04-02 04:18:01.000000000 +0800
-+++ c_check	2023-04-03 13:42:05.000000000 +0800
-@@ -257,7 +257,7 @@
+https://github.com/xianyi/OpenBLAS/pull/4028
+
+diff --git c_check c_check
+index 232adba6..8b633ab8 100755
+--- c_check
++++ c_check
+@@ -168,7 +168,7 @@ fi
+ 
+ no_msa=0
+ if [ "$architecture" = "mips" ] || [ "$architecture" = "mips64" ]; then
+-    tmpd="$(mktemp -d)"
++    tmpd=$(mktemp -d 2>/dev/null || mktemp -d -t 'OBC')
+     tmpf="$tmpd/a.c"
+     code='"addvi.b $w0, $w1, 1"'
+     msa_flags='-mmsa -mfp64 -mload-store-pairs'
+@@ -208,7 +208,7 @@ esac
+ 
+ no_avx512=0
+ if [ "$architecture" = "x86" ] || [ "$architecture" = "x86_64" ]; then
+-    tmpd=`mktemp -d`
++    tmpd=$(mktemp -d 2>/dev/null || mktemp -d -t 'OBC')
+     tmpf="$tmpd/a.c"
+     code='"vbroadcastss -4 * 4(%rsi), %zmm2"'
+     printf "#include <immintrin.h>\n\nint main(void){ __asm__ volatile(%s); }\n" "$code" >> "$tmpf"
+@@ -229,7 +229,7 @@ fi
+ 
+ no_rv64gv=0
+ if [ "$architecture" = "riscv64" ]; then
+-    tmpd=`mktemp -d`
++    tmpd=$(mktemp -d 2>/dev/null || mktemp -d -t 'OBC')
+     tmpf="$tmpd/a.c"
+     code='"vsetvli    zero, zero, e8, m1\n"'
+     printf "int main(void){ __asm__ volatile(%s); }\n" "$code" >> "$tmpf"
+@@ -245,7 +245,7 @@ fi
+ 
+ no_sve=0
+ if [ "$architecture" = "arm64" ]; then
+-    tmpd=`mktemp -d`
++    tmpd=$(mktemp -d 2>/dev/null || mktemp -d -t 'OBC')
+     tmpf="$tmpd/a.c"
+     printf "#include <arm_sve.h>\n\n int main(void){}\n">> "$tmpf"
+     args=" -march=armv8-a+sve -c -o $tmpf.o $tmpf"
+@@ -261,7 +261,7 @@ fi
  c11_atomics=0
  case "$data" in
      *HAVE_C11*)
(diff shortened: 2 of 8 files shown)


catap:
@ryandesign here it is

math/OpenBLAS/Portfile
```diff
@@ -6,11 +6,12 @@ PortGroup           muniversal 1.1
 PortGroup           compilers 1.0
 PortGroup           xcode_workaround 1.0
 PortGroup           makefile 1.0
+PortGroup           compiler_blacklist_versions 1.0
```
ryandesign, line 9: You're not using this portgroup.
```suggestion
```

math/OpenBLAS/Portfile
```diff
@@ -19,6 +20,9 @@ compiler.cxx_standard 2011
 compiler.c_standard   2011
 compiler.thread_local_storage yes
 
+# error: invalid instruction mnemonic 'cvtsi2ssl'
+compiler.blacklist-append {macports-clang-15}
```
ryandesign, line 24: I'm surprised that this would be the fix for this issue. Can you add a URL with a bug report or more information confirming that this error is a bug in clang 15? The Trac ticket only talks about the `mktemp` problem.

Curly braces are not required here.

math/mumps/Portfile
```diff
@@ -6,6 +6,7 @@ PortGroup                   linear_algebra 1.0
 PortGroup                   muniversal 1.0
 PortGroup                   makefile 1.0
 PortGroup                   conflicts_build 1.0
+PortGroup                   compiler_blacklist_versions 1.0
```
ryandesign, line 9: Same:
```suggestion
```

math/mumps/Portfile
```diff
@@ -52,6 +53,9 @@ configure.optflags          -O3
 configure.cflags-append     -fPIC
 configure.fflags-append     -fPIC
 
+# error: invalid instruction mnemonic 'cvtsi2ssl'
+compiler.blacklist-append   {macports-clang-15}
```
ryandesign, line 57: Same.

math/polymake/Portfile
```diff
@@ -20,6 +20,9 @@ checksums               rmd160  be170dff2b2838c9b65d48b08f4b679f60bbd670 \
                         sha256  6b6f81a42f8b748b489bdf04dedc6b8838e424acb26454424342725fec77d34d \
                         size    47480264
 
+compiler.cxx_standard   2014
+compiler.c_standard     2014
```
ryandesign, line 24: There's no such thing as C 2014.

math/scip/Portfile
```diff
@@ -27,6 +27,11 @@ checksums           rmd160  faa34bb47c67cdb7d5789de2e59fc8b09c1cd1f1 \
 
 compilers.setup     require_fortran
 
+# Block compilers that do not support thread-local storage.
+compiler.cxx_standard 2011
+compiler.c_standard   2011
+compiler.thread_local_storage yes
```
ryandesign, line 33: The comment doesn't seem necessary since the code is self-explanatory.

Looking at the `-std` flags used during the build, it seems to be using C 1999 and C++ 2014, doesn't it?

I already added `compiler.thread_local_storage yes` below.

```suggestion
compiler.c_standard   1999
compiler.cxx_standard 2014
```

math/OpenBLAS/files/patch-MacOSX-mktemp.diff
```diff
@@ -1,6 +1,17 @@
...
+diff --git c_check c_check
+index e8f90e1..d9adb23 100755
+--- c_check
++++ c_check
+@@ -204,7 +204,7 @@ esac
+ 
+ no_avx512=0
+ if [ "$architecture" = "x86" ] || [ "$architecture" = "x86_64" ]; then
+-    tmpd=`mktemp -d`
++    tmpd=`gmktemp -d`
```
ryandesign, line 10: Can you add a URL at the top of the patchfile to the bug report about this problem? Upstream should be correcting this by using `mktemp` in a portable manner, rather than in a GNU-specific way. We don't want to maintain this patch forever.

math/OpenBLAS/Portfile
catap: @ryandesign I haven't find any bug report. I simple tried another clang (14) and it works :)

Yes, I've reproduced it on my machine before fixing it.

math/OpenBLAS/files/patch-MacOSX-mktemp.diff
catap: I haven't filed any issue, I simple adjusted the patch to the actual version.

catap: Backported to upstream

catap:
Trying to make something cleaner that blacklist of clang-15, or at least understand wtf

catap:
I figured it out. Clang-15 isn't a problem, but once I blacklisted it, I covered a real problem.

After I started testing each port by building as `sudo port -f deactivate installed && sudo port -N install --unrequested XXX` to get a clean system, I've caught the root cause and fixed it.

The cause is a mix between `clang`, `gfortan` and `cctools`. The last one doesn't know anything about `llvm-15` and it leads to quite strange behavior that was reported.

My blacklisting of `clang-15` simply switched it to `clang-14` / `llvm-14` which is supported by `cctools` and everything works.

Once `llvm-14` is installed, it works fine with `clang-15` as the compiler. Once I've added support for `llvm-15` in `cctools`, it works fine without `llvm-14` :)

To avoid such problems in the future, I've added a note that `cctools` should also be changed if `compilers-1.0.tcl` is touched.

Mumps is a different story. It doesn't require a specific compiler version, and it can use a system one. So, it may faces the issue with cause similar with `OpenBLAS` If it goes by road with `clang-15`, or similar with `fpm`. So, I've reused the blacklist condition from that port here.

The only bad news is that `cctools` is currently broken and this PR is blocked by https://github.com/macports/macports-ports/pull/18443 where I'm fixing it.

kencu:
thank you for taking the time to find the "real" problem.

It is far too common to see people do a local "worked for me" fix without doing the proper analysis. That usually just leads to more trouble down the road as someone runs into trouble again later, and has to undo the previous fixes to finally get it properly sorted out.

I know people get frustrated when the "Well, it fixed it for me!" solution is not committed. But doing the proper analysis pays off in the end.

catap:
It is ready for review. @ryandesign, @kencu ?

kencu:
your cmake fix didn't actually add clang-16, but the addition of clang-15 looks fine to me.

catap:
@kencu thanks to catch that.

catap:
Added link to ticket https://trac.macports.org/ticket/67350 which should be fixed by this PR

catap:
@mascguy may I ask you to review and probably merge this one? Thanks!

mascguy (approved):

catap:
Thanks, Christopher!

PR #18463, opened 2023-04-29
texlive: update to 2023

drkp:
This is a mostly complete update of the texlive ports to the 2023 release of TeX Live.

The only remaining blocker that I'm aware of is that it doesn't yet properly build ConTeXt with the new luametatex engine. There is a port included for luametatex, which successfully builds that, but I haven't yet worked out how to build the formats.

@mojca, do you have any info on how to get ConTeXt built here? I saw that there was some discussion on the tl-distro list but haven't had time to sort through the latest.

Beyond that, if anyone feels like reviewing or testing, the update should be otherwise complete.

### DIFF
--- a/tex/biblatex-biber/Portfile
+++ b/tex/biblatex-biber/Portfile
@@ -12,9 +12,9 @@ epoch           3
 
 perl5.branches  5.34
 
-perl5.setup     Biber 2.17
+perl5.setup     Biber 2.19
 version         ${perl5.moduleversion}
-revision        1
+
 
 categories      tex
 license         {Artistic-2 GPL}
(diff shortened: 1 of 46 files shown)


mojca:
I don't have my mac with me right now.
Can you perhaps try to make a simple `luametatex` package from https://github.com/contextgarden/luametatex/releases/tag/v2.10.07 and make `texlive-bin` depend on that one?

drkp:
Actually, looks like it was just an issue getting the stub scripts installed for ConTeXt. With that, ConTeXt appears to be working now, though it seems to be using luatex instead of luametatex.

PR #18475, opened 2023-04-30
update vim to revision 1499

ufdup89:
This is necessary to fix a know bug related to syntax highlighting of .sh files

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 13.3.1 22E261 arm64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/editors/vim/Portfile
+++ b/editors/vim/Portfile
@@ -18,12 +18,12 @@ if {${os.platform} eq "darwin" && ${os.major} <= 14} {
                     sha256  7185577d6cd3708b62b2079b69a29215bb6a6048070d34e960d2d565c18b8a9b \
                     size    16851486
 } else {
-    set vim_patchlevel 1276
+    set vim_patchlevel 1499
     set port_revision  0
 
-    checksums       rmd160  a1630ecb5b3e8e68f926f595f85ea519d8bf867d \
-                    sha256  9e0690a0225b622e3b8cca1be866f3472062e4c43d9da47fe4536b33c60e97cf \
-                    size    16955692
+    checksums       rmd160  10d9edf0b19197771722e4a2fcf2e36021c34fa6 \
+                    sha256  3db98ddcbb8628ec9818da50353c4dd08d62938c80c71d38062250205310cdc3 \
+                    size    17010976
 }
 
 github.setup        vim vim ${vim_version}.${vim_patchlevel} v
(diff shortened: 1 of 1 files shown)


raimue:
Do you know which patch exactly you need? As the comment in the Portfile says, we usually follow MacVim releases to ensure to keep them always compatible as both use the same `.vimrc` and plugin files. But if there is a specific bug, we can backport it as a separate patchfile to get it fixed early.

tcspestana:
We need 1362 or later (see https://github.com/vim/vim/issues/12127 and link therein).

Seems like MacVim releases are not as frequent as vim's.

tcspestana:
You're right. I just noticed that updating to this new version breaks one of the plugins I am using. Maybe a better solution is just to apply the patch as described here: https://github.com/vim/vim/issues/11937#issuecomment-1442453014

raimue:
MacVim downgraded the sh.vim syntax highlighting file to v204 to address the problem with https://github.com/macvim-dev/macvim/commit/2e74830e8a15d56acb95d5ff71b34a1bab145723. We could do the same in vim, which will once again keep both ports in sync.

mascguy:
You'll need to rebase with our latest changes. And please keep the fallback logic in place, for 10.10 and earlier.

mascguy:
> You'll need to rebase with our latest changes. And please keep the fallback logic in place, for 10.10 and earlier.

Took care of rebasing, fixing the conflict in the process.

mascguy (approved):

neverpanic:
This change seems to have broken the +python variant:

```
:debug:patch system:  cd "/opt/local/var/macports/build/_opt_dports_editors_vim/vim/work/vim-9.0.1499" && /usr/bin/patch -t -p0 < '/opt/dports/editors/vim/files/patch-python3.diff'
:info:patch patching file 'src/configure.ac'
:info:patch 1 out of 1 hunks failed--saving rejects to 'src/configure.ac.rej'
:info:patch Command failed:  cd "/opt/local/var/macports/build/_opt_dports_editors_vim/vim/work/vim-9.0.1499" && /usr/bin/patch -t -p0 < '/opt/dports/editors/vim/files/patch-python3.diff'
:info:patch Exit code: 1
```

neverpanic:
See #18934.

raimue:
The port is now also out of sync with MacVim, which is still at patchlevel 1276. Both should always be in sync as they are using the same configuration, otherwise they may show different behavior.

PR #18502, opened 2023-05-03
libxkbcommon: fix executable linkages

kencu:
meson.build needs to declare that these executables link against libxkbcommon.0.dylib in order for meson to know to rewrite these linkages on installation.

closes: https://trac.macports.org/ticket/67349

### DIFF
--- a/devel/libxkbcommon/Portfile
+++ b/devel/libxkbcommon/Portfile
@@ -31,6 +31,9 @@ homepage            https://xkbcommon.org
 master_sites        ${homepage}/download/
 use_xz              yes
 
+# https://trac.macports.org/ticket/67349
+patchfiles          patch-libxkbcommon-mesonbuild-linkages.diff
+
 depends_build-append \
                     port:bison \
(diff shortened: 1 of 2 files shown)


kencu:
https://github.com/xkbcommon/libxkbcommon/issues/327

devel/libxkbcommon/Portfile
```diff
@@ -44,15 +47,15 @@ configure.args      -Ddefault_library=both \
...
+    revision        1
 
     depends_lib-append \
                     port:libxml2 \
                     port:xkeyboard-config
 }
 
 subport ${name}-x11 {
-    revision        0
+    revision        1
```
ryandesign, line 58: Did you see the `@rpath` problem with libxkbcommon-x11 too? I didn't, but I don't understand why not.

kencu: I didn't, but decided to revbump it anyway just to be sure all systems were fixed. Perhaps not needed, but I didn't know for sure.

PR #18642, opened 2023-05-12
QuantLib: use boost178, since boost176 has a bug that breaks the build on some systems

barracuda156:
Fixes: https://trac.macports.org/ticket/67353

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/finance/QuantLib/Portfile
+++ b/finance/QuantLib/Portfile
@@ -5,8 +5,10 @@ PortGroup           github 1.0
 PortGroup           compiler_blacklist_versions 1.0
 PortGroup           boost 1.0
 
+boost.version       1.78
+
 github.setup        lballabio QuantLib 1.30 QuantLib-v
-revision            0
+revision            1
 checksums           rmd160  b6df70f4b872b28d6d2d87826caa5c85b4b2843d \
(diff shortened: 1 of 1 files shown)


barracuda156:
@ryandesign Basically, any recent version of Boost will do, aside of 1.76 (due to the bug) and 1.80 (broken with gcc).

barracuda156:
@mascguy Re Boost: choice of 1.76 is quite problematic because of this bug too (happened with two ports at least or more).

And then, it has a bug with spinlock alignment (fixed in 1.81), see `i2pd` port and a long discussion referred there.

Finally, it has a bug in `multiprecision`, which pulls in Intel intrinsics headers on PPC when gcc12 being used (see `libtorrent-rasterbar` case, for example). Also fixed in 1.81.

barracuda156:
@ryandesign As long as we are flexible to change Boost version for individual ports as needed, no issues. (I guess no one ever will test every single port that uses Boost, forget fixing those which happen to fail.)

PR #18689, opened 2023-05-15
zmap: submission new Portfile ZMap version 3.0.0 beta1

artkiver:
* closes https://trac.macports.org/ticket/42895
 * closes https://trac.macports.org/ticket/41986

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.3.1 22E772610a arm64
Xcode 14.3 14E222b

###### Verification 
Have you

### DIFF
--- a/net/zmap/Portfile
+++ b/net/zmap/Portfile
@@ -4,11 +4,9 @@ PortSystem          1.0
 PortGroup           cmake 1.0
 PortGroup           github 1.0
 
-name                zmap
 github.setup        zmap zmap 3.0.0-beta1 v
 
 categories          net
-platforms           darwin
 license             apache-2
 maintainers         {@artkiver gmail.com:artkiver} openmaintainer
 
@@ -22,14 +20,13 @@ checksums           rmd160  4b30c0f33c5e4503a89dcb37808c50075373c4f4 \
                     sha256  61787bd59cb23183a8b5424a148b95812d8254d6249c2d4efb31cbebff9ca383 \
                     size    154851
 
-depends_build       port:byacc \
+depends_build-append port:byacc \
                     port:flex \
                     port:gengetopt \
-                    port:pkgconfig
+                    port:pkgconfig \
+                    port:cmake
 
-depends_build-append port:cmake
-
-depends_lib         port:gmp \
+depends_lib-append  port:gmp \
                     port:libdnet \
                     port:libpcap \
                     port:json-c \
--- a/net/zmap/Portfile
+++ b/net/zmap/Portfile
@@ -1,0 +1,43 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.0
+PortGroup           github 1.0
+
+name                zmap
+github.setup        zmap zmap 3.0.0-beta1 v
+
+categories          net
+platforms           darwin
+license             apache-2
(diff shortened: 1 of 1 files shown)


net/zmap/Portfile
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.0
+PortGroup           github 1.0
+
+name                zmap
```
herbygillot, line 7: This is redundant, since `name` is already being set by `github.setup`

net/zmap/Portfile
```diff
@@ -0,0 +1,43 @@
...
+github.setup        zmap zmap 3.0.0-beta1 v
+
+categories          net
+platforms           darwin
```
herbygillot, line 11: `platforms darwin` is now implicit, so this line can be omitted.

net/zmap/Portfile
```diff
@@ -0,0 +1,43 @@
...
+long_description    designed for Internet-wide network surveys capable of \
+                    scanning the entire IPv4 address space in 5 minutes from \
+                    a 10GbE connection.
+homepage            https://zmap.io/
+
+checksums           rmd160  4b30c0f33c5e4503a89dcb37808c50075373c4f4 \
+                    sha256  61787bd59cb23183a8b5424a148b95812d8254d6249c2d4efb31cbebff9ca383 \
+                    size    154851
+
+depends_build       port:byacc \
```
herbygillot, line 25: Since you're using portgroups, it's always best to use the `-append` form for `depends_*` specifications, just to ensure you're not overwriting depend specifications that are already being set by any included portgroups.

- `depends_build` -> `depends_build-append`
- `depends_lib` -> `depends_lib-append`

artkiver (commented):
Oh, interesting! I did not realize that.

Tested locally, removed the name and platforms lines and still seems to build OK! Looks good to me!

net/zmap/Portfile
artkiver: This tests OK locally as well.

net/zmap/Portfile
artkiver: Ah, thank you for the clarification. 

https://trac.macports.org/ticket/42895#comment:3 mentioned something similar, but the nuance was lost on me.

artkiver:
Hopefully the most recent push accounts for the changes @herbygillot mentioned correctly? (fingers crossed)

artkiver:
I see it still has the first commit from when I submitted this PR, is there a way to remove that or "squash commits" I guess as some would phrase it? Or should I just not worry about it?

herbygillot:
> I see it still has the first commit from when I submitted this PR, is there a way to remove that or "squash commits" I guess as some would phrase it? Or should I just not worry about it?

Cumulatively, the Portfile looks good. But yes, there are 2 commits there, and we'd want them squashed into one.

That said, I think I may be able to squash them through Github's UI on my side, so we'll give that a go.

artkiver:
> > I see it still has the first commit from when I submitted this PR, is there a way to remove that or "squash commits" I guess as some would phrase it? Or should I just not worry about it?
> 
> Cumulatively, the Portfile looks good. But yes, there are 2 commits there, and we'd want them squashed into one.
> 
> That said, I think I may be able to squash them through Github's UI on my side, so we'll give that a go.

Oh awesome! I was looking around for a way to do that without much luck. Thank you for your assistance yet again! 🙏

PR #18690, opened 2023-05-15
fheroes2: use legacysupport.use_mp_libcxx on Mojave and older

aeiouaeiouaeiouaeiouaeiouaeiou:
#### Description

icn2img uses std::filesystem since version 1.0.1: https://github.com/ihhub/fheroes2/commit/275ccfd0b4b7fc6dbf9c7498df2388fa5938541a
Failing build: https://build.macports.org/builders/ports-10.9_x86_64-builder/builds/231789/steps/install-port/logs/stdio

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G14042 x86_64
Xcode 10.1 10B61

###### Verification 
Have you

### DIFF
--- a/games/fheroes2/Portfile
+++ b/games/fheroes2/Portfile
@@ -4,7 +4,14 @@ PortSystem      1.0
 
 PortGroup       github 1.0
 PortGroup       cmake 1.1
+PortGroup       legacysupport 1.1
 PortGroup       compiler_blacklist_versions 1.0
+
+# On <10.15 built-in libc++ has no support for std::filesystem
+legacysupport.newest_darwin_requires_legacy 18
+if {[string match *clang* ${configure.compiler}]} {
+    legacysupport.use_mp_libcxx yes
+}
 
 github.setup    ihhub fheroes2 1.0.4
 


oleg-derevenetz:
Hi @aeiouaeiouaeiouaeiouaeiouaeiou @herbygillot thanks! I looked at legacysupport port group at the time, but but in its [README](https://github.com/macports/macports-legacy-support/blob/master/README.md) there is no mention of `std::filesystem` support :(

kencu:
the legacysupport portgroup already takes into account whether you're using libc++, so you don't need to test for clang being the compiler.

kencu:
like most of macports' portgroups, that change fairly quickly, the documentation for the support for a newer libcxx is in the first paragraphs of the portgroup code (legacysupport-1.1.tcl). 

MacPorts support for a newer libc++ has nothing to do with the macports-legacy-support headers or library, which is why it is not documented there. 

You could say the support for a newer libc++ should never have been added to the legacysupport PortGroup then, if it doesn't use the macports-legacy-support library and there is some logic to that statement, however it was decided to add it there as the concepts are similar (supporting older systems).

If you are using the newer libc++, then the legacysupport-1.1.tcl PortGroup file will automatically add the proper dependency on macports-libcxx for you, as the code to do that has been added there.

PR #18691, opened 2023-05-15
liblinear: update to 2.46

barracuda156:
#### Description

Update, switch to `python311`.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/math/liblinear/Portfile
+++ b/math/liblinear/Portfile
@@ -3,7 +3,7 @@
 PortSystem          1.0
 
 name                liblinear
-version             2.30
+version             2.46
 revision            0
 categories          math
 maintainers         nomaintainer
@@ -17,19 +17,18 @@ long_description    LIBLINEAR is a simple package for solving large-scale regula
                     logistic regression.
 
 homepage            https://www.csie.ntu.edu.tw/~cjlin/liblinear/
-platforms           darwin
 license             BSD
 
 master_sites        ${homepage}
-checksums           rmd160  17526c24e9164e3714c0ef743e9c9baefe28cec9 \
-                    sha256  881c7039c6cf93119c781fb56263de91617b3eca8c3951f2c19a3797de95c6ac \
-                    size    526468
+checksums           rmd160  e1cbb3d4dda070df64e28f4dc76e02a7d88bde7c \
+                    sha256  616dd3a9b5596894de427f141a4902fa99dafc51ff9d18e8d723852fdc0cb53b \
+                    size    569001
 
 patchfiles          patch-Makefile.diff
 
 use_configure       no
 
-variant universal {}
+variant universal   {}
 
 build.target        all
 build.args          CC="${configure.cc}   ${configure.cflags}   [get_canonical_archflags]" \
@@ -39,7 +38,7 @@ build.args          CC="${configure.cc}   ${configure.cflags}   [get_canonical_a
 destroot {
     xinstall -m 755 -W ${worksrcpath} train predict ${destroot}${prefix}/bin
     xinstall -m 644 -W ${worksrcpath} liblinear.dylib ${destroot}${prefix}/lib
-    xinstall -m 644 -W ${worksrcpath} linear.h tron.h ${destroot}${prefix}/include
+    xinstall -m 644 -W ${worksrcpath} linear.h newton.h ${destroot}${prefix}/include
 
     # Copy additional documents.
     set dest_doc ${destroot}${prefix}/share/doc/${name}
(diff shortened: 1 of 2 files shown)


math/liblinear/Portfile
```diff
@@ -54,31 +53,31 @@ destroot {
...
     xinstall -m 755 -d ${dest_py}
-    xinstall -m 644 -W ${worksrcpath}/python \
-        README liblinear.py liblinearutil.py \
-        ${dest_py}
+    xinstall -m 644 -W ${worksrcpath}/python README ${dest_py}
+    xinstall -m 644 -W ${worksrcpath}/python/${name} \
+        liblinear.py liblinearutil.py ${dest_py}
 }
 
 variant python27 description {Install Python 2.7 interface} {
```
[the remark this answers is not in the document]
barracuda156: OK, sure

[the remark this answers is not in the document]
barracuda156: Done.

math/liblinear/files/patch-Makefile.diff
```diff
@@ -1,50 +1,39 @@
...
-! 	rm -f *~ tron.o linear.o train predict liblinear.*
+--- Makefile.orig	2023-02-17 20:01:34.000000000 +0800
++++ Makefile	2023-05-15 18:46:27.000000000 +0800
+@@ -3,18 +3,22 @@
+ CFLAGS = -Wall -Wconversion -O3 -fPIC
+ LIBS = blas/blas.a
+ #LIBS = -lblas
+-SHVER = 5
++#SHVER = 5
++PREFIX ?= /opt/local
```
ryandesign, line 9: You've changed the original patch's `/usr/local` to `/opt/local` here. I guess the idea with patchfiles like these is that they should be written in such a way that upstream could apply them. We should be sending these patches to the developers so that in future we don't have to maintain them anymore. For upstream purposes, /usr/local is a better default than /opt/local. For MacPorts purposes, we override it with a build arg so it doesn't matter.

```suggestion
+PREFIX ?= /usr/local
```

Ideally the patchfile should begin with the URL of the bug report or pull request where the changes have been submitted to upstream, or the URL of the commit where they already applied it.

math/liblinear/files/patch-Makefile.diff
```diff
@@ -1,50 +1,39 @@
...
+ 
+ lib: linear.o newton.o blas/blas.a
+-	$(CXX) $(SHARED_LIB_FLAG) linear.o newton.o blas/blas.a -o liblinear.so.$(SHVER)
++	if [ "$(OS)" = "Darwin" ]; then \
++		LIBEXT=".dylib"; \
++		SHARED_LIB_FLAG="-dynamiclib -install_name $(PREFIX)/lib/liblinear$${LIBEXT}"; \
++	else \
++		LIBEXT=".so"; \
++		SHARED_LIB_FLAG="-shared -Wl,-soname,liblinear$${LIBEXT}"; \
++	fi; \
```
ryandesign, line 29: Why have you removed the major version number from the library's install name? Assuming that the developers assign new major version numbers using the correct criteria, removing it here will lead to problems (programs that use the library crashing or not working properly) next time they increase it if we don't revbump all ports that link with the library at that time. (Granted, I don't see any such ports right now, but there could be in the future, and this also applies to any software outside of MacPorts that the user might have that links with the library.) Having the library filename change, via the change in major version number, makes it clear that a rebuild is needed so I'm very skeptical of you removing it.

Before your changes, the patch removed the major version number in the library filename on Darwin and kept it on all others. After your changes, you've removed it for all OS types. It doesn't seem like this change would be suitable for submission to the developers. I would instead keep it for all OS types.

math/liblinear/files/patch-Makefile.diff
```diff
@@ -1,50 +1,39 @@
...
++		$(CXX) $${SHARED_LIB_FLAG} linear.o newton.o blas/blas.a -o liblinear$${LIBEXT}
```
ryandesign, line 30: This line isn't indented properly.
```suggestion
+	$(CXX) $${SHARED_LIB_FLAG} linear.o newton.o blas/blas.a -o liblinear$${LIBEXT}
```

math/liblinear/files/patch-Makefile.diff
barracuda156: @ryandesign Well, it should either be kept for all OS or removed for all OS. As far as I can see, there is no reason to specifically remove it for Darwin. However in Macports we have only one version of the port and no plans to have multiple `liblinear` in parallel, I believe, so adding a number to dylib name is unnecessary.

reneeotten:
@barracuda156 this fails on the buildbot: https://build.macports.org/builders/ports-13_arm64-builder/builds/32497/steps/install-port/logs/stdio

barracuda156:
> @barracuda156 this fails on the buildbot: https://build.macports.org/builders/ports-13_arm64-builder/builds/32497/steps/install-port/logs/stdio

@reneeotten Does it fail across the board? Details show green aside of this odd fetch error: https://ports.macports.org/port/liblinear/details
But may be those are outdated info.

barracuda156:
@reneeotten Yeah, it worked before, but something got broken. I will look into this.

PR #18696, opened 2023-05-16
icu-devel: fix build on Tiger

evanmiller:
#### Description

The icu 73.1 update does not build on Tiger due to the `--enable-rpath` flag that is passed. But with `--disable-rpath`, the build is broken due to an incorrect install_name.

No rev-bump needed since the compile fails.

See: https://trac.macports.org/ticket/67433
See: https://trac.macports.org/ticket/67428

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/icu-devel/Portfile
+++ b/devel/icu-devel/Portfile
@@ -24,7 +24,7 @@ set my_name         icu
 #  port file all | sort -u | xargs grep -El ':icu( |$)' | rev | cut -d / -f 2 | rev | xargs port info --name --subport | cut -d : -f 2 | tr ',' ' ' | grep -v '\-\-' | tr ' ' '\n' | sort -u
 
 version             73.1
-revision            1
+revision            2
 epoch               1
 subport             ${name}-docs         { revision 0 }
 subport             ${name}-lx           { revision 0 }
@@ -74,6 +74,8 @@ if {${subport} eq ${name} || ${subport} eq "${name}-lx"} {
     # use full pathnames to libraries in tools
     patchfiles-append   patch-i18n-formatted_string_builder.h.diff
 
+    # fix rpath mess
+    patchfiles-append patch-config-mh-darwin.diff
 
     compiler.cxx_standard   2011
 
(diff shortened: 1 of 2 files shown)


catap:
@evanmiller I really don't understand this patch because `--enable-rpath` and `--disable-rpath` only control the condition you broke.

Differences between `--enable-rpath` and `--disable-rpath` with your patch shouldn't be noticeable at all.

catap:
See: https://github.com/macports/macports-ports/pull/18392#issuecomment-1520959007

evanmiller:
@catap  `--enable-rpath` causes `-Wl,-rpath` to be passed. This causes a failed build on Tiger.

https://github.com/unicode-org/icu/blob/ba1c7006b7b5acd734538b9997102d7f46203576/icu4c/source/config/Makefile.inc.in#L62

evanmiller:
See: https://trac.macports.org/ticket/67428#comment:5

catap:
I see.

maybe backport it to upstream isn’t so bad idea, is it?

catap:
@evanmiller would you like to backport it to upstream? :)

catap (approved):

evanmiller:
Per @kencu's suggestion I have updated the patch to 1) fix rpath installs 2) fix no-rpath installs and 3) disable rpath for MacPorts. This gives us a patch that we can send upstream.

catap:
I not sure about (3); I think that using rpath isn't bad if it works :)

I'll take closer look to this PR in next few days.

kencu:
macports has a policy/strong preference to not use rpaths unless there is no option, as it is too hard to debug issues. Almost all users and most devs have no clue how to be certain what libraries are really being used when rpaths are involved.

It’s best to configure all systems the same if possible

but go ahead and start using rpaths in icu if you want to… it’s not my policy, you’ll have to take it up with @ryandesign and the admins

I do know how to see which rpath libraries are really being used, when needed

catap:
After some additional thinking I'd like to agree that `@rpath` here makes things more complicated. Let keep it disabled.

@kencu may I ask you to merge this PR? It seems as ready as it can be :)

catap:
@evanmiller may you also bump a revision?

mascguy (approved):

mascguy:
> @evanmiller may you also bump a revision?

Okay, we'll wait to merge then.

evanmiller:
> @evanmiller may you also bump a revision?

done

mascguy (approved):
LGTM

kencu:
well done all, and thanks for the upstream PR too.

PR #18709, opened 2023-05-17
parallel-hashmap: require compiler.thread_local_storage to fix <10.11

barracuda156:
#### Description

Yosemite build fails due to no TLS: https://build.macports.org/builders/ports-10.10_x86_64-builder/builds/225840/steps/install-port/logs/stdio

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/parallel-hashmap/Portfile
+++ b/devel/parallel-hashmap/Portfile
@@ -23,6 +23,7 @@ patchfiles-append       0001-Warn-on-4-byte-char-not-err-out.patch \
                         0002-Unbreak-build-on-Apple-with-GCC-do-not-use-clangism.patch
 
 compiler.cxx_standard   2011
+compiler.thread_local_storage yes
 
 configure.args-append   -DPHMAP_INSTALL=ON \
                         -DPHMAP_BUILD_TESTS=ON \


devel/parallel-hashmap/Portfile
```diff
@@ -23,6 +23,7 @@ patchfiles-append       0001-Warn-on-4-byte-char-not-err-out.patch \
                         0002-Unbreak-build-on-Apple-with-GCC-do-not-use-clangism.patch
 
 compiler.cxx_standard   2011
+compiler.thread_local_storage yes
```
herbygillot, line 26: This looks enabled for all platforms. If changing compiler options, this means that the port contents would be different, and whenever this is the case, that usually necessitates a `revision` bump.

barracuda156: > This looks enabled for all platforms. If changing compiler options, this means that the port contents would be different, and whenever this is the case, that usually necessitates a `revision` bump.

I think it is not needed, since on later systems compilers support TLS, so nothing will change for them. Same applies to PowerPC (which uses new GCC).

PR #18655, opened 2023-05-13
smake:  Update to 1.7

RobK88:
#### Description

smake:  Update to 1.7

* Add Modeline for editors at the top of the Portfile
* Delete tabs in Portfile
* Add "makefile 1.0" PortGroup
* Add "conflicts_build 1.0" PortGroup
* Add "codeberg 1.0" PortGroup
* Add "muniversal 1.1" PortGroup statement
* Add "codeberg.setup" after name
* Update version from 1.2.5 to 1.7-${codeberg.version}
* Add revision number
* Delete "platforms darwin" statement
* Add myself as a maintainer
* Modify long_description
* Update master_sites
* Add and update checksums
* Remove "use_bzip2"
* Create new bash script, Gmake.smake, for building the smake binary and man pages
* Add post-extract section to copy Gmake.smake into the ${worksrcpath)
* Add patchfile to patch Makefile to run Gmake.smake script
* Add post-patch block to replace "/opt/Schily" with "${prefix}" in the source files
* Add "port_conflict_check" procedure
* Add "port_conflict_check" against cdrtools versions 0 to 3.01_1
* Update "compiler.blacklist"
* Change "compiler.blacklist" to "compiler.blacklist-append"
* Remove "use_configure no"
* Add -lintl to configure.ldflags-append
* Add build dependency - port:gmake
* Add build dependency - port:gettext-runtime
* Add library dependency - port:gettext-runtime
* Change build.args to build.args-append and update the build args
* Add CC_OPT to build.args-append
* Add note regarding disabling parallel builds
* Delete destroot.args
* Update destroot.destdir
* Delete post-destroot
* Add destroot{} code block to manually move smake binary and man pages into ${destroot}
* Delete livecheck related statements

CLOSES:  https://trac.macports.org/ticket/67405

**IMPORTANT - PLEASE READ:**  This PR for `smake` will not build if the old version of `cdrtools`, version 3.01, currently in the MacPorts repo has been installed and is active. To address this issue, a procedure called `port_conflict_check` has been added which checks whether `cdrtools`, version 3.01_1 or earlier has been installed. (Thanks to macsguy for the code).  Once `smake` is built, `cdrtools` can be activated again.

FYI -- The old version of `cdrtools`, version 3.01, in the MacPorts repo installs OLD libschily header files and libraries. It is the existence of these old header files and libraries which prevents this updated port for star from building successfully. So the previous version of `cdrtools` must be uninstalled or deactivated first.

[...]

PR #18616, opened 2023-05-10
star: Update to 1.7.0

RobK88:
#### Description

star:  update to version 1.7.0

* Fix broken port
* Reformat Portfile
* Add "makefile 1.0" PortGroup
* Add "conflicts_build 1.0" PortGroup
* Add "codeberg 1.0" PortGroup
* Add codeberg.setup after name
* Update version to 1.7.0-${codeberg.version}
* Add revision number
* Remove "platforms darwin"
* Add myself as a maintainer
* Change the description of the Port
* Modify long_description
* Remove homepage
* Remove master_sites
* Update checksums
* Remove "use_bzip2"
* Delete the patch file - files/patch-RULES__rules.prg.diff
* Add patch block to change all occurrences of gnutar in the man page to sgnutar
* Add patch block to delete "-noclobber" flag in cdrecord/Makefile.dfl
* Add patch block to replace /opt/schily with ${prefix} in the source code files
* Delete worksrcdir
* Add "port_conflict_check" procedure
* Add "port_conflict_check" procedure call to check against build conflicts with cdrtools versions 0 to 3.01_1
* Add "compiler.blacklist-append"
* Remove "use_configure no"
* Add -lintl to configure.ldflags-append
* Remove build dependency - port:gmake
* Add build dependency - port:smake
* Add build dependency - port:gettext
* Change depends_build to depends_build-append
* Add library dependency "port:gettext-runtime" to depends_lib-append
* Add library dependency "port:libiconv" to depends_lib-append
* Change build.cmd to smake
* Delete "build.jobs -1" from the Portfile
* Add "use_parallel_build      no"
* Update note regarding the disabling of parallel builds
* Add INS_BASE, INS_RBASE DEFOSINCDIRS, CC_OPT to build.args-append in the Portfile
* Add INS_RBASE, DEFINSUSR and DEFINSGRP to destroot.destdir in the Portfile
* Replace all system() calls Tcl code and MacPorts macros in the post-destroot code block
* Replace Tcl commands with preferred MacPorts macros wherever possible in the post-destroot code block
* Rename the gnutar binary to sgnutar in destroot
* Delete the rename of the tartest binary in destroot (i.e. keep the tartest name)
* Delete the rename of the tartest man page in destroot (i.e. keep the tartest name)
* Fix post-destroot block to delete unnecessary binaries, manpages and docs
* Delete livecheck related statements
* Add notes to Portfile informing users that the Schily gnutar binary has been renamed sgnutar

[...]

PR #18741, opened 2023-05-19
cdrtools:  update to version 3.02

RobK88:
#### Description
cdrtools:  update to version 3.02

* Reformat Portfile
* Add "makefile 1.0" PortGroup
* Add "conflicts_build 1.0" PortGroup
* Add "codeberg 1.0" PortGroup
* Add codeberg.setup after name
* Update version to 3.02-${codeberg.version}
* Add revision number
* Add audio to categories
* Add myself as a maintainer
* Remove BSD license
* Remove homepage
* Remove master_sites
* Update the checksums
* Remove "use_bzip2"
* Delete the patch files - patch-include_schily_sha2.h and mkisofs.c.patch
* Add patch block to delete "-noclobber" flag in cdrecord/Makefile.dfl
* Add patch block to replace "/opt/schily" with "${prefix}" in the source code files
* Add "port_conflict_check" procedure
* Add "port_conflict_check" procedure call to check against build conflicts with cdrtools versions 0 to 3.01_1
* Remove "use_configure no"
* Add "universal_variant no" to Portfile
* Change compiler.blacklist to compiler.backlist-append
* Remove "configure.cflags-append"
* Add "-lintl" to configure.ldflags-append
* Change depends_build to depends_build-append
* Change depends_lib to depends_lib-append
* Remove build dependency - port:smake
* Add build dependency - port:smake
* Add build dependency - port:gettext
* Add build dependency - port:libiconv
* Add library dependency - port:gettext-runtime
* Add library dependency - port:libiconv
* Change build.cmd to smake
* Delete "build.jobs -1" from the Portfile
* Add "use_parallel_build      no"
* Add note regarding the disabling of parallel builds
* Remove CC, COPTX, LDOPTX and LDPATH from build.args
* Change build.args to build.args-append
* Add INS_BASE, INS_RBASE, CC_OPT to build.args-append in the Portfile
* Remove post-build {} code block
* Add INS_RBASE to destroot.destdir in the Portfile
* Reformat post-distroot code block
* Fix post-destroot block to delete unnecessary binaries, manpages and docs
* Delete livecheck related statements

**IMPORTANT - PLEASE READ:**  This PR for `cdrtools` will not build if the old version of `cdrtools`, version 3.01, currently in the MacPorts repo has been installed and is active. To address this issue, a procedure called `port_conflict_check` has been added which checks whether `cdrtools`, version 3.01_1 or earlier has been installed. (Thanks to @mascguy for the code).  Once `smake` is built, `cdrtools` can be activated again.

FYI -- The old version of `cdrtools`, version 3.01, in the MacPorts repo installs OLD libschily header files and libraries. It is the existence of these old header files and libraries which prevents this updated port for `cdrtools` from building successfully. So the previous version of `cdrtools` must be uninstalled or deactivated first.

This PR for `cdrtools` no longer installs the libshily header files and libraries since they are not needed to run the binaries.  

Please see the following related Pull Requests:

(smake)  https://github.com/macports/macports-ports/pull/18655
(star)      https://github.com/macports/macports-ports/pull/18616

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G14042 x86_64
Xcode 10.1 10B61

###### Verification 
Have you

### DIFF
--- a/sysutils/cdrtools/Portfile
+++ b/sysutils/cdrtools/Portfile
@@ -1,98 +1,194 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
-PortSystem          1.0
-PortGroup           compiler_blacklist_versions 1.0
+PortSystem              1.0
+PortGroup               makefile 1.0
+PortGroup               compiler_blacklist_versions 1.0
+PortGroup               conflicts_build 1.0
+PortGroup               codeberg 1.0
 
-name                cdrtools
-version             3.01
-revision            1
-categories          sysutils
-# listing all these licenses is somewhat redundant when taken as a whole,
-# but different separable components are licensed differently
-license             CDDL-1 BSD LGPL-2.1 GPL-2+ GPL-2
-platforms           darwin
-conflicts           dvdrtools
+name                    cdrtools
+codeberg.setup          schilytools schilytools 2023-04-19
+version                 3.02-${codeberg.version}
+revision                0
+categories              sysutils audio
+conflicts               dvdrtools
+maintainers             {hotmail.com:amtor @RobK88} \
+                        openmaintainer
+license                 CDDL-1 LGPL-2.1 GPL-2+ GPL-2
 
-description         ISO 9660 filesystem and CD creation tools
-long_description    The cdrtools software includes programs to create \
-                    and/or extract ISO 9660 filesystems, verify their \
-                    integrity, and write them to a disc.
-homepage            http://cdrecord.org
-maintainers         nomaintainer
-master_sites        sourceforge:project/cdrtools
+description             ISO 9660 filesystem and CD creation tools
+long_description        The cdrtools software includes programs to create \
+                        and/or extract ISO 9660 filesystems, verify their \
+                        integrity, and write them to a disc.
 
-checksums           rmd160  3696b1bc502905cbe7c06e492b42d2ffe2dba982 \
-                    sha256  ed282eb6276c4154ce6a0b5dee0bdb81940d0cbbfc7d03f769c4735ef5f5860f \
-                    size    2087416
+checksums               rmd160  e154278ecbe7d778bc1d6766ed163c9963b1cc82 \
+                        sha256  a4270cdcca5dd69c0114079277b06e5efad260b0a099c9c09d31e16e99a23ff5 \
+                        size    5896292
 
-use_bzip2           yes
+patch {
+                        reinplace -locale C "s|-noclobber| |g" \
+                        ${worksrcpath}/cdrecord/Makefile.dfl
 
-depends_build       port:smake
-
-depends_lib         port:gettext
-
-patchfiles          patch-include_schily_sha2.h
-patchfiles-append   mkisofs.c.patch
-post-patch {
-    reinplace -locale C "s|/opt/schily|${prefix}|g" \
-        ${worksrcpath}/DEFAULTS/Defaults.darwin \
-        ${worksrcpath}/DEFAULTS/Defaults.mac-os10 \
-        ${worksrcpath}/DEFAULTS_ENG/Defaults.darwin \
-        ${worksrcpath}/DEFAULTS_ENG/Defaults.mac-os10 \
-        ${worksrcpath}/libfind/find.c \
-        ${worksrcpath}/libfind/find_main.c \
-        ${worksrcpath}/librscg/scsi-remote.c \
-        ${worksrcpath}/TEMPLATES/Defaults.gcc
+                        reinplace -locale C "s|/opt/schily|${prefix}|g" \
+                        ${worksrcpath}/btcflash/btcflash.1 \
+                        ${worksrcpath}/cdda2wav/cdda2wav.1 \
+                        ${worksrcpath}/cdrecord/cdrecord.1 \
+                        ${worksrcpath}/DEFAULTS/Defaults.darwin \
+                        ${worksrcpath}/DEFAULTS/Defaults.mac-os10 \
+                        ${worksrcpath}/DEFAULTS_ENG/Defaults.darwin \
+                        ${worksrcpath}/DEFAULTS_ENG/Defaults.mac-os10 \
+                        ${worksrcpath}/libfind/find.c \
+                        ${worksrcpath}/libfind/find_main.c \
+                        ${worksrcpath}/librscg/scsi-remote.c \
+                        ${worksrcpath}/mkisofs/diag/isoinfo.8 \
+                        ${worksrcpath}/patch/tests/random/cmptest.sh \
+                        ${worksrcpath}/patch/tests/random/gentest.sh \
+                        ${worksrcpath}/readcd/readcd.1 \
+                        ${worksrcpath}/rmt/rmt.1 \
+                        ${worksrcpath}/rscsi/default-rscsi.sample \
+                        ${worksrcpath}/rscsi/rscsi.1 \
+                        ${worksrcpath}/scgskeleton/scgskeleton.1 \
+                        ${worksrcpath}/sformat/datio.c \
+                        ${worksrcpath}/smake/job.c \
+                        ${worksrcpath}/smake/make.c \
+                        ${worksrcpath}/sunpro/Make/bin/make/common/misc.cc \
+                        ${worksrcpath}/sunpro/Make/bin/make/common/read2.cc \
+                        ${worksrcpath}/TEMPLATES/Defaults.gcc \
+                        ${worksrcpath}/TEMPLATES/Defaults.clang \
+                        ${worksrcpath}/TEMPLATES/Defaults.xcc \
+                        ${worksrcpath}/ved/ved.h
 }
 
-use_configure       no
+proc port_conflict_check {p_port_name p_conflict_ver_min p_conflict_ver_max} {
+    ui_info "Checking for conflict against port: ${p_port_name}"
 
-# hangs the same way as smake itself with both llvm-gcc-4.2 and Xcode 4.1's clang
-# https://trac.macports.org/ticket/30310
-compiler.blacklist  llvm-gcc-4.2 macports-llvm-gcc-4.2 {clang < 300}
+    if { ![catch {set port_conflict_ver_info [lindex [registry_active ${p_port_name}] 0]}] } {
+        set port_conflict_ver [lindex ${port_conflict_ver_info} 1]_[lindex ${port_conflict_ver_info} 2]
+        ui_info "${p_port_name} active version: ${port_conflict_ver}"
 
-configure.cflags-append \
-                    -DNO_SCANSTACK \
-                    -Wno-error=implicit-function-declaration
+        if { [vercmp ${port_conflict_ver} ${p_conflict_ver_min}] >= 0
+            && [vercmp ${port_conflict_ver} ${p_conflict_ver_max}] <= 0 } {
 
-build.cmd           smake
-build.args          CC=${configure.cc} \
-                    COPTX="${configure.cflags} [get_canonical_archflags cc]" \
-                    DEFOSINCDIRS=${prefix}/include \
-                    LDOPTX="${configure.ldflags} [get_canonical_archflags ld]" \
-                    LDPATH=
-
-# smake does not support -j
-# Set build.jobs to -1 to disable MacPorts from adding the -j flag
-build.jobs          -1
-
-post-build {
-    # The build script doesn't notice when a program fails to build, so
-    # we must verify it ourselves. See
-    # https://trac.macports.org/ticket/34823
-    set progs {btcflash cdda2wav cdrecord devdump isodebug isodump isoinfo isovfy mkisofs readcd scgcheck scgskeleton}
-    fs-traverse path ${worksrcpath} {
-        if {[file isfile ${path}] && [string match {*/OBJ/*} ${path}]} {
-            set file [file tail ${path}]
-            set prog_index [lsearch ${progs} ${file}]
-            if {${prog_index} != -1} {
-                set progs [lreplace ${progs} ${prog_index} ${prog_index}]
-            }
+            ui_info "${p_port_name} conflicts; declare build conflict"
+            conflicts_build-append \
+                ${p_port_name}
         }
     }
-    if {[llength ${progs}] > 0} {
-        ui_error "The following programs did not build: [join ${progs} {, }]"
-        return -code error "build failed"
-    }
 }
 
-destroot.args       DEFINSUSR=${install.user} \
-                    DEFINSGRP=${install.group} \
-                    INS_BASE=${prefix}
+port_conflict_check     cdrtools 0.0 3.01_1
+
+universal_variant       no
+
+compiler.blacklist-append \
+                        llvm-gcc-4.2 macports-llvm-gcc-4.2 {clang < 300}
+
+configure.ldflags-append -lintl
+
+depends_build-append    port:smake \
+                        port:gettext
+
+depends_lib-append      port:gettext-runtime \
+                        port:libiconv
+
+build.cmd               smake
+
+#
+# Must specify INS_BASE and INS_RBASE in both the build and destroot phases
+#
+
+build.args-append       INS_BASE="${destroot}${prefix}" \
+                        INS_RBASE="${destroot}${prefix}" \
+                        DEFOSINCDIRS="${prefix}/include" \
+                        CC_OPT=${configure.optflags}
+
+destroot.destdir        INS_BASE="${destroot}${prefix}" \
+                        INS_RBASE="${destroot}${prefix}" \
+                        MANDIR=man \
+                        DEFINSUSR=${install.user} \
+                        DEFINSGRP=${install.group}
+
+#
+# smake does not support parallel building
+# i.e. smake does not support the -j flag
+#
+use_parallel_build      no
 
 post-destroot {
-    move ${destroot}/etc/default ${destroot}${prefix}/etc/
-}
+                        set keepPrograms  [ list    ${destroot}${prefix}/bin/btcflash \
+                                                    ${destroot}${prefix}/bin/cdda2mp3 \
+                                                    ${destroot}${prefix}/bin/cdda2ogg \
+                                                    ${destroot}${prefix}/bin/cdda2wav \
+                                                    ${destroot}${prefix}/bin/cdrecord \
+                                                    ${destroot}${prefix}/bin/devdump \
+                                                    ${destroot}${prefix}/bin/isodebug \
+                                                    ${destroot}${prefix}/bin/isodump \
+                                                    ${destroot}${prefix}/bin/isoinfo \
+                                                    ${destroot}${prefix}/bin/isovfy \
+                                                    ${destroot}${prefix}/bin/mkhybrid \
+                                                    ${destroot}${prefix}/bin/mkisofs \
+                                                    ${destroot}${prefix}/bin/readcd \
+                                                    ${destroot}${prefix}/bin/scgcheck \
+                                                    ${destroot}${prefix}/bin/scgskeleton ]
 
-livecheck.url       http://sourceforge.net/projects/cdrtools/files/
-livecheck.regex     ${name}-(\[0-9.\]+)${extract.suffix}
+                        set allPrograms [ glob ${destroot}${prefix}/bin/* ]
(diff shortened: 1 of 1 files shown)


RobK88:
Please note that all the static libraries and header files have been removed in this upgrade to the `cdrtools` port.  If you think they are needed, please let me know.

sysutils/cdrtools/Portfile
```diff
@@ -1,98 +1,211 @@
...
+# Need a recent version of git to fetch the source while developing this port
+# on my old Mac running High Sierra.
+# Otherwise the fetch will fail with a SSL certificate expired error.
+#
+# Will switch to a standard tarball download for this port once development has finished.
+#
+
+depends_fetch           port:git
+
+fetch.type              git
```
catap, line 34: You don't need it here. I suggest to use technic similar with https://github.com/macports/macports-ports/pull/18655

RobK88: @catap - Since I now know how I can test this port locally on my Mac running High Sierra to overcome the Lets Encrypt cert expired bug (see https://trac.macports.org/wiki/ProblemHotlist), I have just updated this port, `cdrtools` to download the source tarball instead of using `git` as the fetch method.

RobK88:
@catap - As with the PR for `smake`, using `${git_tag}` in the macro definition for `version` does not work:

```
$ port outdated
The following installed ports are outdated:
cdrtools                       3.01_1 < 3.02-{git_tag}_0   
```
The solution is to eliminate `${git_tag}` in the `version` macro definition or just eliminate `${git_tag}` altogether and use `${livecheck.version}` instead.  I will try the latter.

RobK88:
We also have another bigger problem.  One cannot upgrade an older version of `cdrtools` (such as v3.01) to the most recent version of `cdrtools` without first uninstalling or deactivating the previous version.

If one tries, the build will fail and one will see the same errors I was experiencing when I was trying to get the port for the most recent version of `smake` to work on my Mac running High Sierra.

e.g.

```
searchinpath.c:116:19: error: use of undeclared identifier 'SIP_ONLY_PATH'
        if ((file_mode & SIP_ONLY_PATH) == 0 &&
                         ^
searchinpath.c:125:33: error: use of undeclared identifier 'SIP_NO_STRIPBIN'
                                file_mode & (SIP_PLAIN_FILE|SIP_NO_STRIPBIN)
```

@Dave-Allured found the problem and suggested that I uninstall all my previsions versions of the Schily Tools programs like `cdrtools`, version 3.01.  Once I did that I was able to install the most recent recent of `smake` and the most recent recent of `cdrtools` using the updated port in this Pull Request.

The issue is with the previous version of `cdrtools` (which is current version in the MacPorts repo).  the previous version installs unnecessary header files for libschily which are just too old to be used when building the most recent version of `cdrtools`.

What do we do here.  Do we somehow display a notice to users when they try to upgrade `cdrtools` telling them to first deactivate or uninstall their previous version before upgrading or installing `cdrtools`?

catap:
> If there is a better way to deal with this issue, please let me know.

You may use `conflicts_build` PG for such case. For example mumps can't be build when another mumps is activated.

See: https://github.com/macports/macports-ports/blob/master/math/mumps/Portfile

RobK88:
@capat - I keep learning something new.  I learned how to set a global constant in a Portfile with `set` and I learned three new commands `git.cmd`, `extract.rename` and now `conflicts_build`.  It has been a great leaning experience!

With `conflicts_build`, can one set a version?  (e.g. version 3.01 and older?)  I fixed the problem in this updated port.  `cdrtools`, version  3,02 no longer installs unneeded static libraries and header files.  So once you have successfully installed `cdrtools` version 3.02, one no longer needs to uninstall or deactivate it before installing a newer version.

RobK88:
Nope.  Neither `conflicts_build   cdrtools@3.01` or `conflicts_build   cdrtools@3.01_1` work.  I guess I will just have to use `conflicts_build   cdrtools` in the Portfile.  And after a few newer releases, I may just remove `conflicts_build` altogether from the Portfile since most people would have already upgraded to a release greater than 3.01.

RobK88:
P.S.  I suspect there is a way to only use `conflicts_build` with only certain releases by using a pre-build { } block and some tcl code.  Maybe something like:
```

pre-build {
      if { $version < 3.01 } {
             conflicts_build  cdrtools
     }
}
```

catap:
@RobK88 are you sure that such conflict won't appear in the future update?

RobK88:
Yes.  For some reason, the old port for `cdrtools` installed a lot of bloat including the libschily header files and static libraries.  Both of which were needed to build the `cdrtools` like `cdrecord` and `cdda2wav`.  But were are not needed to run them afterwards.

Years ago, the developer created its own source package for `cdrtools` which has not been updated for years.  To get the most recent version of `cdrtools`, one must now use the source in the Schily Tools tarball.  In other words, one must now compile and install all the Schily Tools which also include the `cdrtools`.

So in this newly written port for `cdrtools`, the port builds all the binaries, libraries etc from the latest version of the Schily Tools source tarball and then deletes all the unneeded binaries, header files, libraries, manpages etc in `destroot`.  I deleted all the Schily library header files and libraries in `destroot`.  

It is the existence of the old header files and libraries in `/opt/local/include/schily` that causes the problem when building newer versions of `cdrtools`.  That should not be a problem when one upgrades `cdrtools` version 3.02 to a newer release since these header files and libraries should no longer exist.  But upgrading `cdrtools` v 3.01 or earlier to version 3.02 or newer is a different matter and will fail.

P.S.  FYI -- In this rewritten port, I deleted all the bloat in `destroot` that was created in `/opt/local/lib` and `/opt/local/include`.  None of the libraries created are dynamic so they are not used to actually run the `cdrtools` like `cdrecord`, `cdda2wav` etc.  If you think they really should be included, please let me know.  In my view, if developers want the latest Schily libraries and header files for a software project that they are developing, they can compile all the Schily Tools themselves (or maybe a new port should be created for all the Schily Tools.  That would be easy to do now).

catap:
@RobK88 I suggest to keep things simpler: just add `conflicts_build  cdrtools` with comment that it is transitional, and the next update should remove it, or in couple of months. Or a year if you would like to be on very safe side.

RobK88:
Until I can figure out how to prevent `cdrtools` from building only when certain versions of the `cdrtools` port are present, I added the following to the Portfile for `cdrtools`:

PortGroup                           conflicts_build 1.0
conflicts_build-append              cdrtools

I also reformatted the post-destroot code block and replaced the Tcl commands with preferred MacPorts macros wherever possible in the the post-destroot code block. I do not think replacing file `delete -force` with just `delete` will make much difference. But apparently, it is much better to use the MacPorts macro `move` instead of the Tcl `file rename` since the MacPorts macro will correctly handle renames that only change the case of a file on a case-insensitive filesystem and the Tcl command may not.

RobK88:
@CAPAT - Please see the updated description of the PR at the top of this webpage. I did my best to explain the situation regarding the problems created by the old incompatible versions of `smake` and `cdrtools` in the MacPorts repo.

catap:
@RobK88 after some thoughts I suggest a different approach.

Create a `Portfile` for a project `schilytools` with subports for each subproject from this repository. Like `smake` or `star`.

Broken or out of updated ports can be replaced by `schilytools-xxx` ports.

As nice side effect it allows to avoid a hack with `git_tag` and use simple version which is matched upstream git tag :)

What do you think?

catap:
Also, see https://github.com/macports/macports-ports/pull/18753 where I've introduced a PG for codeberg

RobK88:
@catap - Many thanks for the great news. The "codeberg 1.0" PortGroup will defeiitely simplify things. I have rebased my `cdrtools` branch to the latest master. I have made the changes to the Portfile to use the codeberg PortGroup, rebased and force pushed.

RobK88:
@catap - I am not sure how to turn my three PR's into one.  And to be honest I do not have the time for at least a week to turn my attention to such a task.  I will be travelling soon.

IMHO, the real issue is the PR for `smake`.   Once the `smake` port has been merged into the MacPorts repo, I can finalize the PR's for `star` and `cdrtools` by making `smake` a build dependency.  (With `smake` a build dependency, it will no longer be nececcsary for the `star` and `cdrtools` ports to build a boostrapped version of `smake`.  This will speed up the builds for `star` and `cdrtools` significantly).

RobK88:
@catap - P.S.  I have also developed a port for all the Schily Tools.  (I did this to make sure I did not miss anything when developing the ports for `star` and `cdrtools`).  

The development of the Schily Tools port  was a LOT of work since the Schily Tools include a lot of tools with the SAME names as existing BSD and GNU tools (e.g.  cal, copy, compare, gnutar etc).  So I had to rename many of them with an 's' prefix (like was done with the GNU coreutils port).  And to be honest I do not know why Mac users would really want to make use of any of the Schily Tools other than `cdrtools` and `star`.  So I am not sure whether I will submit the port for all the Schily Tools.

catap:
> @catap - P.S. I have also developed a port for all the Schily Tools. (I did this to make sure I did not miss anything when developing the ports for `star` and `cdrtools`).
> 
> The development of the Schily Tools port was a LOT of work since the Schily Tools include a lot of tools with the SAME names as existing BSD and GNU tools (e.g. cal, copy, compare, gnutar etc). So I had to rename many of them with an 's' prefix (like was done with the GNU coreutils port). And to be honest I do not know why Mac users would really want to make use of any of the Schily Tools other than `cdrtools` and `star`. So I am not sure whether I will submit the port for all the Schily Tools.

Well.. This can be useful for some users. Specially if you make a lot of efforts to make such port I not sure that it is good idea to waste it.

For example I supports [beets](https://ports.macports.org/port/beets/details/) which isn't that popular and quite large if we speak about port size. And statistic says that at least 2 users are using it since I've introduced it :)

The same applies for [retdec](https://ports.macports.org/port/retdec/details/) which is also quite complicated and long to build, and which is very special software ;)

sysutils/cdrtools/Portfile
```diff
@@ -1,98 +1,205 @@
...
+                        openmaintainer
+license                 CDDL-1 LGPL-2.1 GPL-2+ GPL-2
+
+description             ISO 9660 filesystem and CD creation tools
+long_description        The cdrtools software includes programs to create \
+                        and/or extract ISO 9660 filesystems, verify their \
+                        integrity, and write them to a disc.
+
+homepage                https://codeberg.org/schilytools/schilytools
+master_sites            ${homepage}/archive/
```
mascguy, line 27: pg `codeberg` sets `homepage` and `master_sites`. So go ahead and remove these two lines.

sysutils/cdrtools/Portfile
```diff
@@ -1,98 +1,205 @@
...
-        ${worksrcpath}/TEMPLATES/Defaults.gcc
+PortSystem              1.0
+PortGroup               makefile 1.0
+PortGroup               compiler_blacklist_versions 1.0
+PortGroup               conflicts_build 1.0
+PortGroup               codeberg 1.0
+
+codeberg.setup          schilytools schilytools 2023-04-19
+
+name                    cdrtools
```
mascguy, line 11: Ensure `name` is set prior to `codeberg.setup`, to ensure `dist_subdir` and friends are set correctly.

sysutils/cdrtools/Portfile
```diff
@@ -1,98 +1,205 @@
...
+# Replace "port:gmake" with "port:smake" as a build dependency
+# and change build.cmd to "smake" and delete the pre-destroot block
+#
+
+# depends_build-append    port:smake \
+                          port:gettext-runtime \
+                          port:libiconv
+
+depends_build-append    port:gmake \
+                        port:gettext-runtime \
```
mascguy, line 88: Change the build dep to `gettext`.

sysutils/cdrtools/Portfile
```diff
@@ -1,98 +1,205 @@
...
+                        port:libiconv
```
mascguy, line 89: Since `libiconv` is already a lib dep, you don't need to declare a build dep for it.

sysutils/cdrtools/Portfile
```diff
@@ -1,98 +1,205 @@
...
+
+# build.cmd               ${prefix}/bin/smake
+
+#
+# Use the ./Gmake script provided by the developer to build the port.
+# ./Gmake first builds a bootstrapped version of smake which is used to build
+# the rest of the port.
+#
+
+build.cmd               ${worksrcpath}/Gmake
```
mascguy, line 102: Just to verify, is this the correct case for the bundled `gmake`? (If not, this will fail on our buildbots, as we use a case-sensitive filesystem there.)

mascguy: > Just to verify, is this the correct case for the bundled `gmake`? (If not, this will fail on our buildbots, as we use a case-sensitive filesystem there.)

Checked the distfiles, and sure enough, `Gmake` is correct. All good!

RobK88: @macsguy -- I am updating this port given all your great feedback on the `smake` PR.  And yes the capital G in `./Gmake` is correct.  It is a script provided by the developer.  In any event, this statement will change!  since you just merged the `smake` PR.  

Because MacPorts now has an up to date version of `smake` in its repo,  this line will change to `build.cmd   smake`.   The `port.smake` is also being added as a build dependency and `port:gmake` is being removed.  I will do a rebase and force push VERY soon.

sysutils/cdrtools/Portfile
mascguy: > Since `libiconv` is already a lib dep, you don't need to declare a build dep for it.

it looks like this was the only thing you missed, with your latest push.

RobK88:
@mascguy - Just did a force push incorporating all your suggested changed from the previous `smake` port.  This PR for `cdrtools` has also been updated to use the recently merged `smake` port to build all the `cdrtools`.  The Portfile should be easier to understand now.  I am currently building the +univesral variant to see if that will work.

sysutils/cdrtools/Portfile
RobK88: Are you sure?  I see "localization" of various files during both the build and destroot stages.  

I soon discovered that I needed to add the "-lintl" flag to "configure.ldflags-append".  Otherwise the build would fail on Big Sur and above.

Are you telling me that I just need the "-intl" flag?  And I do not need the "libiconv" as a build dependency?  Is that right?

sysutils/cdrtools/Portfile
RobK88: Done.

sysutils/cdrtools/Portfile
RobK88: Done.

sysutils/cdrtools/Portfile
RobK88: Done!

sysutils/cdrtools/Portfile
mascguy: > Are you sure? I see "localization" of various files during both the build and destroot stages.
> 
> I soon discovered that I needed to add the "-lintl" flag to "configure.ldflags-append". Otherwise the build would fail on Big Sur and above.
> 
> Are you telling me that I just need the "-intl" flag? And I do not need the "libiconv" as a build dependency? Is that right?

In short, all lib deps are also build deps. So if it's declared as a lib dep, it will also be pulled in for the build. (You'll still need your linker flag though, obviously.)

RobK88:
The universal build failed.  I guess I will need to remove the "muniversal 1.1" PortGroup and add "universal_variant no".

catap:
> The universal build failed. I guess I will need to remove the `muniversal 1.1` PortGroup and add `universal_variant no` to the Portfile

I may take a look on this and suggest a patch; but in an hour or two

RobK88:
I cannot even build the non-universal x86_64 binaries now.  The build fails in the install stage.  Before I could build the +universal variant of `cdrtools`, I had to install the `+universal` variant of `smake`.  Maybe it is the `+universal` variant of `smake` that is the problem....  Will see.  Time for testing.

mascguy (approved):
Looks great at this point. But we'll wait to merge, until you're done with your Universal testing.

RobK88:
@mascguy - Thanks!  I hope to be finished testing later tonight.  Making +universal builds is VERY slow.  I will let you know when I am done.

RobK88:
@catap - No.  I can build the +universal variant for `smake` which is needed to build the +universal variant for `cdrtools`.  But I cannot build the +universal variant for `cdrtools` using the `muniversal 1.1` Portgroup.  It fails every time.  See attached log.

[cdrtools universal variant Build Log.txt](https://github.com/macports/macports-ports/files/11559581/cdrtools.universal.variant.Build.Log.txt)

catap:
@RobK88 the log file contins:
```
--->  Running post destroot cdrtools for architecture x86_64
Error: Failed to destroot cdrtools: no files matched glob pattern "/opt/local/var/macports/build/_Users_grinch_Development_MacPorts_local-repo_ports_sysutils_cdrtools/cdrtools/work/destroot-x86_64/opt/local/bin/*"
```
that means that your `post-destroot` phase failed. My bet that it fails on line:
```
                        set allPrograms [ glob ${destroot}${prefix}/bin/* ]
```

RobK88:
> that means that your `post-destroot` phase failed. My bet that it fails on line:
> 
> ```
>                         set allPrograms [ glob ${destroot}${prefix}/bin/* ]
> ```

Yep.  You are most likely right.  We might be able to fix it but we would probably have to add a lot of code in `post-destroot` when the universal variant is being built. (i.e.  `if [variant_isset "universal"] {  lots of new code.... }`

Or maybe you can suggest an easier fix.  Or we just go with `universal_variant no` in the Portfile.  

Let me build the `+universal` variant again for `cdrtools` and see exactly what is being built in `destroot`.

catap:
@RobK88 or you may simplify that by installing everything into `${prefix}/libexec/cdrtools` and adding a symlinks from `${prefix}/bin/cdda2mp3` to `${prefix}/libexec/cdrtools/bin/cdda2mp3` and so on.

RobK88:
@catap -- It is not the post-destroot phase that failed.  It is the destroot phase that is supposed to create the universal binaries that is failing.  The `munivesral` PortGroup creates a numbers of folders in the ${worksrcdir}:

```
destroot
destroot-i386
destroot-x86-64
```

If you look in the `/opt/local/bin` sub-directories inside the `destroot-i386` or `destroot-x86-64` folders, you will see NO binaries.  But if you look in the `/opt/local/bin` folder inside the `destroot` folder, you will see binaries.  But all of them are just 64 bit.  No way to create universal binaries. 

IMHO, the `muniversal` PortGroup is juts not working right with this port.  

You also see a number of warnings that you ONLY see when making the universal build:
`NOTICE: Partial source (SRCROOT/lib) missing`

My recommendation:  Just add `universal_variant no` to the Portfile for `cdrtools`.

RobK88:
P.S.  If you inside the `${destroot}/2023-04-19-i386` folder, you also only see 64 bit binaries -- NO 32-bit binaries.  It does not look like any 32 binaries are being built.   It is probably a problem with `smake` not providing the right `${arch}` flag to the compiler.  I might be able to fix it by providing the arch info to the compiler via a `$COPTX` in `build.args-append`.

I would leave `universal_variant no` in the Portfile for now.  I will see if I can fix it next week when I am back home.  I have to start packing for my trip back North.

mascguy:
> I would leave `universal_variant no` in the Portfile for now. I will see if I can fix it next week when I am back home. I have to start packing for my trip back North.

Works for me! I'll go ahead and merge, and you can revisit on your return.

Great work on this Rob!

RobK88:
@mascgiy -- Thanks.  

P.S.  I doubt adding the following to `build.args-append` would help with the universal build.:

```
COPTX="${configure.cflags} [get_canonical_archflags cc]" \
LDOPTX="${configure.ldflags} [get_canonical_archflags ld]"
```

Why?  The archflag is NOT set when you try to install a port with the +universal variant.  You will see:

```
sudo port -v install cdrtools +universal
Checking for conflict against port: cdrtools
Error: universal: merge: architecture is not set
Error: Unable to open port: unknown architecture (empty)
```
You probably could fix it by hacking the `muniversal` PortGroup to ensure that the right arch flag is being sent to `smake` during universal builds.  Maybe create a new PortGroup called `smakeFixUniversal` that will do the trick.

catap:
Ah, its clear.

Let see if someone really need that tools as universal and I agree that switching it requires too much efforts and probably for nothing.

RobK88:
I have been able to make this port universal.  When I get a chance I will submit a PR next week.

RobK88:
FYI -- I submitted a PR for the +universal variant for `cdrtools`
see
https://github.com/macports/macports-ports/pull/20704

PR #18766, opened 2023-05-21
ipfs: pin version for macOS up to 10.12

catap:
#### Description

I haven't bump revision or increase epoch because it downgrade to the version which can be compiled by pinned go version.

Closes: https://trac.macports.org/ticket/66833

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.12.6 16G2136 x86_64
Xcode 9.2 9C40b

###### Verification 
Have you

### DIFF
--- a/net/ipfs/Portfile
+++ b/net/ipfs/Portfile
@@ -3,11 +3,28 @@
 PortSystem          1.0
 PortGroup           golang 1.0
 
-go.setup            github.com/ipfs/kubo 0.20.0 v
+set version         0.20.0
+revision            0
+
+checksums           rmd160  057d1ee189055af41b1e02c43ebaaef9194bfd54 \
+                    sha256  fbe469b4b3d70a171a94cc86e4aa32b4e3ccef65ba2440421d3bd3ffeff66eb2 \
+                    size    15477176
+
+# the last IPFS which can be build 1.17.13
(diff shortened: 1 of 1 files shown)


catap:
@herbygillot let's move forward with `if` solution :)

mascguy (approved):
LGM. @herbygillot ?

herbygillot:
I think we need to increase the `epoch` inside of the `if` block, since those macOS versions will be seeing a version _decrement_.

catap:
@herbygillot not sure because it is the last version which was possible to build on 10.12 and before.

catap:
But let's do it. To be on safe side.

mascguy:
> @herbygillot not sure because it is the last version which was possible to build on 10.12 and before.

Yep, agreed. But it's fine to increase it, so no worries.

herbygillot:
I understand, but it's a question of the last version indexed by MacPorts on that macOS version, not so much what version last built successfully.

The previous version of `ipfs` last indexed by MacPorts on these older macOS systems would see this new version as a _lower_ version, and thus I don't think MacPorts would see it as an upgrade.

This is why I think we need to increment `epoch` just for the older macOS versions.

catap:
@herbygillot it's done. Let's merge it!

herbygillot:
Thanks @catap

PR #18787, opened 2023-05-23
diffutils:  fix bug with +universal variant

RobK88:
#### Description

* Bump revision number
* Add pre-configure{} code block to add the "--disable-year2038" configure flag for +universal builds
* Change "configure.args" to "configure.args-append"
* Add "--disable-silent-rules" flag to "configure.args-append"
* Reformat Portfile

CLOSES:  https://trac.macports.org/ticket/67475

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G14042 x86_64
Xcode 10.1 10B61

###### Verification 
Have you

### DIFF
--- a/sysutils/diffutils/Portfile
+++ b/sysutils/diffutils/Portfile
@@ -1,32 +1,31 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
-PortSystem          1.0
+PortSystem                  1.0
 
-PortGroup           compiler_blacklist_versions 1.0
+PortGroup                   compiler_blacklist_versions 1.0
 
-name                diffutils
-version             3.10
-revision            0
+name                        diffutils
+version                     3.10
+revision                    1
 
-categories          sysutils textproc devel
-license             GPL-3+
+categories                  sysutils textproc devel
+license                     GPL-3+
 
-maintainers         nomaintainer
+maintainers                 nomaintainer
 
-description         GNU diff utilities
-long_description \
-    diffutils contains the GNU diff, diff3, sdiff, and cmp utilities.   \
-    Their features are a superset of the Unix features and they are     \
-    significantly faster.
+description                 GNU diff utilities
+long_description            diffutils contains the GNU diff, diff3, sdiff, and cmp \
+                            utilities.  Their features are a superset of the Unix \
+                            features and they are significantly faster.
 
-homepage            https://www.gnu.org/software/diffutils/
-master_sites        gnu
+homepage                    https://www.gnu.org/software/diffutils/
+master_sites                gnu
 
-checksums           rmd160  395b91c09409302f7e893ea362116737a3c01a01 \
-                    sha256  90e5e93cc724e4ebe12ede80df1634063c7a855692685919bfe60b556c9bd09e \
-                    size    1624240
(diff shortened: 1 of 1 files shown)


ryandesign:
Please don't reformat the portfile and make functional changes all within one commit. It's impossible to see what functional changes you're proposing when you do that. If you must reformat, do so in a separate commit.

Don't use `--disable-year2038` only in the universal variant. It is needed for pure 32-bit builds as well. It should be harmless to use this flag for all builds. All this flag does is prevent the configure script from complaining when it detects a 32-bit build is being performed.

There is no need to increase the revision since it does not change how the port installs, in those situations where it could install before.

ryandesign:
> CLOSES: https://trac.macports.org/ticket/67475

Please add `Closes: https://trac.macports.org/ticket/67475` to the commit message (not just the PR description) so that Trac will close the ticket when this PR is merged.

mascguy (approved):
LGTM. I'll update the commit message on merge, per Ryan's previous comment.

RobK88:
@masguy @ryandesign - Thanks for the feedback.. I will remember to add "CLOSES: PR#" to my commit messages in the future and will try to avoid reformatting other people's Portfiles.

sysutils/diffutils/Portfile
```diff
@@ -1,32 +1,31 @@
...
 
-PortGroup           compiler_blacklist_versions 1.0
+PortGroup                   compiler_blacklist_versions 1.0
 
-name                diffutils
-version             3.10
-revision            0
+name                        diffutils
+version                     3.10
+revision                    1
```
jmroot, line 9: The revision shouldn't be increased when fixing a problem that prevented installation.

sysutils/diffutils/Portfile
```diff
@@ -51,20 +50,26 @@ use_xz              yes
...
+depends_lib-append          port:gettext-runtime
 
-patchfiles          patch-src_system.h.diff
-configure.args      --program-prefix=g \
-                    --infodir=${prefix}/share/info \
-                    --mandir=${prefix}/share/man
+patchfiles                  patch-src_system.h.diff
 
+# Bypass year 2038 awareness code that causes the universal variant to fail.
+pre-configure {
```
jmroot, line 62: There's no need to do this in pre-configure. As Ryan pointed out, the flag can be used unconditionally, but even if it should be used only when building universal, the appropriate check is:
```
if {$universal_possible && [variant_isset universal]} {
…
}
```

sysutils/diffutils/Portfile
```diff
@@ -51,20 +50,26 @@ use_xz              yes
...
+                            if [variant_isset "universal"] {
```
jmroot, line 63: Remember to [brace your expressions](https://wiki.tcl-lang.org/page/Brace+your+expr%2Dessions).

sysutils/diffutils/Portfile
RobK88: Thanks for the feedback.  I was not sure how this additional new flag would affect 64-bit builds.  That was the main reason I attempted to limit it to universal builds.  And thank you for the code.  I will remember it next time I ever need to limit my code to universal builds.

sysutils/diffutils/Portfile
RobK88: Thanks.  I am still learning tcl.

RobK88:
FYI -- I submitted another PR to fix 32-builds as noted by @ryandesign 

See
https://github.com/macports/macports-ports/pull/18890

PR #18792, opened 2023-05-24
GLC_lib: fix build on 10.6: old Xcode gcc-4.* does not support required cxx_standard

barracuda156:
#### Description

Fix the port on 10.6 (and earlier).

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/graphics/GLC_lib/Portfile
+++ b/graphics/GLC_lib/Portfile
@@ -9,6 +9,7 @@ platforms           darwin
 name                GLC_lib
 version             2.2.0
 revision            0
+license             LGPL-2.1
 maintainers         nomaintainer
 description         C++ class library that enables the quick creation of an OpenGL application based on QT4.
 long_description    ${description}
@@ -36,8 +37,8 @@ post-patch {
 configure.args      GLC_lib.pro
 
 # https://github.com/laumaya/GLC_lib/issues/30
-compiler.cxx_standard 2003
-configure.cxxflags-append -std=c++03
+compiler.cxx_standard 1998
+configure.cxxflags-append -std=c++98
 
 livecheck.type      regex
 livecheck.url       http://sourceforge.net/api/file/index/project-id/153150/mtime/desc/rss?path=%2Fglc-lib


graphics/GLC_lib/Portfile
```diff
@@ -38,6 +38,8 @@ configure.args      GLC_lib.pro
 # https://github.com/laumaya/GLC_lib/issues/30
 compiler.cxx_standard 2003
 configure.cxxflags-append -std=c++03
+# Setting 2003 standard does not prevent Xcode gcc-4.* from being picked, but they do not acknowledge -std=c++03 flag.
+compiler.blacklist-append *gcc-4.0 *gcc-4.1 *gcc-4.2
```
chrstphrchvz, line 42: The `-std=c++03` flag was first recognized by GCC 4.7, although it appears to be equivalent to `-std=c++98`: https://github.com/gcc-mirror/gcc/commit/c7dc8804a86017317cd7d3e5f65f715d5c4d5da2

Is this build error merely due to the flag not being recognized, such that using `-std=c++98` instead would work? Or does this port need C++03 behavior only available in GCC ≥ 4.3?

If the latter, then ideally that can be handled in MacPorts base (which currently treats `compiler.cxx_standard 2003` the same as `compiler.cxx_standard 1998`) and mentioned on https://trac.macports.org/wiki/CompilerSelection

barracuda156: @chrstphrchvz Thank you, I will try that.

barracuda156: UPD. Yeah, seems to work fine.

barracuda156:
@chrstphrchvz Is this good to merge now?

graphics/GLC_lib/Portfile
```diff
@@ -37,7 +38,12 @@ configure.args      GLC_lib.pro
 
 # https://github.com/laumaya/GLC_lib/issues/30
 compiler.cxx_standard 2003
-configure.cxxflags-append -std=c++03
+
+if {${configure.compiler} in [list apple-gcc-4.0 apple-gcc-4.2 gcc-4.0 gcc-4.1 gcc-4.2 llvm-gcc-4.2]} {
+    configure.cxxflags-append -std=c++98
+} else {
+    configure.cxxflags-append -std=c++03
+}
```
chrstphrchvz, line 46: Since there is apparently no need for C++03 rather than C++98, I would prefer to always use `-std=c++98` and set `compiler.cxx_standard` to `1998`.

```suggestion
configure.cxxflags-append -std=c++98
```

chrstphrchvz (changes requested):

kencu:
this set of commits now says the port requires cxx03 and yet sets it to c++98, so more fixing up to do

graphics/GLC_lib/Portfile
```diff
@@ -37,7 +38,7 @@ configure.args      GLC_lib.pro
 
 # https://github.com/laumaya/GLC_lib/issues/30
 compiler.cxx_standard 2003
```
chrstphrchvz, line 40: As previously discussed, C++03 is seemingly not required, and `compiler.cxx_standard` currently treats `2003` the same as `1998`. So let’s use:
```suggestion
compiler.cxx_standard 1998
```
for consistency with `-std=c++98`.

barracuda156:
@chrstphrchvz Sure, done!

barracuda156:
@chrstphrchvz Yes, good to merge.

reneeotten (changes requested):
Please clean-up the commits here

barracuda156:
> Please clean-up the commits here

@reneeotten Done, rebased to the master.

barracuda156:
@reneeotten Is it mergeable now?

PR #18806, opened 2023-05-25
MoltenVK: Update to v1.2.4

Gcenx:
#### Description

As MoltenVK repo now provides prebuilt artifacts that support macOS High Sierra and later lets make use of them.

For macOS Sierra & OS X El Capitan fallback to the VulkanSDK that provided MoltenVK v1.1.9 and by making use of `darwin any` the package built on 10.12 buildbot will be usable for 10.11.

<br>

While not ideal this gets some version of MoltenVK on all supported systems, the reason the MoltenVK repo artifacts are 10.13 is due to changes in Xcode14.

###### Tested on

macOS 13.3.1 22E772610a x86_64
Xcode 14.3 14E222b

###### Verification 

### DIFF
--- a/graphics/MoltenVK/Portfile
+++ b/graphics/MoltenVK/Portfile
@@ -3,14 +3,12 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        KhronosGroup MoltenVK 1.1.9 v
-epoch               1
-revision            1
+github.setup        KhronosGroup MoltenVK 1.2.4 v
+github.tarball_from releases
 
-set sdkversion      1.3.211.0
 categories          graphics
-maintainers         {ryandesign @ryandesign} openmaintainer
-platforms           macosx
+maintainers         {@gcenx gmail.com:gcenx83} openmaintainer
+platforms           {macosx any >= 15}
 license             Apache-2
 
 # MoltenVK would build for i386, but it uses Metal which only works on x86_64 and arm64
@@ -22,29 +20,21 @@ long_description    ${name} is an implementation of the high-performance, \
                     industry-standard Vulkan graphics and compute API, that \
                     runs on Apple's Metal graphics framework.
 
-distname            vulkansdk-macos-${sdkversion}
-use_dmg             yes
+dist_subdir         ${name}/${version}
+distname            MoltenVK-macos
+use_tar             yes
 
-# url only works for the latest avalible SDK, older versions will 404
-master_sites        https://sdk.lunarg.com/sdk/download/${sdkversion}/mac/
-
-checksums           sha256  bfe654af00030b6e65521f834f0830f15e18c828594226865f15c92a9ea68363 \
-                    rmd160  3674f67b37f7bcc1746d55c3baa975d0cca9dbaa \
-                    size    275553243
-
-depends_build       port:p7zip
-depends_skip_archcheck p7zip
+checksums           sha256  168462e4b26a31184497ac7c4a6393708fe23ee4d90da347f00372ed09460714 \
+                    rmd160  c5b41ac440db29e524c01063c9f3cf8554ccbc4a \
+                    size    46776320
 
 variant universal   {}
 use_configure       no
 
-build {
-    # bypass the installer that requires macOS 10.13
-    system "${prefix}/bin/7z x -aoa ${worksrcpath}/InstallVulkan.app/Contents/Resources/installer.dat -o${workpath}/VulkanSDK"
-}
+build {}
 
 destroot {
-    set output_dir ${workpath}/VulkanSDK
+    set output_dir ${workpath}/MoltenVK
 
     # Xcode11 and later are required to use "xcframework"
     # Headers currently break build due to Xcode 12 ProcessXCFramework bug:
@@ -53,27 +43,70 @@ destroot {
         file copy ${output_dir}/MoltenVK/MoltenVK.xcframework ${destroot}${frameworks_dir}
     }
 
-    file copy ${output_dir}/macOS/bin/MoltenVKShaderConverter ${destroot}${prefix}/bin
-    file attributes ${destroot}${prefix}/bin/MoltenVKShaderConverter -permissions +x
-    file copy ${output_dir}/macOS/lib/libMoltenVK.dylib ${destroot}${prefix}/lib
+    file copy ${output_dir}/MoltenVK/dylib/macOS/libMoltenVK.dylib ${destroot}${prefix}/lib
 
     # vulkan and vk_video are provided via vulkan-headers
     file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
 
-    if {(![variant_isset universal])} {
+    if {![variant_isset universal] || ![variant_exists universal]} {
         system -W ${destroot}${prefix}/lib "lipo -thin ${configure.build_arch} libMoltenVK.dylib -o libMoltenVK.dylib 2> /dev/null"
     }
 }
 
-platform darwin {
-    if {${os.major} < 16} {
-        archive_sites
-        distfiles
-        depends_build
-        known_fail  yes
-        pre-fetch {
-            ui_error "${subport} @${version} requires macOS Sierra or later"
-            return -code error "incompatible OS X version"
+if {${subport} eq ${name}} {
+    PortGroup           stub 1.0
+    version             1.0
+    epoch               2
+    if {${os.major} < 17} {
+        depends_run     port:MoltenVK-1.1.9
+    } else {
+        depends_run     port:MoltenVK-latest
+    }
+}
+
+# GitHub artifacts deployment target is 10.13 (Xcode14+)
+subport MoltenVK-latest {
+    platforms           {macosx any >= 17}
+    conflicts           MoltenVK-1.1.9
+}
+
+subport MoltenVK-1.1.9 {
+    github.setup        KhronosGroup MoltenVK 1.1.9 v
+
+    conflicts           MoltenVK-latest
+    set sdkversion      1.3.211.0
+
+    distname            vulkansdk-macos-${sdkversion}
+    use_dmg             yes
+    use_tar             no
+
+    # url only works for the latest avalible SDK, older versions will 404
+    master_sites        https://sdk.lunarg.com/sdk/download/${sdkversion}/mac/
+
+    checksums           sha256  bfe654af00030b6e65521f834f0830f15e18c828594226865f15c92a9ea68363 \
+                        rmd160  3674f67b37f7bcc1746d55c3baa975d0cca9dbaa \
+                        size    275553243
+
+    depends_build       port:p7zip
+    depends_skip_archcheck p7zip
+
+    build {
+        # bypass the installer that requires macOS 10.13
+        system "${prefix}/bin/7z x -aoa ${worksrcpath}/InstallVulkan.app/Contents/Resources/installer.dat -o${workpath}/VulkanSDK"
+    }
+
+    destroot {
+        set output_dir ${workpath}/VulkanSDK
+
+        file copy ${output_dir}/macOS/bin/MoltenVKShaderConverter ${destroot}${prefix}/bin
+        file attributes ${destroot}${prefix}/bin/MoltenVKShaderConverter -permissions +x
+        file copy ${output_dir}/macOS/lib/libMoltenVK.dylib ${destroot}${prefix}/lib
+
+        # vulkan and vk_video are provided via vulkan-headers
+        file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
+
+        if {![variant_isset universal] || ![variant_exists universal]} {
+            system -W ${destroot}${prefix}/lib "lipo -thin ${configure.build_arch} libMoltenVK.dylib -o libMoltenVK.dylib 2> /dev/null"
         }
     }
 }


graphics/MoltenVK/Portfile
```diff
@@ -3,11 +3,12 @@
...
 
-github.setup        KhronosGroup MoltenVK 1.1.9 v
+github.setup        KhronosGroup MoltenVK 1.2.4 v
+github.tarball_from releases
 epoch               1
-revision            1
+revision            0
 
-set sdkversion      1.3.211.0
+platforms           {darwin any >= 15}
```
ryandesign, line 11: If I'm not mistaken it should be:
```suggestion
platforms           {macosx any} {macosx >= 15}
```
And remove the `platforms macosx` line 3 lines down which would otherwise override it.

graphics/MoltenVK/Portfile
```diff
@@ -22,29 +23,20 @@ long_description    ${name} is an implementation of the high-performance, \
                     industry-standard Vulkan graphics and compute API, that \
                     runs on Apple's Metal graphics framework.
 
-distname            vulkansdk-macos-${sdkversion}
-use_dmg             yes
+distname            MoltenVK-macos
+use_tar             yes
```
ryandesign, line 27: They distribute... an uncompressed tar file? What is the matter with these people?

graphics/MoltenVK/Portfile
```diff
@@ -65,15 +55,49 @@ destroot {
...
+        set output_dir ${workpath}/VulkanSDK
+
+        file copy ${output_dir}/macOS/bin/MoltenVKShaderConverter ${destroot}${prefix}/bin
+        file attributes ${destroot}${prefix}/bin/MoltenVKShaderConverter -permissions +x
+        file copy ${output_dir}/macOS/lib/libMoltenVK.dylib ${destroot}${prefix}/lib
+
+        # vulkan and vk_video are provided via vulkan-headers
+        file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
+
+        if {(![variant_isset universal])} {
```
ryandesign, line 99:
```suggestion
        if {![variant_isset universal]} {
```

graphics/MoltenVK/Portfile
Gcenx: _Sigh_ guess I wasn’t so careful when making that commit…

I believe the current usage would be correct as the package built via the build would be viable for 10.11 and later versions of macOS, only thing would be is the usage of darwin or macosx as the platform.

jasonliu--:
@ryandesign , @Gcenx , is this PR ready for merging? I'd love to see whether I'm able to get some of the Vulkan-related ports working on El Capitan, which this PR might finally enable. 🤞

Gcenx:
@jasonliu-- that’s upto @ryandesign

graphics/MoltenVK/Portfile
ryandesign: > only thing would be is the usage of darwin or macosx as the platform

Well the port previously indicated that it required macOS. Now you've changed it to indicate that it also works on pure Darwin, which seems unlikely given that it requires Metal which I assume is only in macOS and not in pure Darwin. So I suggest you change `darwin` back to `macosx`.

ryandesign: > I believe the current usage would be correct

I was just going based off the documentation @jmroot left in his commit message at https://github.com/macports/macports-base/commit/28666188b01ef181a93b36b7468807ae558c5d3b. Your syntax `{darwin any >= 15}` is not one of the examples Josh gave so I don't know if it works.

ryandesign:
@Gcenx you know more about this software and this port than I do at this point. If you'd like to take over maintenance, you can change the `maintainers` line to yourself, otherwise you can change it to `nomaintainer` (or I'll do so later).

graphics/MoltenVK/Portfile
jasonliu--: > > only thing would be is the usage of darwin or macosx as the platform
> 
> Well the port previously indicated that it required macOS. Now you've changed it to indicate that it also works on pure Darwin, which seems unlikely given that it requires Metal which I assume is only in macOS and not in pure Darwin. So I suggest you change `darwin` back to `macosx`.

As far as I'm aware, Apple's graphics frameworks, such as Quartz and Metal, are closed source, and only exist as a part of macOS, but don't exist as a part of the open source Darwin OS. So I think that @ryandesign is correct on this one.

Gcenx: Changed back to macosx

jmroot: > I was just going based off the documentation @jmroot left in his commit message at [macports/macports-base@2866618](https://github.com/macports/macports-base/commit/28666188b01ef181a93b36b7468807ae558c5d3b). Your syntax `{darwin any >= 15}` is not one of the examples Josh gave so I don't know if it works.

Support for this combination was added a bit later in 3d9856b. It both sets the platform segment of the archive filename to `darwin_any` so the same archive will be used by all darwin versions, and sets `known_fail yes` on darwin < 15.

This port goes a little bit outside the anticipated usage, since the same archive is not used by all darwin versions. But since the older darwin versions also get a different port version, I guess it works.

As Ryan knows, my preference is to never have a port change its version depending on context like the OS version, but instead have a different port (or subport) for each version we want to provide.

Gcenx: @jmroot so you’d prefer having say `moltenvk_legacy` or similar

I’d issues with the libsdl2 port that did this forcing another port for 10.6 support, as other portfiles do this already I’d assumed this was fine.

graphics/MoltenVK/Portfile
```diff
@@ -65,15 +55,49 @@ destroot {
...
+        set output_dir ${workpath}/VulkanSDK
+
+        file copy ${output_dir}/macOS/bin/MoltenVKShaderConverter ${destroot}${prefix}/bin
+        file attributes ${destroot}${prefix}/bin/MoltenVKShaderConverter -permissions +x
+        file copy ${output_dir}/macOS/lib/libMoltenVK.dylib ${destroot}${prefix}/lib
+
+        # vulkan and vk_video are provided via vulkan-headers
+        file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
+
+        if {![variant_isset universal]} {
```
jmroot, line 99: Needs to also check `[variant_exists universal]`, since the universal variant will be removed when using an SDK that only supports a single arch. Same on line 53 (which can also have its parens removed).

graphics/MoltenVK/Portfile
jmroot: > @jmroot so you’d prefer having say `moltenvk_legacy` or similar

Or `MoltenVK-1.1.9`, or whatever.

> I’d issues with the libsdl2 port that did this forcing another port for 10.6 support, as other portfiles do this already I’d assumed this was fine.

Yes, dynamic versions are currently a fairly widespread practice, which is unfortunate IMO. Not everyone agrees.

Gcenx: @jmroot thinking about it having a versions subport seems more ideal as who knows when/if this will change again.

I’ll make MoltenVK into a stub and add a runtime dep of `MoltenVK-1.1.9` or `MoltenVK-latest` depending on macOS version that should be better and can always be expanded later.

graphics/MoltenVK/Portfile
Gcenx: I’ll fix this while splitting into stub/subports

graphics/MoltenVK/Portfile
jasonliu--: Do you all think that MoltenVK is going to be one of those ports that needs to be split into version-suffixed subports? The Khronos Group has released Mac packages for MoltenVK that are able to be run on macOS all the way back to 10.11 El Capitan (which is the version of macOS where Apple introduced Metal) since some of MoltenVK's earliest releases. Unless the Khronos Group makes some kind of major update to Vulkan that would break backwards-compatibility, I can't really think of a reason why MoltenVK would need to be split into multiple subports. Or am I missing previous discussions about MoltenVK that I'm unaware of?

Gcenx: @jasonliu-- the VulkanSDK is packaged in the dumbest way possible with a dmg that can’t be mounted on older macOS versions and then a custom installer. The installed also changed.

They’ve refused to change how the VulkanSDK are packaged.

We can’t build from source using buildbots as modern Xcode versions lowest deployment target is 10.13.

The only viable solution is making use of the newly provided archives from GitHub releases.

jasonliu--: > @jasonliu-- the VulkanSDK is packaged in the dumbest way possible with a dmg that can’t be mounted on older macOS versions and then a custom installer. The installed also changed.
>
> They’ve refused to change how the VulkanSDK are packaged.

Now that you mention it, I do remember that happening to certain DMGs after APFS came along, although I think most projects figured out how to create packages that avoided the issue. Sad to hear that the people at Khronos Group aren't doing it.

> We can’t build from source using buildbots as modern Xcode versions lowest deployment target is 10.13.

And I take it that building from source is failing on 10.12? I know that attempting to build from source failed for me on my 10.11 machine.

So are you implying that it's already impossible to get MoltenVK 1.2.4 to work on 10.11? Or maybe you are anticipating newer versions becoming impossible to work on 10.12 and 10.11 in future releases of MoltenVK? If that's true, then I would understand why you would want to split into version-numbered subports now, to be able to keep a "newest version that still worked on macOSes older than _x_" available into the future.

graphics/MoltenVK/Portfile
Gcenx: That change doesn’t really apply to the way this port works.

graphics/MoltenVK/Portfile
```diff
@@ -53,27 +45,75 @@ destroot {
...
-    file copy ${output_dir}/macOS/bin/MoltenVKShaderConverter ${destroot}${prefix}/bin
-    file attributes ${destroot}${prefix}/bin/MoltenVKShaderConverter -permissions +x
-    file copy ${output_dir}/macOS/lib/libMoltenVK.dylib ${destroot}${prefix}/lib
+    file copy ${output_dir}/MoltenVK/dylib/macOS/libMoltenVK.dylib ${destroot}${prefix}/lib
 
     # vulkan and vk_video are provided via vulkan-headers
     file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
 
-    if {(![variant_isset universal])} {
+    if ![variant_isset universal] {
```
jmroot, line 53: The braces are still desirable. Only the parentheses were redundant.

graphics/MoltenVK/Portfile
```diff
@@ -3,14 +3,14 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        KhronosGroup MoltenVK 1.1.9 v
+github.setup        KhronosGroup MoltenVK 1.2.4 v
+github.tarball_from releases
 epoch               1
```
jmroot, line 8: No need to set epoch for the new subports.

graphics/MoltenVK/Portfile
```diff
@@ -53,27 +45,75 @@ destroot {
...
+# GitHub artifacts deployment target is 10.13 (Xcode14+)
+subport MoltenVK-latest {
+    platforms           {macosx any >= 17}
+    conflicts           MoltenVK-1.1.9
+}
+
+subport MoltenVK-1.1.9 {
+    github.setup        KhronosGroup MoltenVK 1.1.9 v
+    epoch               1
+    revision            1
```
jmroot, line 78: This is a new subport, so the revision (as well as epoch as mentioned above) can start at 0.

graphics/MoltenVK/Portfile
jmroot: I don't see why not. It decides whether to install universal files based on the universal variant. That this is done by either thinning the binary or not doesn't make any difference to the final result.

Come to think of it though, the archs provided by the binary don't necessarily match the ones in `configure.universal_archs`, so more effort is needed to make sure the archs of the installed files match those recorded in the registry for the port.

graphics/MoltenVK/Portfile
Gcenx: Agreed this doesn’t require it

graphics/MoltenVK/Portfile
Gcenx: I’ve removed the epoch from here but retained it for the stub as I’d changed the versioning for the stub

graphics/MoltenVK/Portfile
Gcenx: I’d added a “universal” variant manually, removed that now and did the changes you’ve mentioned.

> Come to think of it though, the archs provided by the binary don't necessarily match the ones in `configure.universal_archs`, so more effort is needed to make sure the archs of the installed files match those recorded in the registry for the port.

I’m having it run thin against `configure.build_arch` when the system supports arm64 & x86_64 that steps skipped shouldn’t that be enough?

Gcenx: When it supports both arm64 & x86_64 and +universal is set the thin stage is skipped

jmroot: > I’d added a “universal” variant manually, removed that now and did the changes you’ve mentioned.

Whether the universal variant is added manually or not makes no difference, it is removed if `configure.universal_archs` doesn't end up containing 2 or more archs. That will be the case on systems like 10.14 where universal building isn't possible, or if `universal_archs` is set to an empty value in macports.conf.

jmroot: > I’m having it run thin against `configure.build_arch` when the system supports arm64 & x86_64 that steps skipped shouldn’t that be enough?

OK, so yes, the `supported_archs` will ensure that the universal variant is removed when unsupported universal archs are configured (like on 10.13 where the default is `i386 x86_64`.) But a universal variant never exists now, because you removed it. If you put it back, all should be well.

graphics/MoltenVK/Portfile
```diff
@@ -53,27 +43,72 @@ destroot {
...
-    file copy ${output_dir}/macOS/bin/MoltenVKShaderConverter ${destroot}${prefix}/bin
-    file attributes ${destroot}${prefix}/bin/MoltenVKShaderConverter -permissions +x
-    file copy ${output_dir}/macOS/lib/libMoltenVK.dylib ${destroot}${prefix}/lib
+    file copy ${output_dir}/MoltenVK/dylib/macOS/libMoltenVK.dylib ${destroot}${prefix}/lib
 
     # vulkan and vk_video are provided via vulkan-headers
     file copy ${output_dir}/MoltenVK/include/MoltenVK ${destroot}${prefix}/include
 
-    if {(![variant_isset universal])} {
+    if {[variant_exists universal] && ![variant_isset universal] || ![variant_exists universal]} {
```
Gcenx, line 51: @jmroot assuming you meant doing the check like this?

jmroot: That can be simplified to `![variant_isset universal] || ![variant_exists universal]`.

Gcenx: @jmroot done

jmroot:
I guess since Ryan OK'd the change of maintainership it's OK to merge this.

PR #18808, opened 2023-05-25
Update a few python ports

catap:
#### Description

I'm a maintainer or co-maintainer of all this ports.

It is simple and clean update.

###### Type(s)

- [x] enhancement

###### Tested on

macOS 12.6.3 21G419 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/python/py-pathvalidate/Portfile
+++ b/python/py-pathvalidate/Portfile
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           python 1.0
 
 name                py-pathvalidate
-version             2.5.2
+version             3.0.0
 revision            0
 
 categories-append   devel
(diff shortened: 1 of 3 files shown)


python/py-pathvalidate/Portfile
```diff
@@ -4,7 +4,7 @@ PortSystem          1.0
 PortGroup           python 1.0
 
 name                py-pathvalidate
-version             2.5.2
+version             3.0.0
```
raimue, line 7: I didn't check, but I guess this will no longer be API compatible. Do we have dependents?

raimue (commented):
Update to osc looks good. Thanks for the port update! :+1:

python/py-pathvalidate/Portfile
catap: @raimue only one port `qobuz-dl` and it works :)

catap: BTW, based on https://github.com/thombashi/pathvalidate/releases/tag/v3.0.0 changes aren't that huge and should be safe

PR #18857, opened 2023-05-28
*: prefer configure.build_arch over build_arch

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/R/R-gdata/Portfile
+++ b/R/R-gdata/Portfile
@@ -15,7 +15,7 @@ checksums           rmd160  c6258ef24a745c9571a7873cc71c931e62fa94fd \
 supported_archs     noarch
 
 platform darwin {
-    if {${os.major} < 10 || (${os.major} == 10 && ${build_arch} eq "ppc")} {
+    if {${os.major} < 10 || (${os.major} == 10 && ${configure.build_arch} eq "ppc")} {
         depends_run-append \
                     port:perl5
         configure.perl \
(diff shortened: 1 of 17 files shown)


barracuda156:
@MarcusCalhoun-Lopez This still remains a mystery: what is the difference between the two? )

MarcusCalhoun-Lopez:
Just to give some examples:
    `build_arch` and `configure.build_arch`
    `cxx_stdlib`  and `configure.cxx_stdlib`
    `universal_archs` and `configure.universal_archs`

The basic idea is that the first is what the user or system has requested and the latter is what is given to the Portfiles after some additional processing.
Naturally, they are often the same but it is not guaranteed.

For example, a user could set `build_arch` to ppc64 in macports.conf but `configure.build_arch` could end up as ppc if ppc64 is not in `supported_archs`.

If `cxx_stdlib` is libstdc++, `configure.cxx_stdlib` could be macports-libstdc++ if ${compiler.cxx_standard} >= 2011.

The gory details can be found in the [base code](https://github.com/macports/macports-base/blob/master/src/port1.0/portconfigure.tcl).

mascguy (approved):
LGTM. Anyone else?

tobypeterson:
Makes sense to me; should there be a base change to prevent Portfiles from directly using `build_arch` etc?

jmroot:
> Makes sense to me; should there be a base change to prevent Portfiles from directly using `build_arch` etc?

In some rare cases it does make sense to look at the original macports.conf values.

mascguy:
Marcus, given the number of remaining ports that still use `build_arch` vs. `configure.build_arch`, I'm guessing this is just the first PR of many? (It certainly makes sense to not try to fix everything via one monster PR anyway, given CI's limitations.)

MarcusCalhoun-Lopez:
> (Should I test all of my ports affected or this is supposed to be of no consequence?)

I would say these changes are of *minimal* consequence.
The biggest risk is that I misspelled `configure.` somewhere.

As @jmroot mentioned, there are cases where you do not want `configure.`.
If that is the case for your ports, then this PR needs to be modified.

PR #18882, opened 2023-05-30
Add support for lua-c++ & update gpick

MarcusCalhoun-Lopez:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/graphics/gpick/Portfile
+++ b/graphics/gpick/Portfile
@@ -2,12 +2,13 @@
 
 PortSystem              1.0
 PortGroup               boost 1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               lua 1.0
 
-name                    gpick
-version                 0.2.5
-revision                3
+github.setup            thezbyg gpick 0.3 v
+revision                0
 categories              graphics
-platforms               darwin
 maintainers             nomaintainer
 license                 BSD
 
@@ -18,41 +19,29 @@ long_description        Gpick is a program used to pick colors from anywhere \
                         shades and tints and export palettes to common file \
                         formats or simply copy them to the clipboard.
 
-homepage                https://code.google.com/p/gpick/
-master_sites            googlecode
-distname                ${name}_${version}
+github.tarball_from     releases
 
-checksums               rmd160  a7c65539780b1eb74f44538b165dbff923762beb \
-                        sha256  d57dcbc55dcd1b10a103de8ae6a7028063a4ec8934372c587107dd772da5397f \
-                        size    337112
+checksums               rmd160  19ebed9cb8229e319d8b68013b46c87a6842cffc \
+                        sha256  90b03ce10d2f3e211a3f37d262596e4ae40d05cbd633aa0560d0891562c93474 \
+                        size    368100
 
-boost.version           1.71
-boost.depends_type      build
+patchfiles-append       patch-getExecutablePath.diff \
+                        patch-bool_vector_fix.diff
 
 depends_build-append    port:pkgconfig \
-                        port:scons
+                        port:ragel
 
 depends_lib-append      port:expat \
                         port:gettext \
-                        path:lib/pkgconfig/gtk+-2.0.pc:gtk2 \
-                        port:lua
+                        path:lib/pkgconfig/gtk+-3.0.pc:gtk3
 
-use_configure           no
+# see https://github.com/thezbyg/gpick/blob/dd27232a4dd08cf6271ecc2a7e96da25f8071ed5/CMakeLists.txt#L33
+lua.version             5.4
 
-variant universal {}
+# latest version of Boost that produced no configure warnings
+boost.version           1.80
+boost.depends_type      build
 
-# https://code.google.com/p/gpick/issues/detail?id=112
-configure.ldflags-append -lexpat
+compiler.cxx_standard   2017
 
-build.cmd               ${prefix}/bin/scons
-build.target            build
-build.env               CC=${configure.cc} \
-                        CXX=${configure.cxx} \
-                        "CFLAGS=${configure.cflags} [get_canonical_archflags cc]" \
-                        "CXXFLAGS=${configure.cxxflags} [get_canonical_archflags cxx]" \
-                        "LDFLAGS=${configure.ldflags} [get_canonical_archflags ld]"
-
-destroot.env            {*}${build.env}
-
-# https://code.google.com/p/gpick/issues/detail?id=109
-destroot.destdir        DESTDIR=${destroot}${prefix}
+installs_libs           no
--- a/graphics/gpick/files/patch-bool_vector_fix.diff
+++ b/graphics/gpick/files/patch-bool_vector_fix.diff
@@ -1,0 +1,13 @@
+See https://github.com/thezbyg/gpick/pull/219
+
+--- source/dynv/Xml.cpp.orig	2022-05-08 03:20:36
++++ source/dynv/Xml.cpp	2023-05-30 10:53:32
+@@ -75,7 +75,7 @@
+ 		for (const auto &i: values) {
+ 			if (!writeStart(stream, "li"s))
+ 				return false;
+-			if (!types::xml::write(stream, i))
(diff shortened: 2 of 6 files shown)


graphics/gpick/Portfile
```diff
@@ -2,12 +2,14 @@
...
 PortGroup               boost 1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               lua 1.0
 
-name                    gpick
-version                 0.2.5
-revision                3
+github.setup            thezbyg gpick 0.3 v
+version                 0.3
```
ryandesign, line 10: `github.setup` sets `version` for you.
```suggestion
```

graphics/gpick/Portfile
```diff
@@ -18,41 +20,27 @@ long_description        Gpick is a program used to pick colors from anywhere \
...
+github.tarball_from     releases
 
-checksums               rmd160  a7c65539780b1eb74f44538b165dbff923762beb \
-                        sha256  d57dcbc55dcd1b10a103de8ae6a7028063a4ec8934372c587107dd772da5397f \
-                        size    337112
+checksums               rmd160  19ebed9cb8229e319d8b68013b46c87a6842cffc \
+                        sha256  90b03ce10d2f3e211a3f37d262596e4ae40d05cbd633aa0560d0891562c93474 \
+                        size    368100
 
-boost.version           1.71
```
ryandesign, line 29: Shouldn't every port that uses the boost portgroup specify which `boost.version` to use, since the point is that boost versions are often not compatible with one another?

I was going with 1.80 in the update I was working on, since trying to use 1.81 produced a warning message.

graphics/gpick/Portfile
```diff
@@ -2,12 +2,14 @@
 
 PortSystem              1.0
 PortGroup               boost 1.0
+PortGroup               cmake 1.1
+PortGroup               github 1.0
+PortGroup               lua 1.0
```
ryandesign, line 7: I'm not familiar with the lua portgroup. Which lua version dependency will be added? Will it correctly tell the build system to use that lua version? gpick's CMakeLists.txt looks for lua 5.4 first, then 5.3, then 5.2; therefore, in my update, I was having the port use lua 5.4.

ryandesign (commented):
Oh thanks, I had been working on a gpick update myself but had not gotten this far due to the lua-c++ requirement.

Updating gpick resolves https://trac.macports.org/ticket/65641.

CI is failing.

graphics/gpick/Portfile
MarcusCalhoun-Lopez: Thank you for noticing.
That line was a vestige from earlier testing.

graphics/gpick/Portfile
MarcusCalhoun-Lopez: subport lua installs to `${prefix}/{include,lib}`
subport lua5x installs to `${prefix}/libexec/lua5x/{include,lib}`
Unless `lua.version ...` is specified, plain lua, whose version changes like any other port, is used.

The CMakefile file eventually looks for [lua-c++.pc](https://github.com/thezbyg/gpick/blob/dd27232a4dd08cf6271ecc2a7e96da25f8071ed5/CMakeLists.txt#L33) (no version).
The command `lua.finds_versioned  yes`  shows that the versioned pkgconfig files will be found first, so add a build conflict.

I have no strong feelings about this.
I have just tried to use plain lua unless there was a reason to use the versioned version.

We can easily replace `lua.finds_versioned yes` with `lua.version 5.4` to force version 5.4.

graphics/gpick/Portfile
MarcusCalhoun-Lopez: I am not well versed in Boost version incompatibilities.
I just assumed the default boost version was somehow significant, especially since the default Boost version is different for different OS versions.
I am more that happy to set a different version of Boost if that is more appropriate.

graphics/gpick/Portfile
ryandesign: I would consider the addition of build conflicts to be undesirable, and the use of the latest compatible version of lua to be desirable, so I'd say let's use 5.4.

graphics/gpick/Portfile
ryandesign: > I am not well versed in Boost version incompatibilities.

Back when we only had the one boost port, practically every time we updated it to a new major version, it would break some port. The developers of boost do not maintain backward compatibility in new versions of boost; they [claim](https://www.boost.org/users/faq.html) this "isn't always possible" and recommend "freezing the version of the Boost libraries used by your project". Projects should probably bundle whatever version of boost they are compatible with. For those that don't, we now have many boost ports so each port can depend on whichever version is best for that project.

> I just assumed the default boost version was somehow significant, especially since the default Boost version is different for different OS versions.

Well that's true, I'm not sure what to do about the different default boost versions on different OS versions though it sounds like we want to get rid of that difference; see the [comments on this commit](https://github.com/macports/macports-ports/commit/e5316596d177c9f6370d5817f94b94879a939405). I think which boost version the boost portgroup uses as its default is not necessarily significant anymore. It's not something we can ever change anymore—not without also either revbumping and testing all ports that don't specify the boost version, or changing them all to specify the formerly-default boost version. In my opinion it's an argument for having the boost portgroup not specify any default boost version and requiring each port to specify it. See [ticket 66927](https://trac.macports.org/ticket/66927).

ryandesign: > I was going with 1.80 in the update I was working on, since trying to use 1.81 produced a warning message.

My motivation in using a newer boost was that I remembered there were tickets like [66927](https://trac.macports.org/ticket/66927) where it had been requested that we update ports to newer boost versions since old boost versions did not build on some systems. However I hadn't remembered that 1.81 was the version in which that problem was fixed.

When trying to use 1.81, cmake printed this warning:

```
CMake Warning at /opt/local/share/cmake-3.24/Modules/FindBoost.cmake:1384 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
```

When I looked this up, I understood that cmake prints this warning when the boost version is newer than was available at the time that this version of cmake was released. I guess boost 1.80 was the latest version available when cmake 3.24.4 was released, and I guess there's the possibility that new boost versions need changes to the cmake files to work properly. But if we want to try using 1.81 and it works fine, we can add `-DBoost_NO_WARN_NEW_VERSIONS=ON` to configure.args to suppress the warning or maybe we should [update cmake](https://trac.macports.org/ticket/67540) first.

Looks like a [small change for boost 1.81 support](https://github.com/Kitware/CMake/commit/a97032c4e7b1efb4c5535e5f705ef8ffd5fc2dd3) was made in cmake 3.25.2 and there was another [small change for boost 1.82 support](https://github.com/Kitware/CMake/commit/5cbbe55de85bdd0d6d90241a9b18683f13375fc1) committed to cmake a couple hours ago and not yet part of any release.

graphics/gpick/Portfile
MarcusCalhoun-Lopez: I update the PR with `lua.version  5.4`.

graphics/gpick/Portfile
MarcusCalhoun-Lopez: I update the PR to include `boost.version 1.80` since, as you say, it is the latest version to produce no warnings.
Since Boost is only a build dependency, it should be relatively easy to increase the version if it is decided that a more recent Boost version works as well.

ryandesign:
Thanks for working on this!

Buildbot builds are failing with the same errors we saw in CI on this PR: undefined lua symbols, starting with:

```
    "luaL_error(lua_State*, char const*, ...)", referenced from:
        error(lua_State*) in Script.cpp.o
```

MarcusCalhoun-Lopez:
It would seem I neglected to increase the revisions of the Lua subports.
That part should be [corrected](https://github.com/macports/macports-ports/commit/779e2fa3954b456df09582efaac35b9accf20c37).
I also tried to force the buildbots to build gpick.
Hopefully, that fixes the problem.
Sorry for the mistake.

PR #18890, opened 2023-05-31
diffutils:  fix bug with 32-bit builds

RobK88:
#### Description

* change patchfiles to patchfiles-append
* Remove pre-configure {} code block
* Remove "--disable-silent-rules" from configure,args-append
* Add "--disable-year2038" to configure,args-append for ALL builds

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.13.6 17G14042 x86_64
Xcode 10.1 10B61

###### Verification 
Have you


jmroot:
> * Remove "--disable-silent-rules" from configure.args-append

Why? Does this no longer work, or are you deliberately enabling silent make rules for some reason?

RobK88:
@jmroot - Joshua the original Portfile that I modified a week ago, never had the flag "--disable-silent-rules".  I added the flag for testing.  I can add the flag back in if you like.  Should "--disable-silent-rules" be added to most ports?  I often do not see the flag being used.

RobK88:
Here is a question.  With 64-bit only builds, I am not sure whether specifying `--disable-year2038` will build the 64-bit binary using 32-bit time_t which has the year 2038 limitation.  If that is the case, maybe, it would be better to use `pre-configure {}` code block where the `--disable-year2038` flag is only used when the build is NOT 64-bit only.  Something like:

```
pre-configure {

     if {build != 64bit) { configure.args-append --disable-year2038 }
}
```

The test `{build != 64bit)` is psuedo code.  I do not know how to make the real test.  Is there a global variable available that specifying whether the build is 32-bit or 64-bit?

jmroot:
We always prefer to not use silent make rules, because they make logs much less useful. So if this flag is needed to disable them, it should be kept.

`time_t` is always 64-bit when used from 64-bit code on macOS, so don't worry about that.

RobK88:
Thanks Joshua.  I added the "--disable-silent-rules" flag to configure.args-append.
It looks like this PR is ready to be merged.

PR #18898, opened 2023-05-31
qgis3: update to 3.28.7

nilason:
#### Description

Update QGIS 3 to version 3.28.7.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.6.5 21G531 arm64
Xcode 14.2 14C18

###### Verification 
Have you

cc/ @i0ntempest

### DIFF
--- a/gis/gdal-grass/Portfile
+++ b/gis/gdal-grass/Portfile
@@ -19,8 +19,6 @@ long_description    This plugin allows GDAL to read GRASS raster and vector file
 
 license             MIT
 
-platforms           darwin
-
 depends_lib         port:gdal \
                     port:grass
 
--- a/gis/grass/Portfile
+++ b/gis/grass/Portfile
@@ -10,7 +10,7 @@ github.setup        OSGeo grass 8.2.1
 name                grass
 epoch               1
 set main_version    [join [lrange [split ${version} "."] 0 1] ""]
-revision            3
+revision            4
 set realVersion     ${version}
 distname            grass-${realVersion}
 maintainers         nomaintainer
@@ -43,7 +43,7 @@ depends_lib         port:bzip2 \
                     port:zlib \
                     port:readline
 
-patchfiles          patch-dyld-link.diff
+patchfiles          01_patch-dyld-link.diff 02_grass-python.diff
 
 configure.cflags-append  -DGL_SILENCE_DEPRECATION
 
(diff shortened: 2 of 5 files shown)


nilason:
As `qgis3` is has a dependency to grass (via variant), I took the liberty to add to this PR a fix for https://trac.macports.org/ticket/67537.

gis/gdal-grass/Portfile
```diff
@@ -10,7 +10,7 @@ categories          gis
 # do not (re-)publish 1.4.3_0, 2.1.0_0, 2.1.0_1, 2.2.0_0, 2.2.0_1, or 2.2.0_2
 github.setup        OSGeo gdal-grass 1.0.2
 github.tarball_from archive
-revision            1
+revision            2
```
ryandesign, line 13: Is this revision increase really needed?

ryandesign: After fixing grass for 10.15 and earlier, we do want to reattempt the build of gdal-grass on the buildbot, but we can either schedule that build manually, or you can make *any* change to the gdal-grass Portfile to cause it to happen. For example, consider removing the `platforms darwin` line since that is now the default.

nilason: Thanks for the explanation, I'm slowly starting to understand the MP infrastructure.
I removed the rev bump and instead removed `platforms darwin` for gdal-grass.

nilason:
> As `qgis3` is has a dependency to grass (via variant), I took the liberty to add to this PR a fix for https://trac.macports.org/ticket/67537.

I now reimplemented the grass fix with patch files (instead of using reinplace) as suggested by https://trac.macports.org/ticket/67537#comment:4. I believe this may be merged now.

nilason:
@mascguy , @i0ntempest or @reneeotten 
I believe this may be merged now.

i0ntempest (approved):

i0ntempest:
Merging in 12h

gis/grass/Portfile
```diff
@@ -43,7 +43,7 @@ depends_lib         port:bzip2 \
                     port:zlib \
                     port:readline
 
-patchfiles          patch-dyld-link.diff
+patchfiles          01_patch-dyld-link.diff 02_grass-python.diff
```
mascguy, line 46: My only suggestion for future PRs - and this is a bit of a personal choice, so don't take this as gospel (or required) - is to only apply one patch per line. Likewise, always use `patchfiles-append`.

The benefit is that you can quickly comment/uncomment individual patches during testing, without having to change multiple lines, deal with line continuation, etc.

So this is a better way to do it:

```
patchfiles-append   01_patch-dyld-link.diff
patchfiles-append   02_grass-python.diff
```

No need to change this though, as it's not worth going through another CI cycle.

mascguy (approved):

gis/grass/Portfile
nilason: Noted, thanks!

nilason:
Thank you all!

PR #18920, opened 2023-06-01
openal-soft: fix a patch

barracuda156:
#### Description

A small but necessary fix to the patch. What was overlooked is that CMakeLists check uses `cinttypes`, while the code uses `inttypes.h`, and this does not work as intended with GCC, which does not need `__STDC_FORMAT_MACROS` with the first, but does need with the second header.
Reported to upstream also: https://github.com/kcat/openal-soft/commit/118c729680d6664f793f8d88ff0b7548137847d3

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/audio/openal-soft/Portfile
+++ b/audio/openal-soft/Portfile
@@ -10,7 +10,7 @@ legacysupport.newest_darwin_requires_legacy 14
 
 name                    openal-soft
 version                 1.23.1
-revision                0
+revision                1
 checksums               rmd160  5ca2531cc9e7476d122ddc9ab3a1b3e568d62175 \
                         sha256  796f4b89134c4e57270b7f0d755f0fa3435b90da437b745160a49bd41c845b21 \
                         size    699330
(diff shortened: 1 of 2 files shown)


barracuda156:
@ryandesign Where are we on this? Fixing the patch has the only effect of fixing build with GCC. If that is still controversial, maybe add a flag via portfile? Having this broken leaves all dependents broken, among which my port `toxic`.

ryandesign:
I would prefer to use [the upstream fix for this issue](https://github.com/kcat/openal-soft/commit/6752d5516c8b30fe7db559c7f3a1423705d2a4fd). That way it will become clear, when updating the port to the next version and the patch fails to apply, that the patch can be removed at that time.

barracuda156:
@ryandesign Ok, sure, I will rebase today.

barracuda156:
@ryandesign Done, should be good now. Verified it builds on my PPC system.

barracuda156:
@ryandesign @reneeotten Could we merge this finally?

PR #18934, opened 2023-06-02
vim: Fix +python3 variant, follow-up to #18475

neverpanic:
#### Description

It seems upstream now has a patch that separately tracks `PYTHONFRAMEWORKPREFIX` and adds it as necessary. I tested this by checking the output of `otool -L` on the vim binary and running

```
:py3 print(sys.version_info)
```

in the resulting binary.

The python 2 variants may also have been broken by the recent changes, but I don't use those and python 2 is deprecated anyway. We should probably just remove the python 2 variants from vim.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 13.4 22F66 x86_64
Xcode 14.3 14E222b

###### Verification 
Have you

### DIFF
--- a/editors/vim/Portfile
+++ b/editors/vim/Portfile
@@ -146,7 +146,12 @@ foreach s ${pythons_suffixes} {
             patchfiles-append       patch-python.diff
         } else {
             configure.args-append   --enable-python3interp --with-python3-command=${prefix}/bin/python${v}
-            patchfiles-append       patch-python3.diff
+            if {[vercmp \"${vim_version}.${vim_patchlevel}\" 9.0.1441] < 0} {
+                ui_info \"Vim patchlevel < 1441, python3 patch needed\"
+                patchfiles-append   patch-python3.diff
+            } else {
+                ui_info \"Vim patchlevel >= 1441, no python3 patch needed\"
+            }
         }
         depends_lib-append      port:${p}
         use_autoconf yes


mascguy (approved):
LGTM. Rainer?

ryandesign:
This may fix https://trac.macports.org/ticket/67567

raimue:
After the last update, the vim port is out of sync with MacVim, which is still at patchlevel 9.0.1276. Both should always be in sync as they are using the same configuration, otherwise they may show different behavior.

I understand the vim port is now broken and needs to be fixed. When we remove the +python3 patch with this PR now, we diverge the two ports even more. It is hard to keep track of such changes and these will have to be repeated on the MacVim port with the next update. Blindly increasing the patchlevel is not the way to maintain these ports as the patches in the variants always need manual testing.

Regarding this PR, by removing the +python3 patch completely, we will break the older version which is currently pinned for macOS 10.10 at the top (due to #67209). So we need another conditional here when to apply the patch. I would recommend we base it on the patch level.

I tracked the change to vim patch level 1441 (https://github.com/vim/vim/commit/b6ebe5af74a965d703eb8ee66a07c97b0bbe6e58).

My untested proposal would be the following snippet. I am avoiding `${version}` as then we can use the exact same snippet in the MacVim port to keep them in sync:
```
if {[vercmp "${vim_version}.${vim_patchlevel}" 9.0.1441] < 0} {
    patchfiles-append       patch-python3.diff
}
```

mascguy:
> Regarding this PR, by removing the +python3 patch completely, we will break the older version which is currently pinned for macOS 10.10 at the top (due to #67209). So we need another conditional here when to apply the patch. I would recommend we base it on the patch level.

Your logic works perfectly; update pushed.

> After the last update, the vim port is out of sync with MacVim, which is still at patchlevel 9.0.1276. Both should always be in sync as they are using the same configuration, otherwise they may show different behavior.

This could be a bit problematic, as MacVim upstream hasn't provided a new release since then. (Apart from `prerelease-1.76.1`, which only covers patchlevel 1403.) Even master is only at 1474, with the latest commit dated 2023-04-22.

https://github.com/macvim-dev/macvim/commits/master

How should we handle this situation...?

jrabinow:
I just ran into this bug on my end. I'm not sure why this broke now, it seems like the vim port was updated more than a month ago.
My suggestion is we could add the conditional patch to macvim port.

We still have the issue of keeping vim and macvim versions synced up - for the time being, it seems like macvim updated to 9.0.1609 2 days ago so we could update both vim and macvim patchlevels, then throw in the conditional source code patch. That'll unblock us this time here, but in terms of generally keeping things in sync, I'm wondering if there isn't some kind of linting rule that could be added that would ensure both vim and macvim are in sync

markmentovai:
This fixes https://trac.macports.org/ticket/67567 for me. Is there a reason it hasn’t been merged yet?

mascguy (approved):

mascguy:
> This fixes https://trac.macports.org/ticket/67567 for me. Is there a reason it hasn’t been merged yet?

Sorry Mark, we wanted to give Rainer a chance to provide a final thumbs-up. But didn't realize this PR is already two weeks old. Yikes!

Merging now.

markmentovai:
Thanks!

PR #19011, opened 2023-06-07
edid-decode: new Portfile

wwalexander:
This commit adds a port for the edid-decode utility, which decodes binary EDID monitor description data. There doesn't appear to be any tags or branches to pin versions to, so I'm just using commit SHAs.

###### Type(s)
- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 13.4 22F66 arm64
Xcode 14.3.1 14E300c

###### Verification 
Have you

### DIFF
--- a/sysutils/edid-decode/Portfile
+++ b/sysutils/edid-decode/Portfile
@@ -1,0 +1,19 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+PortSystem              1.0
+PortGroup               makefile 1.0
+
+name                    edid-decode
+version                 20230616
+categories              sysutils
+maintainers             @wwalexander openmaintainer
+license                 MIT
+description             Decode EDID data in human-readable format
+long_description        edid-decode decodes EDID monitor description data in human-readable format
+set domain              linuxtv.org
+homepage                https://git.${domain}/${name}.git
+fetch.type              git
+git.url                 git://${domain}/${name}.git
+git.branch              a31e680
+
+destroot.args-append    bindir=${prefix}/bin \
+                        mandir=${prefix}/share/man


sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
+PortSystem       1.0
+
+name             edid-decode
+version          git-e48fb38
+categories       sysutils
+platforms        darwin
+maintainers      @wwalexander
```
herbygillot, line 7: This should also be marked `openmaintainer` unless there is a reason why specifically only you should be allowed to modify this port.  Additionally, your Github tag and optional email address (if included) should be surrounded by curly braces.

```suggestion
maintainers      {@wwalexander} openmaintainer
```

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
+PortSystem       1.0
+
+name             edid-decode
+version          git-e48fb38
+categories       sysutils
+platforms        darwin
```
herbygillot, line 6: `platforms darwin` is implicit in MacPorts now, and can be completely dropped.

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
+PortSystem       1.0
+
+name             edid-decode
+version          git-e48fb38
```
herbygillot, line 4: Since there are no actual tagged versions you can use, it's recommended to set version to the datestamp of the commit.

```suggestion
version          20230601
```

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
+PortSystem       1.0
```
herbygillot, line 1: This Portfile is a good candidate for the `makefile` portgroup.

Portgroups provide default functionality to a Portfile for the way a software project is structured, such as the `golang` or `cargo` (Rust) portgroups.

The `makefile` portgroup will add the right C compiler flags for the given macOS version and platform.

You can add this by deleting the `use_configure no` line below (since the `makefile` PG does this by default), and adding the following:

```suggestion
PortSystem       1.0
PortGroup        makefile 1.0
```

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
...
+
+name             edid-decode
+version          git-e48fb38
+categories       sysutils
+platforms        darwin
+maintainers      @wwalexander
+license          MIT
+description      Decode EDID data in human-readable format
+long_description edid-decode decodes EDID monitor description data in human-readable format
+homepage         https://git.linuxtv.org/edid-decode.git
```
herbygillot, line 11: You can probably do something like:

```
set domain  linuxtv.org
homepage    https://git.${domain}/edid-decode.git
git.url     git://${domain}/edid-decode.git
```

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
...
+master_sites     git://linuxtv.org/edid-decode.git
```
herbygillot, line 12: Since you are using `fetch.type git`, `master_sites` does not need to be set and can be omitted.

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,17 @@
...
+patchfiles       patch-destdir-variable-fix.diff
+fetch.type       git
+git.url          git://linuxtv.org/edid-decode.git
+git.branch       e48fb38
+use_configure    no
```
herbygillot, line 17: As mentioned above, this line can be dropped after using the `makefile` portgroup.

sysutils/edid-decode/files/patch-destdir-variable-fix.diff
```diff
@@ -0,0 +1,13 @@
+--- Makefile.orig	2022-05-17 21:54:35.000000000 -0400
```
herbygillot, line 1: This patch is not working for MacPorts installations in locations other than `/opt/local`.

Perhaps something like this:
```
--- ./Makefile	2023-06-07 21:20:51.000000000 -0400
+++ ./Makefile	2023-06-07 21:51:31.000000000 -0400
@@ -1,17 +1,16 @@
+
 ifeq ($(OS),Windows_NT)
-	bindir ?= /usr/bin
-	mandir ?= /usr/share/man
+        PREFIX ?= /usr
 else
 	UNAME_S := $(shell uname -s)
 	ifeq ($(UNAME_S),Darwin)
-		bindir ?= /usr/local/sbin
-		mandir ?= /usr/local/share/man
-	else
-		bindir ?= /usr/bin
-		mandir ?= /usr/share/man
+		PREFIX ?= /opt/local
 	endif
 endif

+bindir ?= $(PREFIX)/bin
+mandir ?= $(PREFIX)/share/man
+
 EMXX ?= em++

 SOURCES = $(wildcard *.cpp)
```

herbygillot:
Thanks for this submission @wwalexander. I've made a number of comments.  Please make all changes by amending the same commit you have. (MacPorts wants to see all new submissions or updates as a single commit.)

herbygillot:
Also, Portfile spacing should be in groups of 4 spaces.

sysutils/edid-decode/Portfile
ryandesign: Only multi-value entries (such as those having both a GitHub handle and an obfuscated email address) should be surrounded by curly braces.

sysutils/edid-decode/files/patch-destdir-variable-fix.diff
ryandesign: The reason why Herby's proposal would work in other prefixes, even though it still says /opt/local, is that the makefile portgroup will set the PREFIX environment variable to the correct prefix.

However, ideally, patches such as these should be written with an eye toward submitting them to the developers for inclusion upstream, so that we don't need to maintain the patches forever, and this revised patch as written would not be suitable for upstreaming because it deletes the non-Windows-NT non-Darwin code path so PREFIX would be unset on those systems, and also because it defaults the Darwin installation prefix to /opt/local; we would not want non-MacPorts Darwin users to install this software there. /usr/local seems to me like a reasonable default prefix for all Unix-like systems, but they may prefer to continue to use /usr for non-Darwin Unix-like systems for whatever reason.

ryandesign: The Makefile also mistakenly uses `CFLAGS` instead of `CXXFLAGS`.

sysutils/edid-decode/Portfile
wwalexander: Should I also use ${name} in these URLs, e.g.

```
set domain  linuxtv.org
homepage    https://git.${domain}/${name}.git
git.url     git://${domain}/${name}.git
```

or would that be a bad practice?

sysutils/edid-decode/Portfile
wwalexander: I fixed this, omitting the curly braces (d73a890).

sysutils/edid-decode/Portfile
wwalexander: Fixed (d73a890).

sysutils/edid-decode/Portfile
wwalexander: Fixed (d73a890).

sysutils/edid-decode/Portfile
wwalexander: Fixed (d73a890).

sysutils/edid-decode/Portfile
wwalexander: I added this (d73a890), using `${name}`; let me know if I should revise.

sysutils/edid-decode/Portfile
wwalexander: Fixed (d73a890).

sysutils/edid-decode/Portfile
wwalexander: Fixed (d73a890).

sysutils/edid-decode/files/patch-destdir-variable-fix.diff
wwalexander: I attempted to make a more general-purpose patch (d73a890), which as far as I can tell preserves the original behavior across platforms but allows MacPorts to override `PREFIX`:

```Makefile
ifeq ($(OS),Windows_NT)
        PREFIX ?= /usr
else
        UNAME_S := $(shell uname -s)
        ifeq ($(UNAME_S),Darwin)
                PREFIX ?= /usr/local
        else
                PREFIX ?= /usr
        endif
endif

bindir ?= $(PREFIX)/bin
mandir ?= $(PREFIX)/share/man
```

The patch also swaps `CFLAGS` for `CXXFLAGS`.

wwalexander: I just realized the same result can be accomplished without a patch by simply setting the variables in `destroot.args`:

```
destroot.args-append bindir=${prefix}/bin \
                     mandir=${prefix}/share/man
```

wwalexander:
@herbygillot there's a new commit as of 2023-06-16. Should I amend with the new commit or wait til this PR is merged in to update?

herbygillot:
> @herbygillot there's a new commit as of 2023-06-16. Should I amend with the new commit or wait til this PR is merged in to update?

You can amend this commit to the new version. Afterwards I'll review.

wwalexander:
Added the new version.

herbygillot:
> Added the new version.

Can you double check? Version looks unchanged.

wwalexander:
> Can you double check? Version looks unchanged.

Oops 😅 should be there now

herbygillot:
Ok thanks. We'll wait for CI.

sysutils/edid-decode/Portfile
```diff
@@ -0,0 +1,18 @@
+PortSystem              1.0
```
herbygillot, line 1: I just realized you're missing the modeline: https://guide.macports.org/#development.creating-portfile

The first line in every Portfile should be what's called the modeline, used by editors to set consistent formatting for all Portfiles.

```suggestion
# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
PortSystem              1.0
```

wwalexander: Added. Thank you so much for all the guidance!

herbygillot:
Looks good and merged. Thank you for getting this submission into MacPorts 👍🏾

PR #19034, opened 2023-06-08
bart: Tweaks to Portfile (no bump).

eborisch:
Maintainer update.

### DIFF
--- a/science/bart/Portfile
+++ b/science/bart/Portfile
@@ -17,7 +17,13 @@ checksums \
     sha256  25c0ba9ff2b8ba07ab3aa54566a7be9755a2ea59a6957fcea06b19a899da00da \
     size    879025
 
-default_variants-append +${compilers.gcc_default}
+if { ![c_variant_isset] } {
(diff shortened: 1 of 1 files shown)


gctwnl:
I'm putting the question here as well as I really need to know if this is going to force me to do a lot of work now.

The previous portfile commented: "Not planning to update past this version for minio port" 

It seems however, minio has been updated to  the 2210 version. 

Curious minds want to know: is this thus a backwards incompatible update and installing it will turn an existing minio installation inoperable? And if so, should the Notes not reflect that?

eborisch:
> I'm putting the question here as well as I really need to know if this is going to force me to do a lot of work now.
> 
> The previous portfile commented: "Not planning to update past this version for minio port"
> 
> It seems however, minio has been updated to the 2210 version.
> 
> Curious minds want to know: is this thus a backwards incompatible update and installing it will turn an existing minio installation inoperable? And if so, should the Notes not reflect that?

This is on the wrong commit, but for the record, I’m holding MINIO at the version that should not require their significant “upgrade” process.

PR #19047, opened 2023-06-10
rawhide: fix legacysupport linking

aeiouaeiouaeiouaeiouaeiouaeiou:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

###### Verification 
Have you


barracuda156:
Could you please also add `configure.cflags-append -std=c99` and then close my ticket: https://trac.macports.org/ticket/67604

macportsraf:
`configure.cflags-append -std=c99` won't work. The configure script isn't a GNU configure script. It won't recognize that option. Perhaps this is was why the makefile port group was being used. I had copied it from the daemon Portfile without understanding it. But before I removed it, rawhide was linking (but not working) with legacysupport. I think that restoring the makefile port group is probably a better fix as it won't require patching the Makefile.

barracuda156:
> `configure.cflags-append -std=c99` won't work. The configure script isn't a GNU configure script. It won't recognize that option. Perhaps this was why the makefile port group was being used. I had copied it from the daemon Portfile without understanding it (which also uses a bespoke non-GNU configure script). But before I removed it, rawhide was linking (but not working) with legacysupport. I think that restoring the makefile port group is probably a better fix as it won't require patching the Makefile.

@macportsraf It works because CFLAGS are then passed to build.env:
```
configure.cflags-append \
                    -std=c99
build.target        rh
build.env           CC=${configure.cc} \
                    "CFLAGS=${configure.cflags} [get_canonical_archflags cc]" \
                    "LDFLAGS=${configure.ldflags} [get_canonical_archflags ld]"
```

barracuda156:
I have obviously checked that before advising here. Re-built again now:
```
10:flock-0.4.0 svacchanda$ sudo port -v -n build rawhide
--->  Computing dependencies for rawhide.
--->  Fetching distfiles for rawhide
--->  Verifying checksums for rawhide
--->  Checksumming rawhide-3.2.tar.gz
--->  Extracting rawhide
--->  Extracting rawhide-3.2.tar.gz
Executing:  cd "/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work" && /usr/bin/gzip -dc '/opt/local/var/macports/distfiles/rawhide/rawhide-3.2.tar.gz' | /usr/bin/gnutar --no-same-owner -xf - 
--->  Configuring rawhide
Executing:  cd "/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2" && ./configure --prefix=/opt/local --macports 
Configuring for macos
Configuring CC as /usr/bin/gcc-4.2
Configuring --enable-acl (default)
Configuring --enable-ea (default)
Configuring --enable-attr (default)
Configuring --enable-pcre2 (default when present)
Configuring --enable-magic (default)
Configuring --prefix=/opt/local
Configuring --macports
--->  Building rawhide
Executing:  cd "/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2" && /usr/bin/make -j8 -w rh 
make: Entering directory `/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2'
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhcmds.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rh.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhparse.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhdir.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhdata.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhstr.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rherr.c
/usr/bin/gcc-4.2  -Os -I/opt/local/include/LegacySupport -std=c99 -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -o rh rhcmds.o rh.o rhparse.o rhdir.o rhdata.o rhstr.o rherr.o -L/opt/local/lib -Wl,-headerpad_max_install_names -lMacportsLegacySupport -arch ppc -L/opt/local/lib -lpcre2-8     -lmagic 
make: Leaving directory `/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2'
```

macportsraf:
OK. Sorry, but I thought you might not have realised that the configure script isn't a GNU autoconf configure script. Anyway, the main thing is that this can be fixed by just restoring the line that said `PortGroup makefile 1.0` after the `PortGroup github 1.0` line. The original version of the Portfile had that, and it linked on 10.6.8. I was told that use_configure and the makefile port group aren't usually used together, so I removed it. But it seems that it is needed because the configure script is not a GNU autoconf configure script, and so whatever the makefile port group does is still required. So putting that line back in should be all that is needed to make it link again. No doubt, what you have done is the equivalent of what it does, just not as concise. I just checked again on 10.6.8, and adding the makefile port group line back in does fix the linking.

Having said all that, the program itself still doesn't actually work for me on 10.6.8, even though it compiles and links successfully. The functions that are provided by legacysupport aren't working correctly for me, but maybe they'll work on other people's old systems. The mailing list didn't come up with any theories as to what the problem is.

[Portfile.patch](https://github.com/macports/macports-ports/files/11714197/Portfile.patch)

barracuda156:
@macportsraf `legacysupport` PG should come above `makefile` (not for alphabetic reasons, it explicitly asks for it), but otherwise you are right, adding `makefile` PG works fine without manual flags passing.

One thing I am unsure about is c90 warnings:
```
--->  Configuring rawhide
Executing:  cd "/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2" && ./configure --prefix=/opt/local --macports 
Configuring for macos
Configuring CC as /usr/bin/gcc-4.2
Configuring --enable-acl (default)
Configuring --enable-ea (default)
Configuring --enable-attr (default)
Configuring --enable-pcre2 (default when present)
Configuring --enable-magic (default)
Configuring --prefix=/opt/local
Configuring --macports
--->  Building rawhide
Executing:  cd "/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2" && /usr/bin/make -j8 -w rh CC="/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2" CXX="/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cxx/usr/bin/g++-4.2" OBJC="/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/objc/usr/bin/gcc-4.2" OBJCXX="/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/objcxx/usr/bin/g++-4.2" INSTALL="/usr/bin/install -c" 
make: Entering directory `/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2'
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhcmds.c
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rh.c
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhparse.c
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhdir.c
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhdata.c
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rhstr.c
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -c rherr.c
In file included from rhstr.c:62:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
In file included from rherr.c:34:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
In file included from rhdata.c:36:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
In file included from rh.c:40:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
In file included from rhparse.c:41:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
In file included from rhdir.c:63:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
rh.c: In function ‘help_message’:
rh.c:152: warning: ISO C90 forbids mixed declarations and code
rhparse.c:1234:3: warning: use of C99 long long integer constant
rhparse.c:1234:14: warning: use of C99 long long integer constant
rhparse.c:1234:25: warning: use of C99 long long integer constant
rhparse.c:1234:36: warning: use of C99 long long integer constant
rhparse.c:1234:47: warning: use of C99 long long integer constant
rhparse.c:1234:58: warning: use of C99 long long integer constant
rhparse.c:1235:3: warning: use of C99 long long integer constant
rhparse.c:1235:11: warning: use of C99 long long integer constant
rhparse.c:1235:22: warning: use of C99 long long integer constant
rhparse.c:1235:36: warning: use of C99 long long integer constant
rhparse.c:1235:53: warning: use of C99 long long integer constant
rhparse.c:1235:73: warning: use of C99 long long integer constant
rhparse.c: In function ‘show_token’:
rhparse.c:1727: warning: ISO C90 does not support the ‘ll’ printf length modifier
In file included from rhcmds.c:78:
rh.h:87: warning: ISO C90 does not support ‘long long’
rh.h:88: warning: ISO C90 does not support ‘long long’
rhdir.c: In function ‘visitf_long’:
rhdir.c:1390: warning: ISO C90 does not support the ‘%z’ strftime format
rhdir.c:1390: warning: ISO C90 does not support the ‘%e’ strftime format
rhdir.c:1399: warning: ISO C90 does not support the ‘%z’ strftime format
rhdir.c:1399: warning: ISO C90 does not support the ‘%e’ strftime format
rhdir.c:1408: warning: ISO C90 does not support the ‘%z’ strftime format
rhdir.c:1408: warning: ISO C90 does not support the ‘%e’ strftime format
rhdir.c: In function ‘json’:
rhdir.c:2346: warning: ISO C90 does not support the ‘%z’ strftime format
rhdir.c:2347: warning: ISO C90 does not support the ‘%z’ strftime format
rhdir.c:2348: warning: ISO C90 does not support the ‘%z’ strftime format
rhdir.c: In function ‘visitf_format’:
rhdir.c:2491: warning: ISO C90 does not support the ‘%e’ strftime format
rhdir.c:2549: warning: ISO C90 does not support the ‘%e’ strftime format
rhdir.c:2829: warning: ISO C90 does not support the ‘%e’ strftime format
/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/compwrap/cc/usr/bin/gcc-4.2 -Os -I/opt/local/include/LegacySupport -arch ppc -O3 -g -Wall -pedantic -DETCDIR=\"/opt/local/etc\" -DRAWHIDE_NAME=\"rawhide\" -DRAWHIDE_PROG_NAME=\"rh\" -DRAWHIDE_VERSION=\"3.2\" -DRAWHIDE_DATE=\"20230609\"  -DHAVE_PCRE2=1 -DHAVE_ACL=1 -DHAVE_POSIX_ACL=1 -DHAVE_MACOS_ACL=1 -DHAVE_EA=1 -DHAVE_MACOS_EA=1  -DHAVE_FLAGS=1 -DHAVE_MAGIC=1    -I/opt/local/include        -o rh rhcmds.o rh.o rhparse.o rhdir.o rhdata.o rhstr.o rherr.o -L/opt/local/lib -Wl,-headerpad_max_install_names -lMacportsLegacySupport -arch ppc -L/opt/local/lib -lpcre2-8     -lmagic 
make: Leaving directory `/opt/local/var/macports/build/_opt_PPCRosettaPorts_sysutils_rawhide/rawhide/work/rawhide-3.2'
```

Adding `-std=c99` flag fixes those, though I am not sure if that is of some practical consequence.

macportsraf:
That doesn't matter. It's just warning in case I need to comply with C90 (because I told it to be pedantic). The compiler issuing the warnings does support the things that the warnings are about. They do work.

aeiouaeiouaeiouaeiouaeiouaeiou:
@barracuda156 can you run `rh -l` in a large directory (e.g. `macports-ports` repository)? I am wondering if the "bad file descriptor" error occurs on the PowerPC architecture.

barracuda156:
> @barracuda156 can you run `rh -l` in a large directory (e.g. `macports-ports` repository)? I am wondering if the "bad file descriptor" error occurs on the PowerPC architecture.

Yes it does:
```
drwxr-xr-x    23 svacchanda staff     21 Jul 18 03:22:51 2022 ./xfce
rh: fstatat ./xfce/exo: Bad file descriptor
rh: fstatat ./xfce/garcon: Bad file descriptor
rh: fstatat ./xfce/gtk-xfce-engine: Bad file descriptor
rh: fstatat ./xfce/libxfce4ui: Bad file descriptor
rh: fstatat ./xfce/libxfce4util: Bad file descriptor
rh: fstatat ./xfce/mousepad: Bad file descriptor
rh: fstatat ./xfce/orage: Bad file descriptor
rh: fstatat ./xfce/ristretto: Bad file descriptor
rh: fstatat ./xfce/squeeze: Bad file descriptor
rh: fstatat ./xfce/Thunar: Bad file descriptor
rh: fstatat ./xfce/tumbler: Bad file descriptor
rh: fstatat ./xfce/xfce: Bad file descriptor
rh: fstatat ./xfce/xfce4-appfinder: Bad file descriptor
rh: fstatat ./xfce/xfce4-dev-tools: Bad file descriptor
rh: fstatat ./xfce/xfce4-panel: Bad file descriptor
rh: fstatat ./xfce/xfce4-session: Bad file descriptor
rh: fstatat ./xfce/xfce4-settings: Bad file descriptor
rh: fstatat ./xfce/xfce4-terminal: Bad file descriptor
rh: fstatat ./xfce/xfconf: Bad file descriptor
rh: fstatat ./xfce/xfdesktop: Bad file descriptor
rh: fstatat ./xfce/xfwm4: Bad file descriptor
```
(This is on Rosetta, should be identical.)

barracuda156:
(CI are broken, nothing will pass now.)

ryandesign:
CI should be fixed; rerunning jobs.

sysutils/rawhide/files/patch-Makefile.diff
```diff
@@ -0,0 +1,17 @@
+Do not overwrite MacPorts's CFLAGS.
+--- Makefile
++++ Makefile
+@@ -135,11 +135,11 @@ RAWHIDE_DEFINES = \
+ #UBSAN_CFLAGS = -fsanitize=undefined
+ #UBSAN_LDFLAGS = -fsanitize=undefined
+ 
+-CC = cc
++#CC = cc
```
ryandesign, line 9: You can just remove the line. It's simpler, shorter, easier to read than commenting it out. Or, if you use the makefile portgroup, I don't think you need it, since the makefile portgroup overrides select Makefile variables, and `CC` is one of them (see `makefile.override` in makefile-1.0.tcl).

sysutils/rawhide/files/patch-Makefile.diff
```diff
@@ -0,0 +1,17 @@
...
+ #CC = gcc
+ #CC = other
+ CPPFLAGS = $(RAWHIDE_DEFINES) $(PCRE2_DEFINES) $(ACL_DEFINES) $(EA_DEFINES) $(ATTR_DEFINES) $(FLAG_DEFINES) $(MAGIC_DEFINES) $(DEBUG_DEFINES) $(HAVE_SYS_SYSMACROS) $(HAVE_SYS_MKDEV)
+-ALL_CFLAGS = $(CFLAGS) -O3 -g -Wall -pedantic $(CPPFLAGS) $(PCRE2_CFLAGS) $(ACL_CFLAGS) $(EA_CFLAGS) $(ATTR_CFLAGS) $(FLAG_CFLAGS) $(MAGIC_CFLAGS) $(GCOV_CFLAGS) $(UBSAN_CFLAGS)
++ALL_CFLAGS = $(CFLAGS) $(CPPFLAGS) $(PCRE2_CFLAGS) $(ACL_CFLAGS) $(EA_CFLAGS) $(ATTR_CFLAGS) $(FLAG_CFLAGS) $(MAGIC_CFLAGS) $(GCOV_CFLAGS) $(UBSAN_CFLAGS)
```
ryandesign, line 14: I agree that we want to use our optimization flags (which are within `CFLAGS`) rather than `-O3`, however you probably don't want to remove the rest of the flags the makefile is using here (`-g -Wall -pedantic`). You can just change the order: change `$(CFLAGS) -O3 -g -Wall -pedantic` to `-O3 -g -Wall -pedantic $(CFLAGS)`.

ryandesign (commented):
I've left some comments below. Also consider switching to the legacysupport 1.1 portgroup.

sysutils/rawhide/Portfile
```diff
@@ -20,16 +20,21 @@ license             GPL-3+
 categories          sysutils
 depends_lib         port:pcre2 port:libmagic
 maintainers         {raf.org:raf @macportsraf} openmaintainer
-revision            0
+revision            1
 
 github.setup        raforg ${name} ${version} v
```
ryandesign, line 25: This line should read:
```suggestion
github.setup        raforg rawhide 3.2 v
```
and then the separate `name` and `version` lines above should be removed.

Right after the `github.setup` line (or the `version` line, for ports that use that) is a good place to put the `revision` line.

ryandesign: The `github.setup` line typically goes near the top of the portfile, where you currently have the `name` and `version` lines.

aeiouaeiouaeiouaeiouaeiouaeiou: I'm working on it, however what to do about the broken `fstatat` now (or just fix the legacysupport linking this time)?

herbygillot: @ryandesign: @macportsraf explained his preference for why he kept separate `name` and `version` lines aside from `github.setup` here: https://github.com/macports/macports-ports/pull/19042#discussion_r1225035115

sysutils/rawhide/files/patch-Makefile.diff
macportsraf: That CC line is modified by the configure script when $CC is set. I don't know if that changes your assessment. But I don't think that it needs to be commented out or deleted. At least, I can't see any reason why it should be. It's probably harmless if CC is also set when make is run, because it would have the same value either way, but it's unnecessary to remove it.

sysutils/rawhide/files/patch-Makefile.diff
macportsraf: I was wondering if that would be a better ordering.

sysutils/rawhide/Portfile
macportsraf: Regarding fstatat() and the other POSIX 2008 functions: The code works on Linux, FreeBSD, OpenBSD, NetBSD, Solaris, later macOS (that doesn't need legacysupport), and Cygwin, so I don't think I'm using the functions incorrectly. fstatat() reports a bad file descriptor, but it's a positive file descriptor that was returned by openat(). I don't know what could be wrong with it. It's set to close on exec, but it's unlikely that that matters.

barracuda156: @catap Maybe you have an idea here? The issue is not ppc-specific, so Intel can be tested.

barracuda156:
@macportsraf Apparently either unlinkat is implemented wrongly in LegacySupport or it is inherently broken. I am getting the same Bad file descriptor errors now with another port despite linking to LegacySupport: build is fine, but related test cases fail.

barracuda156:
P. S. Or maybe it is DT_UNKNOWN fails to work but not unlinkat.

kencu:
observation, possible clue:

running "rh" gives a "Bad File Descriptor" error on any file that is in a subdirectory of where rh is run.

files (or folders) that are in the same directory do not show the error.

macportsraf:
It's not just unlinkat(). The error is happening when traversing the directory, not only when unlinking. And I don't think the problem is readdir() returning items with d_type == DT_UNKNOWN (although that might be happening).

openat() (on sub-directories) is returning a positive integer (directory) file descriptor that fstatat() reports is a bad file descriptor. It doesn't get as far as fdopendir() or readdir() (in sub-directories). But perhaps you mean the readdir() that lead to the openat(). Anyway, DT_UNKNOWN might not be important according to:

https://stackoverflow.com/questions/47078417/readdir-returning-dirent-with-d-type-dt-unknown-for-directories-and

DT_UNKNOWN doesn't indicate a problem.

By the way, the bfs port uses legacysupport and seems to work fine. But I don't know if it's using fstatat(). Its Portfile says that it needs legacysupport for "openat(), unlinkat(), fdopendir()" but I don't know if that's a complete list.

I'll provide a small test program that fails, hopefully later today.

When I originally tried to debug this on 10.6.8, I replaced fstatat() with stat() if it failed with EBADF but then a subsequent function failed instead. So I made that failover to a less secure alternative, and something else failed. I saw that failing over to less secure functions wasn't going to work.

kencu:
we have some (not enough) testing code in legacysupport. Here's a relevant but not exact test for you to look at:

https://github.com/macports/macports-legacy-support/blob/master/test/test_fstatat64.c

if you can add another test or two that brings out / demonstrates the error, perhaps we can find a way to fix it. Or perhaps your test will show things are working properly, and there is some other issue to sort out.

barracuda156:
@kencu @macportsraf Ok, let me be specific. The following code fails for me: https://github.com/capnproto/capnproto/blob/8512695db33531aba4e916ede54aecd322e124c4/c%2B%2B/src/kj/filesystem-disk-unix.c%2B%2B#L187-L201
I get the following error when running tests:
```
failed: unlinkat(fd, entry->d_name, 0): Bad file descriptor
```
A number of tests from `kj/filesystem-disk-test.c++` fail with the same error on the same spot.

kencu:
> @kencu @macportsraf Ok, let me be specific. The following code fails for me: https://github.com/capnproto/capnproto/blob/8512695db33531aba4e916ede54aecd322e124c4/c%2B%2B/src/kj/filesystem-disk-unix.c%2B%2B#L187-L201 I get the following error when running tests:
> 
> ```
> failed: unlinkat(fd, entry->d_name, 0): Bad file descriptor
> ```
> 
> A number of tests from `kj/filesystem-disk-test.c++` fail with the same error on the same spot.

that code does some fancy footwork, calling syscall directly itself…

I don’t know how that would work out using legacysupport, which provides the functions to call but does not override syscall in so doing and so does not reproduce the actual syscalls…

raforg:
Here's a directory traversal test program that fails in two different ways. Here's the explanation that is included at the top of the source:

This test shows two failure modes, depending on whether
the traversal starts in a named directory, or in the current
directory. In the first case, fstatat fails with
"No such file or directory". In the second case, fstatat
fails with "Bad file descriptor". In both cases, the failure
happens in a subdirectory of the starting directory.

The "Bad files descriptor" is the dirfd returned by openat().

It creates "dir", "dir/subdir" and "dir/subdir/file".
With no arguments, it traverses the "dir" directory.
With an argument, it chdirs into "dir" and then traverses "."
and tries to chdir to ".." afterwards. In both cases, it
tries to delete "dir" and its contents.

But in the case where the chdir happens, the chdir back to ".."
seems to succeed (it returns 0), but the cwd doesn't change,
and so the subsequent calls to unlink and rmdir fail.

So whatever is going wrong seems to be affecting a subsequent call
to chdir which seems impossible.

Note that on macos-10.14, compilation is just:

  gcc -o test_traverse test_traverse.c

But on macos-10.6, the Makefile is needed for legacysupport:

  make

The output should look something like:

  $ test_traverse
  fstatat(parent_fd=-2, dir) ok
  openat(parent_fd=-2, dir) = dir_fd=3 ok
  fdopendir(dir_fd=3) ok
  entry subdir
  fstatat(parent_fd=3, subdir) ok
  openat(parent_fd=3, subdir) = dir_fd=4 ok
  fdopendir(dir_fd=4) ok
  entry file
  fstatat(parent_fd=4, file) ok

  $ test_traverse .
  fstatat(parent_fd=-2, .) ok
  openat(parent_fd=-2, .) = dir_fd=3 ok
  fdopendir(dir_fd=3) ok
  entry subdir
  fstatat(parent_fd=3, subdir) ok
  openat(parent_fd=3, subdir) = dir_fd=4 ok
  fdopendir(dir_fd=4) ok
  entry file
  fstatat(parent_fd=4, file) ok
  cwd (before cd ..) /Users/raf/src/test_traverse/dir
  cwd (after cd ..) /Users/raf/src/test_traverse

But on my 10.6.8 laptop, it looks like:

  $ test_traverse
  fstatat(parent_fd=-2, dir) ok
  openat(parent_fd=-2, dir) = dir_fd=3 ok
  fdopendir(dir_fd=3) ok
  entry subdir
  fstatat(parent_fd=3, subdir) failed: No such file or directory

  $ test_traverse .
  fstatat(parent_fd=-2, .) ok
  openat(parent_fd=-2, .) = dir_fd=3 ok
  fdopendir(dir_fd=3) ok
  entry subdir
  fstatat(parent_fd=3, subdir) ok
  openat(parent_fd=3, subdir) = dir_fd=4 ok
  fdopendir(dir_fd=4) ok
  entry file
  fstatat(parent_fd=4, file) failed: Bad file descriptor
  cwd (before cd ..) /Users/raf/src/test_traverse/dir
  cwd (after cd ..) /Users/raf/src/test_traverse/dir
  unlink dir/subdir/file failed: No such file or directory
  rmdir dir/subdir failed: No such file or directory
  rmdir dir failed: No such file or directory

  $ rm -r dir

[Makefile.txt](https://github.com/macports/macports-ports/files/12046459/Makefile.txt)
[test_traverse.c.txt](https://github.com/macports/macports-ports/files/12046460/test_traverse.c.txt)

If I copy it into the `macports-legacy-support/test` directory and `make test`, it fails in the first way. If you like I could create a pull request in that repository that adds two new test programs to demonstrate both failure modes separately.

macportsraf:
Here's a directory traversal test program that fails in two different ways. Here's the explanation that is included at the top of the source:

This test shows two failure modes, depending on whether
the traversal starts in a named directory, or in the current
directory. In the first case, fstatat fails with
"No such file or directory". In the second case, fstatat
fails with "Bad file descriptor". In both cases, the failure
happens in a subdirectory of the starting directory.

The "Bad files descriptor" is the dirfd returned by openat().

It creates "dir", "dir/subdir" and "dir/subdir/file".
With no arguments, it traverses the "dir" directory.
With an argument, it chdirs into "dir" and then traverses "."
and tries to chdir to ".." afterwards. In both cases, it
tries to delete "dir" and its contents.

But in the case where the chdir happens, the chdir back to ".."
seems to succeed (it returns 0), but the cwd doesn't change,
and so the subsequent calls to unlink and rmdir fail.

So whatever is going wrong seems to be affecting a subsequent call
to chdir which seems impossible.

Note that on macos-10.14, compilation is just:

gcc -o test_traverse test_traverse.c

But on macos-10.6, the Makefile is needed for legacysupport:

make

The output should look something like:

$ test_traverse
fstatat(parent_fd=-2, dir) ok
openat(parent_fd=-2, dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=4 ok
fdopendir(dir_fd=4) ok
entry file
fstatat(parent_fd=4, file) ok

$ test_traverse .
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=4 ok
fdopendir(dir_fd=4) ok
entry file
fstatat(parent_fd=4, file) ok
cwd (before cd ..) /Users/raf/src/test_traverse/dir
cwd (after cd ..) /Users/raf/src/test_traverse

But on my 10.6.8 laptop, it looks like:

$ test_traverse
fstatat(parent_fd=-2, dir) ok
openat(parent_fd=-2, dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) failed: No such file or directory

$ test_traverse .
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=4 ok
fdopendir(dir_fd=4) ok
entry file
fstatat(parent_fd=4, file) failed: Bad file descriptor
cwd (before cd ..) /Users/raf/src/test_traverse/dir
cwd (after cd ..) /Users/raf/src/test_traverse/dir
unlink dir/subdir/file failed: No such file or directory
rmdir dir/subdir failed: No such file or directory
rmdir dir failed: No such file or directory

$ rm -r dir

Makefile.txt
test_traverse.c.txt

If I copy it into the `macports-legacy-support/test` directory and `make test`, it fails in the first way. If you like I could create a pull request in that repository that adds two new test programs to demonstrate both failure modes separately.

[Makefile.txt](https://github.com/macports/macports-ports/files/12046497/Makefile.txt)
[test_traverse.c.txt](https://github.com/macports/macports-ports/files/12046498/test_traverse.c.txt)

macportsraf:
Here's a directory traversal test program that fails in two different ways. Here's the explanation that is included at the top of the source:

This test shows two failure modes, depending on whether
the traversal starts in a named directory, or in the current
directory. In the first case, fstatat fails with
"No such file or directory". In the second case, fstatat
fails with "Bad file descriptor". In both cases, the failure
happens in a subdirectory of the starting directory.

The "Bad files descriptor" is the dirfd returned by openat().

It creates "dir", "dir/subdir" and "dir/subdir/file".
With no arguments, it traverses the "dir" directory.
With an argument, it chdirs into "dir" and then traverses "."
and tries to chdir to ".." afterwards. In both cases, it
tries to delete "dir" and its contents.

But in the case where the chdir happens, the chdir back to ".."
seems to succeed (it returns 0), but the cwd doesn't change,
and so the subsequent calls to unlink and rmdir fail.

So whatever is going wrong seems to be affecting a subsequent call
to chdir which seems impossible.

Note that on macos-10.14, compilation is just:

gcc -o test_traverse test_traverse.c

But on macos-10.6, the Makefile is needed for legacysupport:

make

The output should look something like:

$ test_traverse
fstatat(parent_fd=-2, dir) ok
openat(parent_fd=-2, dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=4 ok
fdopendir(dir_fd=4) ok
entry file
fstatat(parent_fd=4, file) ok

$ test_traverse .
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=4 ok
fdopendir(dir_fd=4) ok
entry file
fstatat(parent_fd=4, file) ok
cwd (before cd ..) /Users/raf/src/test_traverse/dir
cwd (after cd ..) /Users/raf/src/test_traverse

But on my 10.6.8 laptop, it looks like:

$ test_traverse
fstatat(parent_fd=-2, dir) ok
openat(parent_fd=-2, dir) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) failed: No such file or directory

$ test_traverse .
fstatat(parent_fd=-2, .) ok
openat(parent_fd=-2, .) = dir_fd=3 ok
fdopendir(dir_fd=3) ok
entry subdir
fstatat(parent_fd=3, subdir) ok
openat(parent_fd=3, subdir) = dir_fd=4 ok
fdopendir(dir_fd=4) ok
entry file
fstatat(parent_fd=4, file) failed: Bad file descriptor
cwd (before cd ..) /Users/raf/src/test_traverse/dir
cwd (after cd ..) /Users/raf/src/test_traverse/dir
unlink dir/subdir/file failed: No such file or directory
rmdir dir/subdir failed: No such file or directory
rmdir dir failed: No such file or directory

$ rm -r dir

If I copy it into the macports-legacy-support/test directory and make test, it fails in the first way. If you like I could create a pull request in that repository that adds two new test programs to demonstrate both failure modes separately.

[Makefile.txt](https://github.com/macports/macports-ports/files/12046512/Makefile.txt)
[test_traverse.c.txt](https://github.com/macports/macports-ports/files/12046513/test_traverse.c.txt)

kencu:
good work…

I guess next we need to debug it and see what’s actually going on

tavianator:
I believe I know the issue, at least in your `test_traverse.c` program.  You do

```c
		if (!(dir = fdopendir(dir_fd)))
		{
			...
			return EXIT_FAILURE;
		}
		...
		while ((entry = readdir(dir)))
		{
			...
			if (traverse(dir_fd, entry->d_name) == EXIT_FAILURE)
				return EXIT_FAILURE;
		}
```

But the [legacysupport](https://github.com/macports/macports-legacy-support/blob/b570b7f97476e5377ee2397a7a3a71638b72bdb5/src/fdopendir.c) implementation of `fdopendir()` is basically

```c
DIR *fdopendir(int dirfd) {
    int oldCWD = open(".", O_RDONLY);
    fchdir(dirfd);
    DIR *dir = opendir(".");
    fchdir(oldCWD);
    close(oldCWD);
    close(dirfd);
    return dir;
}
```

In other words, there's no guarantee that `dir_fd` is still open or refers to the same directory after the `fdopendir()` call.  You could work around in `test_traverse.c` with

```c
		if (!(dir = fdopendir(dir_fd)))
		{
			...
			return EXIT_FAILURE;
		}
		dir_fd = dirfd(dir);
```

I don't know if `dirfd(fdopendir(fd)) != fd` violates the spec, but it's certainly weird.  I believe `bfs` is also broken if that happens.

aeiouaeiouaeiouaeiouaeiouaeiou:
[rebase to sync with upstream branch]

kencu:
closing the dirfd was added here

https://github.com/macports/macports-legacy-support/commit/3c9d3c2fc4efb9d1e58e9fb9d44899463391a276

by an interested user who found leaving it open caused leaks.

kencu:
Based on this:

“
The fdopendir() function is like opendir(), but returns a directory stream for the directory referred to by the open file descriptor fd. After a successful call to fdopendir(), fd is used internally by the implementation, and should not otherwise be used by the application.
“

it appears legacysupport may not be wrong here to close fd.

tavianator:
Well the [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fdopendir.html) says

> Upon successful return from fdopendir(), the file descriptor is under the control of the system, and if any attempt is made to close the file descriptor, or to modify the state of the associated description, other than by means of [closedir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/closedir.html), [readdir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/readdir.html), [readdir_r()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/readdir_r.html), [rewinddir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/rewinddir.html), or [seekdir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/seekdir.html), the behavior is undefined. Upon calling [closedir()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/closedir.html) the file descriptor shall be closed.

kencu:
The man page says this:

After a successful call to fdopendir(), fd is used internally by the implementation, and should not otherwise be used by the application.

Which is what the legacysupport implementation does...

You are constrained to a degree by what is possible. The system on 10.6 and less does not support O_CLOEXEC. So the original workaround was invented, and worked pretty well.

Then someone discovered a memory leak, and found we had to close the original fd during fdopendir, which is OK, as it is "under control of the system" anyway, so we can do what we want with it (being "the system" at this point).

I'm open to ideas, but it looks to me like legacysupport may be OK, and the software (rawhide, the test program, etc) should not be using "fd" after the call to fdopendir....

Otherwise ... I have no ideas.

kencu:
Well, we could keep a custom list of open fds I guess, and patch closedir, readdir, etc to close the fds and call the original system call ...

That might mimic what the newer OSs are practically doing .... although still, it seems to me that software is not supposed to assume fd remains valid after a call to fdopendir... "After a successful call to fdopendir(), fd is used internally by the implementation, and should not otherwise be used by the application."

macportsraf:
I disagree that legacysupport may be OK. I was definitely under the impression that it's closedir() that closes the dirfd, not fdopendir(). Code that works everywhere else (e.g. Linux, FreeBSD, OpenBSD, NetBSD, Solaris, Cygwin, later macOS) doesn't work with legacysupport. So, even if it isn't non-compliant, it is incompatible. I suppose what matters is how many affected ports would need to be patched. Maybe it's not many.

But having said that, if you disagree, or if it's too much hassle, I could add a dup() which should fix it. Hmm, the obvious attempt fixes the test case where it starts with a directory name to traverse, but it doesn't fix the case where it starts with the current directory.

Actually, even in that case, all the fstatat() calls work, but the chdir("..") afterwards to delete the test directory still fails to actually do anything (while still returning 0):

    cd dir
    fstatat(parent_fd=-2, .) ok
    openat(parent_fd=-2, .) = dir_fd=3 ok
    fdopendir(dir_fd=3) ok
    set dir_fd = (duped) 4
    entry subdir
    fstatat(parent_fd=4, subdir) ok
    openat(parent_fd=4, subdir) = dir_fd=5 ok
    fdopendir(dir_fd=5) ok
    set dir_fd = (duped) 3
    entry file
    fstatat(parent_fd=3, file) ok
    cwd (before cd ..) /Users/raf/src/macports-legacy-support/dir
    cwd (after cd ..)  /Users/raf/src/macports-legacy-support/dir
    unlink dir/subdir/file failed: No such file or directory
    rmdir dir/subdir failed: No such file or directory
    rmdir dir failed: No such file or directory

Here's a new version of the test program that dup()s dirfd. Any idea how the subsequent chdir("..") could be failing to have any effect? This matters for rawhide because chdir/fchdir is used a lot in some cases.

If I patch rawhide itself, just traversing works, but executing a command on matching files (when it involves chdir/fchdir) doesn't work. It reports "No such file or directory" for everything in subdirectories.

Also, how can a Portfile apply a patch only when legacysupport is in effect?

[test_traverse.c.txt](https://github.com/macports/macports-ports/files/12063059/test_traverse.c.txt)

kencu:
I'm not an fdopendir expert by any stretch, but if you (the royal "you") all feel that this statement is wrong "After a successful call to fdopendir(), fd is used internally by the implementation, and should not otherwise be used by the application." and it seems that fd should still remain valid to use, then I don't see any fix, other than stripping out https://github.com/macports/macports-legacy-support/commit/3c9d3c2fc4efb9d1e58e9fb9d44899463391a276 and letting it leak again, or perhaps  someone doing a big project to patch all the functions I mentioned above.

Maybe someone could revert https://github.com/macports/macports-legacy-support/commit/3c9d3c2fc4efb9d1e58e9fb9d44899463391a276 and see if things work the way you expect them to.

If that does change it back ... well, we're done then I guess. Certain ports won't be compatible with legacysupport's version of fdopendir.

That's fine... nobody promised it would be, and it does seem to work for the vast majority of software... so still a win.

macportsraf:
I think the statement from the POSIX spec that prohibits specific things should override the statement from the manual entry that prohibits all unspecified things. The POSIX spec quote above says:

    if any attempt is made to close the file descriptor, or to modify the state
    of the associated description, other than by means of closedir(), readdir(),
    readdir_r(), rewinddir(), or seekdir(), the behavior is undefined.

The application isn't attempting to close the file descriptor or to modify its state. It is only using it in certain syscalls to indicate the relevant directory. I don't think that that constitutes a state change. It's not read from or seeked or closed.

I think the quote from the manual entry goes a lot further than the POSIX spec in what it prohibits (even though the additional prohibition doesn't apply to any other implementations). Perhaps whoever wrote it just wasn't thinking about syscalls like openat() when they said "should not otherwise be used by the application".

Ideally, the DIR structure would include a field to store the file descriptor so that closedir() could close it. Then there would be no file descriptor leak, and the behaviour would match that of other systems. But I don't know if there would be a big problem with adding an int field to the DIR structure. Perhaps it's not feasible but it sounds easy in theory. struct DIR would have to redefined in legacysupport's dirent.h to extend the system DIR with an int field.

Anyway, if there is no desire to fix this (or to agree that there is anything to fix), I'm happy to patch the application, but some help with understanding the wierd behaviour of the subsequent call to chdir() would be greatly appreciated. Without fixing that, it still won't work.

If that's not possible, I'll have to just indicate in the Portfile that rawhide doesn't work on platforms that require legacysupport. What would that look like? Would it be:

    platforms {darwin > 13}

I'm guessing that from `legacysupport.newest_darwin_requires_legacy 13` in the Portfile which I copied from the bfs Portfile.

But if anyone can see why chdir() is failing to have any effect, and what can be done about it, that would be great.

kencu:
> Anyway, if there is no desire to fix this (or to agree that there is anything to fix), I'm happy to patch the application, but some help with understanding the wierd behaviour of the subsequent call to chdir() would be greatly appreciated. Without fixing that, it still won't work.

I don't think anyone said that.

Once it is fully sorted out where the error is, if there is an agreed-upon error in legacysupport that can be fixed, I think there are users out that that would appreciate your fix, assuming that such a fix is possible.

So PR away!

kencu:
> If that's not possible, I'll have to just indicate in the Portfile that rawhide doesn't work on platforms that require legacysupport. What would that look like? Would it be:
> 
> ```
> platforms {darwin > 13}
> ```

I suppose you would just take legacysupport out completely then, and mark it broken, if you find it can't be fixed.

There is a new platforms syntax for that, indeed, if that is the conclusion.

kencu:
In the meantime, we might as well convert this to a draft (or just close it) as even if legacysupport is effectively added, it doesn't fix the software so we don't actually want to do that now anyway.

barracuda156:
@kencu I will try dropping that commit from LegacySupport and see how tests in `capnproto` behave. Gonna update on this today.

barracuda156:
@macportsraf I think it is verily desirable to have LegacySupport fixed rather than making a per-case fix to the specific port, since this issue matters for other ports and affects, at least potentially, many users.

barracuda156:
@macportsraf @kencu Definitely it is LegacySupport is broken. I reverted the commit quoted above, and all filesystem-related tests of `capnproto` now pass.

kencu:
thanks for confirming.

we can’t just revert that commit for the main legacysupport project, as it will leak again.

so fdopendir in legacysupport will have to stay as is until someone puts together a more comprehensive fix as described above.

barracuda156:
@kencu Isn’t it better to have it at least working, if imperfectly, than broken, like it is now?

P. S. It is not an area I understand, so cannot rewrite the thing, unfortunately.

kencu:
> @kencu Isn’t it better to have it at least working, if imperfectly, than broken, like it is now?
> 
> P. S. It is not an area I understand, so cannot rewrite the thing, unfortunately.

It is working now.

But when edge cases, like these couple, are found, then someone needs to dig deeper.

If you revert the commit it leaks again, and you just broke it another way.

barracuda156:
@macportsraf Maybe you could look into fixing `legacysupport` implementation, so that both `rawhide` and `capnproto` are fixed?

kencu:
let's open a trac ticket about this, and ... I guess ... close this PR as nobody will ever find the legacysupport issue in here.

kencu:
https://trac.macports.org/ticket/67754

barracuda156:
> let's open a trac ticket about this, and ... I guess ... close this PR as nobody will ever find the legacysupport issue in here.

Maybe wait until @macportsraf responds here?

kencu:
This PR is about rawhide linking against legacysupport.

Discussion about legacysupport's implementation of fdopendir should logically be in the ticket to keep things coherent - otherwise it gets like those issues in Iain's repo with 1000 different topics in them, where nobody can find any relevant info later and it's just a mess.

But as you wish, close it or don't.

macportsraf:
Hi, thanks everyone. I'll create a PR for legacysupport to delay closing the supplied file descriptor until closedir(). I think it just needs a wrapper data structure and macro for DIR and wrapper functions and/or macros for all of the functions that take a DIR* argument. I suspect that I might still have a problem with subsequent calls to chdir (even though that problem makes no sense at all). We'll see. But even if that's the case, it will help ports that don't require chdir. Even rawhide would mostly work. It's -X option (and "cmd".sh predicate) wouldn't function with legacysupport and I can add a note to the Portfile about that. But hopefully that problem will go away as well.

macportsraf:
Update: Exellent progress is being made with legacysupport's implementation of fdopendir. I suspect that it will soon work properly on macos-10.5+ (but macos-10.4 is still being difficult). It's working for me on macos-10.6.

kencu:
can’t thank you two enough… much appreciated

barracuda156:
@macportsraf Awesome news!

macportsraf:
Update: It's working on 10.4 now as well! Yay!
It makes me want to track down a working power cord for my old PowerBook G4. :-)
Best keyboard I've ever used.

kencu:
won't work "right" until the new legacysupport release, but this PR is OK.

barracuda156:
@macportsraf Could you please take a look? https://github.com/vectorgraphics/asymptote/issues/386#issuecomment-1648615746
Something got broken with `seekdir`, which worked fine previously.

PR #19055, opened 2023-06-11
zfp: update to upstream, fix tests on 32-bit

barracuda156:
#### Description

Update and a fix from upstream for 32-bit compilation and tests.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6.8 Server
Xcode 3.2.6

###### Verification 
Have you

### DIFF
--- a/science/zfp/Portfile
+++ b/science/zfp/Portfile
@@ -2,10 +2,11 @@
 
 PortSystem          1.0
 PortGroup           cmake 1.1
+PortGroup           compiler_blacklist_versions 1.0
 PortGroup           github 1.0
 
-github.setup        LLNL zfp f15d00720e454cd70206a85d7013d8240ad38585
-version             2023.02.11
+github.setup        LLNL zfp 300e77d12f25d3eaa4ba9461d937fb17a71d45f6
+version             2023.04.30
 revision            0
 categories          science devel
 license             BSD
@@ -14,14 +15,14 @@ description         zfp is a compressed format for representing multi-dimensiona
 long_description    {*}${description}
 homepage            https://zfp.llnl.gov
 
-checksums           rmd160  cac426354ae3f5f7ffd15442d044a0dbad876bba \
-                    sha256  d9ceb27be5a8591c204b2599e60f178edeaf0139da24aeee5d3f1768fe3cc3e2 \
-                    size    504306
+checksums           rmd160  6427229673f288ae8e3d8ca0d90ec8264ab4c9dd \
+                    sha256  69754125d2b25f5a35574aac20af32dea2c17f7268fcb6cc26d0e1205467bf74 \
+                    size    504422
 
-compiler.c_standard 1999
 # index.hpp: error: integer constant is too large for ‘unsigned long’ type
 compiler.blacklist-append \
-                    *gcc-4.*
+                    {clang < 400} *gcc-4.*
+compiler.c_standard 1999
 
 configure.args-append \
                     -DBUILD_EXAMPLES=OFF \
@@ -44,6 +45,17 @@ variant openmp description {use OpenMP} {
     }
 }
 
-# Note, that tests are likely to fail on 32-bit platforms due to insufficient precision:
-# https://github.com/LLNL/zfp/issues/205
+# See: https://github.com/LLNL/zfp/issues/205
+if {${configure.build_arch} in [list i386 ppc]} {
+    configure.args-append \
+                    -DZFP_INT64='long long' \
+                    -DZFP_INT64_SUFFIX=ll \
+                    -DZFP_UINT64='unsigned long long' \
+                    -DZFP_UINT64_SUFFIX=ull
+}
+
 test.run            yes
+# Drop once someone finally removes this from cmake PG:
+configure.pre_args-replace \
+                    -DCMAKE_BUILD_WITH_INSTALL_RPATH:BOOL=ON \
+                    -DCMAKE_BUILD_WITH_INSTALL_RPATH:BOOL=OFF


barracuda156:
```
--->  Testing zfp
Executing:  cd "/opt/local/var/macports/build/_opt_PPCRosettaPorts_science_zfp/zfp/work/build" && DYLD_LIBRARY_PATH=/opt/local/var/macports/build/_opt_PPCRosettaPorts_science_zfp/zfp/work/build/lib /usr/bin/make test 
Running tests...
/opt/local/bin/ctest --force-new-ctest-process 
Test project /opt/local/var/macports/build/_opt_PPCRosettaPorts_science_zfp/zfp/work/build
    Start 1: testzfp
1/2 Test #1: testzfp ..........................   Passed    1.43 sec
    Start 2: testviews
2/2 Test #2: testviews ........................   Passed    0.26 sec

100% tests passed, 0 tests failed out of 2
```

science/zfp/Portfile
```diff
@@ -44,6 +45,14 @@ variant openmp description {use OpenMP} {
...
+if {${configure.build_arch} in [list i386 ppc]} {
+    configure.args-append \
+                    -DZFP_INT64='long long' \
+                    -DZFP_INT64_SUFFIX=ll \
+                    -DZFP_UINT64='unsigned long long' \
+                    -DZFP_UINT64_SUFFIX=ull
+}
+
 test.run            yes
+test.cmd-prepend    DYLD_LIBRARY_PATH=${cmake.build_dir}/lib
```
ryandesign, line 58: You don't want to set t his in `test.env` instead?

kencu:
most always, the better fix for cmake tests is to toggle this off

https://github.com/macports/macports-ports/blob/8189f16db57801d6533bd33e3dbcbac4580a80f3/_resources/port1.0/group/cmake-1.1.tcl#L111

kencu:
we have talked before about having that off as the default so tests would always work… I’m not sure why it is on in the portgroup

ryandesign:
> we have talked before about having that off as the default so tests would always work… I’m not sure why it is on in the portgroup

I think this has caused me problems in my ports too. Is there a PR or ticket for that previous discussion? I don't recall.

barracuda156:
> we have talked before about having that off as the default so tests would always work… I’m not sure why it is on in the portgroup

@kencu @ryandesign It is safe to just always turn that option off?

science/zfp/Portfile
barracuda156: I think there was some case which led me to generally prefer this form, but I cannot recall now. Here it is perhaps equivalent. (Are these supposed to differ somehow, in theory, or Macports does the same with either?)

kencu:
> 

Just fyi It’s also a one-line fix in the Portfile… so not onerous.

barracuda156:
> Just fyi It’s also a one-line fix in the Portfile… so not onerous.

My reasoning is that I am not really sure about potential consequences of changing settings for the build, while `DYLD_LIBRARY_PATH` passed to tests affects only tests and nothing else. So as long as the latter works perfectly fine, why not just use that (at least until CMake PR is changed, which I leave for someone more competent to deal with).

kencu:
There are dozens if other ports that fix the cmake tests as I outlined, fyi. 

Just letting you know what the correct way to fix this is. 

Cmake is designed to allow testing to work, after all. MacPorts breaks testing on purpose — no idea why.

barracuda156:
@kencu I have no objections to that way, but maybe someone with commit access can finally fix it in PG, and then we can drop all those dozens of fixes from portfiles?

kencu:
and look, @barracuda156 -- you used the proper method yourself a few weeks ago here!

https://github.com/macports/macports-ports/commit/81a7f66213c10d160c69566194800e482942628c

barracuda156:
> and look, @barracuda156 -- you used the proper method yourself a few weeks ago here!
> 
> [81a7f66](https://github.com/macports/macports-ports/commit/81a7f66213c10d160c69566194800e482942628c)

@kencu Because it is inside a variant ;)

I don’t really get why this port arbitrarily was picked to make an issue of this, when numerous ports use `DYLD_LIBRARY_PATH`, for example @ryandesign own `minizip-ng` https://github.com/macports/macports-ports/blob/fb48eaa07990d6a640b2b8974778a4d7e800eed4/archivers/minizip-ng/Portfile#L83-L85 and `taglib` https://github.com/macports/macports-ports/blob/fb48eaa07990d6a640b2b8974778a4d7e800eed4/audio/taglib/Portfile#L62-L64 – and apparently over hundred other ports in Macports repo (I am not gonna go through every reference).

I would appreciate an explanation why that is wrong – whether in principle or in a specific context. If it is, I will change not only this port, but a few other of my ports which use a similar syntax. I do not want to have a wrong code, here or elsewhere.

If it is not wrong, then, at the very minimum, changing would require extra testing, while everything already works :)

barracuda156:
P. S. @kencu And you are right, I did use your syntax elsewhere (not only in `kokkos-devel`), and there it was tested and confirmed to work. And in other instances `DYLD_LIBRARY_PATH` was tested and confirmed to work (like here).

kencu:
by default, to allow testing, cmake (and meson) build software such that the executables and libraries in the built products refer to the libraries in the build tree. This way, when the test suites are run, they reference the newly built libraries rather than any already installed libraries that might exist on the user's system.

When the software is installed, these temporary build-tree references are rewritten to final installed paths, so that things are correct when they actually get installed.

cmake (but not meson) came up with an option to override this behaviour, and build the software with references to the final installed paths instead. This is certain to break testing (and does) because the executables in the build tree cannot find any of the libraries when things are built this way. 

Why exactly cmake ever thought this option was a useful or good idea is pretty unclear. It seems pretty useless to me. Maybe there is some situation where it is needed that I just can't think of. meson doesn't even have such an option (and good thing, really).

cmake does not make this option a default option.

Somewhere along the line, whoever wrote up the cmake portgroup thought it sounded like a good idea to enable this option. I don't know why. Enabling `-DCMAKE_BUILD_WITH_INSTALL_RPATH:BOOL=ON` only has the effect of breaking all cmake testing, and (AFAICT) does no useful thing.

At this point, it seems everyone is afraid to change the default to ```-DCMAKE_BUILD_WITH_INSTALL_RPATH:BOOL=OFF``` because they don't want to be responsible for some unforeseen effect of doing that that will tick somebody off. So they leave it.

People often don't seem to know what the option does exactly -- they think it might have something to do with rpath, maybe, and that sounds kind of baddish -- so they leave it, and work around it with the brute force method of setting DYLD_LIBRARY_PATH. 

That kinda can work around it, but if there are multiple libraries and executables in the cmake build tree, it doesn't work well at all, as it's too brutish of an approach. Ergo [81a7f66](https://github.com/macports/macports-ports/commit/81a7f66213c10d160c69566194800e482942628c) which fixed it right, in that case.

kencu:
reading through the cmake documentation on this option is, by the way, anything but enlightening.

barracuda156:
Rebased to master. If no one is really insisting on switching to CMake tweak, let’s merge. Otherwise I will do that, don’t want to have it stalled indefinitely.

kencu:
I see you used the cmake fix again in the new clamav port you just had committed :>

barracuda156:
@kencu Done.

barracuda156:
@herbygillot Thank you!

P. S. I am in the process of enabling MPI variant for `mfem`, and it uses `zfp`. Will need to make another fix to `strumpack` too to ensure MPI variant is actually built with MPI and needed headers are installed.

kencu:
> @kencu
> 
> > I see you used the cmake fix again in the new clamav port you just had committed.
> 
> In `clamav` only non-functional changes are mine (lint etc.). For `clamav-legacy` I used an earlier version of `clamav`. So…
> 
> However, I am tired of arguing on this. Fine, I will test now for your way of fixing tests, and gonna rebase the PR, provided that way works. _Then only some 100+ ports of every maintainer are left to change._

Oh, come now -- deep, deep down inside, you know you feel good by fixing it the right way ;>

PR #19062, opened 2023-06-12
sshdo: new port (version 1.1.1)

macportsraf:
#### Description

sshdo - controls which commands may be executed via incoming ssh

sshdo provides an easily configurable way of controlling which commands may be executed via incoming ssh connections by specific users and specific keys. Training mode allows a key to be used for any command, while logging them. Then sshdo can learn from the logs to know what commands to allow. It can also unlearn commands that are no longer in use. This mitigates against private ssh key compromise with very little effort.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G6032 x86_64
Xcode 11.3 11C29

macOS 10.6.8 10K549 i386
Xcode 3.2.6 10M2518

###### Verification
Have you

### DIFF
--- a/security/sshdo/Portfile
+++ b/security/sshdo/Portfile
@@ -1,0 +1,70 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           makefile 1.0
+
+github.setup        raforg sshdo c27b4a0
+version             1.1.1
+revision            0
+
+description         controls which commands may be executed via incoming ssh
+long_description    {*}${description}. \
+                    \n\nsshdo provides an easily configurable way of controlling which \
+                    commands may be executed via incoming ssh connections by specific \
+                    users and specific keys. Training mode allows a key to be used for \
+                    any command, while logging them. Then sshdo can learn from the logs \
+                    to know what commands to allow. It can also unlearn commands that \
+                    are no longer in use. This mitigates against private ssh key compromise \
+                    with very little effort.
+license             GPL-2+
+categories          security
+supported_archs     noarch
+platforms           {darwin any}
+maintainers         {raf.org:raf @macportsraf} openmaintainer
+
+homepage            https://raf.org/sshdo/
+checksums           rmd160  e597acb80e58da62d0432ab3638eddda813ed653 \
+                    sha256  edd8b082c508390d597b728fb8ad6c4f6c432672406e0b395ae4617e2e57b0d9 \
+                    size    76694
+
+
+# Any of Python 2.6, 2.7, or 3.3+ will do.
+# Update the python3 version every year or so.
+set python_branch   3.11
+set python_version  [string map {. {}} ${python_branch}]
+depends_lib-append  port:python${python_version}
+
+# Note: Not a GNU autoconf configure script
+use_configure       yes
+configure.args      --prefix="${prefix}" \
+                    --etcdir="${prefix}"/etc \
+                    --with-python="${prefix}"/bin/python${python_branch} \
+                    --enable-mangz
+build {}
+
+test.run            yes
+
+destroot.keepdirs   ${destroot}${prefix}/etc/sshdoers.d
+
+notes {
+    sshdo relies on syslog messages to learn and unlearn policy.
+
+    On recent macOS systems with the unified logging system,\
+    syslog messages don't look the same as elsewhere, and they no\
+    longer appear in /var/log/system.log where sshdo expects them,\
+    so you'll need to use the log(1) command to feed log messages\
+    into sshdo via its standard input (stdin). See man sshdo.
+
+    Use the following command as the basis for your log queries:
+
+        log show --style syslog --info --predicate '
+            process == "Python" and
+            composedMessage contains "type=" and
+            composedMessage contains "user=" and
+            composedMessage contains "command="
+        ' | sed -E -e 's/Python/sshdo/' -e 's/: \([^)]+\)/:/'
+
+    You will also need to incorporate the appropriate time range\
+    into the search criteria (e.g., --last 7d). See man log.
+}


security/sshdo/Portfile
```diff
@@ -0,0 +1,57 @@
...
+github.tarball_from releases
+checksums           rmd160 6de62e1182526b5c245b6b7a75974fffd0ffeda1 \
+                    sha256 04a36696538ddbf188f86c357c1d2eaca21b2a09dd555bae072ed8660400c75a \
+                    size 78652
+
+patchfiles          macports.patch
+
+# There is nothing to build
+use_configure       no
+build.target        help
```
herbygillot, line 33: If your intent is to have `make` do nothing by just running a useless make target, you can simply use:

```
build {}
```

This will effectively disable the build phase.

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,131 @@
...
+@@ -45,6 +45,6 @@
+ #
+ # training
+ 
+-# Add command authorization directives here or in /etc/sshdoers.d/*
++# Add command authorization directives here or in /opt/local/etc/sshdoers.d/*
+ # (e.g. "user1 user2/label +group3 -user3: cmd arg").
+ 
+--- test_sshdo.orig	2021-09-08 21:02:25.000000000 +1000
++++ test_sshdo	2023-06-12 16:08:39.000000000 +1000
```
herbygillot, line 98: I wonder how much of this patch should actually be part of the upstream source?

security/sshdo/Portfile
macportsraf: Thanks. Done.

security/sshdo/files/macports.patch
macportsraf: Probably most of it. I can add a configure script for the next version and make the patch unnecessary or much smaller, but it's what's needed for v1.1.

security/sshdo/Portfile
herbygillot: The `configure` phase is separate from the `build` phase, so you may still need `use_configure no` alongside `build {}`.

security/sshdo/Portfile
```diff
@@ -0,0 +1,55 @@
...
+long_description    {*}${description}. \
+                    \n\nsshdo provides an easily configurable way of controlling which \
+                    commands may be executed via incoming ssh connections by specific \
+                    users and specific keys. Training mode allows a key to be used for \
+                    any command, while logging them. Then sshdo can learn from the logs \
+                    to know what commands to allow. It can also unlearn commands that \
+                    are no longer in use. This mitigates against private ssh key compromise \
+                    with very little effort.
+
+github.setup        raforg ${name} ${version} v
```
reneeotten, line 23: move this to the top of the Portfile and incorporate the "name" and "version" fields in the "github.setup" line

reneeotten (changes requested):

security/sshdo/Portfile
macportsraf: Thanks, but I'd really rather not. Unless there's a compelling reason.

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,131 @@
...
+ install: man
+-	[ -d $(ETCDIR)/$(SSHDOERS).d ] || mkdir -m 755 $(ETCDIR)/$(SSHDOERS).d
+-	[ -f $(ETCDIR)/$(SSHDOERS) ] || install -o root -g root -m 644 $(SSHDOERS) $(ETCDIR)
+-	[ -f $(ETCDIR)/$(SSHDO_BANNER) ] || install -o root -g root -m 644 $(SSHDO_BANNER) $(ETCDIR)
+-	install -o root -g root -m 755 $(BIN) $(BINDIR)
++	[ -d $(ETCDIR)/$(SSHDOERS).d ] || mkdir -p -m 755 $(ETCDIR)/$(SSHDOERS).d
++	[ -f $(ETCDIR)/$(SSHDOERS) ] || install -o root -g wheel -m 644 $(SSHDOERS) $(ETCDIR)
++	[ -f $(ETCDIR)/$(SSHDO_BANNER) ] || install -o root -g wheel -m 644 $(SSHDO_BANNER) $(ETCDIR)
++	install -o root -g wheel -m 755 $(BIN) $(BINDIR)
+ 	[ -z "`which python3`" -a -n "`which python`" ] && sed 's/env python3$$/env python/' < $(BIN) > $(BINDIR)/$(BIN) || true
```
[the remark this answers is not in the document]
macportsraf: Thanks. Why would it be better? There are no non-standard python modules needed. I think it's better to work with whatever version of python the local system already has. There's no need to download and install a different version. I also think it it's better to reduce the prerequisites as much as possible.

But if there's a good reason to require a particular macports python version, which one to pick? It works with 2.6, 2.7, and 3.3+. Will I have to keep changing it every so often? Or is there a port that means the current default version of python3?

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,131 @@
+--- Makefile.orig	2021-09-09 19:58:05.000000000 +1000
++++ Makefile	2023-06-12 13:16:16.000000000 +1000
+@@ -23,10 +23,10 @@
+ ID := $(NAME)-$(VERSION)
+ 
+ DESTDIR :=
+-PREFIX := $(DESTDIR)/usr
++PREFIX := $(DESTDIR)/opt/local
```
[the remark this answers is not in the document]
macportsraf: Thanks. How do I do that? I assumed patch files were static.

security/sshdo/Portfile
macportsraf: OK. I've changed it to download from the homepage instead. Is that OK?

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
reneeotten: for example because older OSes do not have "python3" and newer ones eventually will not have Python pre-installed if I am not mistaken. 

Typically, one uses the default as is set in the "python" PG. that version is incremented every year on January 1.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
reneeotten: I haven't tried this myself but it should be done in a "post-patch" block. At that point the should exist. 

Typically, you would add a place holder (e.g., "@@MACPORTS@@") in the patchfile and then do a reinplace of that.

But I agree with @herbygillot, you should just change the Makefile such that it will take values from the environment if set; in that case MacPorts will likely set most of things to their correct values, potentially with needing the "makefile" PortGroup.

security/sshdo/Portfile
macportsraf: Aesthetics. Never mind. I thought it was only conventions relating to the use of the github port group that concerned you, and that by not using that port group, we could both get what we want. I've changed it so that the github.setup is near the top.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
macportsraf: Thanks. It works with python 2 or 3. But if there isn't going to be a system python3 in future, that would be a problem. I've changed it to depend on port python311. That seems to be the current one. I'm not using the python port group because it seems to assume that setup.py will be used for installation, but this program uses a Makefile.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
macportsraf: Thanks. I found example uses of reinplace and it's now working.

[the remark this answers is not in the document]
macportsraf: Apologies in advance for the things you already know, but this is public so others might read it:

Makefile macros can always be overridden via the command line, rather than via the environment. Overriding via the environment only works if there is no value in the Makefile itself. For some macros, like CFLAGS, it can make sense to set them via the environment (only), because they might not strictly be necessary, so they can be left empty in the Makefile (although I would have thought that most Makefiles for C programs would set CFLAGS in the Makefile unless they are anticipating something like the macports legacysupport environment). But for some macros, like PREFIX, some non-empty value needs to be in the Makefile, so the best way to override those is via the command line (or a configure script that changes the Makefile). The make install command in macports overrides DESTDIR via the command line (that might be due to the makefile port group, not sure). It could/might also do it for PREFIX, but I haven't seen it do that (possibly because the prefix was set by configure (in a different port with use_configure yes)).

I've change the Portfile to include the makefile port group, because there is a Makefile, and I've changed the Makefile so that the PREFIX no longer includes DESTDIR (that was wrong). Now they can be overridden independently of each other. So, if the PREFIX were overridden on the make install command line, it should be fine without the need for reinplace. But it would probably require overriding the make install command so as to include PREFIX=${prefix}. However, the reinplace changes more than just the Makefile (i.e., the config file and sshdo itself). So it's probably best to keep using it for now. The next version of sshdo will have a configure script that will be able to do everything that the reinplace is doing now for v1.1.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
macportsraf: Thanks. How do I do that?

security/sshdo/Portfile
```diff
@@ -0,0 +1,78 @@
...
+                    any command, while logging them. Then sshdo can learn from the logs \
+                    to know what commands to allow. It can also unlearn commands that \
+                    are no longer in use. This mitigates against private ssh key compromise \
+                    with very little effort.
+# Any of Python 2.6, 2.7, 3.3+ will do
+depends_lib         port:python311
+
+homepage            https://raf.org/sshdo/
+github.tarball_from releases
+master_sites        {*}${master_sites} \
```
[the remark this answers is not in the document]
macportsraf: OK, but out of interest, what's wrong with multiple download sites? The macports guide says that when there are multiple download sites, they will be sorted by ping times, and the closest one chosen. That sounds like a good thing.

security/sshdo/Portfile
```diff
@@ -0,0 +1,78 @@
...
+
+patchfiles          macports.patch
+
+post-patch {
+    # Set the installation prefix
+    # Suppress a runtime reinplace warning when ${prefix} is /opt/local (the default)
+    # Note: \ is used here to defuse a false positive port lint error
+    if {![string match \/opt\/local ${prefix}]} {
+        reinplace -W ${worksrcpath} "s|/opt/local|${prefix}|g" Makefile sshdo sshdoers
+    }
```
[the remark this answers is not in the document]
macportsraf: Sorry. I didn't realise that was an instruction. It wasn't worded like one. There will be a replacement pull request for the new version (1.1.1) which won't need the reinplace.

security/sshdo/Portfile
```diff
@@ -0,0 +1,78 @@
...
+        ' | sed -E -e 's/Python/sshdo/' -e 's/: \([^)]+\)/:/'
+
+    You will also need to incorporate the appropriate time range
+    into the search criteria (e.g., --last 7d). See man log.
+}
+
+livecheck.type      regex
+livecheck.url       ${homepage}download/
+livecheck.regex     ${name}-(\\d+(?:\\.\\d+)*)${extract.suffix}
+
```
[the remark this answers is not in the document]
macportsraf: OK.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
macportsraf: If the user has already selected some version of python3 (which I expect would usually be the case), I don't think a port should go against the user's wishes and change the selected python3 version to some other version. That seems rude. But perhaps I'm misinterpreting, and that's not what you are saying. If the port is required to depend on a particular version of python, and it's not allowed to assume that the user has selected their preferred version of python3, maybe I should just create a patch to change the python3 to python3.11. Would that be an accpetable approach?

security/sshdo/files/macports.patch
raforg: I've upgraded it to v1.1.1 which has a configure script that eliminates the patch that was needed for v1.1.

security/sshdo/Portfile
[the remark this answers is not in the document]
raforg: I've upgraded it to v1.1.1 which has a configure script that eliminates the need for reinplace.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
raforg: I've just done it that way. Now it won't matter which version of python3 is selected or whether any version is selected.

security/sshdo/Portfile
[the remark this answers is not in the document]
ryandesign: Listing multiple `master_sites` is fine. It was more useful back in the aughts before we had distfile mirror servers.

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
ryandesign: Yes, MacPorts base overrides a Makefile's DESTDIR at destroot time via the `destroot.destdir` variable.

MacPorts base doesn't do anything with a PREFIX variable. The makefile portgroup does set the PREFIX environment variable. It does not by default override such a variable. The Makefile can be written such that that is not necessary, for example:

```
PREFIX ?= /usr/local
```

If the Makefile has not been written that way, then the makefile portgroup can be asked to override the PREFIX environment variable by using:

```
makefile.override-append PREFIX
```

security/sshdo/Portfile
```diff
@@ -0,0 +1,67 @@
...
+                    \n\nsshdo provides an easily configurable way of controlling which \
+                    commands may be executed via incoming ssh connections by specific \
+                    users and specific keys. Training mode allows a key to be used for \
+                    any command, while logging them. Then sshdo can learn from the logs \
+                    to know what commands to allow. It can also unlearn commands that \
+                    are no longer in use. This mitigates against private ssh key compromise \
+                    with very little effort.
+# Any of Python 2.6, 2.7, or 3.3+ will do.
+# When updating the python3 version, update the corresponding patch below as well.
+# Update the python3 version every year (or until there's a meta port:python3).
```
ryandesign, line 24: There is no plan for there ever to be a meta python3 port. The `sudo port select` mechanism is the way that users would cause a `python3` symlink to exist.

You can make it easier to update the python version each year by setting it in a variable (well, two variables):

```
set python_branch       3.11
set python_version      [string map {. {}} ${python_branch}]
```

Then you only have to change a single line to update the python version.

In the patchfile, everywhere where you currently wrote `3.11` literally, instead use a placeholder like `@PYTHON_BRANCH@`. Then you can use the above variables wherever they're needed, e.g.:

```
depends_lib-append port:python${python_version}

post-patch {
    reinplace -W ${worksrcpath} "s|@PYTHON_BRANCH@|${python_branch}|g" Makefile sshdo test_sshdo
}
```

security/sshdo/Portfile
```diff
@@ -0,0 +1,67 @@
...
+                    sha256 9b22e14fec786b692e8d6431a81e9b17ad8f6e62ae5d6ef9e9c643690bf33f90 \
+                    size 79485
+
+# Replace "python3" with the specific version depended-on above.
+# This means the user doesn't have to select a default python3 version.
+patchfiles          macports.patch
+
+# Note: Not a GNU autoconf configure script
+use_configure       yes
+configure.args      --macports-prefix=${prefix}
```
ryandesign, line 39: I'm curious why this project has a `--macports-prefix` flag. Why doesn't it use the standard `--prefix` flag that MacPorts already sets in `configure.pre_args`?

security/sshdo/Portfile
```diff
@@ -0,0 +1,67 @@
...
+build {}
```
ryandesign, line 40: Same comment regarding the disabling of the build phase here as I left for the [danectl PR](https://github.com/macports/macports-ports/pull/19124).

security/sshdo/Portfile
```diff
@@ -0,0 +1,67 @@
...
+            process == "Python" and
+            composedMessage contains "type=" and
+            composedMessage contains "user=" and
+            composedMessage contains "command="
+        ' | sed -E -e 's/Python/sshdo/' -e 's/: \([^)]+\)/:/'
+
+    You will also need to incorporate the appropriate time range
+    into the search criteria (e.g., --last 7d). See man log.
+}
+
```
ryandesign, line 67: Instead of adding newlines into paragraphs manually, please allow MacPorts to add them automatically based on the user's terminal width:

```suggestion
notes {
    sshdo relies on syslog messages to learn and unlearn policy.

    On recent macOS systems with the unified logging system,\
    syslog messages don't look the same as elsewhere, and they no\
    longer appear in /var/log/system.log where sshdo expects them,\
    so you'll need to use the log(1) command to feed log messages\
    into sshdo via its standard input (stdin). See man sshdo.

    Use the following command as the basis for your log queries:

        log show --style syslog --info --predicate '
            process == "Python" and
            composedMessage contains "type=" and
            composedMessage contains "user=" and
            composedMessage contains "command="
        ' | sed -E -e 's/Python/sshdo/' -e 's/: \([^)]+\)/:/'

    You will also need to incorporate the appropriate time range\
    into the search criteria (e.g., --last 7d). See man log.
}

```

security/sshdo/files/macports.patch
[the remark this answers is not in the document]
macportsraf: Thanks. The new-since-writing-the-above (v1.1.1) configure script sets the prefix in the Makefile to ${prefix}. The Portfile makes that happen with `configure.args --macports-prefix=${prefix}`. It also changes the prefix in some other files that need to match.

security/sshdo/Portfile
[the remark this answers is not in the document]
macportsraf: Thanks.

security/sshdo/Portfile
macportsraf: Thanks. That's much better. It's done.

security/sshdo/Portfile
macportsraf: The short answer is that the configure script isn't a standard GNU autoconf configure script. The long answer is:

It does have a --prefix argument, but that only changes the PREFIX in the Makefile which, by default, only affects where executables and manual entries are installed. It doesn't affect where configuration files are installed. That's based on the convention that I'm used to on Linux and Solaris, where the PREFIX might be e.g. /usr or /usr/local, but /etc is used for config files (i.e., not /usr/etc or /usr/local/etc). Placing /etc under the PREFIX as well seems to be more of a BSD thing.

So --prefix alone isn't suitable for macports. The configure script also has a --macports argument which does everything that's needed for macports but it assumes that the prefix is always /opt/local, which I've been informed isn't always the case. So I added --macports-prefix which is the same as --macports but with a non-default prefix supplied.

The configure script without any arguments will put config files in the right place for different systems, but I followed the Linux/Solaris convention on macOS by default, and only switch to the BSD convention with --macports or --macports-prefix.

And it's a little more complicated, because sshdo needs to read config files from under ${prefix}/etc, and one of those files needs to refer to another config file in the same directory, so it's more than just the Makefile that needs to be modified. The sshdo script itself, and the main config file also need to be changed know that /etc files are in ${prefix}/etc. If sshdo were a C program, CPPFLAGS in the Makefile could be used so that the program source itself didn't need to be modified, but it's a shell script. The configure script also changes the default source of auth.* syslog messages to /var/log/system.log (even though that's no longer used for that on new macOS systems, it is used for auth.* syslog messages on older macOS versions).

security/sshdo/Portfile
macportsraf: Thanks. I've added `supported_archs noarch`. But I haven't added `platforms any` because the installed files are different between macports/darwin and other BSD platforms (i.e., references in the installed files to /usr/local on FreeBSD and OpenBSD and references to /usr/pkg on NetBSD, and the references to the python executable. I think the default `darwin` platform is correct here. Does that sound right?

security/sshdo/Portfile
macportsraf: Thanks. It's done.

macportsraf:
Hi, It's been a couple of months since the last activity. I'm under the impression that I've made all of the required changes. Have I missed something? Is there still anything wrong with this Portfile?

security/sshdo/Portfile
ryandesign: Are there any differences in what the port installs for the different macOS versions? If not, then you should use:
```
platforms {darwin any}
```

security/sshdo/Portfile
ryandesign: We have over 36,000 ports in MacPorts. Not all of them use configure scripts, of course, but I can't recall any where setting just `--prefix` didn't also use that value to set the sysconfdir.

I recommend you strive to eliminate anything MacPorts specific from this software. Anything you need to do for MacPorts, you probably also need to do for some other package managers or some other subset of users, so find a generic way to make those features available in your configure script or build system. For example, if you need to use a different logfile on macOS, you can have your configure script do so automatically when building for Darwin—in fact it looks like your configure script already does that. And it looks like your configure script already supports an `--etcdir` flag, so that can be used in combination with `--prefix` without any need for a separate `--macports-prefix` option.

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,38 @@
+--- Makefile.orig	2023-06-19 15:23:20.000000000 +1000
++++ Makefile	2023-06-19 15:58:11.000000000 +1000
+@@ -58,8 +58,8 @@
+ 	@echo "make diff      - Show differences between source and installed versions"
+ 
+ test:
+-	@[ -z "`which python3 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3$$/env python/' < ./sshdo > ./sshdo2 && chmod 755 ./sshdo2 && mv ./sshdo2 ./sshdo || true
+-	@[ -z "`which python3 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3$$/env python/' < ./test_sshdo > ./test_sshdo2 && chmod 755 ./test_sshdo2 && mv ./test_sshdo2 ./test_sshdo || true
++	@[ -z "`which python@PYTHON_BRANCH@ 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3\.11$$/env python/' < ./sshdo > ./sshdo2 && chmod 755 ./sshdo2 && mv ./sshdo2 ./sshdo || true
++	@[ -z "`which python@PYTHON_BRANCH@ 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3\.11$$/env python/' < ./test_sshdo > ./test_sshdo2 && chmod 755 ./test_sshdo2 && mv ./test_sshdo2 ./test_sshdo || true
```
ryandesign, line 10: Two hardcoded references to `3\.11` remain here (and another below).

You shouldn't need to run `which` on the python binary to see if it exists; you can be certain it exists because the port declares a dependency on it.

`which` isn't the best way to determine if a program exists; `command -v` is.

It looks like the original code in the Makefile is trying to do this: if the user does not have `python3`, but does have `python`, then it replaces `env python3` with `env python` in the sshdo and test_sshdo scripts. For MacPorts, we would always want to replace `/usr/bin/env python3` with the absolute path of the MacPorts python that we're using. For example (I have not tested this):

```suggestion
+	@sed '1s,#!/usr/bin/env python3$$,#!@PREFIX@/bin/python@PYTHON_BRANCH@,' < ./sshdo > ./sshdo2 && chmod 755 ./sshdo2 && mv ./sshdo2 ./sshdo || true
+	@sed '1s,#!/usr/bin/env python3$$,#!@PREFIX@/bin/python@PYTHON_BRANCH@,' < ./test_sshdo > ./test_sshdo2 && chmod 755 ./test_sshdo2 && mv ./test_sshdo2 ./test_sshdo || true
```

However, you're also patching sshdo and test_sshdo later in the patchfile; only one or the other is needed, and patching the shebang line with placeholders and reinplacing them in the portfile is the usual method. In that case:

```suggestion
```

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,38 @@
...
+ 	@./test_sshdo
+ 
+ check: test
+@@ -69,7 +69,7 @@
+ 	if [ ! -f $(DEST_ETCDIR)/$(SSHDOERS) ]; then cp $(SSHDOERS) $(DEST_ETCDIR); chmod 644 $(DEST_ETCDIR)/$(SSHDOERS); fi
+ 	if [ ! -f $(DEST_ETCDIR)/$(SSHDO_BANNER) ]; then cp $(SSHDO_BANNER) $(DEST_ETCDIR); chmod 644 $(DEST_ETCDIR)/$(SSHDO_BANNER); fi
+ 	cp $(BIN) $(DEST_BINDIR); chmod 755 $(DEST_BINDIR)/$(BIN)
```
ryandesign, line 17: The `install` program exists to copy files and set their permissions, so probably most instances of `cp` (with or without `chmod`) in the install target could be replaced with `install` with the appropriate flags.

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,38 @@
...
+-	[ -z "`which python3 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3$$/env python/' < $(BIN) > $(DEST_BINDIR)/$(BIN) || true
++	[ -z "`which python@PYTHON_BRANCH@ 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3\.11$$/env python/' < $(BIN) > $(DEST_BINDIR)/$(BIN) || true
+ 	[ -d $(DEST_MANDIR)/man8 ] || mkdir -m 755 $(DEST_MANDIR)/man8
+ 	[ -d $(DEST_MANDIR)/man5 ] || mkdir -m 755 $(DEST_MANDIR)/man5
+ 	cp sshdo.8 $(DEST_MANDIR)/man8; chmod 644 $(DEST_MANDIR)/man8/sshdo.8
+--- sshdo.orig	2023-06-19 15:23:20.000000000 +1000
++++ sshdo	2023-06-19 15:56:00.000000000 +1000
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python3
++#!/usr/bin/env python@PYTHON_BRANCH@
```
ryandesign, line 27:
```suggestion
+#!@PREFIX@/bin/python@PYTHON_BRANCH@
```

The `reinplace` in the Portfile would have to be expanded to also handle the `@PREFIX@` placeholder, e.g. (untested):

```tcl
post-patch {
     reinplace -W ${worksrcpath} "s|@PYTHON_BRANCH@|${python_branch}|g" Makefile
     reinplace -W ${worksrcpath} "s|@PYTHON_BRANCH@|${python_branch}|g;s|@PREFIX@|${prefix}|g" sshdo test_sshdo
 }
```

security/sshdo/files/macports.patch
```diff
@@ -0,0 +1,38 @@
...
++	@[ -z "`which python@PYTHON_BRANCH@ 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3\.11$$/env python/' < ./test_sshdo > ./test_sshdo2 && chmod 755 ./test_sshdo2 && mv ./test_sshdo2 ./test_sshdo || true
+ 	@./test_sshdo
+ 
+ check: test
+@@ -69,7 +69,7 @@
+ 	if [ ! -f $(DEST_ETCDIR)/$(SSHDOERS) ]; then cp $(SSHDOERS) $(DEST_ETCDIR); chmod 644 $(DEST_ETCDIR)/$(SSHDOERS); fi
+ 	if [ ! -f $(DEST_ETCDIR)/$(SSHDO_BANNER) ]; then cp $(SSHDO_BANNER) $(DEST_ETCDIR); chmod 644 $(DEST_ETCDIR)/$(SSHDO_BANNER); fi
+ 	cp $(BIN) $(DEST_BINDIR); chmod 755 $(DEST_BINDIR)/$(BIN)
+-	[ -z "`which python3 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3$$/env python/' < $(BIN) > $(DEST_BINDIR)/$(BIN) || true
++	[ -z "`which python@PYTHON_BRANCH@ 2>/dev/null | grep '^/'`" -a -n "`which python 2>/dev/null | grep '^/'`" ] && sed 's/env python3\.11$$/env python/' < $(BIN) > $(DEST_BINDIR)/$(BIN) || true
```
ryandesign, line 19: It occurs to me that it would be better to do this in the configure script. Give the configure script a flag that lets the user specify what python they want to use. If they don't specify one, then you can guess based on whether they have `python3` or `python`, and you can complain with a good error message if they don't have any. Then the configure script makes the adjustments in the other files.

Typically, files that need adjustment use names that end with `.in`, e.g. sshdo.in. The configure script would then use `sed` to convert sshdo.in to sshdo, replacing any placeholders along the way.

security/sshdo/Portfile
macportsraf: We don't need to use the `--macports-prefix` option. That's just there for convenience. I've changed it to use the equivalent set of other options explicitly:

    --prefix="${prefix}" --etcdir="${prefix}"/etc --enable-mangz

security/sshdo/Portfile
macportsraf: Thanks.

There are no differences between different versions of macOS, just differences between
different platforms. I've changed it to `{darwin any}`.

I had remembered this from the macports guide:

	The special value any can also be used to indicate that a port will
	install identical files across platforms or platform versions

But not this:

    Ports that install identical files on any Darwin version, but may
	install different files on other platforms, should use:
    platforms           {darwin any}

security/sshdo/files/macports.patch
macportsraf: Actually, those rewrites can completely disappear for macports because it is known that python311 will be present. I've removed them.

security/sshdo/files/macports.patch
macportsraf: That's true for some (admittedly most) operating systems. But on Solaris, the install program is very different. That's why I tend to use mkdir/cp/chown. It's more portable.

Is that OK, or do you need me to patch it?

security/sshdo/files/macports.patch
macportsraf: It's done.

security/sshdo/files/macports.patch
macportsraf: These lines have been deleted (see above).

macportsraf: I've added a `--with-python=` argument to configure, and there's no patch now.

PR #19109, opened 2023-06-15
clang_dependency PG: switch to clang-11-bootstrap

catap:
#### Description

Instead of blacklisting all clangs I've enforced to use clang-11-bootstrap on all systems which ships with something older which can't be used by system to build port which is kept inside clang bootstrap cycle.

Anyway, I've replaced compiler only if desire one isn't available and if desired was MacPorts clang in accepted list of versions.

This dramatically decrease number of compilers which needs to be build on old system.

Thus, clang-11-bootstrap contains all necessary patches to be able to build LLVM/Clang up to 15 (tested), and probably 16 (untested).

P.S. I've used this approach on my test machines for almost a year, and seems that it works very well.

###### Type(s)

- [x] enhancement

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

macOS 10.6.8 10K549 x86_64
Xcode 3.2.6 10M2518

macOS 10.7.5 11G63 x86_64
Xcode 4.6.3 4H1503

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/group/clang_dependency-1.0.tcl
+++ b/_resources/port1.0/group/clang_dependency-1.0.tcl
@@ -1,14 +1,45 @@
 # -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
 
 # This portgroup is to help prevent circular dependencies for ports that
-# recent clang builds depend on, by blacklisting all 5+ clang compilers.
+# recent clang builds depend on, switching it to clang-11-bootstrap
 
-# call this if an older clang version depends on the port
-proc clang_dependency.extra_versions {versions} {
-    global prefix
-    foreach ver $versions {
-        compiler.blacklist-append macports-clang-${ver}
+proc clang_dependency.clang_bootstrap {} {
+    global os.platform os.major prefix configure.compiler
+
+    # When MacPorts decided to use MacPorts clang replace it to
+    # clang-11-bootstrap when it possible, to break circular dependencies.
+    if {[regexp {macports-clang-(.*)} ${configure.compiler} -> clang_v]} {
+
+        # detects when decided clang is already installed
+        if {![catch {lindex [registry_active clang-${clang_v}] 0}]} {
+            ui_debug "${configure.compiler} is installed, do not replace by clang-11-bootstrap"
+            return
+        }
+
+        # clang-11-bootstrap can build clang up to 15,
+        # future clang / llvm may requires future patches.
+        if {[vercmp ${clang_v} 15] >= 0} {
+            ui_debug "clang-11-bootstrap can't replace ${configure.compiler}"
+            return
+        }
+
+        ui_debug "Replace ${configure.compiler} by clang-11-bootstrap"
+
+        depends_build-delete    port:clang-${clang_v}
+        depends_build-append    port:clang-11-bootstrap
+        depends_skip_archcheck-append \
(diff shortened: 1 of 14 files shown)


catap:
@mascguy may I ask you that should I do with it?

catap:
@mascguy, @herbygillot  may I ask what should I do with it?

herbygillot:
It seems reasonable. I guess let's rebase on top of latest master to start.

@kencu, any comments?

kencu:
this was always my background plan with clang11-bootstrap, so it's on the path I had in mind.

There are certain to be errors and issues discovered, I can basically 100% guarantee that. How upset @jmroot and @ryandesign might be about these inevitable issues that will crop up is uncertain. It would be best for at least one of them to buy in and approve this before pushing it, to avoid later disagreements about it.

I see that in cases this PR would build with an installed clang compiler if it exists.

```
# detects when decided clang is already installed
         if {![catch {lindex [registry_active clang-${clang_v}] 0}]} {
             ui_debug "${configure.compiler} is installed, do not replace by clang-11-bootstrap"
             return
         }
```

this has, in the past, always turned out to be a bad idea. It causes different things to build different ways on different people's systems, for one thing. It also can lead to the chosen compiler being broken, even though it is installed, and resulting in build errors that way. So I would not bother getting too fancy doing that, and just use clang11-bootstrap all the time, if that is what is going to be used.

Integrating this into a bootstrap to a newer cctools and ld64 and libtapi than what the system tools can build on older systems would be next -- that probably involves building a 'throwaway' interim set of tools on each system that is built by the system tools, and used by clang11-bootstrap to build the current cctools and ld64 and llvm and libtapi that will be installed.

catap:
@kencu, thanks for support. Quoted part of code was added to make everyone less nervous, and I'd love to avoid that... 

So, let make build consistent.

@herbygillot sure, I just need a few days of rebuilding after rebase to confirm that everything seems ok :)

herbygillot:
@kencu, thank you for taking the time to weigh in.

catap:
> Integrating this into a bootstrap to a newer cctools and ld64 and libtapi than what the system tools can build on older systems would be next -- that probably involves building a 'throwaway' interim set of tools on each system that is built by the system tools, and used by clang11-bootstrap to build the current cctools and ld64 and llvm and libtapi that will be installed.

We may start from use clang-11-bootstrap as LLVM for cctools instead of a lot of different llvm per OS it should simplify things to be honest.

catap:
FYI: @Gcenx I'll try to push an updated version of this PR in weeks. Need more time to test.

pmetzger:
@catap Howdy! What's needed to get this unstuck?

catap:
@pmetzger a week or two of my time to rebase it and retest it 🫣

pmetzger:
@catap Do you want us to keep it open?

catap:
@pmetzger yes, it like a todo for me

pmetzger:
There is a price for that. Just today I found a reasonably committable patch that no one had responded to in six months because the queue is a giant mess. I wonder if that person will ever work on MacPorts again because they were ignored for so long.

catap:
@pmetzger yeah, I see that you mean.

One of the reason why I'm using MacPorts and not nix for example is that here PRs are merged. Eventually.

Unfortunately to many peoples take this things too personal and seems to forgot that we're here to make fun :)

pmetzger:
@catap Just a poke. This is nearly the oldest remaining PR in the system.

pmetzger:
I'm going to close this for now. At the point at which someone wants to work on it, it can be re-opened.

PR #19124, opened 2023-06-17
danectl: new port (version 0.8.4)

macportsraf:
#### Description

DNSSEC DANE implementation manager

DANE is "DNS-based Authentication of Named Entities".  It means securely letting the world know in advance what  your public encryption keys are by publishing them as  DNS records (TLSA, SSHFP, OPENPGPKEY, SMIMEA) in your  DNSSEC-enabled internet domain zone. This is the simplest  and most secure way to let the world know what keys to expect when connecting to your servers. This can apply to  TLS keys, SSH host keys, and OpenPGP and S/MIME keys. This makes it possible to prevent impersonation or  man-in-the-middle attacks. It's mostly used with mail  servers. Eventually, it could render certificate  authorities unnecessary. DNSSEC has become very easy.

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G6032 x86_64
Xcode 11.3 11C29

###### Verification

### DIFF
--- a/security/danectl/Portfile
+++ b/security/danectl/Portfile
@@ -1,0 +1,89 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem       1.0
+PortGroup        makefile 1.0
+
+name             danectl
+version          0.8.4
+revision         0
+
+categories       security
+license          GPL-2+
+maintainers      {raf.org:raf @macportsraf} openmaintainer
+description      DNSSEC DANE implementation manager
+long_description {*}${description}. \
+                 \n\nDANE is \"DNS-based Authentication of Named Entities\". \
+                 It means securely letting the world know in advance what \
+                 your public encryption keys are by publishing them as \
+                 DNS records (TLSA, SSHFP, OPENPGPKEY, SMIMEA) in your \
+                 DNSSEC-enabled internet domain zone. This is the simplest \
+                 and most secure way to let the world know what keys to \
+                 expect when connecting to your servers. This can apply to \
+                 TLS keys, SSH host keys, and OpenPGP and S/MIME keys. \
+                 This makes it possible to prevent impersonation or \
+                 man-in-the-middle attacks. It's mostly used with mail \
+                 servers. Eventually, it could render certificate \
+                 authorities unnecessary. DNSSEC has become very easy.
+depends_lib      port:libidn2
+supported_archs  noarch
+platforms        any
+
+default_variants +tlsa +sshfp +openpgpkey +smimea
+
+variant tlsa description {Add TLSA support for TLS servers} {
+    depends_run-append \
+        port:certbot \
+        port:openssl \
+        port:coreutils
+}
+
+variant sshfp description {Add SSHFP support for ssh servers} {
+    depends_run-append \
+        port:openssh
+}
+
+variant openpgpkey description {Add OPENPGPKEY support for GnuPG keys} {
+    depends_run-append \
+        port:gnupg2
+}
+
+variant smimea description {Add SMIMEA support for S/MIME keys} {
+    # Suppress a port lint warning about port:openssl appearing twice
+    if {![variant_isset tlsa]} {
+        depends_run-append \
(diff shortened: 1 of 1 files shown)


security/danectl/Portfile
```diff
@@ -0,0 +1,87 @@
...
+    # Certbot is definitely a macports port (tlsa variant dependency)
+    reinplace -W ${worksrcpath} "s|/etc/letsencrypt|${prefix}/etc/letsencrypt|g" danectl danectl.1
+    # Assume apache2 or nginx would also be macports ports, if present
+    reinplace -W ${worksrcpath} "s|/etc/apache2|${prefix}/etc/apache2|g" danectl danectl.1
+    reinplace -W ${worksrcpath} "s|/etc/nginx|${prefix}/etc/nginx|g" danectl danectl.1
+    # Don't require users to put ${prefix}/libexec/gnubin in their $PATH
+    reinplace -W ${worksrcpath} "s|sha256sum|gsha256sum|g" danectl danectl.1
+}
+
+use_configure       no
```
ryandesign, line 79: The makefile portgroup sets this for you.
```suggestion
```

security/danectl/Portfile
```diff
@@ -0,0 +1,87 @@
...
+build {}
```
ryandesign, line 80: Does this port only install scripts and other files that are not compiled for a specific processor? If so, indicate that with:
```
supported_archs noarch
```

Assuming that is the case, do any of the installed files vary based on the OS version? If not, indicate that with:

```
platforms {darwin any}
```

security/danectl/Portfile
rafmod: Thanks. It's removed.

security/danectl/Portfile
rafmod: Thanks. That's correct. It's a shell script. It handles various operating systems without the files themselves changing. I've added the supported_archs and platforms directives.

security/danectl/Portfile
macportsraf: Thanks. It's removed.

security/danectl/Portfile
macportsraf: Thanks. That's correct. It's a shell script. It handles various operating systems without the files themselves changing. I've added the supported_archs and platforms directives.

security/danectl/Portfile
```diff
@@ -0,0 +1,89 @@
...
+    if {![string match /usr/local ${prefix}]} {
+        reinplace -W ${worksrcpath} "s|/usr/local|${prefix}|g" Makefile
+    }
+    # Certbot is definitely a macports port (tlsa variant dependency)
+    reinplace -W ${worksrcpath} "s|/etc/letsencrypt|${prefix}/etc/letsencrypt|g" danectl danectl.1
+    # Assume apache2 or nginx would also be macports ports, if present
+    reinplace -W ${worksrcpath} "s|/etc/apache2|${prefix}/etc/apache2|g" danectl danectl.1
+    reinplace -W ${worksrcpath} "s|/etc/nginx|${prefix}/etc/nginx|g" danectl danectl.1
+    # Don't require users to put ${prefix}/libexec/gnubin in their $PATH
+    reinplace -W ${worksrcpath} "s|sha256sum|gsha256sum|g" danectl danectl.1
```
mascguy, line 79: While this is a personal preference - and port authors are free to use whatever style they prefer - it might be good to wrap these `reinplace` statements in a `foreach` block. (Since they're all modifying the same pair of files.)

```
foreach f {danectl danectl.1} {
    # Certbot is definitely a macports port (tlsa variant dependency)
    reinplace -W ${worksrcpath} "s|/etc/letsencrypt|${prefix}/etc/letsencrypt|g" ${f}
    # Assume apache2 or nginx would also be macports ports, if present
    reinplace -W ${worksrcpath} "s|/etc/apache2|${prefix}/etc/apache2|g" ${f}
    reinplace -W ${worksrcpath} "s|/etc/nginx|${prefix}/etc/nginx|g" ${f}
    # Don't require users to put ${prefix}/libexec/gnubin in their $PATH
    reinplace -W ${worksrcpath} "s|sha256sum|gsha256sum|g" ${f}
}
```

But I'm fine with it as-is, so no need to change it unless you're interested.

mascguy (approved):

PR #19146, opened 2023-06-18
gnutls: reduce default install footprint, via new variant docs

mascguy:
### Description
- Reduce default install footprint, via new variant docs
- Manpages still installed by default
- Supplemental docs are developer-focused, and require gtk-doc; latter pulls in significant number of deps

### Change Details
- For default installation (without `+docs`), file count reduced from 1275 to 62

### DIFF
--- a/devel/gnutls/Portfile
+++ b/devel/gnutls/Portfile
@@ -10,7 +10,7 @@ PortGroup       muniversal 1.0
 
 name            gnutls
 version         3.7.9
-revision        2
+revision        3
 categories      devel security
 # yes, some of the libs are GPL only
 license         LGPL-2.1+ GPL-3+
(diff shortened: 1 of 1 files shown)


Schamschula:
`gnutls-devel` also needs a revision bump!

Unfortunately, contrary to what the description and  your change to the `configure.args` might imply, no man pages are installed w/o the `+docs` variant.

It looks like the only intended change is it that the info pages should no longer be generated. However, in order to get the man pages you will also get the info pages.

Schamschula:
OK. I do see the `man1` pages.

mascguy:
> OK. I do see the `man1` pages.

Yep, the additional config argument enables those regardless. Also, just pushed an update to rev-bump `gnutls-devel` too.

Schamschula:
It seems the config arguments are a little ambiguous: When you run `./configure --help` it doesn't tell you that it will only install the `man1` pages, rather than all the man pages.

However, I do see your point in terms of install space.

PR #19158, opened 2023-06-19
graphite2: use py311

mascguy:
### Description

- Migrate port to py311; no other changes

### DIFF
--- a/graphics/graphite2/Portfile
+++ b/graphics/graphite2/Portfile
@@ -31,7 +31,7 @@ extract.suffix      .tgz
 compiler.cxx_standard \
                     2011
 
-set py_ver          3.10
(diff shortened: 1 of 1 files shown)


ryandesign:
Thank you! It looks like python is only a test dependency though, so the revision shouldn't need to be increased?

mascguy:
Meanwhile, it looks like my latest push included changes to characters in the description/long_description. But that wasn't intentional.

Will push a fix momentarily...

mascguy:
> Meanwhile, it looks like my latest push included changes to characters in the description/long_description. But that wasn't intentional.

I have absolutely no idea how that happened, though perhaps I need to check my Vim settings. Fixed now though.

PR #19207, opened 2023-06-23
libsdl2: patch to revert upstream fix for sdl issue 6948

mascguy:
### Description

- Fixes builds for 10.7/10.8
- Closes: https://trac.macports.org/ticket/67666


mascguy:
Josh, this is just a quick-and-dirty revert. But I suppose a better approach would be to re-add the macOS version checks...?

Up to you, let me know how you'd like to handle this.

mascguy:
Maintainer timeout, merging.

jmroot:
> Josh, this is just a quick-and-dirty revert. But I suppose a better approach would be to re-add the macOS version checks [with the new code]...?

Yes, it would be preferable to put together something that could be accepted upstream.

PR #19234, opened 2023-06-25
libcerf: fix build on powerpc

barracuda156:
#### Description

Upstream recommends at least gcc-4.3, and since Intel now uses Macports clang on 10.6, this only affects PPC. Also, old Apple compilers do not support `native`.
Then, CMakeLists set wrong optflags which break the build. Fix it.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/math/libcerf/files/patch-powerpc.diff
+++ b/math/libcerf/files/patch-powerpc.diff
@@ -1,0 +1,15 @@
+--- CMakeLists.txt.orig	2023-01-10 16:10:08.000000000 +0800
++++ CMakeLists.txt	2023-06-26 02:28:37.000000000 +0800
+@@ -76,7 +76,11 @@
+     if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "GNU" )
+         option(PORTABLE "Build a portable binary without host-specific optimization" OFF)
+         if(NOT PORTABLE)
+-            add_compile_options(-march=native)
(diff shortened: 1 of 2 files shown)


barracuda156:
@mojca If you are in touch with upstream, maybe you could pass the patch for CMakeLists? It is relevant for any PPC, not just Apple. (_Perhaps_ for non-Mac *BSDs and Linux it should be "ppc|ppc64|powerpc|powerpc64" for `CMAKE_SYSTEM_PROCESSOR` or similar.)
Their repo requires some obscure registration method even to open an issue.

math/libcerf/files/patch-powerpc.diff
```diff
@@ -0,0 +1,15 @@
+--- CMakeLists.txt.orig	2023-01-10 16:10:08.000000000 +0800
++++ CMakeLists.txt	2023-06-26 02:28:37.000000000 +0800
+@@ -76,7 +76,11 @@
+     if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_C_COMPILER_ID STREQUAL "GNU" )
+         option(PORTABLE "Build a portable binary without host-specific optimization" OFF)
+         if(NOT PORTABLE)
```
catap, line 6: @barracuda156 I suggest to switch on `PORTABLE` here, otherwice it an artifact from build bot might not work on user's machine.

barracuda156: UPD. In fact no need, since that option is only used for GCC, and Macports has no buildbots using GCC:
https://jugit.fz-juelich.de/mlz/libcerf/-/blob/main/CMakeLists.txt#L76

barracuda156:
@mojca @reneeotten Bump

PR #19260, opened 2023-06-28
socnetv: new port

szhorvat:
#### Description

New port: SocNetV

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 13.4.1 22F82 x86_64
Xcode 14.3.1 14E300c

###### Verification 
Have you

### DIFF
--- a/science/socnetv/Portfile
+++ b/science/socnetv/Portfile
@@ -1,0 +1,35 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           qt6 1.0
+
+name                socnetv
+github.setup        socnetv app 3.1 v
+revision            0
+
+categories          science
+license             GPL-3+
+maintainers         nomaintainer
+
+description         Social Network Visualizer
+long_description    Social Network Visualizer (SocNetV) is a cross-platform, \
+                    user-friendly free graphical software application for social \
+                    network analysis and visualization.
+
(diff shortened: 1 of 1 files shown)


szhorvat:
CC @MarcusCalhoun-Lopez 

While this compiles and works on my system, I am not confident that it will build fine in all environments. Making sure that it does would effectively require replicating what a qmake6 port group would do. Hence this request: https://trac.macports.org/ticket/67697

science/socnetv/Portfile
```diff
@@ -0,0 +1,36 @@
...
+PortSystem          1.0
+PortGroup           github 1.0
+PortGroup           qt6 1.0
+
+name                socnetv
+github.setup        socnetv app 3.1 v
+revision            0
+
+categories          science
+platforms           darwin
```
herbygillot, line 12: Looks good. `platforms darwin` is now implicit in MacPorts, so this line can be dropped.

szhorvat:
@herbygillot To be explicit about what's missing: 

 - There's a simply replacement of the `--prefix` configure argument to make it suitable for qmake, but other arguments are not handled/transformed. I don't know under what circumstances MacPorts might pass other arguments.
 - No effort is made to respect the selected compiler, or to make sure that `ccache` is actually used when selected.

All these are handled by the `qmake` (Qt 4) and `qmake5` (Qt 5) port groups. I don't have the expertise to re-implement all this for Qt 6.

science/socnetv/Portfile
szhorvat: I have now removed this line.

herbygillot:
I think that's fine for now.  I don't even think `Qt6` is compiling on every single macOS version.  This looks good as it is now, and is building successfully in CI, so we'll merge this for now.

You can commit fixes or improvements down the line as it becomes necessary to do so.

PR #19272, opened 2023-06-30
uacme: new port

sideeffect42:
#### Description

A new port for [uacme](https://github.com/ndilieto/uacme).

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 12.6.6 21G646 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/security/uacme/Portfile
+++ b/security/uacme/Portfile
@@ -1,0 +1,56 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        ndilieto uacme 1.7.4 v
+revision            0
+categories          security
+license             GPL-3+
+platforms           any
+maintainers         nomaintainer
+description         ACMEv2 client written in plain C with minimal dependencies
+long_description    lightweight client for the RFC8555 ACMEv2 protocol, \
+                    written in plain C with minimal dependencies \
+                    (libcurl and OpenSSL).
+
+checksums           rmd160  5d6bb1001849f28ae5da01913c2ddce9cb20a502 \
+                    sha256  5dd8ecd6411b1b18548c2660b05acd273e0878c60d8762c412aea56e9de2f727 \
+                    size    257908
+
+depends_build-append \
+                    port:asciidoc
+# dependencies of autoreconf
+depends_build-append \
+                    port:pkgconfig \
+                    port:autoconf \
+                    port:automake \
+
+depends_lib-append  port:curl \
+                    port:libev \
+                    port:openssl
+
+patchfiles-append   autoconf.patch
+
+configure.args      --with-libcurl=${prefix} \
+                    --with-openssl=${prefix} \
+                    --without-mbedtls \
+                    --without-gnutls
+
+compiler.c_standard 1999
+
+platform darwin {
+    # ualpn does not build on Darwin because Darwin does not implement unnamed
+    # semaphores
+    # https://github.com/ndilieto/uacme/issues/49
+    configure.args-append \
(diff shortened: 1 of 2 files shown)


security/uacme/Portfile
```diff
@@ -0,0 +1,55 @@
...
+license             GPL-3+
+platforms           any
+maintainers         nomaintainer
+description         ACMEv2 client written in plain C with minimal dependencies
+long_description    lightweight client for the RFC8555 ACMEv2 protocol, \
+                    written in plain C with minimal dependencies \
+                    (libcurl and OpenSSL).
+homepage            https://github.com/ndilieto/uacme
+
+github.setup        ndilieto uacme 1.7.4 v
```
herbygillot, line 17: `github.setup` implicitly sets `name`, `version`, and `homepage`.  You can move `github.setup` to the top of the line, and delete the `name`, `version` and `homepage` lines.

security/uacme/Portfile
```diff
@@ -0,0 +1,55 @@
...
+
+checksums           rmd160  5d6bb1001849f28ae5da01913c2ddce9cb20a502 \
+                    sha256  5dd8ecd6411b1b18548c2660b05acd273e0878c60d8762c412aea56e9de2f727 \
+                    size    257908
+
+depends_build       port:pkgconfig \
```
herbygillot, line 23: This would be better as `depends_build-append`

security/uacme/Portfile
```diff
@@ -0,0 +1,55 @@
...
+                    port:autoconf \
+                    port:automake \
+                    port:asciidoc
+
+depends_lib         port:curl \
```
herbygillot, line 28: Would be better as `depends_lib-append`

security/uacme/Portfile
sideeffect42: @herbygillot Can you explain what is better with `depends_x-append`?

herbygillot: In a Portfile, certain statements like `depends_build`, `depends_libs`, and `patchfiles` are arrays. They are a list of values.

When you import a portgroup at the beginning of your Portfile, it may make some implicit additions to any of these.

So by using `...-append`, you ensure that you are never overwriting anything by mistake.

sideeffect42: Thanks for the explanation. So should I also use `patchfiles-append` instead?

herbygillot: You don't _have_ to, but it's good practice as well.

security/uacme/Portfile
```diff
@@ -0,0 +1,54 @@
+PortSystem          1.0
```
herbygillot, line 1: The Portfile is missing the modeline - that should be the first line in every Portfile:

```

PortSystem          1.0
```

That's used by editors to set the right formatting and spacing for Portfiles.

sideeffect42:
I was just testig this port on 10.5.8 and noticed that with `port -vst install` I get this error by `autoreconf`:
```
autom4te: error: need GNU m4 1.4 or later: /opt/local/bin/gm4
```

Dropping `-t` makes `autoreconf` finish. 
@herbygillot Do you by chance know if this is a bug of this port?

herbygillot:
> I was just testig this port on 10.5.8 and noticed that with `port -vst install` I get this error by `autoreconf`:
> 
> ```
> autom4te: error: need GNU m4 1.4 or later: /opt/local/bin/gm4
> ```
> 
> Dropping `-t` makes `autoreconf` finish. @herbygillot Do you by chance know if this is a bug of this port?

That's entirely fine.  This is more an issue with MacPorts rather than your port.  `-t` is trace mode, which in my experience has not worked properly in a consistent way across various macOS versions.

herbygillot:
Looks good and merged. Thank you for your contribution and quick turnaround, @sideeffect42.

sideeffect42:
@herbygillot Thank you for your review and quickly merging.

PR #19275, opened 2023-06-30
sane-backends: add snmp variant

sideeffect42:
#### Description

On Mac OS X 10.5.8 the build of sane-backends currently fails because net-snmp does not build on it.

As I don't need SNMP auto-detection of my scanners and I can imagine few people do, I think it makes sense to add a variant to enable SNMP when needed.

While at it I also restore the C++ standard version when the `no_local` variant is used.
As far as I know a C++11 compatible compiler is only required for the genesys backend.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

Mac OS X 10.5.8
Xcode 3.1.4

###### Verification 
Have you

### DIFF
--- a/graphics/sane-backends/Portfile
+++ b/graphics/sane-backends/Portfile
@@ -4,7 +4,7 @@ PortSystem                  1.0
 
 name                        sane-backends
 version                     1.2.1
-revision                    0
+revision                    1
 set upload_tag              110fc43336d0fb5e514f1fdc7360dd87
 categories                  graphics
 platforms                   darwin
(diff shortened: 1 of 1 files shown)


pmetzger:
Sorry no one has replied before now. I think you should reverse this; for better or worse, what you're doing now changes the default behavior (and without bumping the package version number). Adding a variant that removes the snmp behavior is harmless, silently changing the behavior of the existing package is not.

sideeffect42:
@pmetzger I see your point. Would making `snmp` a default variant solve this issue?

Alternatively, of course, `snmp` could be converted to `no_snmp` like `no_local`, however I find the inverted variants quite confusing, TBH.

pmetzger:
LGTM, so I'm merging. Thanks, @sideeffect42!

PR #19283, opened 2023-07-01
qemu: update to 8.0.2

herbygillot:
- support numerous image formats (qcow1, parallels, vdi, etc.) by default
- build with compression formats (lzo, snappy, lzfse & zstd) by default
- build with slirp support
- use Python 3.11 Sphinx to build docs
- disable linking against the jack audio library

Fixes: https://trac.macports.org/ticket/67409
Fixes: https://trac.macports.org/ticket/64246
See: https://trac.macports.org/ticket/67549

###### Verification 
Have you

### DIFF
--- a/emulators/qemu/Portfile
+++ b/emulators/qemu/Portfile
@@ -8,8 +8,8 @@ PortGroup legacysupport 1.1
 legacysupport.newest_darwin_requires_legacy 16
 
 name                    qemu
-version                 7.1.0
-revision                1
+version                 8.0.2
+revision                0
 categories              emulators
 license                 GPL-2+
 maintainers             {raimue @raimue} \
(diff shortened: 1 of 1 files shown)


emulators/qemu/Portfile
```diff
@@ -169,7 +181,7 @@ pre-configure {
 # disable silent rules
 build.args-append       V=1
 
-default_variants        +usb
+default_variants        +ssh +usb
```
raimue, line 184: Why? I do not think the average user will ever need this...

Another reason libssh was in a variant was the OpenSSL dependency that prevented binary archives. But I guess with openssl3 this shouldn't be an issue any longer.

emulators/qemu/Portfile
```diff
@@ -265,11 +272,6 @@ variant snappy description {Support Snappy compression} {
     depends_lib-append      port:snappy
 }
 
-variant zstd description {Support zstd compression} {
-    configure.args-replace  --disable-zstd --enable-zstd
-    depends_lib-append      port:zstd
-}
-
```
raimue, line 272: I agree that zstd is seeing more widespread use and I am fine with adding zstd as a new default as that is more likely to be already installed anway.

However, I have not encountered any image compressed with lzo yet. Is it really that common? Switching to port:lzo2 seems reasonable, though. But I do not see lzo to be any more important than lzfse or snappy and we keep them as variants.

raimue:
Thank you for the update and making the switch to libslirp, which was previously holding back an update. Using python3.11 is also a good choice by now. :+1:

emulators/qemu/Portfile
herbygillot: I think that ideally we should have all the compression formats as built-in.  `qemu` no longer builds with `lzo`, but now requires `lzo2`.

I think maybe it makes sense to add `snappy` to this list as well, and that should be fine.

emulators/qemu/Portfile
herbygillot: I guess you're right: reverted.

herbygillot:
@raimue This final version enables all the compression formats by default (snappy, lzo, snappy & zstd), and also enables a few more image formats by default.  I've removed libssh from the default variants.

If this looks good, we'll merge. Thanks.

emulators/qemu/Portfile
raimue: I still doubt that the average user will ever need lzfse or snappy with qemu. That adds a few more dependencies, although these are not that big and are only updated infrequently. But okay, to make this simpler, let's include all compression formats by default.

raimue (approved):

raimue:
Thank you once again for the port update and for resolving the review comments. Keep up the good work. :+1:

PR #19327, opened 2023-07-03
Add btop port

Samasaur1:
As the comment in the Portfile notes, this port currently uses the latest commit and not the latest tagged version, because there has not been a tagged version that can be built with Clang and libc++20. Upon the release of the next tagged version, I will update this Portfile

#### Description

Add [`btop`](https://github.com/aristocratos/btop) port.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 13.4.1 22F82 arm64
Xcode 14.3.1 14E300c

###### Verification 
Have you

### DIFF
--- a/sysutils/btop/Portfile
+++ b/sysutils/btop/Portfile
@@ -2,6 +2,8 @@
 
 PortSystem          1.0
 PortGroup           github 1.0
+PortGroup           makefile 1.0
+PortGroup compiler_blacklist_versions 1.0
 
 github.setup        aristocratos btop 7e50b03e391a346b2b6ce92be332c58d91f75a00
 version             20230615
@@ -20,10 +22,10 @@ categories          sysutils
 license             Apache-2
 maintainers         {samasaur.com:sam @samasaur1} openmaintainer
 
-use_configure       no
-depends_build       port:coreutils port:gmake port:clang-16
-destroot.args       PREFIX=${prefix}
-build.args          CXX=clang++-mp-16
+depends_build       port:coreutils port:gmake
+compiler.cxx_standard   2020
+compiler.blacklist-append {clang < 1600} {*gcc-[4-9]*} {*clang-[3-9]*} \
+    *clang-10 *clang-11 *clang-12 *clang-13 *clang-14 *clang-15
 
 # ---
 # The above Portfile is based off of a Git commit, because the latest tagged
--- a/sysutils/btop/Portfile
+++ b/sysutils/btop/Portfile
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        aristocratos btop 7e50b03e391a346b2b6ce92be332c58d91f75a00
+version             20230615
+
+fetch.type          git
+
+post-fetch {
+    system -W ${worksrcpath} "git submodule update --init"
+}
+
+description         A monitor of resources
+
+long_description    Resource monitor that shows usage and stats for processor, memory, disks, network and processes.
+
+categories          sysutils
+license             Apache-2
+maintainers         {samasaur.com:sam @samasaur1} openmaintainer
+
+use_configure       no
+depends_build       port:coreutils port:gmake port:clang-16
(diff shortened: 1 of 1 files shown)


sysutils/btop/Portfile
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           github 1.0
+
+github.setup        aristocratos btop 7e50b03e391a346b2b6ce92be332c58d91f75a00
+version             20230615
+
+checksums           rmd160  4262c3ae2af812132a5ca62f24cb307d86db1e0f \
```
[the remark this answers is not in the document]
Samasaur1: Cool, I removed it!

Samasaur1:
This port dynamically links against libc++20, but I'm only requiring clang 16 as a build-time dependency. I don't know how to ensure that a new enough version of libc++20 is available at runtime — or I could set this port to statically link against the C++ standard library.

The installed version of this port did work even after I ran `sudo port deactivate clang-16` and the same with LLVM 16, so I'm not actually sure where my version of libc++ is coming from.

i0ntempest:
There's a makefile portgroup that can help you better manage this kind of builds (no configure script). See port `crunch` for an example. Also see [here](https://trac.macports.org/wiki/CompilerSelection) and [here](https://trac.macports.org/wiki/UsingTheRightCompiler) for compiler selection and overriding in portfiles.

sysutils/btop/Portfile
```diff
@@ -0,0 +1,46 @@
...
+long_description    Resource monitor that shows usage and stats for processor, memory, disks, network and processes.
+
+categories          sysutils
+license             Apache-2
+maintainers         {samasaur.com:sam @samasaur1} openmaintainer
+
+use_configure       no
+depends_build       port:coreutils port:gmake port:clang-16
+destroot.args       PREFIX=${prefix}
+build.args          CXX=clang++-mp-16
```
ryandesign, line 26: Don't manually set the compiler like this. If you require a specific language standard (such as C++20) specify that (e.g. `compiler.cxx_standard 2020`) and let MacPorts pick a compatible compiler. Using the makefile portgroup as @i0ntempest mentioned would be a prerequisite to getting that to work with a port like this that doesn't use a standard configure phase.

ryandesign: And if it is C++20 that you require, MacPorts base 2.8.1 doesn't actually know what the requirements of C++20 are yet (see https://trac.macports.org/ticket/62677) so until there is a new MacPorts base release you'll have to add that information to the port:
```
```
```
compiler.blacklist-append {clang < 1400}
```

Samasaur1:
Hi all,

Thanks for helping me with this! The latest commit I pushed uses the `makefile` and `compiler_blacklist_versions` PortGroups, and works with a clean install.

However, it no longer works with GCC, since it attempts to use the libc++ on my machine:
```
CXXFLAGS   +| $(REQFLAGS) $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS) -Os -isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX13.sdk -arch arm64 -stdlib=libc++
LDFLAGS    +| $(LDCXXFLAGS) $(OPTFLAGS) $(WARNFLAGS) -L/opt/local/lib -Wl,-headerpad_max_install_names -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX13.sdk -arch arm64
```
which do not support some features of C++20 that btop requires.

Should I blacklist GCC as well for now? Or is it better to allow btop to compile with GCC and use libstdc++?

Samasaur1:
btop also recommends running `sudo make setuid` after installing, which runs:
```
chown root:wheel $(DESTDIR)$(PREFIX)/bin/btop
chmod u+s $(DESTDIR)$(PREFIX)/bin/btop
```
Is this something packages in MacPorts are allowed to do? I'm not super clear on why some of my installed binaries are owned by macports and some are owned by root.

If I am allowed to do this, would it be better to run the rule in the Makefile or to manipulate these attributes directly from the Portfile?

pmetzger:
@ryandesign Do you have an opinion on the stuff above?

pmetzger:
@herbygillot and @ryandesign opinions? Otherwise I'm just going to commit this.

PR #19340, opened 2023-07-05
boost181: Apply upstream patch for c++17-compat.

eborisch:
#### Description

Applies [upstream patch](https://github.com/boostorg/functional/commit/6a573e4) for boost181 to compile under c++17 (default on clang-16).

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 11.7.8 20G1351 x86_64
Xcode 13.2.1 13C100

### DIFF
--- a/devel/boost181/Portfile
+++ b/devel/boost181/Portfile
@@ -115,7 +115,7 @@ patchfiles-append patch-boost-context-asm.diff
 patchfiles-append patch-fix-fiber.diff
 
 # Upstream c++17 fix: https://github.com/boostorg/functional/commit/6a573e4
-patchfiles-append patch-include-boost-functional.diff
+patchfiles-append patch-boost-functional.diff
 
 # Availability.h -> AvailabilityMacros.h on Tiger
 # The attempted fix:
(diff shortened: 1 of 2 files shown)


mascguy:
Eric, is this applicable only to 181...?

eborisch:
> Eric, is this applicable only to 181...?

The file in question was previously last [changed in 2016](https://github.com/boostorg/functional/commits/develop/include/boost/functional.hpp), so I would expect the patch would apply cleanly to most of the versions of boost we have set up.

mascguy (approved):

mascguy:
> > Eric, is this applicable only to 181...?
> 
> The file in question was previously last [changed in 2016](https://github.com/boostorg/functional/commits/develop/include/boost/functional.hpp), so I would expect the patch would apply cleanly to most of the versions of boost we have set up.

Sounds good, I'll take care of patching older versions. Thanks Eric!

mascguy:
> > > Eric, is this applicable only to 181...?
> > 
> > The file in question was previously last [changed in 2016](https://github.com/boostorg/functional/commits/develop/include/boost/functional.hpp), so I would expect the patch would apply cleanly to most of the versions of boost we have set up.
> 
> Sounds good, I'll take care of patching older versions. Thanks Eric!

Haven't forgotten about this. I'm presently pruning our set of Boost ports, to reduce the maintenance burden. Once that's done - and we're almost there - I'll take care of this for our remaining ones.

PR #19349, opened 2023-07-06
ghc-ppc-bootstrap: do not hardcode prefix and version

barracuda156:
Fixes: https://trac.macports.org/ticket/67696

#### Description

@ryandesign Could you please take a look?

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/lang/ghc-ppc-bootstrap/Portfile
+++ b/lang/ghc-ppc-bootstrap/Portfile
@@ -5,7 +5,7 @@ PortSystem          1.0
 name                ghc-ppc-bootstrap
 set canonicalname   ghc
 version             7.0.4
-revision            1
+revision            2
 categories          lang haskell
 maintainers         {@barracuda156 gmail.com:vital.had} openmaintainer
 license             BSD
@@ -17,6 +17,7 @@ description         The Glorious Glasgow Haskell Compilation System
 long_description    This is a package that installs a binary bootstrap GHC compiler.
 
 homepage            https://haskell.org/${canonicalname}
+# This address is unique, hardcoding version here is appropriate:
 master_sites        https://downloads.haskell.org/ghc/7.0.4/krabby/
 distname            GHC-${version}-powerpc
 distfiles           ${distname}.pkg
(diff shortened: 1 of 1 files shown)


barracuda156:
@ryandesign Bump

lang/ghc-ppc-bootstrap/Portfile
```diff
@@ -17,6 +17,7 @@ description         The Glorious Glasgow Haskell Compilation System
 long_description    This is a package that installs a binary bootstrap GHC compiler.
 
 homepage            https://haskell.org/${canonicalname}
+# This address is unique, hardcoding version here is appropriate:
 master_sites        https://downloads.haskell.org/ghc/7.0.4/krabby/
```
ryandesign, line 21: Doesn't it seem likely that if the port were ever updated to some other version, the version number in this URL would need to change too?

lang/ghc-ppc-bootstrap/Portfile
```diff
@@ -51,8 +52,8 @@ extract {
     system -W ${worksrcpath}/files/ghc.pkg "cat Payload | gunzip -dc | cpio -i"
     
 if {![variant_isset framework]} {
-    system -W ${worksrcpath}/files/ghc.pkg "mv GHC.framework/Versions/7.0.4-powerpc/usr ${worksrcpath}/${name}"
-    system -W ${worksrcpath}/files/ghc.pkg "cp -R GHC.framework/Versions/7.0.4-powerpc/Tools ${worksrcpath}/${name}/share"
+    system -W ${worksrcpath}/files/ghc.pkg "mv GHC.framework/Versions/${version}-powerpc/usr ${worksrcpath}/${name}"
+    system -W ${worksrcpath}/files/ghc.pkg "cp -R GHC.framework/Versions/${version}-powerpc/Tools ${worksrcpath}/${name}/share"
```
ryandesign, line 56: Not related to the changes you're making here of course, but MacPorts does have `move` and `copy` commands that should usually be used instead of shell commands.

ryandesign (approved):
The changes look reasonable! I haven't tried them on my own system.

lang/ghc-ppc-bootstrap/Portfile
barracuda156: `krabby/` is a unique folder (apparently of someone who made that particular distribution). I do not think we want to create an illusion that this URL is something constant across versions. (It will not work for 7.0.1, for instance.)

lang/ghc-ppc-bootstrap/Portfile
barracuda156: This code was largely based upon existing previously in Macports. Can be simplified, but then it makes sense to do for `ghc-bootstrap` as well, and that is quite a huge port to rebuild (and also rather fragile). Unless it is a matter of principle for someone, I rather leave it as it is, at least until some major refactoring is done for both ports.

barracuda156: @pmetzger I agree.

pmetzger: Okay @barracuda156 I'm going to commit. Please go through and fix the tcl in the portfiles to use the move and copy commands and do a distinct pull request for it.

PR #19395, opened 2023-07-09
ocp: new port @0.2.105

kencu:
Open Cubic Player
closes: https://trac.macports.org/ticket/67733

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.4.1 22F82 arm64
Xcode 14.3.1 14E300c

###### Verification 
Have you

### DIFF
--- a/audio/ocp/Portfile
+++ b/audio/ocp/Portfile
@@ -1,0 +1,44 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem              1.0
+PortGroup               github 1.0
+
+github.setup            mywave82 opencubicplayer 0.2.105 v
+revision                0
+name                    ocp
+distname                ${name}-${version}
+github.tarball_from     releases
+
+license                 GPL-2+
+categories              audio
+maintainers             nomaintainer
+
+description             OpenCubicPlayer - Music visualizer
+
+long_description        {*}${description} for various tracked music formats \
+                        (amiga modules, S3M, IT), chiptunes and other formats \
+                        related to demoscene
+
+checksums               rmd160  ed44efddc2e2627c4f31af41d4812fc9288aae22 \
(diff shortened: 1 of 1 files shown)


kencu:
@ryandesign -- I'll leave this here if you're interested

barracuda156:
@kencu It does not support `libsdl` as a fallback, right? (Some currently developed ports still do.)

kencu:
> @kencu It does not support `libsdl` as a fallback, right? (Some currently developed ports still do.)

I believe it can. That might be a nice enhancement for the systems that don’t support libsdl2.

Also sdl is not a hard requirement.

I didn’t try older systems … it uses core audio so there could be issues.

ryandesign:
> I believe it can. That might be a nice enhancement for the systems that don’t support libsdl2.

I'm adding sdl1 and sdl2 variants. Both work for me on Monterey. I think sdl2 works in MacPorts back to Snow Leopard at least but if we need to default to sdl1 on older systems we can.

ryandesign:
> I am not sure about 10.6 Intel

There is a separate [libsdl2-snowleopard port](https://ports.macports.org/port/libsdl2-snowleopard) (which the libsdl2 port uses on Snow Leopard and earlier) so I assume at least the intention is that this would work on Snow Leopard and earlier.

barracuda156:
> > I am not sure about 10.6 Intel
> 
> There is a separate [libsdl2-snowleopard port](https://ports.macports.org/port/libsdl2-snowleopard) (which the libsdl2 port uses on Snow Leopard and earlier) so I assume at least the intention is that this would work on Snow Leopard and earlier.

Thank you, I will try it. Somehow missed that one.

kencu:
libsdl2-snowleopard works nicely for SL Intel.

Defaulting MacPorts SnowLeopard support to what SnowLeopard-PowerPC can support would be an unfortunate approach.

barracuda156:
> libsdl2-snowleopard works nicely for SL Intel.
> 
> Defaulting MacPorts SnowLeopard support to what SnowLeopard-PowerPC can support would be an unfortunate approach.

Thankfully, we have arch settings and do not have to coerce every arch to behave identically :)

kencu:
> Thankfully, we have arch settings and do not have to coerce every arch to behave identically :)

and yet, look what happened here.

PR #19440, opened 2023-07-16
libvirt: enable qemu driver always

mohd-akram:
#### Description

Required to [run a server](https://libvirt.org/macos.html).

###### Type(s)

- [x] enhancement

###### Tested on
macOS 13.4.1 22F770820d x86_64
Xcode 14.3.1 14E300c

###### Verification
Have you

### DIFF
--- a/sysutils/libvirt/Portfile
+++ b/sysutils/libvirt/Portfile
@@ -12,7 +12,7 @@ legacysupport.newest_darwin_requires_legacy 12
 # Remember to update libvirt and py-libvirt at the same time.
 name                libvirt
 version             9.8.0
-revision            0
+revision            1
 checksums           rmd160  f0044cda4ac4aa6a434701de3a2bbb75619f298c \
                     sha256  7aa90d133bb301e94663a45c36176f46f4a9fc1b34d77d2e22b7a2517106f506 \
                     size    9307064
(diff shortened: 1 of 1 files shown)


sysutils/libvirt/Portfile
```diff
@@ -135,10 +135,13 @@ variant sasl description {Use Cyrus SASL for authentication} {
...
-    depends_lib-append      port:qemu
-    configure.args-replace  -Ddriver_qemu=disabled -Ddriver_qemu=enabled
-}
+startupitem.create      yes
+startupitem.executable  ${prefix}/sbin/libvirtd
+
+startupitems        name        libvirt-system \
+                    location    LaunchDaemons \
+                    name        libvirt-session \
+                    location    LaunchAgents
```
ryandesign, line 144: Does the user need to start both of these, or are you giving the user the choice of whether to start it as root or as their user account?

ryandesign (commented):
The qemu variant you're removing had added a dependency on the qemu port. But the port builds successfully without it so I guess the qemu port wasn't needed?

In your commit message when you say "by default" you mean "unconditionally". "By default" implies the user has a way to override that default, such as via a variant, but your commit actually removes the user's ability to do that. If you wanted to enable the qemu driver by default and give the user the option of disabling it, you would retain the qemu variant and add the line `default_variants-append +qemu`. But i building with qemu support doesn't have a dependency on qemu, then I guess there's no reason to offer the choice anymore.

I see this in the log:

```
    Privileges
      QEMU               : root:wheel !!! running QEMU as root is strongly discouraged !!!
```

I don't know if we should be concerned about that.

sysutils/libvirt/Portfile
mohd-akram: It turns out the startup item wasn't working properly because it cannot find the qemu binary. The `${prefix}` isn't in the `PATH` when running `libvirtd` via launchd. I reverted the change for now as I'm not sure what's a nice way to solve it (the .plist does support `EnvironmentVariables` but it seems MacPorts does not expose that).

mohd-akram:
@ryandesign Is this good to merge?

PR #19445, opened 2023-07-16
spice-server: update to 0.15.2

mohd-akram:
###### Tested on
macOS 13.4.1 22F770820d x86_64
Xcode 14.3.1 14E300c

###### Verification 
Have you


devel/spice-server/Portfile
```diff
@@ -18,41 +17,20 @@ homepage            https://www.spice-space.org/
...
-    path:lib/pkgconfig/glib-2.0.pc:glib2 \
-    path:lib/pkgconfig/pixman-1.pc:libpixman \
-    path:include/turbojpeg.h:libjpeg-turbo \
-    port:gstreamer1 port:gstreamer1-gst-plugins-base \
-    port:libopus
+checksums           rmd160  a540a901e256ad2f806a25aa8861bcf7fdea507e \
+                    sha256  6d9eb6117f03917471c4bc10004abecff48a79fb85eb85a1c45f023377015b81 \
+                    size    1663326
+
+depends_build       port:pkgconfig
```
herbygillot, line 24: Do prefer the `-append` suffix for the `depends_` statements:

- `depends_build-append`
- `depends_lib-append`

devel/spice-server/Portfile
```diff
@@ -18,41 +17,20 @@ homepage            https://www.spice-space.org/
...
+
+depends_lib         port:lz4 \
```
herbygillot, line 26: See comment above.

devel/spice-server/Portfile
mohd-akram: It's ugly though. The meson port group is immune to this, it registers a callback.

herbygillot: 😆 Didn't think there would be an objection based on aesthetics :) It's more a best practice, it's not any kind of hard rule.

mohd-akram: Haha. Port groups should be transparent IMO, it should be handled at that level where possible.

mohd-akram:
Thanks!

PR #19477, opened 2023-07-18
spdlog: add variant tests; set cmake build type; reduce warnings; don't build examples

mascguy:
### Description

* Add new varint `tests`
* Set CMake build type appropriately, for debug/non-debug
* Reduce compiler warnings
* Don't build examples; they aren't installed, and cause build errors for some macOS releases


judaew:
Thank you for the contribution @mascguy ✨

PR #19493, opened 2023-07-18
MoltenVK: Add subport MoltenVK-1.0.41

Gcenx:
Closes https://trac.macports.org/ticket/67759

###### Type(s)

- [x] bugfix

###### Tested on

macOS 14.0 23A5286i x86_64
Xcode 15.0 15A5195m

###### Verification 

[skip notification]

### DIFF
--- a/graphics/MoltenVK/Portfile
+++ b/graphics/MoltenVK/Portfile
@@ -20,7 +20,6 @@ long_description    ${name} is an implementation of the high-performance, \
                     industry-standard Vulkan graphics and compute API, that \
                     runs on Apple's Metal graphics framework.
 
-dist_subdir         ${name}/${version}
 distname            MoltenVK-macos
 use_tar             yes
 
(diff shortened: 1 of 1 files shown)


jasonliu--:
> ... once we have the 1.0.41 archive on macports mirror I’d like to then force using that over archive.org
> 
> The last failure was due to archive.org blocking the traffic.

Once the distfile has been downloaded and placed into distfiles.macports.org, all future fetches should occur from there, as far as I'm aware. There shouldn't be any need to force using it over archive.org. So if you were thinking that you need to submit another PR to change the `master_sites` over to distfile.macports.org instead of archive.org, I don't think you do. In fact, it's probably better to keep the archive.org URL, so that we know where the file in distfiles.macports.org came from.

mascguy:
> Once the distfile has been downloaded by the MacPorts buildbots and placed into distfiles.macports.org, all future fetches should occur from there, as far as I'm aware. There shouldn't be any need to force using it over archive.org. So if you were thinking that you need to submit another PR to change the `master_sites` over to distfile.macports.org instead of archive.org, I don't think you do. In fact, it's probably better to keep the archive.org URL, so that we know where the file in distfiles.macports.org came from.

Yep, agreed on all counts.

mascguy (approved):

PR #19548, opened 2023-07-22
py-poetry-core: add missing dep py-idna

macportsraf:
#### Description

`sudo port install -vst py-poetry-core` failed because the py-idna package was hidden by the -t option. This pull request adds `port:py${python.version}-idna` to `depends_lib`.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G6032 x86_64
Xcode 11.3 11C29

###### Verification
Have you

### DIFF
--- a/python/py-poetry-core/Portfile
+++ b/python/py-poetry-core/Portfile
@@ -9,7 +9,7 @@ PortGroup           select 1.0
 # both in terms of APIs but also the version pinning.
 name                py-poetry-core
 version             1.6.1
-revision            0
+revision            1
 
 distname            poetry_core-${version}
 
@@ -38,6 +38,8 @@ if {${name} ne ${subport}} {
     if {${python.version} < 38} {
         depends_lib-append  port:py${python.version}-importlib-metadata
     }
+
+    depends_lib-append  port:py${python.version}-idna
 
     post-destroot {
         set docdir ${prefix}/share/doc/${subport}


jmroot (commented):
This is almost certainly incorrect; `poetry-core` vendors its dependencies to minimise bootstrap issues.

mascguy:
> This is almost certainly incorrect; `poetry-core` vendors its dependencies to minimize bootstrap issues.

So this is one of those rare occasions where trace mode interferes with installation. Still, bless you for testing with it Raf!

dgilman:
I see how jsonschema is trying to lazily import idna although it looks like py-poetry-core's use of jsonschema doesn't use the format validation functionality in jsonschema that uses the imported idna.

Is it MacPorts policy that the trace mode should work / is supported? That seems like motivation by itself to accept the patch.

This also makes me wonder about `py-jsonschema` in MacPorts, maybe it should be installed with the extra `jsonschema[format]` dependencies for full functionality with or without trace mode.

jmroot:
> Is it MacPorts policy that the trace mode should work / is supported?

Due to its implementation details, trace mode cannot always work. In this case the bug is probably fixable, but the fix is to improve darwintrace. If a port's files are hidden by trace mode, it should look the same to a program trying to access them as if the port was not installed. That's clearly not the case for Python looking for modules currently.

macportsraf:
Should we close this pull request?

PR #19553, opened 2023-07-22
opendx: fix implicitly declared function errors in compilation

bernstei:
#### Description
Make it possible to compile opendx with modern compilers by ignoring implicitly declared functions (bison/flex-related).

Fixes trac ticket 61842 https://trac.macports.org/ticket/61842

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

```
macOS 11.7.8 20G1351 arm64
Xcode 13.2.1 13C100
```

###### Verification 
Have you

### DIFF
--- a/science/opendx/Portfile
+++ b/science/opendx/Portfile
@@ -84,6 +84,10 @@ configure.pre_args-replace \
 
 configure.args-append   --without-javadx
 
+# ignore implicitly declared functions to allow compilation with modern
+# compilers
+configure.cflags-append -Wno-error=implicit-function-declaration
+
 post-destroot {
     foreach bin [glob -tails -directory ${destroot}${prefix}/libexec/${name}/bin *] {
         ln -s ${prefix}/libexec/${name}/bin/${bin} ${destroot}${prefix}/bin/${bin}


kencu:
The reason that Apple upgraded this to an error is that for arm64 Macs you often need the functions to have proper definitions, especially when building universal.

So MacPorts normally finds and adds the proper definitions for the functions instead of switching the error off.

bernstei:
Well, this port's been broken this way for several years, so macports apparently isn't adding definitions.  It's some subtle issue with functions that are generated by flex, I think. The trac ticket has a bit of the investigation we've done.  No one's found any other workaround that I know of.

bernstei:
Also, note that we've tested this and it works both on an M1 and Intel CPU macs.

kencu:
> Well, this port's been broken this way for several years, so macports apparently isn't adding definitions.

Various people have added hundreds or thousands of them, in fact.

kencu:
> Also, note that we've tested this and it works both on an M1 and Intel CPU macs.

good to hear. this may be quieting a function that is resilient to the arm64/intel issues.

bernstei:
I didn't actively select anything w.r.t. architecture, and I don't see a universal variant. We got the same error on Intel and M1, and the PR fixes it on both machines.  When I check the executable I see
```
$ file  /opt/local/libexec/opendx/dx/bin_macos/dxexec
/opt/local/libexec/opendx/dx/bin_macos/dxexec: Mach-O 64-bit executable arm64
```

kencu:
I see this port's been on life-support since about 2019, so it's something of a special case. 

In this case, simply toggling the error off and moving on may be reasonable, as upstream is apparently long dormant.

If I commit this I'll hear about it, so we'll leave this here for a little bit until we see if someone wants to try a proper fix. If nobody bites after a short while, IMHO it would be reasonable to commit this.

kencu:
ah - this port has no universal variant due to other constraints. Should be fine to disable the error then.

bernstei:
Thanks a lot.  We're really glad you're willing to keep opendx limping along - it's very useful for us.

kencu:
https://trac.macports.org/wiki/WimplicitFunctionDeclaration

for MacPorts official approach to this problem.

bernstei:
Interesting, and good to know for the future. I may try to track it down, but because it appears to be in functions automatically generated by flex/bison, it's going to be hard.  I guess maybe that means that it's a bug in those programs, rather than the opendx source as such, and so may even get fixed properly some day (unless the opendx source includes statically the functions auto-generated by those tools).

PR #19557, opened 2023-07-22
rawtherapee: add missing lib deps

mascguy:
### Description

So happy to see this port working again!

Based on the output from the configure phase, a number of ports need to be added as lib deps. This PR takes care of that.

### DIFF
--- a/graphics/rawtherapee/Portfile
+++ b/graphics/rawtherapee/Portfile
@@ -8,7 +8,7 @@ PortGroup           active_variants 1.1
 github.setup        Beep6581 RawTherapee 5.9
 github.tarball_from releases
 name                rawtherapee
-revision            0
+revision            1
 
 categories          graphics
 supported_archs     x86_64 arm64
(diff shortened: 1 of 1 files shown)


mascguy:
Since this port requires Quartz, CI will not succeed. As such, I cancelled the workflow, to avoid wasting compute cycles.

jasonliu--:
Is there a reason why most of the dependencies that you've added are path-style dependencies, rather than simply being dependent on the port?

mascguy:
> Is there a reason why most of the dependencies that you've added are path-style dependencies, rather than simply being dependent on the port?

Yep, many of these ports have `-devel` variations, so it's important to use the path style.

jasonliu--:
Ah okay, thanks for the explanation!

mascguy:
> Ah okay, thanks for the explanation!

Any concerns with merging?

jasonliu--:
Nope, LGTM! 😃

PR #19566, opened 2023-07-22
ocp: set cxx_standard 2011

barracuda156:
#### Description

With `gcc-4.2` configure fails, complaining about C++ standard.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/audio/ocp/Portfile
+++ b/audio/ocp/Portfile
@@ -52,6 +52,8 @@ configure.args          --with-ncurses \
                         --without-update-mime-database \
                         --without-x11
 
+compiler.cxx_standard   2011
+
 platform darwin {
     configure.args-append \
                         --with-coreaudio


ryandesign:
Does this actually fix the build for you? On what system?

I would not expect this to fix it on 10.7-10.13. On the buildbot, we see these build failures:

```
In file included from filesystem-ancient.cpp:26:
/opt/local/include/ancient/ancient.hpp:29:10: fatal error: 'optional' file not found
#include <optional>
         ^~~~~~~~~~
```

"optional" is a part of C++17, and the "ancient" port does already indicate that it requires C++17.

I thought "optional" required support from libc++ and that using a more recent compiler was not sufficient to fix this for older operating systems that have older libc++. The "ancient" port did build successfully on older OS versions but that might not hold for dependencies like ocp that use libancient.

10.6 and earlier don't have libc++ from Apple at all so MacPorts installs its newer libc++ as the system libc++, thus "optional" support is probably available on 10.6 and earlier but not on 10.7–10.13.

A solution to the "optional" issue is to use a newer MacPorts libc++, which can be done by using a switch in the legacysupport portgroup, however if any of the dependencies already use the standard system version of libc++ the attempted use of two different versions of libc++ within the same process might cause crashes. For this reason, I had not worked on fixing this issue yet. I was anticipating sidestepping the issue by having the ocp port exclude libancient support on OS versions whose libc++ doesn't support "optional".

barracuda156:
@ryandesign It does fix the build, but gcc supports C++17. The port seems to pick C++14 though at compile time. Perhaps I am wrong with setting 2011, and 2017 should be used.

barracuda156:
@ryandesign I tried adding `legacysupport` since I saw the warning re implicit definition of `strnlen`, however then CLOCK_GETTIME gets redefined (`ocp` has its own defined for it). Since it builds with `legacysupport`, I did not add it.

If cxx_standard is undesirable, it will work equally fine for PowerPC to disable `*gcc-4.*`, and that won’t have any consequences for other platforms. Should I rather do that?

barracuda156:
@ryandesign X11 on its own appear to work fine:
ocp_x11

SDL2 loads but rendering is buggy, as of now.

kencu:
When `<optional>` was still in `<experimental/optional>`, where it was for a long time, the landing pad functions were inlined in the header itself, like this:

```
_LIBCPP_BEGIN_NAMESPACE_EXPERIMENTAL
class _LIBCPP_EXCEPTION_ABI _LIBCPP_AVAILABILITY_BAD_OPTIONAL_ACCESS bad_optional_access
    : public std::logic_error
{
public:
	bad_optional_access() : std::logic_error("Bad optional Access") {}

//	Get the key function ~bad_optional_access() into the dylib
    virtual ~bad_optional_access() _NOEXCEPT;
};
```

later on, when they moved <optional> into the mainstream and out of `<experimental>`, they moved the landing pad functions into libc++.dylib.

However, it is certainly possible for those landing pad functions to be put back into the <optional> header directly as inline functions again (for clang/llvm), and have the availability adjusted. Then optional (and two other similar c++17 features that also were similarly handled) would work even on the existing libc++.dylib s found on < 10.13.

I have been meaning to try that, but so far just haven't got round to it.

`<filesystem>` is different, and does require library support. That used to exist as filesystem.a, but was then rolled into libc++.dylib. filesystem.a could be resuscitated as well, for a motivated soul.

ryandesign:
> Should I rather do that?

As I said, my plan is to set `compiler.cxx_standard 2017` when using libancient since it requires that, and to make the port not use libancient on systems that currently don't support "optional".

barracuda156:
@ryandesign 

> and to make the port not use libancient on systems that currently don't support "optional"

This perhaps can be made conditional on compiler, since GCC has no problems to build both.

barracuda156:
@ryandesign And then I close this PR, since you plan to make the fix anyway.

PR #19570, opened 2023-07-23
py-graph-tool: Update to version 2.57

essandess:
#### Description

py-graph-tool: Update to version 2.57
    * Add 311, Remove 37
    * Use boost181
    * Bugfix for unary_function C++17 removal

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.4.1 22F770820d arm64
Xcode 14.3.1 14E300c

###### Verification 
Have you

### DIFF
--- a/python/py-graph-tool/Portfile
+++ b/python/py-graph-tool/Portfile
@@ -6,9 +6,10 @@ PortGroup           boost 1.0
 PortGroup           python 1.0
 
 name                py-graph-tool
-version             2.46
+version             2.57
 revision            0
 epoch               20190711
+supported_archs     arm64 x86_64
 
 categories          python science
 license             LGPL-3
@@ -25,13 +26,14 @@ homepage            https://graph-tool.skewed.de
 master_sites        https://downloads.skewed.de/graph-tool/
 use_bzip2           yes
 
-checksums           rmd160  316e8cc9c0b15e066e8ac2388a460ba4866505fe \
-                    sha256  305508f0c2989a150aecd5df010424979cd17e83f67852121e4ab29eb07d3275 \
-                    size    15189665
+checksums           rmd160  6fab092de09dcc24ef14324349a9604400755c82 \
+                    sha256  a20a0e73b78e78f233e960c6ba89d1969f457221b03b66947c71d3044affbb72 \
+                    size    15192391
 
-boost.version       1.76
+# please ensure that this boost.version matches that specified in cgal5
+boost.version       1.81
 
-python.versions     37 38 39 310
+python.versions     38 39 310 311
 
 if {${os.major} <= 12 && ${os.platform} eq "darwin"} {
     version         2.2.26
@@ -70,8 +72,24 @@ if {${name} ne ${subport}} {
                     port:py${python.version}-numpy \
                     port:py${python.version}-scipy
 
+    depends_lib-append \
+                    port:py${python.version}-matplotlib
+
     # graph-tool relies on Boost.Python, so make sure it is installed.
     require_active_variants boost[boost::version_nodot] python${python.version}
+
+    # remove after this issue is fixed:
+    # https://git.skewed.de/count0/graph-tool/-/issues/761
+    post-patch {
+        fs-traverse f ${worksrcpath}/src {
+            if { [file isfile ${f}]
+                && [regexp {\.(cc|hh)$} [file extension ${f}]] } {
+                reinplace -q -E \
+                    {s|:[[:space:]]+public[[:space:]]+std::unary_function<.+,.+>||} \
+                    ${f}
+            }
+        }
+    }
 
     use_configure   yes
     # parallel build starts swapping with 8GB of RAM.
@@ -115,6 +133,7 @@ if {${name} ne ${subport}} {
     destroot.cmd    make
     destroot.destdir \
                     DESTDIR=${destroot}
+    destroot.target install
 
     variant gtk3 description "Enable interactive drawing with gtk3" {
         depends_lib-append \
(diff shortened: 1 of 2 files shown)


python/py-graph-tool/Portfile
```diff
@@ -70,9 +72,25 @@ if {${name} ne ${subport}} {
...
+
     # graph-tool relies on Boost.Python, so make sure it is installed.
     require_active_variants boost[boost::version_nodot] python${python.version}
 
+    # remove after this issue is fixed:
+    # https://git.skewed.de/count0/graph-tool/-/issues/761
+    post-patch {
+        fs-traverse f ${worksrcpath}/src {
+            if { [file isfile ${f}]
+                && [regexp ".(cc|hh)" [file extension ${f}]] } {
```
ryandesign, line 86: In a regular expression `.` means "any character". Did you mean literally a period? If so, escape it with a backslash. Also, did you mean filenames that *contain* `cc` or `hh` or did you mean filenames that *end* with that? If the latter, anchor the regexp to the end of the string with `$`:
```suggestion
                && [regexp "\.(cc|hh)$" [file extension ${f}]] } {
```

python/py-graph-tool/Portfile
```diff
@@ -70,9 +72,25 @@ if {${name} ne ${subport}} {
...
+                reinplace -q -E \
+                    "s|:\[\[:space:]]+public\[\[:space:]]+std::unary_function<.+,.+>||" \
```
ryandesign, line 88: Does upstream agree that this is an appropriate fix?

Since you are not interpolating Tcl variables or procedure calls, you can eliminate the need for escaping by using `{}` quoting characters instead of `""`:

```suggestion
                    {s|:[[:space:]]+public[[:space:]]+std::unary_function<.+,.+>||} \
```

python/py-graph-tool/Portfile
```diff
@@ -115,6 +133,7 @@ if {${name} ne ${subport}} {
     destroot.cmd    make
     destroot.destdir \
                     DESTDIR=${destroot}
+    destroot.target install
```
ryandesign, line 136: Isn't `install` the default value for `destroot.target`?

python/py-graph-tool/Portfile
essandess: Thanks, I used braces: `{\.(cc|hh)$}`. Note that EOL anchor is unnecessary because that's file `file extension` does, but it's good to be explicit for clarity.

python/py-graph-tool/Portfile
essandess: Thanks, done.

python/py-graph-tool/Portfile
essandess: It's overwritten by PG python, so this is necessary here.

python/py-graph-tool/Portfile
essandess: Upstream @count0 is aware of the fix at https://git.skewed.de/count0/graph-tool/-/issues/761. The `unary_function` is unnecessary and has been removed from C++17, so this will certainly work. I'll defer to Tiago how he'd like to address this and have noted in the Portfile that this Portfile-based fix should be removed after the issue is resolved.

essandess:
Note: The CI and buildbots will all fail on this port for `py10-graph-tool` and so forth because of the way the `boost` ports are written: `boost181`, e.g., is always installed with variant `+python11`. There is no way to install the necessary `boost` variants in the Portfile.

cc: @mascguy

mascguy:
> Note: The CI and buildbots will all fail on this port for `py10-graph-tool` and so forth because of the way the `boost` ports are written: `boost181`, e.g., is always installed with variant `+python311`. There is no way to install the necessary `boost` variants in the Portfile.
> 
> cc: @mascguy

I'll try to take a look at a subport approach for 181 soon, as this is definitely a pain...

essandess:
@mascguy May we please merge this PR and get a working `graph-tool` into the field for `arm64` machines?

The `boost` Portfile fixes can be added later.

python/py-graph-tool/Portfile
mascguy: > It's overwritten by PG python, so this is necessary here.

Well, would it be more accurate to say that this is necessary when `python.pep517` is enabled? Which is the default for 3.11 and later.

I'd need to dig into the upstream project a bit, before providing any official recommendations. But if you're working around that by setting `destroot.target`, etc, then perhaps it might be better to simply set `python.pep517` to `no`. (Which might allow you to eliminate at least a few overrides you're setting.)

However, I'll let our more Python-savvy folks chime in with their thoughts and recommendations.

CC: @jmroot
CC: @reneeotten

mascguy: > I'd need to dig into the upstream project a bit, before providing any official recommendations. But if you're working around that by setting `destroot.target`, etc, then perhaps it might be better to simply set `python.pep517` to `no`. (Which might allow you to eliminate at least a few overrides you're setting.)

That being said, I'm not overly concerned, so no need to hold up merging this.

mascguy:
Presently building the 3.11 subport locally on 10.15 [with trace mode], to see whether all is well. It's taking far longer than expected though.

There are also an inordinate number of warnings, enough so to utterly pollute the build log. I may suppress some of them before merging, as they make it difficult to follow the compilation itself...

mascguy:
> There are also an inordinate number of warnings, enough so to utterly pollute the build log. I may suppress some of them before merging, as they make it difficult to follow the compilation itself.

Which is probably due to the project Makefile setting `-Wall` and `-Wextra`. (With no immediately-apparent way to disable that without patching, based on a quick cursory review of the various AutoTools scripts.)

mascguy:
p.s. You inadvertently included an update to `py-absl`, which you also have an open PR for. So I dropped that commit from this branch.

mascguy:
Building a single subport takes a very long time, at least on my older MacPro. So I've only validated `py311-graph-tool`, but that looks good [with trace mode].

Given that there don't appear to be any dependents on this port, it might be worth dropping subports for 3.9 and prior.

mascguy (approved):

PR #19605, opened 2023-07-25
libmovit: update to 1.7.0, fix broken port

barracuda156:
#### Description

This port is currently broken across the board: https://ports.macports.org/port/libmovit/details
Let’s see if it builds now.

UPD. OK, fixed now. X11 and Mesa are needed for Epoxy, otherwise configure fails.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you


barracuda156:
Something failing with autotools:
```
  autoreconf: running: /opt/local/bin/autoconf --force
  configure.ac:22: error: possibly undefined macro: AC_MSG_WARN
        If this token and others are legitimate, please use m4_pattern_allow.
        See the Autoconf documentation.
  autoreconf: error: /opt/local/bin/autoconf failed with exit status: 1
  Command failed:  cd "/opt/local/var/macports/build/_Users_runner_work_macports-ports_macports-ports_ports_multimedia_libmovit/libmovit/work/movit-1.7.0" && autoreconf -fvi 
  Exit code: 1
```

RJVB:
>Something failing with autotools:

I have to ask: you did test this on your own system before making a PR, I hope?

barracuda156:
> Something failing with autotools:
> I have to ask: you did test this on your own system before making a PR, I hope?

@RJVB `pkgconfig` had to be added as a build dependency. Since it was installed on my system, there was no error locally.

RJVB:
And that caused the automake/autoconf error??

barracuda156:
> And that caused the automake/autoconf error??

`pkgconfig` was not a declared dependency in the portfile :)

RJVB:
Testing locally...

I'll have a few requests but I'll save those up until I'm done testing.

2 questions:
- Aren't the dependencies on SDL2, mesa and X11 only for the +install_demo variant?
- I'm surprised by the fact there's a dependency on X11+Mesa yet you patch in a link to Apple's "Cocoa" OpenGL framework. What's up with that?

barracuda156:
@RJVB I needed Cocoa locally to fix the build, but on CI configure wanted Mesa. I can deactivate Mesa and see if I get the same configure error or not.

It is possible that we may need to add linking conditionally.

Re demo: it installs that file, but it gets compiled regardless, from what I saw.

barracuda156:
@RJVB I will appreciate if you determine what exactly needed on Intel systems, and then whatever is specific to PPC and redundant otherwise (if anything), I will move into a condition macro.

RJVB:
I don't have a PPC system but I'll compile all my findings in a change request when I'm done. May take a few days though.

barracuda156:
> I don't have a PPC system but I'll compile all my findings in a change request when I'm done. May take a few days though.

I will take care of PPC side. After I make change you ask for, I will verify the build on PPC is fine, or otherwise add a PPC-only condition with a fix. No issues here.

RJVB:
OK, so here's what I found. Working from top to bottom of the Portfile:

- Add a line `installs_libs yes`, e.g. under the `checksums`
- move the `depends_build` on `port:pkgconfig` to the top of the dependency list, in the common code (i.e. for all platforms; I like to put build dependencies first, before the "normal" ones).
- add the dependencies to `port:fftw3` and `port:libepoxy` to the common code
- in the `post-destroot`, use `xinstall -m 644` instead of `move`
- personally I'd put the `livecheck` instructions at the bottom of the file

Have you tried to run the demo? It fails for me because of 2 shaders that are no longer included (`vs-color.{130,150}.vert` and `color.{130,150}.frag`). It turns out that this is done in `init_hsv_resources()` (widgets.cpp) so I doubt it has to do with the fact I'm testing on an older OS X version and an about equally old Linux distro.

So for my own version of the port I copied the files in question from the libmovit 1.5.0 sources (into `$filespath`) and added this to the post-destroot:
{{{
    foreach check {vs-color.130.vert vs-color.150.vert color.130.frag color.150.frag} {
        if {![file exists ${destroot}${prefix}/share/movit/${check}]} {
            ui_warn "Installing missing ${check}"
            xinstall -m 644 ${filespath}/${check} ${destroot}${prefix}/share/movit/
        }
    }
}}}

(the conditional check is there because I assume you didn't run into this issue and that the files thus must have been generated somehow, though I really don't see how/where.)

barracuda156:
@RJVB Thank you! I will go through these today.
One thing: `installs_libs yes` is unneeded, since that is the default. Only setting it to no (where appropriate) is due.

RJVB:
>One thing: `installs_libs yes` is unneeded, since that is the default. Only setting it to no (where appropriate) is due.

Guess what, the same thing applies to `revision`, and then guess what has become the new normal ;)

For that kind of setting I don't agree that it's a good thing to "confirm the default" (which is self-evident) but I do think it's useful for things like `installs_libs` (where the default is a lot less self-evident). But I'll leave this one up to you.

barracuda156:
@RJVB Revisions are updated, so if not upon a port creation then somewhat later the field will have to be added. I don’t think the value for `installs_libs` changes normally.
But this is your port, and I am fine with anything as long as the build works. I can add that line, no issues.

barracuda156:
@RJVB Ok, I implement all suggestions now and rebase.

barracuda156:
@RJVB Rebased, hopefully it builds fine still.

P. S. Systems besides MacOS do not need `autoreconf` here?

RJVB:
>@RJVB Rebased, hopefully it builds fine still.

Should be unless a dependency was updated to a too-new version.

>P. S. Systems besides MacOS do not need `autoreconf` here?

No, because they don't have patches to the build system. If you want to be really strict you could check if the current patches require the reconf step but it should never hurt.

RJVB:
let's go for it!

BTW, if you feel like taking over maintainership I'd be happy to give you the reigns!

barracuda156:
@RJVB 
> let's go for it!

Great!

> BTW, if you feel like taking over maintainership I'd be happy to give you the reigns!

Thank you for the offer! It will be better though if the maintainer can test Linux targets for this port, since it explicitly supports those. I have no environment for that.
But I will try to support PPC side of things; if you update it later, please tag me!

RJVB:
> It will be better though if the maintainer can test Linux targets for this port, since it explicitly supports those. I have no environment for that.

Fair enough, setting MacPorts up on Linux is not necessarily without pitfalls. I've only done it on 1 or 2 systems and it's partly still a hack.

barracuda156:
> Fair enough, setting MacPorts up on Linux is not necessarily without pitfalls. I've only done it on 1 or 2 systems and it's partly still a hack.

On a side-note, which Linux did you use with Macports?

RJVB:
>On a side-note, which Linux did you use with Macports?

My main rig still runs Ubuntu 14.04 with self-upgraded packages through PPAs and my "LinuxPorts"; I've also used it to some extent on Devuan (Debian without systemd) and a little bit on some version of FreeBSD years ago.

The main reason I'm also letting my Linux ports depend on system packages is that until now I've only run Linux on under-dimensionsed systems (think netbooks, notebooks) on which CPU cycles are (and sometimes space was) at a premium.

RJVB (approved):
LGTM

RJVB:
On Thursday August 03 2023 08:45:46 Sergey Fedorov wrote:
>@RJVB On the PR: could you merge it now, since apparently no further changes needed?

I can't. Github's PR review mechanism is so well hidden that I always forget to use it.

>On a side-note: do you still use FreeBSD? I have seen on Trac someone mentioned Macports segfaults there.

Haven't for a long time, and it probably had patches to base (just like on Linux).

barracuda156:
@herbygillot @reneeotten @harens Could someone please help us to merge this?

PR #19631, opened 2023-07-26
obs-build-helpers: remove useless python27 dependency

catap:
#### Description

This port provides only one bash script which hasn't use python.

###### Type(s)

- [x] bugfix

###### Tested on

macOS 12.6.8 21G725 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/devel/obs-build-helpers/Portfile
+++ b/devel/obs-build-helpers/Portfile
@@ -2,9 +2,9 @@
 
 PortSystem          1.0
 PortGroup           github 1.0
-PortGroup           python 1.0
 
 github.setup        openSUSE obs-build 2.5.0 obs_
+revision            1
 name                obs-build-helpers
 categories          devel python
 platforms           darwin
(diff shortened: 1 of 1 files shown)


raimue (commented):
Totally makes sense... Can we fix the description? It still mentions `unrpm` in addition to `vc` and I don't know what ever happened to that. Was that even ever in the port? It's been a long time since I used this.

catap:
@raimue that mean adding `rpm` as dependency. Frankly speaking `unrpm` is [very simple wrapper](https://github.com/openSUSE/obs-build/blob/obs_2.5.0/unrpm) for `rpm2cpio ... | cpio`

Frankly speaking I haven't see any use case for `vc` from this port and I suggest to at least remove dependency from ocs... remove completley this strange port.

What do you think?

raimue:
Right, I rather meant to edit the description and don't claim that the port provides unrpm.

In the past osc would complain if vc wasn't found. Maybe that changed?

raimue:
There was an explicit request to provide it for `osc vc`: https://trac.macports.org/ticket/49694

If we want to remove the dependency from osc, let's patch `osc vc` to print a helpful message to install obs-build-helpers from MacPorts.

On the other hand, this port is quite small, so it shouldn't do much harm.

catap:
I see. I'm using OpenSuse build service for decade and never used `osc vc` :) Let me update description when.

raimue (approved):
LGTM! :+1:

PR #19636, opened 2023-07-27
PothosCore: improve GCC patch

barracuda156:
#### Description

Addressing: https://github.com/macports/macports-ports/pull/19424
Flags added for Apple build with GCC according to the suggestion by @kencu 
Also Linux flags are retained so that port builds (presumably) on Linux too.

@mascguy @ra1nb0w Please notice that my fix (including [earlier one](https://github.com/macports/macports-ports/commit/173952e1f8755a6233dc557de7520f3dda6994ca)) has an effect only for builds with GCC, which is not the case on Intel in Macports. So I could not have broken anything with its dependencies, when build is done with Clang. At the very worst, I might have broken Linux builds with GCC, when I removed Linux-only flag from GCC options. That should be fixed with this PR.
Original code: https://github.com/pothosware/PothosCore/blob/bee99d002740e1de6a58ee1812d9a01e79d3d56c/cmake/Modules/PothosStandardFlags.cmake#L23-L45

P. S. Neither the last fix to `poco` should have any consequence for `PothosFlow` or other Pothos* ports: https://github.com/macports/macports-ports/commit/68c15a190fdde88aa50983ddce975ce2cb0ed888
Besides fixing tests and setting MongoDB variant, it only passes `-latomic`, only for GCC. There were no patches to `poco` sources.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/science/PothosCore/files/patch-unbreak-gcc-build.diff
+++ b/science/PothosCore/files/patch-unbreak-gcc-build.diff
@@ -1,15 +1,16 @@
 --- a/cmake/Modules/PothosStandardFlags.cmake	2021-01-25 11:32:12.000000000 +0800
-+++ b/cmake/Modules/PothosStandardFlags.cmake	2023-07-13 18:47:50.000000000 +0800
-@@ -27,9 +27,9 @@
-         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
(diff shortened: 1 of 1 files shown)


pmetzger:
This seems harmless enough, it passed the CI, and it's been a month.

PR #19686, opened 2023-07-31
tiledb: new port in databases

barracuda156:
#### Description

New port.

###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/databases/tiledb/Portfile
+++ b/databases/tiledb/Portfile
@@ -1,0 +1,122 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           cmake 1.1
+PortGroup           github 1.0
+PortGroup           legacysupport 1.1
+PortGroup           openssl 1.0
+
+# _getline
+legacysupport.newest_darwin_requires_legacy 10
+
+github.setup        TileDB-Inc TileDB 2.16.1
+name                tiledb
+revision            0
+categories          databases
+license             MIT
+maintainers         {@barracuda156 gmail.com:vital.had} openmaintainer
+description         The universal storage engine
+long_description    {*}${description}
+homepage            https://tiledb.com
+checksums           rmd160  38c2a9cd08aa9511354aef0b110974c92c5bad00 \
+                    sha256  2a7785422290b8a0ec4a957bd5e10e0780d954a56f6478509b71d75444dd4b1d \
+                    size    4268969
+
+# Notice that 32-bit platforms are not supported by upstream, and there are reports elsewhere
+# that some targets are broken or fail tests. It may or may not work at any given moment,
+# but seems to require substantial effort to have it properly fixed.
+# ppc64 works much better than ppc32, though some tests fail.
+# I do not disable 32-bit archs in order to facilitate testing and possible fix-ups.
+# Do not expect it to work as-is. See also: https://github.com/TileDB-Inc/TileDB/issues/4096
+
+cmake.module_path-append \
+                    ${prefix}/lib/libfmt10/cmake
+
+depends_lib-append  port:blosc \
+                    port:bzip2 \
+                    port:curl \
+                    port:libfmt10 \
+                    port:libpng \
+                    port:lz4 \
+                    port:spdlog \
+                    port:zlib \
+                    port:zstd
+
+depends_build-append \
+                    port:gmake
+
+if {${os.platform} eq "darwin" && ${os.major} < 13} {
+    # Lion+ (with Xcode 4.1+) have git; earlier need to bring their own.
+    # On 10.8.5 git fetch fails with ssl error.
+    depends_build-append    port:git
+    git.cmd                 ${prefix}/bin/git
+}
+
+if {${os.platform} eq "darwin" && ${os.major} < 11} {
+    # This trickery is intended. Normal passing of flags to CMake fails to have a desired effect.
+    # Non-Apple OS on ppc will need libatomic linking too, but leave it until tested.
+    patchfiles-append \
+                    patch-linking.diff \
+                    0001-Add-a-fix-for-32-bit.patch
+
+    post-patch {
+        reinplace "s,@PREFIX@,${prefix},g" ${worksrcpath}/cmake/Modules/FindMagic_EP.cmake
+    }
+}
+
+compiler.cxx_standard   2017
+
+# config modules are defunct.
+# https://github.com/TileDB-Inc/TileDB/issues/4070
+configure.args-append \
+                    -DCATCH_INCLUDE_DIR:STRING=${prefix}/include \
+                    -DTILEDB_ABSEIL=OFF \
+                    -DTILEDB_ASSERIONS=OFF \
+                    -DTILEDB_AZURE=OFF \
+                    -DTILEDB_CPP_API=ON \
+                    -DTILEDB_EXPERIMENTAL_FEATURES=OFF \
+                    -DTILEDB_SERIALIZATION=OFF \
+                    -DTILEDB_TESTS=OFF \
+                    -DTILEDB_TOOLS=OFF \
+                    -DTILEDB_VERBOSE=ON \
+                    -DTILEDB_WEBP=OFF \
+                    -DTILEDB_WERROR=OFF
+
+configure.cflags-append \
+                    -DSPDLOG_FMT_EXTERNAL=1
+
+configure.ldflags-append \
+                    -L${prefix}/lib/libfmt10 -lfmt
+
+platform darwin {
+    if {${os.major} < 11 && [string match *gcc* ${configure.compiler}]} {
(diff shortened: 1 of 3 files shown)


databases/tiledb/files/patch-linking.diff
```diff
@@ -0,0 +1,47 @@
...
+ if (TILEDB_VCPKG)
+   # vcpkg uses SPDLOG_FMT_EXTERNAL=1
+@@ -654,6 +655,13 @@
+   )
+ endif()
+ 
++if(CMAKE_OSX_ARCHITECTURES STREQUAL "ppc")
++  target_link_libraries(TILEDB_CORE_OBJECTS_ILIB
++    INTERFACE
++    atomic
```
kencu, line 18: the need for the atomic library here has to do with gcc being used as the compiler, not related to ppc as the arch.

databases/tiledb/files/patch-linking.diff
```diff
@@ -0,0 +1,47 @@
...
+ target_link_libraries(tiledb_unit PRIVATE $<BUILD_INTERFACE:common>)
+
+--- cmake/Modules/FindMagic_EP.cmake.orig	2023-04-26 09:28:54.000000000 +0800
++++ cmake/Modules/FindMagic_EP.cmake	2023-05-18 11:33:39.000000000 +0800
+@@ -134,7 +134,7 @@
+         -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+         -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+         -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+-        "-DCMAKE_C_FLAGS=${CFLAGS_DEF}"
++        "-DCMAKE_C_FLAGS=${CFLAGS_DEF} -isystem@PREFIX@/include/LegacySupport -Wl,@PREFIX@/lib/libMacportsLegacySupport.a"
```
kencu, line 44: we don't generally static link this library unless it is something fragile like a compiler. Any subsequent fixes to legacysupport will require knowing about this static link and revbumping this port.

databases/tiledb/Portfile
```diff
@@ -0,0 +1,122 @@
...
+configure.cflags-append \
+                    -DSPDLOG_FMT_EXTERNAL=1
+
+configure.ldflags-append \
+                    -L${prefix}/lib/libfmt10 -lfmt
+
+platform darwin {
+    if {${os.major} < 11 && [string match *gcc* ${configure.compiler}]} {
+        configure.args-append \
+                    -DCOMPILER_SUPPORTS_AVX2=OFF
```
kencu, line 94: gcc building on Intel will be able to use AVX2 ... but it requires MacPorts cctools and a reasonably recently clang to be installed for it to assemble, so I guess just switching it off is the easier path.

databases/tiledb/files/patch-linking.diff
barracuda156: I have been told a number of times that `i386` does not need it. Is that wrong?

64-bit archs do not need it with GCC. (Yes, I know that we usually do not bother to write a complete condition, since few if anyone gonna use GCC on x86_64 or aarch64, and no one uses ppc64, or if they do, they can survive with a redundant linking.)

databases/tiledb/files/patch-linking.diff
barracuda156: @kencu It is generally safe to assume that if I do something this non-standard, there are (or at least were at the time) reasons for that. I am a lazy person, and whenever not, there are more entertaining stuff to do than patching legacy-support linking :)

Admittedly, I have this port building for quite a while, and I did not try to fix legacy-support in a better way with every update. When I did, it failed to work normally. Perhaps because the build system is weird, and builds external components in separate goes.

But if you have a better suggestion, I am eager to try. I do not like the way it had to be done, it is bothersome to carry this patch.

databases/tiledb/Portfile
barracuda156: Here it gets turned off only for building with GCC (which I tried on 10.6.8, and had to add this to fix the build). Clang should be fine.

kencu: gcc building on Intel will be able to use AVX2 ... but it requires MacPorts cctools and a reasonably recently clang to be installed for it to assemble, so I guess just switching it off is the easier path.

databases/tiledb/files/patch-linking.diff
kencu: there was no reason to link legacysupport statically, AFAICT.

barracuda156: Just swap the flag there to dynamic one? I can try that.

databases/tiledb/files/patch-linking.diff
kencu: well, last I tried it. things can change. I'd have to run a test with gcc building i386 using atomics to stake my life on it.

databases/tiledb/Portfile
barracuda156: Got it, thank you for clarifying. By the way, how does it work together with clang for assembling? Never tried. Will it work automatically once Clang/LLVM is installed?

databases/tiledb/files/patch-linking.diff
barracuda156: Ok, I make no claims re `i386` here. What happened is that I mentioned it a number of times in different repos that 32-bit archs need linking to `libatomic` (with GCC), but got replies that `i386` does not. It could have been that persons commenting only used Clang and assumed it extrapolates to GCC, or w/e.

kencu: nothing has changed. All gcc builds fail, 32bit and 64 bit, on Intel archs unless libatomic is added.

So all those folks were off base.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81358

barracuda156:
@kencu Any idea why older Intel fails? This is with clang-16:
```
  "std::bad_optional_access::~bad_optional_access()", referenced from:
      __ZNSt3__127__throw_bad_optional_accessB7v160006Ev in config.cc.o
  "typeinfo for std::bad_optional_access", referenced from:
      __ZNSt3__127__throw_bad_optional_accessB7v160006Ev in config.cc.o
  "vtable for std::bad_optional_access", referenced from:
      __ZNSt3__127__throw_bad_optional_accessB7v160006Ev in config.cc.o
  NOTE: a missing vtable usually means the first non-inline virtual member function has no definition.
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
https://build.macports.org/builders/ports-10.11_x86_64-builder/builds/230228/steps/install-port/logs/stdio
Or:
https://build.macports.org/builders/ports-10.8_x86_64-builder/builds/142471/steps/install-port/logs/stdio
etc.
https://ports.macports.org/port/tiledb/details

kencu:
yes, those functions are not in the system library before 10.13

for now, MP usually uses macports-libcxx, which I created for this

databases/tiledb/files/patch-linking.diff
barracuda156: @kencu If I understand correctly, @iains says only 32-bit needs it:

> There are no 64 bit (or larger) lock-free atomics on PPC 32 - this is a constraint of the ISA, not the software.

See: https://github.com/iains/darwin-toolchains-start-here/discussions/30#discussioncomment-4242172

kencu: would never disagree with Iain, but this code fails on 64bit Intel unless libatomic is added, so --- 

```
#include <stdio.h>
#include <complex.h>
#include <stdatomic.h>

int main(void)
{
    _Atomic long double _Complex x = 0.0 + 0.0*I;
    for (int i=0; i<1000; i++) {
        x += 1.0;
    }
    printf("(%Lf,%Lf)\n",creall(x), cimagl(x));
    printf("%zu\n",sizeof(x));
    return 0;
}

$ gcc-7 -std=c11 a.c -latomic
$ gcc-7 -std=c11 a.c         
  "___atomic_feraiseexcept", referenced from:
      _main in cceeEpua.o
ld: symbol(s) not found for architecture x86_64
```

(I also tried newer gccs).

barracuda156: > this basic info is that the need for libatomic is imposed by using gcc

@kencu Then I do not really get why upstream does not fix this. If every arch requires `libatomic` and not just non-Intel 32-bit ones, the library must be linked unconditionally.

Sure, then let us change it to be used with any GCC build. We certainly do not want to have x86_64 or ppc64 broken. (I may not be interested to invest significant efforts in fixing Intel builds specifically, but I do prefer them fixed rather than not.)

kencu: (I'm not sure, exactly, how the presence or absence of 64 bit lock free atomics on PPC32 is relevant to needing libatomic when building on Intel anyway, but ... )

barracuda156:
@kencu I will make a PR with suggested fixes tonight.

PR #19691, opened 2023-08-01
jq: update to new Github project

herbygillot:
`jq` is now at a new home on Github, though no checksums have changed.
This PR updates the port to reflect that.

### DIFF
--- a/sysutils/jq/Portfile
+++ b/sysutils/jq/Portfile
@@ -3,13 +3,14 @@
 PortSystem          1.0
 PortGroup           github 1.0
 
-github.setup        stedolan jq 1.6 jq-
+github.setup        jqlang jq 1.6 jq-
+github.tarball_from releases
 revision            4
+
 categories          sysutils
-platforms           darwin
+license             MIT
 maintainers         {raimue @raimue} \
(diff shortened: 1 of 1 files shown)


raimue (approved):

sysutils/jq/Portfile
```diff
@@ -3,22 +3,22 @@
...
 
 description         A lightweight and flexible command-line JSON processor
 long_description\
     jq is like sed for JSON data – you can use it to slice and filter and \
     map and transform structured data with the same ease that sed, awk, \
     grep and friends let you play with text.
 
-homepage            https://stedolan.github.io/jq/
-github.tarball_from releases
+homepage            https://jqlang.github.io/jq
```
ryandesign, line 21: When you get the opportunity, the slash at the end of this URL should be restored. GitHub Pages directory URLs without a slash get redirected to have a slash, so we should use that canonical with-a-slash URL in the Portfile.

herbygillot: Done in https://github.com/macports/macports-ports/commit/4946089fa18b57e1d6fd1e138307d293f44786c2

PR #19730, opened 2023-08-02
clang compilers: never use clang on Darwin/PowerPC

catap:
#### Description

This trivial changes should decrease number of `if` when it blacklists clang on PowerPC. 

###### Type(s)

- [x] enhancement

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/_resources/port1.0/compilers/clang_compilers.tcl
+++ b/_resources/port1.0/compilers/clang_compilers.tcl
@@ -7,6 +7,11 @@
 
 global os.major os.platform
 
+# clang is useless on Darwin / PowerPC, let disable it globally
+if {${os.platform} eq "darwin" && [option configure.build_arch] in [list ppc ppc64]} {
+    return
+}
+
 if {${os.major} >= 11 || ${os.platform} ne "darwin"} {
     if {[option compiler.cxx_standard] >= 2014} {
         # For now limit exposure of clang-16 to ports needing c++14 or newer


catap:
FYI: @kencu , @barracuda156

barracuda156:
Since clangs are known to be broken on PPC, there is no positive side in having them, while a lack of blacklist occasionally causes unpleasant build failures. And yes, a global setting frees us from a need for local hacks.
I am totally for it!

kencu:
I proposed this a year or two ago, but it was disallowed by admins as unnecessary. Basically, PPC systems will never be asked to use clang as a compiler unless there is something to be fixed in the Portfile as the opinion, which is correct, of course.

So we'll see if it flies this time.

catap:
@mascguy, @herbygillot may I ask you to review this trivial changes which should reduce number of `if` which we have in ports :)

Dramatically.

barracuda156:
That would certainly make life easier, clang never works on Darwin PPC, not even Iain’s branch (from his own words), and as long as Macports can pick clang on any feasible ppc build, that can only introduce otherwise avoidable failures. (It has nothing to do with OS version indeed, building on a modern macOS against 10.5 SDK for ppc with clang won’t work either. It should just be disabled for all ppc/ppc64.)

mascguy:
@kencu Ken, any concerns with this change?

barracuda156:
@mascguy As I recall, Ken was the first to suggest doing this.

_resources/port1.0/compilers/clang_compilers.tcl
```diff
@@ -7,6 +7,11 @@
 
 global os.major os.platform
 
+# clang is useless on Darwin / PowerPC, let disable it globally
+if {${os.platform} eq "darwin" && [option build_arch] in [list ppc ppc64]} {
```
mascguy, line 11: Just to confirm, we want to use `build_arch`, rather than `configure.build_arch`?

kencu: Indeed, it should be configure.build_arch ...

kencu:
> @kencu Ken, any concerns with this change?

It is logical, as it prevents clang from ever coming up as a fallback compiler when building for PowerPC. 

Josh rightly pointed out last year that clang is only being put forward now if there is mismanagement in the compiler blacklisting (ie all the gcc compilers are being blacklisted during a build on PowerPC) and that is what should be fixed. So ideally this should not be needed.

That is true, but ... it's more of a pain to see clang blacklisting being added to all the Portfiles when the build is on 10.6 and the configure.build_arch is ppc/ppc64 and that is what is being done now.

So rather than have more of that, I'd rather see this, and have all those removed from the Portfiles instead.

catap:
Here the switch to `configure.build_arch`. @barracuda156 may you test it on PPC? Do it makes things better?

mascguy (approved):

PR #19764, opened 2023-08-05
gnupg1: end conflict (coexist) with gnupg2

macportsraf:
#### Description

gnupg1 is currently not co-installable with gnupg2. This is unfortunate. Very few people will require gnupg1, but probably everybody requires gnupg2. Many ports require gnupg2. For those few who do require gnupg1, this port not supporting coexistence with gnupg2 prevents these people from benefitting from gnupg2, or requires them to switch backwards and forwards between the two. That is not a desirable situation. This was brought up in https://trac.macports.org/ticket/61245 and my memory is that the maintainer ended up agreeing with me, but the situation was never fixed. I left a working Portfile (named gnupg14) attached to that bugreport.

There has been a note in the Portfile saying that gnupg1 would become co-installable with the modern version in mid-October 2018, but that hasn't happen yet. This pull request implements that intention.

The executable binaries are renamed with "1" postfixed to their original names, and so it is not backwards-compatible. There is a note to explain to users that if they have any local scripts that reference "gpg" or "gpgv" that they must modify those references to "gpg1" or "gpgv1". Also, the two other ports that have variants that depend on gnupg1 are updated to handle the new name (gpg1 rather than gpg), and the gnupg1 and gnupg2 ports no longer conflict with each other.

I've been using my local gnupg14 port alongside gnupg2 for years with no problem. I don't think anyone that requires gnupg 1.4 will mind having to use the name gpg1 rather than gpg, especially if it means that they can have the modern version installed as well.

Another minor change is that the manual entries are installed where man will automatically find them.

###### Type(s)

- [x] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.14.6 18G6032 x86_64
Xcode 11.3 11C29

###### Verification 
Have you

### DIFF
--- a/mail/gnupg1/Portfile
+++ b/mail/gnupg1/Portfile
@@ -5,7 +5,7 @@ PortSystem          1.0
 set my_name         gnupg
 name                ${my_name}1
 version             1.4.23
-revision            2
+revision            3
 categories          mail security
 license             GPL-3+
 installs_libs       no
@@ -19,7 +19,6 @@ homepage            https://www.gnupg.org
 platforms           darwin freebsd sunos
 distname            ${my_name}-${version}
 master_sites        ${my_name}:${my_name}
-conflicts           ${my_name}2
 
 use_bzip2           yes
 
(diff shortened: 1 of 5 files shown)


pmetzger:
LGTM, and this has been sitting in the queue for three weeks.

PR #19767, opened 2023-08-05
dotnet-cli: @7.0.9 fix checksums for arm64

silverbzez:
###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.5 22G74 arm64
Command Line Tools 14.3.1.0.1.1683849156

###### Verification 
Have you

### DIFF
--- a/dotnet/dotnet-cli/Portfile
+++ b/dotnet/dotnet-cli/Portfile
@@ -37,9 +37,9 @@ switch ${build_arch} {
     }
     arm64 {
         distname            dotnet-runtime-${version}-osx-arm64
-        checksums           rmd160  7b61acf0e1a9c38ab1d1379a8ebe87ace914e42a \
-                            sha256  3c5256ad903d08099e2e807980ea708ce311ce05becbd890ef9c897fd4a3d6ae \
-                            size    30555333
+        checksums           rmd160  fc3bdede5916e20de22e36788a62be94d28e001f \
+                            sha256  77d005ed2638b1c8b46be8ca20ddb3470a8c2b6f7e0c6092767030701b712581 \
+                            size    29177947
     }
     default {
         known_fail yes


silverbzez:
That's a huge failure but nobody has ever reported.

The failed checksums (on arm64 only) matched with dotnet-runtime-devel ones instead of dotnet-runtime-7's. Should we create a new port in this case, calling 'dotnet-cli-devel' to provide prerelease dotnet-cli?

judaew:
This is my failure, I apologize. I update checksums manually for ports of the .NET framework (`aspnetcore-runtime`, `dotnet-runtime`, `dotnet-sdk` for the 6, 7 and devel versions, for the x86_64 and arm64 platform, and also `dotnet-cli`). It takes a lot of time :(

judaew (approved):
LGTM, thank you for the contribution!

silverbzez:
> It takes a lot of time

I see, you are also maintaining other ports, I could take over the maintainance for dotnet ports if you want. I have an arm64 machine on my hand and GitHub Actions check on x86_64 platform so such a mistake won't happen again.



> What will be the benefits from this?

There's no benefit tbh. There's even no benefit for having a separate ```dotnet-cli``` port. No one could guarantee that ```dotnet``` command is backward-compatible nor forward-compatible with all these sdks/runtimes, though it looks like so currently.

tsabirgaliev:
The initial dotnet port group design was mine, and I have very limited expertise in dotnet. 

Folks in dotnet forums/chatgroups told me (and my experiments then confirmed) that Microsoft installers put dotnet-cli in a common shared folder and update/maintain the latest version among those ever installed on a particular system.

So I decided that it is safe to use the latest supported version of dotnet-cli as long as legacy LTS'es of runtimes/SDKs are also up to date.

judaew:
I agree with you, `dotnet-cli` is more of a wrapper. The user will select a version with a command like this, and then `dotnet-cli` will simply use that version:

```
dotnet new globaljson --sdk-version 7.0.0 --roll-forward latestMinor --force
```

Upd: @tsabirgaliev Thank you for this group of ports!

silverbzez:
Thank you @tsabirgaliev for all the works on these ports.

I'm also not an expertise either, I'm only a developer using dotnet to build web applications. 

I totally agree with your initial motivation of separating out this ```dotnet-cli``` port, but here's the thing: 

We are currently working on an old project built with asp.net core 2.1, which has been out of support for years, but right now it requires ```dotnet``` command from dotnet 7, the latest verison.

It's really hard to tell if they are compatible, right? In this case it's not always safe to use the latest command avaliable for out-of-support versions as long as we are labeling ```dotnet-cli``` as a dependency.

TBH I can't think of a better solution to this, the workaround can be to implement macports' select system and let users select their perferred command version. Homebrew besides does much simpler: they only support the latest dotnet version to be installed and linked. Others are keg-only, or only be installed but not activated in macports' term. I don't really wanna go that way, and that's why I dropped homebrew and switched to macports.


> The user will select a version with a command like this

@judaew Nah it's not only about selecting the sdk verisons. That command also acts as a package manager and a devtool plugin manager (Microsoft's convention ofc to disobey KISS). And if any feature breaks it would be hard to debug.

silverbzez:
NVM, I tried to setup a similar environment on a Debian server and here are all the packages that Microsoft provides among a sdk-3.1 installation:

```
$ dpkg -l | grep dotnet
ii  dotnet-apphost-pack-3.1                          3.1.32-1                       amd64        Microsoft.NETCore.App.Host 3.1.32
ii  dotnet-host                                      7.0.9-1                        amd64        Microsoft .NET Host - 7.0.9
ii  dotnet-hostfxr-3.1                               3.1.32-1                       amd64        Microsoft .NET Core Host FX Resolver - 3.1.32 3.1.32
ii  dotnet-runtime-3.1                               3.1.32-1                       amd64        Microsoft .NET Core Runtime - 3.1.32 Microsoft.NETCore.App 3.1.32
ii  dotnet-runtime-deps-3.1                          3.1.32-1                       amd64        dotnet-runtime-deps-3.1 3.1.32
ii  dotnet-sdk-3.1                                   3.1.426-1                      amd64        Microsoft .NET Core SDK 3.1.426
ii  dotnet-targeting-pack-3.1                        3.1.0-1                        amd64        Microsoft.NETCore.App.Ref 3.1.0
```

The pack `dotnet-host` provides the `dotnet` command just as `dotnet-cli` port does, and it is the latest version, so my worries are totally nonsense. However clearly hostfxr is missing from current port installations, I will try to fix that.

Anyway huge thanks to you twos, thank you so much for all the works you have done. Wish you all the best.

PR #19780, opened 2023-08-06
fpc: bugfix trial re extract

kamischi:
###### Type(s)

- [ ] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/lang/fpc/Portfile
+++ b/lang/fpc/Portfile
@@ -230,6 +230,7 @@ if {${subport} eq "${name}"} {
 #    }
 
     extract.only-append ${pp}
+    use_bzip2           yes
 
     post-patch {
         # adjust the path "codfilepath" for plex from /usr/local to MacPort's prefix.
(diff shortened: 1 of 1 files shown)


barracuda156:
@kamischi We may get weird errors if CI try building PowerPC code, which they should obviously not do. However I recall Ryan mentioning recently in a Trac ticket that some other port did try. Not sure what was the reason though.

Here PowerPC implementation uses a rather ugly workaround to extract the required binary from the dmg – since apparently there are no bootstrap binaries from upstream (but full compiler is there!). It works for its purpose locally (no CI for PowerPC), but I can imagine detaching dmg or related stuff may go wrong.

If there is a suspicion that this could be the reason, we could modify logic a bit, perhaps making it uglier, but more robust ))

kamischi:
I think I found a solution. It does not affect your usage of the disk image.

barracuda156:
Awesome!

PR #19815, opened 2023-08-09
go: update to 1.21.0

herbygillot:
###### Verification 
Have you

### DIFF
--- a/lang/go/Portfile
+++ b/lang/go/Portfile
@@ -15,12 +15,12 @@ epoch               3
 # tested the candidate version using the `go-devel` port, to verify how it
 # builds against current and older versions of macOS.
 
-# Go 1.20 does not build on macOS 10.12 and older
+# Go 1.21 does not build on macOS 10.12 and older
 if {${os.platform} eq "darwin" && ${os.major} < 17} {
     version         1.17.13
     set unsupported_macos true
 } else {
-    version         1.20.7
+    version         1.21.0
     set unsupported_macos false
 }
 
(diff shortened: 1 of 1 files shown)


jpeeler:
@herbygillot I noticed that go.env was missing in /opt/local/lib/go. After putting the file in place, GOPROXY functioned as expected: https://github.com/golang/go/issues/61928.

saj:
I'm seeing the same problem on `go@1.21.0_0`.

```
% go mod init foo
go: creating new go.mod: module foo
% go get golang.org/x/exp/rand
go: golang.org/x/exp/rand: GOPROXY list is not the empty string, but contains no entries
```

To elaborate on what jpeeler wrote above...

```
curl -LSsf -o /opt/local/lib/go/go.env https://raw.githubusercontent.com/golang/go/go1.21.0/go.env
```

`go.env` was added in upstream CR [462198](https://go-review.googlesource.com/c/go/+/462198) and landed in [7aa85e](https://github.com/golang/go/commit/7aa85e01376d840acc8bb931156d607a00b64a60); first appeared in go1.21rc1.

golang/go#61928 was closed by the reporter but I wonder whether this is an upstream bug.
[`readEnvFile`](https://github.com/golang/go/blob/aa5d483f25e89f3bf399362ccc9a5b146cae5c79/src/cmd/go/internal/cfg/cfg.go#L327-L330) tolerates `ENOENT`; it is only [later](https://github.com/golang/go/blob/aa5d483f25e89f3bf399362ccc9a5b146cae5c79/src/cmd/go/internal/modfetch/proxy.go#L118-L124) when the error is returned.

```
len(proxyOnce.list) == 0
```

Is that not expected if a downstream chooses to disable `GOPROXY=`?

In any case, I think we probably want to copy `go.env` into the destroot.

herbygillot:
@jpeeler @saj Thank you both for catching and commenting on this, tracking this in MacPorts Trac here: https://trac.macports.org/ticket/67961

This PR is meant to fix this: https://github.com/macports/macports-ports/pull/19892

herbygillot:
This issue should now be fixed.

PR #19868, opened 2023-08-13
gpac: fix compilation on Mavericks and Mountain Lion

erikbs:
#### Description

The Mavericks build fails with this error (Mountain Lion fails too, and I assume it is for the same reason):

```
:info:build Undefined symbols for architecture x86_64:
:info:build   "_memrchr", referenced from:
:info:build       _gf_props_parse_value in filter_props.o
:info:build ld: symbol(s) not found for architecture x86_64
```

I am not sure why, as `memrchr` does not appear in the GPAC source code. I tried to grep `/opt/local/include` for `memrchr`, but did not find any obviously helpful results. The problem might lie in the compiler. This port blacklists compilers Clang < 700, so on my system it chooses LLVM/Clang 15. I took a look at `gf_props_parse_value` in filter_props.c and I am still non the wiser (it does reference `strrchr`, but could the compiler for some reason replace that with `memrchr`?). The solution I therefore went with is a patch that pastes in an implementation of `memrchr` (from [sudo](https://opensource.apple.com/source/sudo/sudo-67/src/memrchr.c.auto.html)) on OS X 10.8 and 10.9. I do not view it as optimal, but it is the best I have. Another option might be to add this function to LegacySupport.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on
macOS 10.9.5 13F1911 x86_64
Xcode 6.2 6C131e

###### Verification
Have you

(Basic functionality: I ran the two binary files without any command line option to verify that there were no linker errors. I also ran `MP4Box` with `--version`, `-languages` and `-nodes` and verified that the outputs made sense)

### DIFF
--- a/multimedia/gpac/Portfile
+++ b/multimedia/gpac/Portfile
@@ -57,6 +57,12 @@ depends_lib         port:freetype \
 patchfiles          patch-configure.diff \
                     patch-no-hevc-yosemite.diff
 
+if {${os.platform} eq "darwin" && (${os.major} == 12 || ${os.major} == 13) } {
+    # Fix undefined symbol _memrchr by copying in the implementation from sudo
+    # https://opensource.apple.com/source/sudo/sudo-67/src/memrchr.c.auto.html
+    patchfiles-append patch-memrchr-mavericks-mountain-lion.diff
+}
+
 post-patch {
     reinplace "s|@APPLICATIONS_DIR@|${applications_dir}|g" ${worksrcpath}/configure
 }
--- a/multimedia/gpac/files/patch-memrchr-mavericks-mountain-lion.diff
+++ b/multimedia/gpac/files/patch-memrchr-mavericks-mountain-lion.diff
@@ -1,0 +1,27 @@
+--- src/filter_core/filter_props.c.orig
++++ src/filter_core/filter_props.c
+@@ -30,6 +30,24 @@
+ //for base64 decode
+ #include <gpac/base_coding.h>
+ 
++void *
++memrchr(s, c, n)
++    const void *s;
++    int c;
++    size_t n;
++{
++    const unsigned char *cp;
++
++    if (n != 0) {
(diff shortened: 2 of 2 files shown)


mohd-akram:
I cannot find a reference to memrchr in GPAC or macOS [libc](https://github.com/apple-oss-distributions/Libc). Perhaps try to compile a small program on your system with Clang 15 that uses strrchr and see what happens.

mohd-akram:
This is an LLVM [bug](https://github.com/llvm/llvm-project/issues/62254) with an upstream [patch](https://reviews.llvm.org/D155168). It should be fixed in the llvm port, not here.

erikbs:
> Perhaps try to compile a small program on your system with Clang 15 that uses strrchr and see what happens.

I am back on my computer now and decided to test this idea. I saved this code as `example.c`:

```c
#include <string.h>
#include <stdio.h>

int main(int argc, char** argv)
{
  printf("%s\n", strrchr("abgujfedcehfj", "abcdef"[argc]));
}
```

It compiled just fine with `-O0`, but then with `-O1`:
```
$ /opt/local/bin/clang-mp-15 -O1 -o example example.c
  "_memrchr", referenced from:
      _main in example-c1d79b.o
ld: symbol(s) not found for architecture x86_64
```

So it seems that it only happens with optimisation turned on (regardless of level). I also tested a few Clang versions (13, 14, 15 and 16, as well as 3.5 from Xcode), and the only affected versions were 15 and 16. I will try building these version with the patch you linked to, and then test again.

erikbs:
As suggested I opened a new PR (#19873) to fix the LLVM ports instead, so I am closing this one. Thanks for putting me on the right track!

mohd-akram:
Thanks for looking into this further and finding the affected versions. I nudged the maintainers to merge the fix before LLVM 17 releases.

PR #19899, opened 2023-08-15
xxhash: Fix 10.7 Lion clang segfault

thestr4ng3r:
#### Description

Fixes: https://trac.macports.org/ticket/67963
May supersede https://github.com/macports/macports-ports/pull/19897

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 10.7.5 11G63 x86_64
Xcode 4.6.3 4H1503

###### Verification 
Have you

- [x] tried existing tests with `sudo port test`? (There are none)

### DIFF
--- a/devel/xxhash/Portfile
+++ b/devel/xxhash/Portfile
@@ -22,6 +22,12 @@ if {[string match "*gcc-4.*" ${configure.compiler}]} {
                     patch-Makefile-gcc.diff
 }
 
+if {${os.platform} eq "darwin" && ${os.major} == 11} {
+    # see https://trac.macports.org/ticket/67963
+    # clang: error: unable to execute command: Segmentation fault: 11
+    patchfiles-append lion-clang-segfault-fix.diff
+}
+
 subport ${name} {
     revision        2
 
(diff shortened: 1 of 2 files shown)


thestr4ng3r:
Sent upstream too: https://github.com/Cyan4973/xxHash/pull/878

thestr4ng3r:
Yes, I think it should be possible to do unconditionally. I agree that less control flow in the portfile helps maintenance.

Should we maybe update the revision only for the xxhashlib subport since that is the one that installs the header? The xxhash one should result in identical binaries if it compiled before.

chrstphrchvz:
> Should we maybe update the revision only for the xxhashlib subport since that is the one that installs the header? The xxhash one should result in identical binaries if it compiled before.

I agree with this.

mascguy (approved):
I was able to reproduce the crash on my 10.7 VM, and can confirm that this patch fixes the issue. Nice work!

@Schamschula Marius, this is blocking other ports from building on 10.7, so it would be great to merge this ASAP. Can you take a look when you get a chance? (The changes are trivial, so it's a 2-minute review at most.)

mascguy:
Maintainer timeout, merging.

PR #18956, opened 2023-06-03
Replace opensource.apple.com/tarballs to macports_distfiles

catap:
#### Description

Apple moved all sources from opensource.apple.com/tarballs to github, that leads to a different distfiles which hadn't passed checksum.

Unfortunately they hadn't migrated everything, for example some patches for ld64 are missed.

But, we have it inside distfiles => let use distfiles as mirror.

###### Type(s)

- [x] bugfix

###### Tested on

macOS 10.5.8 9L34 i386
Xcode 3.1.4 9M2809

###### Verification 
Have you

- [x] followed our [Commit Message Guidelines](https://trac.macports.org/wiki/CommitMessages)?
- [x] squashed and [minimized your commits](https://guide.macports.org/#project.github)?
- [x] checked that there aren't other open [pull requests](https://github.com/macports/macports-ports/pulls) for the same change?
- [x] checked your Portfile with `port checksum`?

PR #19918, opened 2023-08-16
unifdef: add conflict to developer_cmds

catap:
#### Description

Closes: https://trac.macports.org/ticket/67970

###### Type(s)

- [x] bugfix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/devel/developer_cmds/Portfile
+++ b/devel/developer_cmds/Portfile
@@ -12,7 +12,8 @@ categories              devel
 maintainers             nomaintainer
 
 # We install BSD ctags, conflicting with Exuberant Ctags.
-conflicts               ctags
+# unifdef and unifdefall is conflicting with unifdef port.
+conflicts               ctags unifdef
 
 homepage                http://opensource.apple.com/source/${name}/
 master_sites            http://opensource.apple.com/tarballs/${name}/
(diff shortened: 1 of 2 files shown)


catap:
@mascguy, @barracuda156 here it is. CI probably will fail because of https://github.com/macports/macports-ports/pull/18956

barracuda156:
@catap By the way, do we have a better `mkdep`? Then I could drop dependency on `developer_cmds` from `mk-configure`. And then also update `developer_cmds` – I could not last time, because the recent version lacks `mkdep`, which I needed.

catap:
> @catap By the way, do we have a better `mkdep`? Then I could drop dependency on `developer_cmds` from `mk-configure`. And then also update `developer_cmds` – I could not last time, because the recent version lacks `mkdep`, which I needed.

as far as I recall `mkdep` is awk script, or awk + bash; someone like NetBSD or OpenBSD has C-reimplementation.

Probably they works the same way.

barracuda156:
@catap In Macports I could only find it in `developer_cmds`. But anyway, this is not a subject of this PR and should not prevent merging it.

mascguy (approved):

PR #20006, opened 2023-08-19
meson ports: migrate to meson.wrap_mode, rather than configure arg '--wrap-mode'

mascguy:
### Description

Meson subproject downloads can now be controlled via a new option, `meson.wrap_mode`.

That eliminates the need to specify `--wrap-mode` as a configure argument. And it provides a formal method for ports to declare how they want subproject downloads to be handled.

Related tracking ticket: [67990 - meson ports: subproject handling: update all, to explicitly declare whether subproject download allowed](https://trac.macports.org/ticket/67990)

CC: @Gcenx

### DIFF
--- a/emulators/dosbox-staging/Portfile
+++ b/emulators/dosbox-staging/Portfile
@@ -11,7 +11,7 @@ legacysupport.newest_darwin_requires_legacy 10
 
 github.setup          dosbox-staging dosbox-staging 0.80.1 v
 github.tarball_from   archive
-revision              0
+revision              1
 
 categories            emulators
 license               GPL-3+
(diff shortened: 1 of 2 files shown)


emulators/dosbox-staging/Portfile
```diff
@@ -56,8 +56,8 @@ if {![variant_isset debug]} {
     configure.args-append -Dbuildtype=release
 }
 
-# Discourage Meson from downloading stuff during the configure process
-configure.args-append --wrap-mode=nofallback --wrap-mode=nodownload
```
mascguy, line 60: Bear in mind that Meson only appears to support one value for `--wrap-mode`, with the last one taking precedence.

So we don't lose anything with the new approach, and we achieve the same result as before.

harens (commented):
Thanks @mascguy and @Gcenx for doing this 👍 Since this achieves the same result as before, why is a revbump required here?

mascguy:
> Thanks @mascguy and @Gcenx for doing this 👍 Since this achieves the same result as before, why is a revbump required here?

It's not strictly required, but we wanted to ensure the ports rebuild properly. So we could remove the rev-bumps, if you folks would prefer. (But both of these ports build very quickly, so there's no concern with tying up the buildbots.)

harens (approved):

mascguy:
Maintainer timeout, merging.

PR #20011, opened 2023-08-20
chicken: remove Xcode requirement

saj:
Hello,

This patch should allow the Chicken Scheme compiler, `csc`, to function on systems without the full Xcode suite.

I filed [ticket/67917](https://trac.macports.org/ticket/67917) a couple of weeks ago wondering whether the hard Xcode dependency -- **at `csc` run time** -- was a bug.  Since then, I found [SummerOfCode/2019#remove-xcode-dep](https://trac.macports.org/wiki/SummerOfCode/2019#remove-xcode-dep) and `use_xcode`; both seem to imply that this port should function with only the CLT (or, rather, either a full Xcode installation or the CLT).

> By default, it is assumed on macOS that ports will not need tools from Xcode.app unless (1) Command Line Tools aren't installed, (2) you are on an old version of Mac OS X that does not support the xcode-select mechanism, or (3) the port uses build.type xcode or includes the xcode PortGroup.

This patch works by removing upstream references to `Xcode.app` and replacing them with libxcselect shims.

I figure this approach should work OK on OS X 10.9 and later.  MacPorts still distribute Chicken for OS X 10.7 and 10.8, however, and I don't think these older systems shipped with the shims.  What is the correct way to maintain compatibility with OS X < 10.9?  (FWIW, mpstats does not show any chicken installations on anything older than [10.12](https://ports.macports.org/port/chicken/stats/).)

`csc` now works with only the CLT.

```
% csc -o hello-world hello-world.scm
% ./hello-world
Hello, world.
% otool -L hello-world
hello-world:
        /opt/local/lib/libchicken.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.0.0)
```

`port test` fails, though I think the failure is a false negative?  I shall paste test output separately, below.

Thank you for MacPorts!

---

#### Description

Indirect via libxcselect shims at runtime.
Allows the chicken compiler, csc, to function with only the CLT.

Closes: https://trac.macports.org/ticket/67917

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 12.6.7 21G651 x86_64
Command Line Tools 14.2.0.0.1.1668646533

###### Verification 
Have you

- [ ] tried existing tests with `sudo port test`?  FAILS - note below

### DIFF
--- a/lang/chicken/Portfile
+++ b/lang/chicken/Portfile
@@ -8,6 +8,7 @@ PortGroup               muniversal 1.0
 
 name                    chicken
 version                 5.3.0
+revision                1
 categories              lang scheme
 license                 BSD public-domain
 maintainers             nomaintainer
--- a/lang/chicken/files/patch-Makefile.macosx.diff
+++ b/lang/chicken/files/patch-Makefile.macosx.diff
@@ -1,6 +1,32 @@
---- Makefile.macosx.orig	2016-05-28 04:30:50.000000000 -0700
-+++ Makefile.macosx	2016-12-11 14:07:16.000000000 -0700
-@@ -66,18 +66,12 @@
+--- Makefile.macosx.orig	2023-08-20 15:07:14.000000000 +1000
++++ Makefile.macosx	2023-08-20 15:07:18.000000000 +1000
+@@ -28,14 +28,12 @@
+ 
+ # platform configuration
+ 
+-XCODE_DEVELOPER ?= /Applications/Xcode.app/Contents/Developer
+-XCODE_TOOL_PATH ?= $(XCODE_DEVELOPER)/Toolchains/XcodeDefault.xctoolchain/usr/bin
+-C_COMPILER ?= $(XCODE_DEVELOPER)/usr/bin/gcc
++C_COMPILER ?= /usr/bin/gcc
+ ARCH ?= $(shell sh $(SRCDIR)/config-arch.sh)
+ 
+ # commands
+ 
+-POSTINSTALL_PROGRAM = $(XCODE_TOOL_PATH)/install_name_tool
++POSTINSTALL_PROGRAM = /usr/bin/install_name_tool
+ 
+ # options
+ 
+@@ -49,7 +47,7 @@
+ C_COMPILER_OPTIMIZATION_OPTIONS ?= -Os -fomit-frame-pointer
+ endif
+ endif
+-LIBRARIAN ?= $(XCODE_TOOL_PATH)/ar
++LIBRARIAN ?= /usr/bin/ar
+ LINKER_LINK_SHARED_LIBRARY_OPTIONS = -dynamiclib -compatibility_version 1 -current_version 1.0 -install_name $@
+ POSTINSTALL_PROGRAM_FLAGS = -change lib$(PROGRAM_PREFIX)chicken$(PROGRAM_SUFFIX)$(SO) $(LIBDIR)/lib$(PROGRAM_PREFIX)chicken$(PROGRAM_SUFFIX)$(SO)
+ LIBRARIAN_OPTIONS = scru
+@@ -66,18 +64,12 @@
  
  # architectures
  


saj:
Here is the `port test` failure.  It looks like the test executable is trying to link with `libchicken.dylib` but the linker is not looking in the correct destroot path.

(Once installed, the port seems to work OK.)

```
:info:test ======================================== csc tests ...
:info:test dyld[43760]: Library not loaded: 'libchicken.dylib'
:info:test   Referenced from: '/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/csc'
:info:test   Reason: tried: 'libchicken.dylib' (no such file), '/usr/local/lib/libchicken.dylib' (no such file), '/usr/lib/libchicken.dylib' (no such file), '/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/tests/libchicken.dylib' (no such file), '/usr/local/lib/libchicken.dylib' (no such file), '/usr/lib/libchicken.dylib' (no such file)
:info:test Error: shell invocation failed with non-zero return status
:info:test "/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/csc -v -I.. -compiler /opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/chicken -libdir .. null.scm -t"
:info:test 6
:info:test      Call history:
:info:test      <syntax>          [csc] (apply run csc-path "-v" "-I.." "-compiler" (realpath chicken-path) "-libdir" ".." args)
:info:test      <syntax>          [csc] (realpath chicken-path)
:info:test      <syntax>          (csc "null.scm" "-t")
:info:test      <eval>    (csc "null.scm" "-t")
:info:test      <eval>    [csc] (apply run csc-path "-v" "-I.." "-compiler" (realpath chicken-path) "-libdir" ".." args)
:info:test      <eval>    [csc] (realpath chicken-path)
:info:test      <eval>    [realpath] (normalize-pathname (make-pathname (current-directory) x))
:info:test      <eval>    [realpath] (make-pathname (current-directory) x)
:info:test      <eval>    [realpath] (current-directory)
:info:test      <eval>    [run] (system* (string-intersperse (cons (realpath x) args)))
:info:test      <eval>    [run] (string-intersperse (cons (realpath x) args))
:info:test      <eval>    [run] (cons (realpath x) args)
:info:test      <eval>    [run] (realpath x)
:info:test      <eval>    [realpath] (normalize-pathname (make-pathname (current-directory) x))
:info:test      <eval>    [realpath] (make-pathname (current-directory) x)
:info:test      <eval>    [realpath] (current-directory)        <--
:info:test gnumake: *** [check] Error 70
:info:test Command failed:  cd "/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0" && /usr/bin/gnumake PLATFORM=macosx check
:info:test Exit code: 2
:error:test Failed to test chicken: command execution failed
:debug:test Error code: CHILDSTATUS 43708 2
:debug:test Backtrace: command execution failed
:debug:test     while executing
:debug:test "system {*}$notty {*}$callback {*}$nice $fullcmdstring"
:debug:test     invoked from within
:debug:test "command_exec -callback portprogress::target_progress_callback test"
:debug:test     (procedure "porttest::test_main" line 4)
:debug:test     invoked from within
:debug:test "$procedure $targetname"
:error:test See /opt/local/var/macports/logs/_Users_saj_ports_lang_chicken/chicken/main.log for details.
```

saj:
This post addresses an unresolved question from the PR message.

> `port test` fails, though I think the failure is a false negative?

(Worth noting that I did not even get this far on trunk.  `build` fails without the CLT, after all. 😉 )

The upstream test suite relies on `DYLD_LIBRARY_PATH` to point the loader at `libchicken.dylib` -- before `libchicken.dylib` is installed to its nominal location.  Coming from more open unix systems, this seemed fine to me.

It took me an embarrassingly long time to realise that Apple drop `DYLD_` environmental knobs when certain 'protected' programs are executed; namely, `/bin/sh`.  Upstream export `DYLD_` knobs from a shell script, which is fine in itself, but `csi` executes back in to `/bin/sh` when evaluating `(system*)` calls in, e.g., `tests/csc-tests.scm`.  `DYLD_LIBRARY_PATH` is silently dropped here by the kernel or some other Apple security machinery.

As a test of this theory, I quickly hacked up `tests/csc-tests.scm` as follows.

```diff
--- csc-tests.scm.orig  2021-09-09 22:12:10.000000000 +1000
+++ csc-tests.scm       2023-08-21 03:48:28.000000000 +1000
@@ -13,7 +13,11 @@
   (normalize-pathname (make-pathname (current-directory) x)))

 (define (run x . args)
-  (system* (string-intersperse (cons (realpath x) args))))
+  (let-values (((pid ok status)
+                (process-wait (process-run (realpath x) args))))
+    (cond
+      [(not ok) (error "ruh roh")]
+      [else 0])))

 (define (csc . args)
   (apply run csc-path "-v" "-I.." "-compiler" (realpath chicken-path) "-libdir" ".." args))
```

(My Scheme is poor.  I was trying to get Chicken installed so that I could learn the language.  Sorry if this is terrible.)

I think this worked.  `port test` was able to get much farther this time -- before eventually failing on a similar loader problem elsewhere in the test suite.

```
:info:test ======================================== csc tests ...
:info:test /usr/bin/env DYLD_LIBRARY_PATH='/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/tests/..' /opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/chicken 'null.scm' -output-file 'null.c' -verbose
:info:test /usr/bin/env DYLD_LIBRARY_PATH='/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/tests/..' '/usr/bin/clang' 'null.c' -o 'null.o' -c  -fno-strict-aliasing -fwrapv -fno-common -DHAVE_CHICKEN_CONFIG_H -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 -DC_ENABLE_PTABLES -Os -fomit-frame-pointer -I.. -I/opt/local/include/chicken
:info:test /usr/bin/env DYLD_LIBRARY_PATH='/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/tests/..' '/usr/bin/clang' 'null.o' -o 'null' -Wl,-syslibroot,/Library/Developer/CommandLineTools/SDKs/MacOSX12.sdk -arch x86_64 -L.. -lchicken -lm
:info:test /usr/bin/env DYLD_LIBRARY_PATH='/opt/local/var/macports/build/_Users_saj_ports_lang_chicken/chicken/work/chicken-5.3.0/tests/..' /usr/bin/install_name_tool -change libchicken.dylib '/opt/local/lib/libchicken.dylib' 'null'
:info:test ======================================== compiler inlining tests  ...
[...]
```

It looks like this test suite has been busted on macOS for a long time.

ryandesign:
Yes, you must set DYLD_LIBRARY_PATH pointing to the directory containing the library you want to use, if you have not installed it yet.

Yes, System Integrity Protection, introduced in OS X El Capitan 10.11, disables the propagation of DYLD_ environment variables to subprocesses from protected executables. You have to set it right at the point of use, not from some parent process.

> MacPorts still distribute Chicken for OS X 10.7 and 10.8, however, and I don't think these older systems shipped with the shims

On 10.8 and earlier, /usr/bin/ar, /usr/bin/gcc, and /usr/bin/install_name_tool are the actual executables (assuming the command line tools have been installed), not shims, so there should be no problem.

saj:
Thanks.  In that case, I think this PR is good to go.  Compat with OS X 10.8 and earlier was my primary concern.

The test failures were not introduced by this patch to the best of my knowledge.  (I merely stumbled on them, and was unsure whether the failures were new or old.)  I just remembered I have another Mac with XCode 14.2.  D'oh.  We can use this machine to provide a definitive answer.  `port test chicken` also fails on this Mac -- on trunk -- on the same linkage problem.  It's an old bug.

I'd like to follow up on the test failures separately, if that is OK with you.  This matter is unrelated to the PR subject.  We could carry test patches for the SIP problem in MacPorts, but I think it would be better to engage with upstream.  The problem may be avoided if the upstream test suite was adjusted to avoid needless use of `sh`.

PR #20044, opened 2023-08-20
qsynth-qt4: new port, legacy version for old systems

barracuda156:
#### Description

I skip CI here, there is nothing to build: `qsynth-qt4` is set for `darwin < 11`, `qsynth` gets non-functional change only: conflict declaration and restriction to `darwin > 9`.

@mascguy @ryandesign Could someone please review this?

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS 10.6
Xcode 3.2

###### Verification 
Have you

### DIFF
--- a/audio/qsynth-qt4/Portfile
+++ b/audio/qsynth-qt4/Portfile
@@ -1,0 +1,53 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem          1.0
+PortGroup           qt4 1.0
+
+# Please keep this pegged, unless a newer version is fixed for Qt4 and confirmed to work.
+# This port is meant only for old systems which do not presently support Qt5.
+name                qsynth-qt4
+set real_name       qsynth
+version             0.5.3
+revision            0
+maintainers         {@barracuda156 gmail.com:vital.had} openmaintainer
+categories          audio
+platforms           {darwin < 11}
+license             GPL-2+
+conflicts           qsynth
+
+description         A fluidsynth GUI front-end application
+long_description    Qsynth is a fluidsynth GUI front-end application, written in C++ using the Qt framework. \
+                    This is a legacy version primarily aimed at PowerPC systems.
+
+homepage            https://qsynth.sourceforge.io
+master_sites        sourceforge:project/${real_name}/${real_name}/${version} \
+                    https://download.sourceforge.net/${real_name}/
+
+distname            ${real_name}-${version}
+
+checksums           rmd160  35468f7b9fdbbcfcea919cd4749d4876de984276 \
+                    sha256  2ce9791f8a927425d29bacc44fdfff575638b24512e0f0b6f2eca85deb67f0c9 \
+                    size    269252
+
+use_autoreconf      yes
+autoreconf.cmd      ./autogen.sh
+autoreconf.args
+
+# The build system is crazy: in fact C++11 is not required, moreover, both configure and environment settings
+# for compiler choice are stubbornly ignored, and the build picks the compiler which was used to build Qt4,
+# which normally would mean gcc-4.2. Yet, just using gcc-4.2 leads to configure failure: Qt4 version
+# is not recognized correctly. Despite this madness, build succeeds and the app works.
+compiler.cxx_standard 2011
+
+depends_lib-append  port:desktop-file-utils \
+                    port:fluidsynth \
+                    port:xorg-libX11 \
+                    port:xorg-libXext
+
+configure.args-append \
+                    --bindir=${applications_dir} \
+                    --enable-qt4
+
+post-activate {
+    system "${prefix}/bin/update-desktop-database ${prefix}/share/applications"
+}
--- a/audio/qsynth/Portfile
+++ b/audio/qsynth/Portfile
@@ -9,8 +9,9 @@ version             0.5.7
 revision            2
 maintainers         {gmail.com:rjvbertin @RJVB} {mojca @mojca} openmaintainer
 categories          audio
-platforms           darwin
+platforms           {darwin > 9}
 license             GPL-2+
+conflicts           qsynth-qt4
 
 description         A fluidsynth GUI front-end application
 long_description    Qsynth is a fluidsynth GUI front-end application, written in C++ using the Qt framework. \


barracuda156:
qsynth

RJVB:
Out of curiosity, don't all those targeted legacy systems have a Qt5 port which is new enough for the same qsynth you're building against Qt4 here? IMHO that would be preferable (and then you'd add a `qsynth-legacy` port)

barracuda156:
> Out of curiosity, don't all those targeted legacy systems have a Qt5 port which is new enough for the same qsynth you're building against Qt4 here? IMHO that would be preferable (and then you'd add a `qsynth-legacy` port)

PowerPC – no. 10.6 Intel – had an archaic version, does not support anything recent.
See, for example: https://trac.macports.org/ticket/64932#comment:10

RJVB:
>PowerPC – no. 10.6 Intel – had an archaic version, does not support anything recent.

You'd probably not need anything recent. The minimum Qt5 version requirement might not be a hard one, or might not be hard to downgrade by re-introducing code that was dropped when the dependency version was bumped. If you were really motivated you could also translate the Qt4 code to "legacy Qt5". That might not even be that hard given that this is a relatively small code base with a simple enough GUI.

I didn't realise Qt5 never supported PPC though. That's of course a bigger hurdle to take.

It's your proposal. If it were me I'd have a good look at the effort required to get the Intel versions to use Qt5 and then set up the port so it builds against Qt4 on PPC. That's almost trivial to do with portfiles and the Qt PortGroups.

barracuda156:
@RJVB My aim is just to provide a working implementation for PPC. There is just no reason to prohibit building it for Intel on old OS, so I do not restrict the arch, but I have no interest to specifically support old Intel. There is a port already with Qt5 GUI, I guess no point to introduce a mess here.

Of course, if some version of Qt5 is ever fixed for PPC, I will see what can build with it. There are not too many such ports, and even less which I specifically added.

As for rewriting: well, it should be feasible, but I guess no one gonna go it for this particular port. I may try that for `transmission-qt` or `retroshare`, which I see more point to spend time on.

RJVB:
They were just suggestions of course.

I have no further reason to interfere with this PR: you're introducing a new port, not a subport of `port:qsynth` and the changes to that port are trivial.

barracuda156:
(Unfortunately, upstream either did not have Qt4 with CMake build system from its beginning or dropped it earlier – that is why I use autotools here, and that was the primary reason to make a separate port.)

mascguy:
The Monterey CI job had been running for nearly four hours, versus 10 minutes or less for Big Sur and Ventura. So I cancelled it.

Haven't looked at the logs yet, but something isn't quite right there. (Likely not caused by your changes though.)

mascguy:
> The Monterey CI job had been running for nearly four hours, versus 10 minutes or less for Big Sur and Ventura. So I cancelled it.
> 
> Haven't looked at the logs yet, but something isn't quite right there. (Likely not caused by your changes though.)

Ah, CI is attempting to build LLVM/Clang, for Monterey. That explains it...

mascguy (approved):
LGTM. @mojca ?

RJVB:
On Sunday August 20 2023 10:18:15 Sergey Fedorov wrote:
>I skip CI here, there is nothing to build: `qsynth-qt4` is set for `darwin &lt; 11`, `qsynth` gets non-functional change only: conflict declaration and restriction to `darwin &gt; 9`.

Then why is the CI system wasting time trying a build that will (presumably) not succeed? Can't you avoid that by setting `known_fail`?

barracuda156:
@RJVB I have no idea why it tried to build something here.

RJVB:
>@RJVB I have no idea why it tried to build something here.

I thought it was obvious the question was not for you ;)

Maybe `known_fail` is set behind the scenes when you do something like `platform darwin<9` but if not and you are aware the build will fail it may be good to use that keyword.As far as I understand it will prevent the build bots from wasting time on trying to create binary install images. And once the port is accepted you'll also not get emails about failing builds ;)

barracuda156:
> Maybe `known_fail` is set behind the scenes when you do something like `platform darwin<9` but if not and you are aware the build will fail it may be good to use that keyword.As far as I understand it will prevent the build bots from wasting time on trying to create binary install images. And once the port is accepted you'll also not get emails about failing builds ;)

AFAIK, `known_fail` is replaced with new platforms settings. It is essentially the same thing but easier and more versatile.

RJVB:
>AFAIK, `known_fail` is replaced with new platforms settings. It is essentially the same thing but easier and more versatile.

I haven't updated my "base" for a  while ('t ain't broke ;) ) but apparently it isn't yet so versatile that it prevents the CI from building your port. Maybe `known_fail` doesn't either.
It certainly does look less cumbersome for the usual case where only a oldest supported and/or newest compatible OS version is required.

barracuda156:
@mojca Could you take a look, please? (CI cancellation on macos-12 is irrelevant.)

audio/qsynth-qt4/Portfile
```diff
@@ -0,0 +1,53 @@
...
+use_autoreconf      yes
+autoreconf.cmd      ./autogen.sh
+autoreconf.args
+
+# The build system is crazy: in fact C++11 is not required, moreover, both configure and environment settings
+# for compiler choice are stubbornly ignored, and the build picks the compiler which was used to build Qt4,
+# which normally would mean gcc-4.2. Yet, just using gcc-4.2 leads to configure failure: Qt4 version
+# is not recognized correctly. Despite this madness, build succeeds and the app works.
+compiler.cxx_standard 2011
+
```
kencu, line 41: qt embeds it’s build compiler into it’s build scripts and uses that data to build software. It also embeds other things into those scripts.

To properly override the embedded qt build info, and use macports’ selected compiler etc, macports provides the qmake4 PortGroup (and others for qt5, etc).

barracuda156: I think I have tried `qmake` PG and it did not work, but I will check again.

RJVB:
>I think I have tried `qmake` PG and it did not work, but I will check again.

It's unlikely to be of much use if you build with the autoconf build system which AFAICT (from a slightly newer qsynth version) is required to prepare the qmake build system (there's a `src/src.pri.in` qmake include file that contains `ac` macros).

There's a CMake-based build system in my version but that one appears to be Qt5 only. Plus CMake would more or less impose the libc++ conversion IIRC, which might not be required if you just stick to autoconf?

barracuda156:
> CMake would more or less impose the libc++ conversion

Not on its own, or if CMakeLists have this bug, that can be fixed, however there is no reason to bother fixing this port for CMake with Qt4, given that no further updates are expected.

RJVB:
On Saturday August 26 2023 08:53:51 Sergey Fedorov wrote:
>> CMake would more or less impose the libc++ conversion
>
>Not on its own, or if CMakeLists have this bug, that can be fixed, however there is no reason to bother fixing this port for CMake with Qt4, given that no further updates are expected.
>
>

That wouldn't be a bug in the CMakeLists file, AFAIK CMake has required C++11 since many versions now. If a way around that has been found for MacPorts than that's great but it's not a bug either way.

Even if the port isn't going to be updated it will still need to be maintained and rebuilt with some kind of regularity when dependencies are updated. Not only is a CMake-based build system easier to maintain (when tweaks are required), it also requires less resources to do the actual build (the compiler isn't called through the *huge* libtool shell script and even more so when you use ninja or samu instead of make). Plus you'll get full control over the compiler and compiler arguments.

barracuda156:
> CMake would more or less impose the libc++ conversion
> AFAIK CMake has required C++11 since many versions now

I can’t put this together. C++11 does not require libc++ and never did.

RJVB:
>I can’t put this together. C++11 does not require libc++ and never did.

I can recall a bootstrap situation for building cmake, and also that the llvm ports long used a legacy build system (patched back in) because of a circular dependency with cmake. I've only followed this from a distance as I was already on 10.9 by the time libc++ started to become an unavoidable dependency. I do know some attempts have been made to use libstdc++ instead; I don't know if and why that was abandoned. (Nor why libc++ on those older systems isn't built like it is on linux, using part of libstdc++ instead of libc++abi ; this would probably have improved compatibility with the system C++ runtime.)

Notice the dependencies on port:libcxx below; for 10.6 (if I'm not mistaken):
```
> port info --version --depends cmake os.major=10 | grep -v ': $'
version: 3.24.4
depends_build: clang-3.7
depends_lib: curl, expat, zlib, bzip2, libarchive, ncurses, libuv, libcxx, legacy-support
```

And 10.4 :
```
> port info --version --depends /tmp/cmake os.major=8 | grep -v ': $'
version: 3.24.4
depends_build: clang-3.4
depends_lib: curl, expat, zlib, bzip2, libarchive, ncurses, libuv, libcxx, legacy-support
```

barracuda156:
@RJVB I can assure you that everything works fine with `libstdc++` up to at least C++20. This applies to PowerPC and Intel alike. (Personally I only test these up to 10.6, but there are no reasons for it not to work on 10.7+ as well.)
What Macports shows you as dependencies is conditional on a chosen runtime library. Of course, as long as `libc++` is picked in Macports configuration, C++11 etc. will need it. Change the choice, and it won’t be needed anymore.

RJVB:
>@RJVB I can assure you that everything works fine with `libstdc++` up to at least C++20. 

Well evidently. I never *really* understood (or don't remember) why the libc++ conversion of older systems was considered better than just keeping a full GCC install up-to-date. Probably something to do with potential ABI compatibility issues between libstdc++ versions. 

Oh well, good for you. Not that it'll be crucial for qsynth but my latest benchmark results of libc++ vs libstdc++ (on Linux) suggest that the latter is considerably faster in certain operations and not really slower in the majority of the other tests. :)

barracuda156:
@herbygillot @pmetzger Could we merge this please?

PR #20091, opened 2023-08-23
texlive-bin: revbump after zlib update

catap:
#### Description

Closes: https://trac.macports.org/ticket/67998

###### Type(s)

- [x] bugfix

###### Tested on

macOS 12.6.8 21G725 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/tex/texlive-bin/Portfile
+++ b/tex/texlive-bin/Portfile
@@ -10,6 +10,7 @@ PortGroup       muniversal 1.1
 
 name            texlive-bin
 version         2023.66589
+revision        1
 
 categories      tex
 maintainers     {dports @drkp}


jmroot:
How does a noarch port link with zlib?

jmroot:
`/opt/local/bin/texlua` is only a symlink. This finds no Mach-O files in the archive:
```
find texlive-basic-66584_0+doc.any_any.noarch/opt/local -type f -exec file {} \; | grep Mach-O
```

jmroot:
I'm sure that does the trick for now, but it would be better to remove the check entirely since zlib is and always has been binary compatible with programs built against previous versions.

catap:
@jmroot not always is harmless. Similar discussion at Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=581818 with link to example of libxml which crashes on different zlib: http://www.mail-archive.com/xml@gnome.org/msg07276.html

catap:
@jmroot yes, and we does it quite often for update of some libraries. Like `icu` or `onetbb` or something else.

jmroot:
I don't know about onetbb, but ICU is quite clear about it when it breaks ABI. OTOH, zlib has been providing `libz.1.dylib` for decades.

Anyway, I'll leave the decision on how to handle this to Dan.

catap:
The code in question looks like:
```
    if (strncmp(version, ZLIB_VERSION, 4))
    {
        lua_pushfstring(L, "zlib library version does not match - header: %s, library: %s", ZLIB_VERSION, version);
        lua_error(L);
    }
```
https://github.com/TeX-Live/texlive-source/blob/trunk/texk/web2c/luatexdir/luazlib/lzlib.c#L548-L553

I feel that this pice should be challenged in upstream of latex. And for us revbump seems like easy win for now.

drkp:
I'm going to go ahead and revbump it for now while I look into whether luazlib really needs to be picky about which version of zlib is installed.

I suspect we _should_ remove the check, but want to confirm that because this is already the relaxed check (the original version even checked that it matched down to patch level!)

ryandesign:
> @jmroot not always is harmless. Similar discussion at Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=581818

Yes there is discussion there but it is just a bug report that the check is too strict, and as a result they relaxed it. I don't see anything in the discussion supporting the existence of the check at all.

> with link to example of libxml which crashes on different zlib: http://www.mail-archive.com/xml@gnome.org/msg07276.html

because libxml used private zlib API that changed, so of course all bets are off in that case.

PR #20152, opened 2023-08-26
asl: new port (version 20230818)

catap:
###### Type(s)

- [x] enhancement

###### Tested on

macOS 12.6.8 21G725 x86_64
Xcode 14.2 14C18

###### Verification 
Have you

### DIFF
--- a/math/asl/Portfile
+++ b/math/asl/Portfile
@@ -1,0 +1,40 @@
+# -*- coding: utf-8; mode: tcl; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*- vim:fenc=utf-8:ft=tcl:et:sw=4:ts=4:sts=4
+
+PortSystem                      1.0
+PortGroup                       cmake 1.1
+PortGroup                       github 1.0
+PortGroup                       muniversal 1.1
+
+github.setup                    ampl asl 94efb9e220c963bddee3d4500c28e86eb560f858
+version                         20230818
+revision                        0
+
+categories                      math
+maintainers                     {@catap korins.ky:kirill} openmaintainer
+
+# License is unclear, see: https://github.com/ampl/asl/issues/7
+license                         Restrictive Unknown
+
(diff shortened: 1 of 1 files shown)


catap:
@herbygillot, @mascguy may I ask you to review and merge it?

It should unblock fix for https://trac.macports.org/ticket/68031

math/asl/Portfile
```diff
@@ -0,0 +1,40 @@
...
+
+github.setup                    ampl asl 94efb9e220c963bddee3d4500c28e86eb560f858
+version                         20230818
+revision                        0
+
+categories                      math
+maintainers                     {@catap korins.ky:kirill} openmaintainer
+
+# It is similar for so-called CWI license which used for Python between 0.9.0 and 1.2
+license                         Unknown
```
mascguy, line 16: Kirill, is the license Permissive-ish, such that we can redistribute binaries? Or should we avoid distribution, just-in-case?

CC: @jmroot

catap: @mascguy as far as I see we can redistirbute it without any reason.

See the license here: https://github.com/ampl/asl/blob/master/src/solvers/asl.h#L1-L17

jmroot: A lot of the files don't have any license header, and there's no blanket license specified anywhere, which makes things questionable at best. https://github.com/ampl/asl/issues/7

Many of the files that do have a license are an MIT variant, but you'd have to go through all of them to make sure there aren't any other licenses in there.

mascguy: > A lot of the files don't have any license header, and there's no blanket license specified anywhere, which makes things questionable at best. [ampl/asl#7](https://github.com/ampl/asl/issues/7)
> 
> Many of the files that do have a license are an MIT variant, but you'd have to go through all of them to make sure there aren't any other licenses in there.

So would `Restrictive` be the safest at this point? Or does `Unknown` provide the same net effect?

catap: To be on safe side I've added `Restrictive` and keep `Unknown`.

mascguy (approved):

PR #20161, opened 2023-08-26
Avoid overlinking abseil in users of grpc and re2

neverpanic:
#### Description

grpc and re2 export a pkg-config file that instructs all users to link against some libraries from libabseil when linking against grpc or re2. Because abseil often changes its soversion, this requires many ports that use grpc or re2 to be revbumped whenever abseil changes, even though they only use symbols from grpc or re2.

Avoid this by using the Requires.private field in the grpc and re2 pkg-config file as documented in [\[1\]][1]. See also [\[2\]][2], which explains the difference in linking.

See also https://github.com/macports/macports-ports/pull/19905#issuecomment-1694478949 for previous discussion.

[1]: https://people.freedesktop.org/~dbn/pkg-config-guide.html#concepts
[2]: https://stackoverflow.com/a/61674208

###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 12.6.8 21G725 arm64
Xcode 14.2 14C18

I built the Bear and mtxclient ports, which depend on grpc and re2 to test these changes; both built successfully and no longer overlink the abseil libraries.

###### Verification 
Have you

### DIFF
--- a/devel/grpc/Portfile
+++ b/devel/grpc/Portfile
@@ -8,7 +8,7 @@ PortGroup           compiler_blacklist_versions 1.0
 
 # NOTE: Also rev-bump 'apache-arrow' when updating this port
 github.setup        grpc grpc 1.48.4 v
-revision            1
+revision            2
 categories          devel
 maintainers         nomaintainer
 license             Apache-2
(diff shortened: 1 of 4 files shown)


mascguy (approved):
Works for me! @herbygillot ?

PR #20260, opened 2023-08-31
py-*: remove old subports without dependencies

catap:
#### Description

Goal is cleaning up unused py27...py37 ports.

How? I've run the script:
```
port echo maintainer:catap | grep '^py[23][0-7]-' | while read port
do
    if [[ -z $(port echo depends:$port) ]]; then
        echo $port
    fi
done
```
and removes all subport which it find. And repeat it a few times, until it returned an empty string.

Here I've touched only ports where I'm a (co)maintainer.

###### Type(s)

- [x] enhancement

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/python/py-about-time/Portfile
+++ b/python/py-about-time/Portfile
@@ -22,7 +22,7 @@ checksums           rmd160  c7e8941d0d92ee11976753dba045dd4df6abf065 \
                     sha256  6a538862d33ce67d997429d14998310e1dbfda6cb7d9bbfbf799c4709847fece \
                     size    15380
 
-python.versions     37 38 39 310 311
+python.versions     38 39 310 311
 
 if {${name} ne ${subport}} {
     depends_build-append \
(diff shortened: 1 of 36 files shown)


catap:
I've tried to use generic version of that script to cleanup from old subports:
```
port echo 'py[23][0-7]-*' | while read port
do
    if [[ -z $(port echo depends:$port) ]]; then
        echo $port
    fi
done
```
but impact are huge and I have no idea how to automatically validate that here no exists a non-default variant which used it :(

jmroot:
> I have no idea how to automatically validate that here no exists a non-default variant which used it :(

Yes, that's a hard problem unfortunately. You can grep the ports tree for the rootname of each module, but that will sometimes give false positives.

xeron:
Sorry I actually don't use macports anymore. I should probably remove myself from maintainers list.

catap:
@xeron np, I'm a co-maintainer and take care of pycryptodome (and pycryptodomex).

catap:
@herbygillot may I ask you to merge also this PR? Thanks!

catap:
Thanks!

PR #20291, opened 2023-09-02
vim, MacVim: update to vim version 9.0.1677

thomasnyman:
#### Description

Updated both Vim and MacVim to 9.0.1677.

- update Vim to v9.0.1677
- update MacVim to release-177
- change homepage to new official website at https://macvim.org
-  remove obsolete patchfile from MacVim as Vim now sets the Python3  library patch relative to the python framework prefix since v9.0.1441 (vim/vim@b6ebe5af74a965d703eb8ee66a07c97b0bbe6e58)

###### Type(s)
- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on
macOS 13.5.1 22G90 arm64
Xcode 14.3.1 14E300c

###### Verification 
Have you

  - Unable to run with trace-mode due to [#66356](https://trac.macports.org/ticket/66358)
  - Tested building and running both Vim and MacVim with:
    - `sudo port -vs install +python311 +tcl` 
    - `sudo port -vs install +python310`
    - `sudo port -vs install +python27`


editors/MacVim/Portfile
```diff
@@ -25,9 +25,9 @@ long_description \
 
 homepage            https://macvim-dev.github.io/macvim/
```
raimue, line 26: Should update the homepage to https://macvim.org/ per release notes.

raimue (changes requested):
Awesome work on the port updates! Thanks for adapting the patches to the new version.

However, I think the changes to the patches patch-python{,3}.diff and patch-tcl.diff that were applied to MacVim are still missing for the vim port.

editors/MacVim/files/patch-python3.diff
```diff
@@ -1,11 +1,11 @@
 --- src/configure.ac	2021-04-05 15:48:50.000000000 +0200
 +++ src/configure.ac	2021-04-05 15:48:56.000000000 +0200
-@@ -1663,7 +1663,7 @@
+@@ -1688,7 +1680,7 @@ __:
  	@echo "python3_LIBS='$(LIBS)'"
  	@echo "python3_SYSLIBS='$(SYSLIBS)'"
  	@echo "python3_DLLLIBRARY='$(DLLLIBRARY)'"
 -	@echo "python3_INSTSONAME='$(INSTSONAME)'"
 +	@echo "python3_INSTSONAME='$(PYTHONFRAMEWORKPREFIX)/$(INSTSONAME)'"
+	@echo "python3_PYTHONFRAMEWORKPREFIX='$(PYTHONFRAMEWORKPREFIX)'"
```
raimue, line 9: Actually I think this patch should no longer be needed at all. It looks like this was finally accepted/implemented upstream: https://github.com/vim/vim/commit/b6ebe5af74a965d703eb8ee66a07c97b0bbe6e58

editors/MacVim/Portfile
thomasnyman: Fixed in thomasnyman@5bca4e078d27901f63dfaba150e134d1f99b71be.

editors/MacVim/files/patch-python3.diff
thomasnyman: Thanks for the review. It does indeed appear that the `patch-python3.diff` isn't needed for MacVim any longer. I have removed the patch file and the corresponding `patchfiles-append` directive from the MacVim Portfile in thomasnyman@5bca4e078d27901f63dfaba150e134d1f99b71be.

The Vim `patch-python3.diff` is still needed for the Vim Portfile, as that now applies is selectively if building a version earlier than v9.0.1441 (which I suppose in practice only means v9.0.0472 on MacOS 10.10 and earlier due to [#67209](https://trac.macports.org/ticket/67209)).

The Vim `patch-python.diff`and `patch-tcl.diff` actually apply cleanly to v9.0.1677 which is why they are not modified in these changes.

raimue (approved):
Thanks for working in the comments. Looks good to me now. Thank you for providing the port updates! :+1:

PR #20360, opened 2023-09-09
MacVim: disable sparkle auto update

i0ntempest:
###### Type(s)

- [ ] bugfix
- [x] enhancement
- [ ] security fix

###### Tested on

macOS x.y

###### Verification 
Have you

### DIFF
--- a/editors/MacVim/Portfile
+++ b/editors/MacVim/Portfile
@@ -10,7 +10,7 @@ PortGroup           compiler_blacklist_versions 1.0
 set vim_version     9.0
 set release         177
 github.setup        macvim-dev macvim ${release} release-
-revision            0
+revision            1
 name                MacVim
 version             ${vim_version}.release${release}
 categories          editors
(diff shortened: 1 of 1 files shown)


raimue:
That means the patch alone was not enough? Or why is that now necessary?

i0ntempest:
Apparantly your patch modified the xib files to remove the entry to check update, but today installed MacVim, launched it a few times (twice maybe) and got a popup asking if I want automatic updates. So I think Sparkle is still there? Removing the appcast URL from Info.plist will render the updater completely useless.

thomasnyman:
Was this the actual updater prompt or the new "What's New"  pane which was added in r177?

image

Commit 35198990c3630519d6356ad3232a0d6176971162 disables the menu entry for the new pane, but I believe it can still be displayed once after updating the port from r176 to r177.

raimue:
You are right, we are removing the UI for the updater. In the past it was possible to remove Sparkle completely from the build, but that gets complicated to maintain over time.

The solution to remove the URL from the Info.plist seems reasonable. Let's go with this.

PR #20375, opened 2023-09-10
lima: fix templates folder not found when creating VM

MidnightRocket:
#### Description
See: lima-vm/lima#1679
The old examples folder is replaced with a symlink, to preserve any existing references, similarly to how it is done in official binary archives.

###### Type(s)

- [x] bugfix
- [ ] enhancement
- [ ] security fix

###### Tested on

macOS 13.5.1 22G90 x86_64
Xcode 14.3.1 14E300c

###### Verification 
Have you

### DIFF
--- a/sysutils/lima/Portfile
+++ b/sysutils/lima/Portfile
@@ -5,7 +5,7 @@ PortGroup           golang 1.0
 
 go.setup            github.com/lima-vm/lima 0.17.2 v
 github.tarball_from archive
-revision            0
+revision            1
 
 homepage            https://lima-vm.io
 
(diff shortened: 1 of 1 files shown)


sysutils/lima/Portfile
```diff
@@ -52,10 +52,12 @@ destroot {
...
         ${destroot}${prefix}/share/doc/lima/docs
 
-    xinstall -m 0755 -d ${destroot}${prefix}/share/lima/examples
+    xinstall -m 0755 -d ${destroot}${prefix}/share/lima/templates
     xinstall -m 0644 \
         {*}[glob ${worksrcpath}/_output/share/doc/lima/examples/*.yaml] \
-        ${destroot}${prefix}/share/lima/examples
+        ${destroot}${prefix}/share/lima/templates
+
+    ln -fs templates ${destroot}${prefix}/share/lima/examples
```
MidnightRocket, line 60: I am unsure whether or not this is the correct way of creating a relative symlink 😅. 
It works as intended on my system, I don't suppose it would be any different on other unix systems. 
But as I am still relative new to writing Portfiles, do suggest better ways if any 😃.

herbygillot: Thanks for this change, @MidnightRocket. I'd say the link should explicitly point to the explicit final path to `templates`:

```suggestion
    ln -fs ${prefix}/share/lima/templates \
            ${destroot}${prefix}/share/lima/examples
```

herbygillot:
Merged. Thank you @MidnightRocket

